PR #81, opened 2026-07-11
Build auth-page URLs from JSON-encoded constants, not raw interpolation

### LINKED ISSUE #79: Encode remaining interpolated values in WebResponse.Generator
The auth-completion page (`WebResponse.Generator`) still interpolates a few server values raw into single-quoted JS/URL string literals, after the `data` value was hardened (#78-follow / PR encoding `data` via JsonSerializer). Harden the rest for defense-in-depth:

- **`punycodeBaseUrl`** (lines ~440, ~465, ~483, ~507), the one genuinely external input. It derives from `GetRequestBase` → `Request.Host.Host` (attacker-supplied Host header) run through `IdnMapping.GetAscii`. **GetAscii is not an ASCII sanitizer** (it passes `'`, `<`, quotes through unchanged for ASCII input). Kestrel's Host-header validation rejects `<`/`>`/space, but a single quote `'` is a URI sub-delim and may pass, a review could not rule out a `'`-via-Host-header breakout from code alone. Encode it (JSON for the JS-string uses; validate/normalize the host) and/or verify Kestrel rejects `'`.
- **`provider`** (lines ~440, ~483), a route segment, but clamped to admin-defined config keys (a non-existent provider 400s before reaching Generator), so admin-controlled. Lower priority; encode as defense-in-depth (parallels the linking.js A-4 change).
- **`mode`** / **`isLinking`**, code-controlled literals; safe, no action.

Best fix: build the URLs/values server-side and pass them into the page as JSON-encoded literals (same approach as the `data` fix), rather than raw concatenation. This is the same DOM/JS-injection class as the linking.js hardening (A-4). Not known to be externally exploitable today, but the Host-derived path is the one worth closing/confirming.

Found during the SonarCloud S5131 review. Login-path change → 4-lens gate on the fix.

iderex:
Closes the last reflected-XSS surface on the SSO auth-completion page, **SonarCloud S5131 (BLOCKER)**. Closes #79.

## Why

After the `data` value was JSON-encoded (PR #80), `WebResponse.Generator` still interpolated two server values raw into single-quoted JS/URL literals:

- **`punycodeBaseUrl`**, derived from the request `Host` header; `IdnMapping.GetAscii` does **not** sanitize ASCII, so a `'` in the Host header could break out of the JS string (the one genuinely external vector; the earlier review could not rule it out from code alone).
- **`provider`**, the route segment (admin-controlled config key).

## What

Emit the three server values once as JSON-encoded JS constants (`ssoBaseUrl`, `ssoProvider`, `ssoMode`) and build every URL from them, with `encodeURIComponent(ssoProvider)` on the path segment. The default `System.Text.Json` encoder escapes `'`, `"`, `<`, `>`, `&`, `\`, and U+2028/U+2029, so no value can break out of the string or the surrounding `<script>`. No raw interpolation of untrusted values remains (only the `isLinking` bool literal).

## Behavior

Byte-identical URLs for normal providers/hosts (`encodeURIComponent` is identity for unreserved names; JSON does not escape `/` or `:`). Special provider names round-trip via ASP.NET route decoding (encode-once / decode-once), matching the linking.js A-4 approach. No lockout, no wire-contract change.

## Review

Build clean (`--warnaserror`), 151 tests pass (+2). **4-lens adversarial review (bypass / correctness / integration / availability), all PASS**, empirically confirming: no residual raw interpolation, hostile Host/provider cannot break out, normal URLs byte-identical, no const TDZ/clash, routes still resolve. E2E note logged: confirm a percent-encoded provider segment round-trips on a real server.

### DIFF
--- a/SSO-Auth.Tests/WebResponseTests.cs
+++ b/SSO-Auth.Tests/WebResponseTests.cs
@@ -41,4 +41,31 @@ public class WebResponseTests
 
         Assert.Contains("var data = \"" + base64 + "\";", html);
     }
+
+    [Fact]
+    public void Generator_EmitsProviderAsEncodedConstant_NotRawInUrl()
+    {
+        // A provider name that would break out of a single-quoted JS/URL literal if interpolated raw.
+        var hostileProvider = "p';alert(1);//";
+
+        var html = WebResponse.Generator("ZGF0YQ==", hostileProvider, "https://jf.example.com", "SAML");
+
+        // The provider is emitted once as a JSON-encoded constant and used through
+        // encodeURIComponent in the URLs, never concatenated raw.
+        Assert.Contains("const ssoProvider = " + JsonSerializer.Serialize(hostileProvider) + ";", html);
+        Assert.Contains("encodeURIComponent(ssoProvider)", html);
+        Assert.DoesNotContain("';alert(1);//", html);
+    }
+
+    [Fact]
+    public void Generator_EmitsBaseUrlAsEncodedConstant()
+    {
+        // The base URL derives from the request host, so it is treated as untrusted and emitted as
+        // a JSON-encoded constant rather than interpolated into the script.
+        var html = WebResponse.Generator("ZGF0YQ==", "keycloak", "https://jf.example.com", "SAML");
+
+        Assert.Contains("const ssoBaseUrl = \"https://jf.example.com\";", html);
+        // URLs are built from the constant, not a raw host string.
+        Assert.Contains("ssoBaseUrl + '/web/index.html'", html);
+    }
 }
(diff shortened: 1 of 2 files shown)


iderex (commented):
owner review (iderex). Copilot left no comments on this PR; confirming it reviewed and found nothing.

Adversarial review beyond Copilot: 4-lens (bypass/correctness/integration/availability) all PASS — every untrusted value (Host-derived base URL, route provider) is now a JSON-encoded JS constant, provider path segment goes through encodeURIComponent, no raw interpolation remains, normal URLs are byte-identical, and special names round-trip via ASP.NET route decoding. Closes the S5131 BLOCKER at the auth-page sink.

Gate: all required checks green and the SonarCloud quality gate is OK on this PR (new Generator tests cover the change). Cleared to merge. Closes #79.

Issue #61, opened 2026-07-11
Resolve open SonarCloud findings and enforce the quality gate


iderex:
SonarCloud reports 71 open findings (OPEN + CONFIRMED) on the current main branch:

- `scripts/check-jellyfin-compat.sh`: 15 shell findings, `[[ ]]` over `[ ]` (S7688), positional parameters into named locals (S7679), error messages to stderr (S7677).
- `SSO-Auth/Config/config.js`, `linking.js`, `linking.html`: `var` → `let`/`const` (S3504), prefer `.dataset` over `get/setAttribute` (S7761), `Boolean` over identity arrow (S7770), unused/useless assignment (S1481/S1854), named function expressions (S7726), single `classList.add` call, top-level await (S7785).
- `SSO-Auth/Views/apiClient.js`: `var` → `let`/`const`, `return value` over `return Promise.resolve(value)` (S7746), mutable `var` export (S6861), promise-aggregator misuse (S4123), stale TODO (S1135), and one minor client-side tainted-URL finding (S8476) to evaluate.
- CSS (`emby-restyle.css`, `style.css`): commented-out code (S125), duplicate selectors (S4666), three text-contrast findings (S7924) to evaluate against the intended theme.
- `configPage.html`: two inputs without an associated label (accessibility bugs).

Beyond the one-time cleanup, the SonarCloud quality gate becomes part of the standing loop: its PR check is required before merge and the findings sweep joins the post-merge security-insights routine (CodeQL, Dependabot, secret scanning, SonarCloud).

iderex:
Outcome after #64 merged and SonarCloud re-analyzed `main`: 71 → 5 open findings.

- 1 new minor (S6582, optional chaining at the reshaped filter), fixed in #65.
- 4 remain open as documented acceptances:
  - 3× S7924 text contrast (`emby-restyle.css`: delete button 3.37:1, its disabled state 2.71:1, WCAG 1.4.3 exempts inactive controls - , filter bubble 2.63:1): meeting 4.5:1 would flip the intended design; evaluated in #64.
  - 1× S7785 (`linking.html`): the flagged inline script is a classic `<script defer>`, not a module; converting to `type="module"` for top-level await would change execution semantics (real deferral, strict mode, module scope). Declined as a behavior change.

These four should be marked **Accepted** in the SonarCloud UI; the rationale above is the review trail. The quality gate itself is green and now a required PR check on `main`.

PR #83, opened 2026-07-11
Tidy pure helpers: LINQ filters, member order, argument guards

iderex:
Behavior-preserving SonarCloud maintainability cleanups on the pure helpers (Refs #61).

- **AuthStateStore.InvalidateExpired**, move the expiry test into a `Where` filter (S3267). The `ConcurrentDictionary` enumeration stays lazy over the same modification-tolerant enumerator, so the removed key set and concurrent-prune behavior are identical.
- **SamlLoginPolicy.IsLoginAllowed**, replace the inner allow-list `foreach` with `Any` (S3267); same short-circuit and null/empty/Ordinal handling.
- **SamlReplayCache**, order the static `ComputeRetention` before the instance `TryConsume` (SA1204).
- **SerializableDictionary**, guard the `IXmlSerializable` reader/writer arguments with `ArgumentNullException.ThrowIfNull` (CA1062).

Build clean (`--warnaserror`), 151 tests pass. A correctness review confirmed the two LINQ rewrites are exactly equivalent, empirically, including the concurrent-prune and null/empty/case paths, and the existing AuthStateStore/SamlLoginPolicy tests pin those paths.

### DIFF
--- a/SSO-Auth/Api/AuthStateStore.cs
+++ b/SSO-Auth/Api/AuthStateStore.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Concurrent;
+using System.Linq;
 
 namespace Jellyfin.Plugin.SSO_Auth.Api;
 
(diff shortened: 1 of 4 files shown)


iderex (commented):

These are behavior-preserving helper cleanups (S3267 LINQ filters, SA1204 member order, CA1062 arg guards). A correctness review empirically confirmed the two LINQ rewrites are exactly equivalent (ConcurrentDictionary lazy-enumeration prune + Any short-circuit with null/empty/Ordinal handling); the AuthStateStore/SamlLoginPolicy tests pin those paths. Build clean, 151 tests pass, gate OK. Cleared to merge.

PR #84, opened 2026-07-11
Tidy Saml.cs: static helper, LINQ audience check, dead comments, guard

iderex:
Behavior-preserving SonarCloud maintainability cleanups on the SAML core (Refs #61). A SAML-domain review confirmed the signature/algorithm-allowlist/time-bound/audience fail-closed semantics are unchanged.

- **IsSignatureAlgorithmAllowed** made `static` (it uses only its `SignedXml` argument and the static allowlist; no instance state), still runs in the same position before `CheckSignature` (S2325/CA1822).
- **IsValid(audience)** audience match uses `Any` instead of a `foreach`, same `Ordinal` comparison, short-circuit, and empty→`false` (S3267).
- Remove two commented-out code blocks in `AuthRequest.GetRequest` (S125); the emitted AuthnRequest XML is unchanged (both were fully commented out).
- **GetRedirectUrl** guards a null `samlEndpoint` with `ArgumentNullException` rather than a bare `NullReferenceException` (CA1062).

Not touched here: the pure member-ordering findings (SA1202/S4136), a strict public-then-private reorder of the crypto file is a separate, dedicated PR to keep this diff behavior-focused and low-churn.

Build clean (`--warnaserror`), 151 tests pass (audience matching is covered by SamlResponseTests).

### DIFF
--- a/SSO-Auth/Saml.cs
+++ b/SSO-Auth/Saml.cs
@@ -10,6 +10,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 using System.Security.Cryptography.Xml;
 using System.Text;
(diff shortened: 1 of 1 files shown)


iderex (commented):

Behavior-preserving SAML-core cleanups (static helper, LINQ audience match, dead-comment removal, null guard). A SAML-domain review confirmed the signature/algorithm-allowlist/time/audience fail-closed semantics are unchanged and the emitted AuthnRequest XML is identical. Build clean, 151 tests pass (audience matching covered), gate OK. Cleared to merge.

PR #88, opened 2026-07-11
Use structured logging templates in two controller log calls

iderex:
Replace a string-interpolated (`$"Is request linking: {isLinking}"`) and a string-concatenated (`"Set default login provider to " + defaultProvider`) `LogInformation` with message templates (**S2629/CA2254**). Same log output; both values are non-request data (a bool and the admin-configured default provider), so no log-forging sanitization change is needed. Build clean, 160 tests pass.

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -318,7 +318,7 @@ public class SSOController : ControllerBase
 
             if (timedState.Valid)
             {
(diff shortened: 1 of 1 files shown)


iderex (commented):

Two LogInformation calls converted to structured message templates (S2629/CA2254). Log output is unchanged; both interpolated values are non-request data (a bool and the admin-configured default provider), so no log-forging sanitization change was needed. Build clean, 160 tests pass, gate OK (green). Cleared to merge.

Issue #89, opened 2026-07-11
Decompose the OID/SAML role-to-privilege mapping in SSOController


iderex:
The last 20 open SonarCloud findings are all concentrated in the role-to-privilege assignment loops of `OidPost`, `SamlAuth`, and `Authenticate`:

- **S3776 (x3)**: cognitive complexity of `OidPost`, `SamlAuth`, `Authenticate`.
- **S3267 (x9)** and **CA1860 (x5)**: the per-role `foreach`/`if` matching loops that set Valid/Admin/Folders/LiveTv (`config.Roles.Any()` on arrays → `.Length > 0`; loops → LINQ).
- **S1066 (x2)**: collapsible nested `if`s in those blocks.
- **S107 (x1)**: `Authenticate` has 10 parameters (candidate for a parameter object).

These were **deliberately deferred** during the maintainability sweep: they are the actual privilege-decision logic on the login path, and the controller has no unit coverage of it yet (that is *why* coverage sits low). The OID and SAML blocks are **not identical** (SAML has no `Valid` computation and different null/Trim handling), so they cannot be merged into one shared helper without a behavior change, a faithful, per-path extraction with characterization tests written first is required.

**Plan (each a separate reviewed PR):**
1. Extract the OID role→privilege mapping into a pure, tested helper (roles + config → Valid/Admin/Folders/LiveTv/LiveTvManagement), preserving every current null/empty/Trim quirk exactly.
2. Same for the SAML path.
3. Convert the now-isolated matching loops to LINQ (clears S3267/CA1860) inside the helpers, covered by the new tests.
4. Consider a parameter object for `Authenticate` (S107).

This also lifts `new_coverage` toward the 80% goal, since the extracted helpers become unit-testable (see the free-plan coverage policy). Recommend doing this **supervised** given it is privilege-assignment logic; the full 4-lens gate (bypass especially) plus characterization tests must pass before merge.

iderex:
Progress: both role→privilege mappings are now extracted into pure, fully-tested helpers and merged.

- **#90**, OID path → `OidcRolePrivilegeMapper.Evaluate(roles, config)` (13 characterization tests).
- **#91**, SAML path → `SamlRolePrivilegeMapper.Evaluate(roles, config)` (14 tests).

[...]

PR #90, opened 2026-07-11
Extract the OID role-to-privilege mapping into a tested pure helper

iderex:
First half of the role-mapping decomposition tracked in #89 (the OID path).

`OidPost` inlined, per claim and per role, the logic that turns roles into privileges, login validity, admin, folder access, Live TV, Live TV management. It was the bulk of `OidPost`'s cognitive complexity (**S3776**) and had no unit coverage.

**Change:** collect the roles across all matching claims, then map them with the new pure `OidcRolePrivilegeMapper.Evaluate(roles, config)` (returns a `RoleGrants` record) and merge into the authorize state, OR the booleans, append the folders. The merge is **monotonic**, which preserves the pre-initialized config defaults exactly.

**Faithfulness:** verified by two independent adversarial reviews (bypass + correctness). Identical guards, ordinal `role.Equals`, the `EnableFolderRoles`/`EnableLiveTvRoles` master switches, folder order/duplicates, and, deliberately, via `foreach` not LINQ `Where`, the **pre-existing NullReferenceException** when `EnableFolderRoles` is set but `FolderRoleMapping` is null (admin misconfiguration; fails closed with a 500). 13 new characterization tests pin every branch, including that NRE.

Clears **S3776/S3267/CA1860/S1066** in `OidPost`. Remaining in #89: the SAML path, the folder-loop S3267 (kept as a faithful `foreach`), and the `Authenticate` parameter object (S107).

Build clean (`--warnaserror`), 173 tests pass. SonarCloud will likely be red on new_coverage only (the controller merge lines), the accepted free-plan state.

### DIFF
--- a/SSO-Auth.Tests/OidcRolePrivilegeMapperTests.cs
+++ b/SSO-Auth.Tests/OidcRolePrivilegeMapperTests.cs
@@ -1,0 +1,202 @@
+using System;
+using System.Collections.Generic;
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Characterization tests for <see cref="OidcRolePrivilegeMapper.Evaluate"/> — the role→privilege
+/// mapping extracted from the OID callback. These pin the exact behavior (including its quirks) so
+/// the extraction is a proven no-op and the mapping can be reasoned about in isolation.
+/// </summary>
+public class OidcRolePrivilegeMapperTests
+{
+    private static OidConfig Config(Action<OidConfig> configure)
+    {
+        var config = new OidConfig();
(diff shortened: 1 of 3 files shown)


iderex (commented):

This is the OID half of the role-mapping decomposition (#89) and the highest-risk change in the sweep (privilege assignment), so it went through TWO independent adversarial reviews:
- bypass-lens: could not construct any (claims + config) where the refactor grants a privilege main withheld or withholds one main granted — exact equivalence on multi-claim aggregation, monotonic OR-merge onto the pre-initialized state, all guards, ordinal comparison, master switches, and folder order/duplicates.
- correctness-lens (independent): confirmed the same, and that the folder-loop NRE on null FolderRoleMapping is faithfully preserved (foreach, not LINQ Where) and pinned by a passing test.

13 new characterization tests cover every branch. Build clean, 173 tests pass, and the SonarCloud quality gate is GREEN on this PR (the fully-tested helper lifts new-code coverage). Cleared to merge.

PR #91, opened 2026-07-11
Extract the SAML role-to-privilege mapping into a tested pure helper

iderex:
Second half of the role-mapping decomposition (#89), the SAML path, pairs with the OID extraction in #90.

`SamlAuth` inlined, per role, the logic that grants admin, folder access, and Live TV. Move it verbatim into the pure `SamlRolePrivilegeMapper.Evaluate(roles, config)` and merge into the pre-initialized locals (OR the booleans, append the folders), the grants are monotonic, so the config defaults are preserved. Login validity stays with `SamlLoginPolicy`.

**Faithful**, verified by two independent adversarial reviews (bypass + correctness). The SAML quirks are preserved exactly and deliberately differ from the OID mapper: the comparison receiver is the configured value (so a null configured entry throws, **even after an earlier match**, kept as a no-break `foreach`, not LINQ `Any` which would short-circuit), `FolderRoleMapping` IS null-checked (no NRE), and the folder role is matched **without** trimming. 14 characterization tests pin every branch and each of those differences.

Clears **S3776** and collapses **S1066** in `SamlAuth`. The matching loops stay as faithful `foreach` (their no-break/receiver exception semantics must be preserved, so they are NOT converted to LINQ, the S3267 there is left deliberately), tracked in #89. Build clean (`--warnaserror`), 187 tests pass.

### DIFF
--- a/SSO-Auth.Tests/SamlRolePrivilegeMapperTests.cs
+++ b/SSO-Auth.Tests/SamlRolePrivilegeMapperTests.cs
@@ -1,0 +1,198 @@
+using System;
+using System.Collections.Generic;
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Characterization tests for <see cref="SamlRolePrivilegeMapper.Evaluate"/> — the role→privilege
+/// mapping extracted from the SAML callback. These pin the exact behavior, including the quirks that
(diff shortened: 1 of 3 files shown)


iderex (commented):

The SAML half of the role-mapping decomposition (#89), and privilege-assignment logic, so it went through two independent adversarial reviews (bypass + correctness). Both confirmed exact equivalence: the monotonic OR/append merge reproduces every field including config-default-true Live TV, and the SAML-specific quirks are preserved verbatim and correctly distinguished from the OID path (config-value receiver → NRE on null entry even after a match, kept as no-break foreach; FolderRoleMapping null-checked; no trim). 14 characterization tests pin every branch and each difference. Build clean, 187 tests pass, gate GREEN. Cleared to merge.

PR #92, opened 2026-07-11
Use Length instead of Any() for the role-claim segment guard

iderex:
`roleClaimSegments` is a non-null `string[]` (either `Array.Empty<string>()` or the split result), so `.Length > 0` is exactly equivalent to `.Any()` and is the idiomatic array check (**CA1860**). Behavior unchanged; the guard is the role-claim-type check, not part of any privilege-decision loop. Build clean, 187 tests pass.

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -226,7 +226,7 @@ public class SSOController : ControllerBase
                 }
 
                 // Collect the roles carried by every claim on the configured role-claim path.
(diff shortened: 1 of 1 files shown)


iderex (commented):
owner review (iderex). Copilot left no comments; confirming it reviewed and found nothing. roleClaimSegments is a non-null string[], so .Length > 0 is exactly equivalent to .Any() (CA1860); behavior unchanged, not in a decision loop. Build clean, 187 tests pass, gate OK. Cleared to merge.

PR #96, opened 2026-07-11
Extract SAML authorize-state derivation into a pure builder

iderex:
## What

Extracts the per-login privilege derivation from the SAML callback (`SamlAuth`) into a pure, testable `SamlAuthorizeStateBuilder.Build(roles, config)`. The callback now reads the returned values into the session parameters instead of computing them inline. This mirrors the OpenID extraction (#94) on the SAML side and clears `SamlAuth`'s cognitive-complexity finding (S3776, `Refs #89`).

## Why it is behavior-preserving

The derived values were plain locals flowing by value into a fresh `SessionParameters` object initializer, no pre-existing mutable state, so moving their computation is a pure value move. The builder reproduces the inline block one-for-one:

- the folder pre-init ternary (`!EnableFolderRoles && EnabledFolders != null` → copy, else empty), then the same `AddRange(grants.Folders)`;
- the boolean seeds (`false` for admin, the config defaults for Live TV) OR-merged with the role grants (`x |= y` on an already-materialized value is exactly `x | y`);
- the same single `SamlRolePrivilegeMapper.Evaluate` call on the same roles (`GetCustomAttributes("Role")` is still invoked exactly twice, at the same program points).

Login validity (`SamlLoginPolicy`) and the username (NameID) are decided elsewhere and stay untouched; the guard ordering (SAML validity → login allow-list → replay consume → derivation → user creation) is unchanged.

## Tests

9 characterization tests pin the derivation in isolation: folder defaulting on/off, admin grant, non-matching role, Live TV config-default pass-through, a positive role-grant test for **each** Live TV flag (so a dropped OR-merge cannot slip through), and the `EnableLiveTvRoles`-off gating.

## Verification

- Build clean (`--warnaserror --no-incremental`, Debug).
- Full suite green: 211/211.
- Two independent adversarial reviews (bypass-lens + correctness-lens) proved exact equivalence: identical folder branches, verbatim boolean seeds, unchanged call counts and guard ordering, `SamlRolePrivilegeMapper` untouched. The one review finding (a missing positive Live TV role-grant test) is fixed in this PR.

Refs #89

### DIFF
--- a/SSO-Auth.Tests/SamlAuthorizeStateBuilderTests.cs
+++ b/SSO-Auth.Tests/SamlAuthorizeStateBuilderTests.cs
@@ -1,0 +1,154 @@
+using System.Collections.Generic;
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Characterization tests for <see cref="SamlAuthorizeStateBuilder.Build"/> — the admin/Live TV/
+/// folder privilege derivation extracted from the SAML callback. These pin the exact behavior (folder
+/// defaulting, the config-default-then-role-grant merge for Live TV, and the admin/folder grants) so
+/// the extraction is a proven no-op. Login validity and the username are decided elsewhere, so they
+/// are not part of the derivation and not tested here.
+/// </summary>
+public class SamlAuthorizeStateBuilderTests
+{
+    private static SamlConfig Config(System.Action<SamlConfig> configure)
+    {
+        var config = new SamlConfig();
+        configure(config);
+        return config;
+    }
+
+    [Fact]
+    public void NoRoles_NoConfigDefaults_YieldsNoPrivileges()
+    {
+        var result = SamlAuthorizeStateBuilder.Build(new List<string>(), Config(_ => { }));
+
+        Assert.False(result.Admin);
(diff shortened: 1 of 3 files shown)


iderex (commented):
pre-merge review (iderex).

**Copilot review status:** Copilot could not review this PR — "user who requested the review has reached their quota limit" (free-plan constraint). There are zero Copilot comments to evaluate; documenting that here per the documented-accept policy rather than leaving the gate silently unmet. The review burden is carried by the adversarial gate below.

**Verification of the extraction:**
- Two independent adversarial reviews (bypass-lens + correctness-lens) proved exact equivalence: the folder pre-init ternary is identical, the boolean seeds are verbatim (`false | Admin`, `config.EnableLiveTv[/Management] | grant` — `|=` on materialized bools is exactly `|`), `GetCustomAttributes("Role")` is still invoked exactly twice at the same program points, and the guard ordering (SAML validity → login allow-list → replay consume → derivation → user creation) is unchanged.
- `SamlRolePrivilegeMapper` is untouched (empty diff vs main); its preserved quirks (config-value receiver `.Equals`, null-checked `FolderRoleMapping`) flow through the builder unchanged.
- The one review finding — no positive role-grant test for the plain `EnableLiveTv` flag — was fixed in this PR before commit (each Live TV flag now has a positive role-grant test, so a dropped OR-merge cannot slip through).
- Build clean (`--warnaserror`), 211/211 tests green, all CI checks pass, SonarCloud gate green.

**Overall verdict: APPROVE (sign-off).** Merging.

PR #97, opened 2026-07-11
Deduplicate the OIDC client construction

iderex:
## What

`OidPost` and `OidChallenge` assembled near-identical `OidcClientOptions` inline (~25 lines each; token-identical apart from `RedirectUri` and `Scope`). This extracts a shared `CreateOidcClient(config, redirectUri, scope)` core plus a thin `CreateCallbackOidcClient(config, provider)` wrapper that owns the callback's two former ternaries (the scopes null-guard and the redirect-path segment). Removes the duplication and takes `OidPost` under the cognitive-complexity threshold (S3776, `Refs #89`).

## Why it is behavior-preserving

- The two original inline blocks were verified **token-identical** in every shared field (both trims, `DisablePushedAuthorization`, `LoadProfile`, the `HttpClientFactory` lambda, the endpoint `Uri`, and the three `Policy.Discovery` negations `RequireHttps`/`ValidateIssuerName`/`ValidateEndpoints`), only `RedirectUri`/`Scope` differed, which are exactly the core's two parameters.
- Property-initializer order and failure points are preserved: a null `OidEndpoint` still throws at the `Uri` constructor after options assembly, in both callers.
- The caller asymmetries stay on their original sides: the callback tolerates a null scopes array (yielding the same `"openid profile  "` join artifact, empirically verified), the challenge still throws on it; the challenge's `NewPath`/`isLinking` handling is untouched.
- The callback's redirect-segment expression is kept verbatim, including its pre-existing always-`"r"` quirk (the `/start/` test never matches a callback path), now documented at the wrapper and tracked separately rather than silently changed.

## Verification

- Build clean (`--warnaserror --no-incremental`, Debug); 211/211 tests green.
- Two independent adversarial reviews (bypass-lens + correctness-lens) proved byte-equivalence, including a mechanical diff of both original blocks against the shared core and empirical probes of the null-scopes and exception-ordering behavior.

Refs #89

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -145,30 +145,7 @@ public class SSOController : ControllerBase
                 return BadRequest("Invalid or expired state");
             }
 
-            var scopes = config.OidScopes == null ? new string[2] : config.OidScopes;
-            var options = new OidcClientOptions
-            {
-                Authority = config.OidEndpoint?.Trim(),
-                ClientId = config.OidClientId?.Trim(),
-                ClientSecret = config.OidSecret?.Trim(),
-                RedirectUri = GetRequestBase(config.SchemeOverride, config.PortOverride) + $"/sso/OID/{(Request.Path.Value.Contains("/start/", StringComparison.InvariantCultureIgnoreCase) ? "redirect" : "r")}/" + provider,
-                Scope = string.Join(" ", scopes.Prepend("openid profile")),
-                DisablePushedAuthorization = config.DisablePushedAuthorization,
-                LoggerFactory = _loggerFactory,
-                LoadProfile = !config.DoNotLoadProfile,
-                HttpClientFactory = o =>
-                {
-                    var client = _httpClientFactory.CreateClient();
-                    client.DefaultRequestHeaders.UserAgent.ParseAdd(UserAgentString);
-                    return client;
-                }
-            };
-            var oidEndpointUri = new Uri(config.OidEndpoint?.Trim());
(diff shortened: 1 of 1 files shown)


iderex (commented):
pre-merge review (iderex).

**Review process note:** per me's decision of 2026-07-11, the mandatory Copilot review is removed from the PR process (quota exhausted, free plan). The independent-review burden is carried by the internal adversarial gate, documented below; Copilot comments, should they appear again, will still be evaluated and answered per the old rule.

**Verification of the dedup:**
- Two independent adversarial reviews (bypass-lens + correctness-lens) proved byte-equivalence: both original inline blocks were token-identical in every shared field (mechanically diffed against `main`); the shared core preserves property order, all three `Policy.Discovery` negations, both trims, and the null-`OidEndpoint` failure point; the caller asymmetries (callback scopes null-guard with its `"openid profile  "` artifact — empirically probed; challenge unguarded) stay on their original sides.
- One review finding was a misleading comment on the wrapper (claimed a redirect mirroring the code does not do) — fixed before commit; the underlying pre-existing always-`"r"` quirk is filed as #98 rather than silently changed in a no-op refactor.
- Build clean (`--warnaserror`), 211/211 tests green, all functional CI checks pass.

**Quality gate:** red on `new_coverage` ONLY (0% — the moved code constructs the OIDC network client and is not unit-testable; verified via `api/qualitygates/project_status`, sole failing condition `new_coverage 0.0 < 80`). Gate red on coverage only — accepted per free-plan policy; merging via the ruleset admin bypass. The single PR finding (SA1202, public-before-private ordering) is the documented-accepted helper-near-caller grouping.

**Overall verdict: APPROVE (sign-off).** Merging.

PR #109, opened 2026-07-11
Use logical OR in the SAML Live TV grant merge

iderex:
Fixes the two SonarCloud S2178 findings (`Correct this '|' to '||'`) in `SamlAuthorizeStateBuilder` from the post-merge sweep.

Both operands are side-effect-free, already-materialized bools (the grants struct is computed before the merge), so the switch to the short-circuit operator is exactly equivalent. Two independent adversarial reviews (bypass + correctness) proved total equivalence, identical results, exceptions, and evaluation order on every input, and both confirmed the characterization tests pin the truth table. The one unpinned row they flagged, (true, true), is added as a test in this PR (an accidental XOR would have survived the suite before).

Build clean (`--warnaserror`), 212/212 tests green.


iderex:
@coderabbitai review

iderex (commented):
pre-merge review (iderex).

Fixes the two S2178 findings from the post-merge sweep: the SAML Live TV grant merge now uses the logical OR. Two independent adversarial reviews (bypass + correctness) proved total equivalence — both operands are side-effect-free, already-materialized bools (auto-property + readonly record-struct field), the grants evaluation precedes both merges, and the guard ordering is untouched. The one review finding (the unpinned true/true truth-table row, where an accidental XOR would have survived) is fixed in this PR with a test. CodeRabbit reviewed on request with no findings. Branch revalidated on top of the analyzer posture from #107; build clean, 212/212 tests, all CI checks green, SonarCloud gate green.

Issue #95, opened 2026-07-11
OID/SAML: fail closed when a login is valid but resolves no username


iderex:
## Summary

An OpenID login can end up **valid with a null/empty username**, which then throws downstream (a 500) instead of taking the clean invalid-login path. Surfaced by the Copilot reviewer on #94 (a behavior-preserving refactor); the behavior is **pre-existing on `main`**, so it is tracked here rather than changed inside that no-op extraction.

## How it happens

In `OidPost` (now via `OidcAuthorizeStateBuilder.Build`), login validity and the username are derived independently:

- `Valid` becomes true if a role matches the allow-list (`grants.Valid`, i.e. `config.Roles` is non-empty and a role claim matches), **or** if the preferred-username/`DefaultUsernameClaim` claim is present while no allow-list is configured.
- `Username` is set **only** by the preferred-username / `DefaultUsernameClaim` branch, or by the `sub` fallback, and the `sub` fallback runs **only when `!Valid`**.

So if the provider grants validity purely through a **role allow-list match** but sends **no** preferred-username / `DefaultUsernameClaim` claim, `Valid == true` while `Username == null`, and the `sub` fallback is skipped (because `Valid` is already true).

## Impact

`OidAuth` then calls `CreateCanonicalLinkAndUserIfNotExist("oid", provider, timedState.Username, …)` with a null name; the create path reaches `_userManager.CreateUserAsync(null)`, which throws → **HTTP 500** instead of a clean invalid-login response. This is **fail-closed** (no unauthorized access is granted, the request errors out), but it is a poor failure mode and a latent correctness bug. The SAML path (`SamlAuth` → `GetNameID()`) should be checked for the analogous null/empty case.

## Proposed fix (separate PR)

In the callback, after deriving the authorize state, treat `Valid && string.IsNullOrEmpty(Username)` as an **invalid login** (return the existing invalid-login path, not a 500). Add negative tests: role-only allow-list validity with no username claim, and the SAML null-NameID analogue. Keep the change fail-closed.

Refs #89

PR #113, opened 2026-07-11
Deduplicate the denial message and drop null-guards the belt made dead

iderex:
Post-merge sweep after #112, the three Sonar findings the hardening introduced:

- **S1192**: the generic 401 denial message appeared as four identical literals → a single `PermissionDeniedMessage` constant, byte-identical (verified against `main`, no client or test string-matches it), deliberately uniform so a rejection does not reveal its cause.
- **S2589 ×2**: the identity belt guards `CreateCanonicalLinkAndUserIfNotExist` unconditionally, so the two null-conditional accesses behind it were provably dead → plain dereferences, sanitizers kept inline.

Two adversarial reviews (correctness + bypass) proved behavior identity: the belt dominates both dereference sites with no path around it, `Valid=true` with a null/whitespace username is unconstructible after #112, all four denial sites keep the same status and bytes, and no sanitizer moved. Build clean (`--warnaserror`), 221/221 tests.

Refs #95

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -41,9 +41,11 @@ namespace Jellyfin.Plugin.SSO_Auth.Api;
 [Route("[controller]")]
 public class SSOController : ControllerBase
 {
-    // Error messages returned when a provider name does not resolve to a configured provider.
+    // Client-facing error messages: unresolved provider names, and the generic denial for any
+    // rejected login (deliberately uniform so the response does not reveal why it was rejected).
     private const string NoMatchingProviderMessage = "No matching provider found";
     private const string ProviderDoesNotExistMessage = "Provider does not exist";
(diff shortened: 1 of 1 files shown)


iderex (commented):
pre-merge review (iderex).

Post-merge sweep for the three Sonar findings #112 introduced. Two adversarial reviews (correctness + bypass) proved behavior identity: the const is byte-identical to all four (and only four) replaced literals with no client/test string-matching it; the two removed null-conditionals are dominated by the unconditional identity belt with no path around it and no reassignment (Valid=true with a null/whitespace username is unconstructible after #112); sanitizers stay inline; all denial sites keep the same status and bytes. One review nit (the const-block comment no longer covering the new constant) fixed before commit. CodeRabbit reviewed with no findings. Build clean (`--warnaserror`), 221/221 tests, all CI checks green, SonarCloud gate green.

PR #114, opened 2026-07-11
Rebuild the OID callback redirect_uri from the callback's own route

### LINKED ISSUE #98: OID callback rebuilds redirect_uri with the wrong path segment on the new-path flow
## Summary

The OID callback rebuilds the client's `redirect_uri` for the token exchange with a path-segment test that can never match, so the new-path flow sends a mismatched `redirect_uri` to the IdP. Pre-existing (inherited from upstream); surfaced by the adversarial review on #97, where the expression was deliberately preserved verbatim.

## Details

`OidChallenge` builds the authorize-request `redirect_uri` as `…/sso/OID/{newPath ? "redirect" : "r"}/{provider}`, so a login started via `/OID/start/{provider}` advertises `…/OID/redirect/{provider}`.

The callback (`OidPost`, now `CreateCallbackOidcClient`) rebuilds the redirect URI for `ProcessResponseAsync` with:

```csharp
Request.Path.Value.Contains("/start/", StringComparison.InvariantCultureIgnoreCase) ? "redirect" : "r"
```

But the callback routes are `OID/r/{provider}` and `OID/redirect/{provider}`, the path never contains `/start/`, so the segment is **always `"r"`**. On the new-path flow the token request therefore carries `redirect_uri=…/OID/r/…` while the authorization request used `…/OID/redirect/…`.

## Impact

RFC 6749 §4.1.3 requires the token-request `redirect_uri` to match the one used in the authorization request; spec-compliant IdPs reject the code exchange. Net effect: the `/start/` (new-path) flow fails against strict providers, **fail-closed** (no security impact), but functionally broken. The classic `/p/` flow (`newPath` false, both sides `"r"`) is unaffected, which is presumably why this survived unnoticed.

## Proposed fix (separate PR)

Test the callback path for the actual route segment (`"/redirect/"`), or better, carry the challenge-side choice through the stored authorize state instead of re-deriving it from the path. Add a characterization test either way.

## Related

`OidChallenge` also persists `config.NewPath = newPath` on non-linking challenges, an unsynchronized write to live shared config noted in the same review; evaluate folding it into the same fix (the stored-state approach would remove the need for it on this path).

iderex:
Fixes #98. The OID callback rebuilt the token-request `redirect_uri` segment with a `/start/` test that never matches a callback route (`OID/r` or `OID/redirect`), so it always emitted `r`. A login started on the new-path route (`OID/start`, which advertises `OID/redirect`) then sent `redirect_uri=.../r/...` to the token endpoint while the authorization request used `.../redirect/...`, RFC 6749 §4.1.3 mismatch, rejected by spec-enforcing IdPs. The classic `/p/` flow (both sides `r`) was unaffected, which is why it went unnoticed.

## Fix

A pure, tested helper `OidcCallbackPath.RedirectSegment` derives the segment from the callback's own route. The IdP delivers the callback on exactly the route the authorization request advertised, so mirroring the callback path makes both `redirect_uri`s match for every flow (including linking, which reuses `config.NewPath`, now no longer read at callback time). The match is **segment-exact** (the element after `OID`), not a substring, so a provider literally named `redirect` cannot flip the classic route even with a trailing slash.

## Verification

- Tests first; 230/230 green, build clean (`--warnaserror`). Nine helper tests pin both routes, case-insensitivity, the provider-named-`redirect` corner (classic and new-path), and null/empty paths.
- Two independent adversarial reviews (bypass + correctness) PASS on the base fix: both traced the full mint chain (nothing mints before a successful token exchange; `redirect_uri` is an IdP-side verification value, so any wrong segment fails closed at the exchange, never a bypass) and empirically probed `%2F`, trailing-slash, query-injection, and null-path manipulation on Kestrel. Both flagged one non-blocking, fail-closed corner (substring match + provider named `redirect` + trailing slash); this PR closes it outright with segment-exact matching and a pinning test.

Closes #98

### DIFF
--- a/SSO-Auth.Tests/OidcCallbackPathTests.cs
+++ b/SSO-Auth.Tests/OidcCallbackPathTests.cs
@@ -1,0 +1,56 @@
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Tests for <see cref="OidcCallbackPath.RedirectSegment"/> — the r/redirect choice for rebuilding
+/// the callback's redirect_uri. The token request's redirect_uri must match the authorization
+/// request's, and the IdP calls back on exactly the advertised route, so the segment must mirror
+/// the callback path. The old inline expression tested for "/start/" (never present on callback
+/// routes) and therefore always produced "r", breaking the new-path flow against spec-enforcing
+/// IdPs (#98).
+/// </summary>
+public class OidcCallbackPathTests
+{
+    [Theory]
+    [InlineData("/sso/OID/redirect/keycloak", "redirect")]
+    [InlineData("/sso/OID/r/keycloak", "r")]
+    [InlineData("/SSO/OID/Redirect/keycloak", "redirect")]
+    public void RedirectSegment_MirrorsTheCallbackRoute(string path, string expected)
+    {
+        Assert.Equal(expected, OidcCallbackPath.RedirectSegment(path));
+    }
+
+    [Theory]
+    // A provider NAMED "redirect" must never flip the classic route — including with a trailing
+    // slash, which a substring "/redirect/" test would wrongly match (the N1 corner both reviews
+    // flagged). Segment-exact matching keys only on the element right after "OID".
+    [InlineData("/sso/OID/r/redirect")]
+    [InlineData("/sso/OID/r/redirect/")]
+    [InlineData("/sso/OID/r/my-redirect-provider")]
+    public void RedirectSegment_ProviderNamedRedirect_OnClassicRoute_StaysR(string path)
+    {
+        Assert.Equal("r", OidcCallbackPath.RedirectSegment(path));
+    }
+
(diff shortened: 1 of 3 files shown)


iderex (commented):
pre-merge review (iderex).

Fixes the RFC 6749 §4.1.3 redirect_uri mismatch that broke the OIDC new-path flow against spec-enforcing IdPs (#98). Two independent adversarial reviews (bypass + correctness) PASS on the base fix: both traced the mint chain (nothing mints before a successful token exchange; redirect_uri is an IdP-side verification value → any wrong segment fails closed, never a bypass), verified the challenge/callback pairing holds for every flow including linking, and empirically probed %2F / trailing-slash / query-injection / null-path on Kestrel. The single non-blocking corner both flagged (substring match + provider named "redirect" + trailing slash, fail-closed) is closed outright in this PR by switching to segment-exact matching, with a pinning test — a strict robustness tightening over the reviewed version (any wrong segment still fails closed regardless of derivation). CodeRabbit reviewed with no findings. Tests-first, 230/230 green, build clean, all CI checks green, SonarCloud gate green.

Issue #104, opened 2026-07-11
Work through the analyzer-adoption backlog (rule family by rule family)


iderex:
The analyzer hardening (Refs #101) lands with the Security category at error and the newly surfaced non-security findings demoted to suggestion in `.editorconfig`, so the net could land green without folding behavior changes into a config PR. This issue tracks working through that backlog deliberately, one rule family per PR, each with the review gate:

**High priority (identity-path semantics, needs its own reviewed PR with a migration consideration):**
- `CA1304`/`CA1310`: culture-dependent `ToLower()`/`IndexOf(string)` on the identity path (`AddCanonicalLink`, `WebResponse.Generator`). Culture-sensitive casing can split or collide identities (Turkish-I class). Fixing changes the canonical-link keys of existing installations, requires a compatibility/migration decision, not a mechanical replace.

**Normal priority:**
- `CA1309`/`MA0006`/`MA0002`/`MA0074`/`MA0011`: explicit ordinal string comparisons, behavior-relevant on the login path, fix with characterization tests.
- `MA0015`: ArgumentException overloads with the parameter name (mechanical, low risk).
- `CA1848`: LoggerMessage delegates, the plugin's log volume is trivial; adopt only with a measured reason or close as accepted.
- `MA0016`, `MA0048`, `MA0051`, `MA0158`, `RCS1194`, `MA0008`: style/structure, batch or accept with reasons.

**Accepted (documented in `.editorconfig`, not planned):**
- `CA1707` (the `SSO_Auth` namespace underscore is load-bearing, persisted `AuthenticationProviderId`).
- `CA1861`/`CA1825` in test code (inline arrays keep cases readable; standing decision).

iderex:
The backlog is now also visible in SonarCloud: the CI scan imports the analyzers' suggestion-severity diagnostics as ~142 external (roslyn) findings on main. They stay open deliberately, this issue is the documented reason, and get closed family by family as the backlog is worked through; they do not fail builds (suggestion severity) and are not expected to trip the PR quality gate on old code.

iderex:
**Re-scoping after a code re-check (2026-07-18): the high-priority migration item is already mitigated.**

The original high-priority concern was a culture-sensitive `ToLower()` identity-path dispatch (Turkish-I collision/split), requiring a canonical-link-key migration decision. Re-reading the current `4.2` code:

[...]

PR #130, opened 2026-07-12
Make string comparisons explicitly ordinal on the auth path

### LINKED ISSUE #129: Make string comparisons explicitly ordinal on the auth path (CA1309/MA0006)
First family from the analyzer-adoption backlog (#104), chosen because it is security-adjacent and behavior-preserving: make the role/claim/scheme string comparisons explicitly `StringComparison.Ordinal`.

C# `string ==`/`!=` and instance `string.Equals(string)` are already ordinal, so this changes no match outcome, but stating it explicitly on the security-critical comparisons (role allow-list, admin/Live-TV role matching, claim-type matching, the avatar SSRF scheme allowlist) prevents a future refactor from silently introducing a culture-sensitive comparison, and it clears the CA1309/MA0006 family (promoted to error in `.editorconfig`).

Care taken: the two role mappers keep the INSTANCE `.Equals(x, Ordinal)` overload (receiver unchanged), preserving the SAML mapper's documented null-configured-entry NRE quirk (pinned by `NullConfiguredAdminRole_Throws`); only the `==`/`!=` operator sites become the null-safe static `string.Equals(...)`. Refs #104.

iderex:
Closes #129 (first analyzer-backlog family from #104, chosen because it is security-adjacent and behavior-preserving).

Converts 18 comparison sites to explicit `StringComparison.Ordinal` and promotes `CA1309` + `MA0006` from suggestion to error. C# `string ==`/`!=` and instance `string.Equals(string)` are already ordinal, so no match outcome changes, but making it explicit on the security-critical comparisons (role allow-list, admin/Live-TV matching, claim-type matching, the avatar SSRF scheme allowlist) prevents a future refactor from silently introducing a culture-sensitive comparison, and it clears the family so it stays enforced.

## Behavior-preserving, carefully

- The role mappers keep the **instance** `.Equals(x, Ordinal)` overload (receiver unchanged), preserving the SAML mapper's documented NRE on a null configured role (pinned by `NullConfiguredAdminRole_Throws`), a switch to static `string.Equals` would have made it null-safe and turned that misconfiguration fail-open. Only the `==`/`!=` operator sites became the null-safe static form.
- The genuinely case-**insensitive** sites (localhost SSRF host check, scheme/port normalization, media-type prefix, OID `r`/`redirect` path segment) are deliberately left `OrdinalIgnoreCase`, converting them would be a real regression (e.g. a `LocalHost` host bypassing the SSRF guard).

## Verification

- Build clean (`--warnaserror` with the two rules at error → proves all sites addressed, none missed), 240/240 tests including the null-quirk facts and the FsCheck monotonicity properties.
- Two adversarial reviews (bypass + correctness) PASS: per-site equivalence proven, instance-vs-static null semantics verified per site, negations logically identical, and the intentional case-insensitive sites confirmed untouched.

Refs #104
Closes #129

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -39,12 +39,12 @@ dotnet_diagnostic.SA1649.severity = none
 # is worked through deliberately in follow-ups (several are behavior-relevant
 # on the login path — string-comparison semantics must not change in a config
 # PR).
-dotnet_diagnostic.CA1309.severity = suggestion
+dotnet_diagnostic.CA1309.severity = error
 dotnet_diagnostic.CA1311.severity = suggestion
 dotnet_diagnostic.CA1848.severity = suggestion
 dotnet_diagnostic.RCS1194.severity = suggestion
 dotnet_diagnostic.MA0002.severity = suggestion
-dotnet_diagnostic.MA0006.severity = suggestion
+dotnet_diagnostic.MA0006.severity = error
 dotnet_diagnostic.MA0011.severity = suggestion
 dotnet_diagnostic.MA0015.severity = suggestion
 dotnet_diagnostic.MA0016.severity = suggestion
--- a/SSO-Auth/Api/AvatarUrlValidator.cs
+++ b/SSO-Auth/Api/AvatarUrlValidator.cs
@@ -26,7 +26,7 @@ internal static class AvatarUrlValidator
             return false;
         }
 
-        if (parsed.Scheme != Uri.UriSchemeHttp && parsed.Scheme != Uri.UriSchemeHttps)
+        if (!string.Equals(parsed.Scheme, Uri.UriSchemeHttp, StringComparison.Ordinal) && !string.Equals(parsed.Scheme, Uri.UriSchemeHttps, StringComparison.Ordinal))
         {
             return false;
         }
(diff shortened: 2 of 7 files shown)


iderex (commented):
pre-merge review (iderex).

First analyzer-backlog family from #104 (CA1309/MA0006), chosen because it is security-adjacent and behavior-preserving. Two adversarial reviews (bypass + correctness) PASS with per-site equivalence proofs: every converted site was already ordinal (C# ==/!= and instance .Equals(string)), so no match outcome changes; the two role mappers correctly keep the INSTANCE .Equals(x, Ordinal) overload preserving the SAML null-configured-role NRE quirk (pinned by NullConfiguredAdminRole_Throws), while the ==/!= sites became the null-safe static form; and the genuinely case-insensitive sites (localhost SSRF host check, scheme/port normalization, OID path segment) were correctly left OrdinalIgnoreCase — converting those would have been a real SSRF/routing regression. Build clean under --warnaserror with both rules at error (proves all sites addressed, none missed), 240/240 tests incl. the null-quirk facts and FsCheck monotonicity properties. CodeRabbit reviewed with no findings. All CI checks green, SonarCloud gate green.

Issue #134, opened 2026-07-12
Validate the OIDC id_token: OidcClient 7.1.0 performs no validation in our configuration


iderex:
With `IdentityTokenValidator` unset and `Policy.RequireIdentityTokenSignature` left at its false default, OidcClient 7.1.0 falls back to `NoValidationIdentityTokenValidator`, which only base64-decodes the JWT payload: no signature check, no iss/aud/azp match, no exp/iat, and at_hash is skipped because the reported algorithm is "none" (source-verified at tag imoc-7.1.0: https://github.com/DuendeSoftware/foss/blob/imoc-7.1.0/identity-model-oidc-client/src/IdentityModel.OidcClient/ResponseProcessor.cs and NoValidationIdentityTokenValidator.cs). OIDC Core 1.0 sect. 3.1.3.7 makes the iss/aud/exp checks unconditional MUSTs (https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation); RFC 9700 (2025, https://datatracker.ietf.org/doc/rfc9700/) reinforces this. Worst case today: with `DisableHttps` set, the login is forgeable on-path. Duende deprecated their validator package and points at the extensibility point (https://www.nuget.org/packages/IdentityModel.OidcClient.IdentityTokenValidator). Fix: set `RequireIdentityTokenSignature = true` and register a custom `IIdentityTokenValidator` on Microsoft.IdentityModel.JsonWebTokens, signature against `ProviderInformation.KeySet`, iss == discovery issuer, aud/azp == client_id, exp/iat with the existing 5-min skew, asymmetric-only algorithm allowlist (RS/PS/ES), matching the SAML allowlist posture.

Acceptance criteria:
- [ ] Login fails closed on an id_token with missing/invalid signature, wrong iss, wrong aud/azp, or expired exp, no fallback path accepts an unvalidated token.
- [ ] `HS256` and `alg=none` tokens are rejected regardless of key material.
- [ ] at_hash validation is active when the response carries an access token.
- [ ] Unit tests cover each rejection path plus a happy path against a static JWKS.

Refs #125, this validator plus the RFC 9207 iss parameter check together provide the only issuer anchoring at response time; neither exists today.

PR #184, opened 2026-07-12
Reflect the id_token validation in the README and fix a config key typo

iderex:
Daily docs review, 2026-07-12 (second pass, after #182 merged).

## What

- **README:** the hardened-login-path feature bullet still described the OpenID side as state-binding only. Since #182 the id_token is fully validated, signature against the discovery JWKS under an asymmetric-only allowlist (RS/PS/ES), issuer, audience, expiry, so the bullet now says so and links the provider-side requirements section that #182 added to providers.md.
- **providers.md:** the Kanidm example used the non-existent config key `AdminsRoles`; the actual property is `AdminRoles` (both provider config classes). Copying the example verbatim would silently map no admin role.

The wiki side (Security Model section for the id_token checks, matching Troubleshooting entry) is updated directly on the wiki, which takes no PRs.

Prettier-clean; docs-only, no code changes.

Refs #134


iderex (commented):
pre-merge review (docs-only change, self-authored PR, so a comment review per repo policy).

Verified:
- README feature-bullet claim checked against the implementation (OidcIdTokenValidator: JWKS-based signature validation under the RS/PS/ES-only allowlist, issuer vs discovery, audience/azp rules, exp/nbf with skew) - the wording matches what the code enforces, no overclaim.
- The providers.md anchor link (#openid-connect-id_token-requirements) resolves.
- AdminsRoles -> AdminRoles confirmed against PluginConfiguration.cs (the property is AdminRoles in both provider config classes); the old key would have been silently ignored.
- Docs-only diff, no code touched, so the adversarial-review gate does not apply.
- Checks: build, prettier, CodeQL, dependency-review, CodeRabbit (no findings), SonarCloud quality gate passed (0 new issues).

Verdict: good to merge.

Issue #161, opened 2026-07-12
Config export/import UI and endpoint


iderex:
User-facing config export/import (the feature; the security redaction/preservation is tracked separately under the P2 config item). Provide an admin UI action and endpoint to export the plugin configuration (redacted) and import it into another instance.

Acceptance criteria:
- [ ] Export produces a redacted, importable document (reuses the P2 redaction work - do not duplicate it).
- [ ] Import validates and merges without wiping server-managed fields.
- [ ] Round-trip documented in providers.md.

Depends on the P2 config export/import redaction issue.

iderex:
Export/import redaction requirement folded in from #157 (that issue delivers the server-managed-field preservation + F-10 locked reads that apply to today's code; there is no export endpoint yet to redact). When this export/import lands, it MUST:
- redact secrets (OidSecret, SAML signing/cert material) from the exported document,
- omit CanonicalLinks (already withheld from JSON responses via [JsonIgnore] as of #157; keep it out of exports too),
- on import, preserve server-managed fields (reuse SSOPlugin.PreserveServerManagedFields) so an import missing them cannot wipe live links (fail closed).

Refs #157.

iderex:
Delivered in #581, merged into the 4.2 feature-dev branch (auto-close does not fire off a non-default branch, so closing manually).

PR #188, opened 2026-07-12
Preserve server-managed canonical links across config saves

### LINKED ISSUE #157: Config export/import redaction and server-managed-field preservation
Three related config-safety gaps from the V2 audit (F-9/F-10) plus export hygiene. (1) A settings save has no server-side merge, so a stale admin form can wipe server-managed fields (canonical links, generated keys) - port V2 PreserveServerManagedFields. (2) Config export must redact secrets (OidSecret, signing keys) and never leak CanonicalLinks off the server. (3) The admin list reads are not lock-guarded and can tear against a concurrent write (F-10); route them through ReadConfiguration.

Acceptance criteria:
- [ ] Saving a provider merges server-managed fields instead of overwriting them; a stale form cannot blank generated keys or canonical links (test).
- [ ] Config export redacts OidSecret and signing keys and omits CanonicalLinks; a redacted export round-trips without wiping the omitted fields.
- [ ] All admin list reads go through the config lock (ReadConfiguration); no unguarded reads remain on the admin path.
- [ ] Fail closed: an import missing server-managed fields preserves the live values rather than clearing them.

Refs #51 (config write race, closed).

iderex:
# What & why

The admin settings page saves the whole plugin configuration through Jellyfin core's `updatePluginConfiguration` from a snapshot taken at page load. A canonical link created by a login since then is absent from the posted config, so the save wiped it (F-9). Account links are also server-managed state that should neither leave the server nor be settable by a config POST. This treats `CanonicalLinks` as a server-managed field and closes the unguarded admin reads (F-10). Closes #157.

- **`[JsonIgnore]` on both `CanonicalLinks` properties**, still persisted in the config XML, withheld from every JSON response. Stops the account-link map leaking off the server, removes the tear from serializing it while a login writes it, and blocks setting links via a config PUT (deserialization ignores them).
- **Re-inject the live links on every save**, a `SSOPlugin.UpdateConfiguration` override copies each provider's live `CanonicalLinks` into the incoming config before persisting (catches the admin-UI core path); `OidAdd`/`SamlAdd` do the same explicitly (the API path, where the override sees the live object and skips). Fail closed: a save missing the links preserves the live values.
- **F-10 locked reads**, the links-by-user endpoints build their result under `ReadConfiguration` and materialize with `ToList` (fixing a deferred-LINQ tear that enumerated the live map during JSON serialization outside any lock); `OID/Get` and `SAML/Get` return a snapshot taken under the lock.

Export/import redaction (part 2 of the issue) is tracked on #161, there is no export endpoint in this repo yet to redact.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: a save that omits server-managed links preserves the live values; links cannot be injected or overwritten via a config POST.
- [x] No secrets logged; `CanonicalLinks` is now withheld from JSON responses (`OidSecret`/`SamlCertificate` remain in admin-only responses as before, the admin edit UI needs them; true export redaction is #161).
- [x] A security review was performed: adversarial gate with 4 lenses (results in the sign-off).
- [x] Log inputs from the IdP are sanitized inline at the log call (unchanged).

## Quality checklist

- [x] No duplicated logic; preservation is one shared `PreserveServerManagedFields`, the links-by-user projection is one shared `BuildLinksByUser`.

## Verification

- [x] `dotnet test` is green (294/294).
- [x] Prettier: no `.js`/`.html`/`.md`/`.css` changed.

## Notes

- The `UpdateConfiguration` override takes `ConfigMutationLock`, which is reentrant, so calls from within `MutateConfiguration` do not deadlock; the `ReferenceEquals` guard skips the self-copy on that path.
- Behavior note (pre-existing): renaming a provider in the settings UI does not carry its links to the new name, the links key on the old provider name. Out of scope here.
- Owed before any release: real-server E2E, edit a provider in the admin UI while a login has just created a link, save, and confirm the link survives; confirm `OID/Get` no longer returns `CanonicalLinks`.

### DIFF
--- a/SSO-Auth.Tests/ConfigPreservationTests.cs
+++ b/SSO-Auth.Tests/ConfigPreservationTests.cs
@@ -1,0 +1,95 @@
+using System;
+using System.Xml.Serialization;
+using Jellyfin.Plugin.SSO_Auth;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Tests for the server-managed-field handling (#157): canonical links are preserved across a save
+/// built from a stale client snapshot (<see cref="SSOPlugin.PreserveServerManagedFields"/>), are
+/// withheld from JSON responses ([JsonIgnore]), and still round-trip through the config XML.
+/// </summary>
+public class ConfigPreservationTests
+{
+    private static readonly Guid User = Guid.Parse("11111111-1111-1111-1111-111111111111");
+
+    [Fact]
+    public void Preserve_ReinjectsLiveLinks_IntoStaleIncomingProvider()
+    {
+        var live = new PluginConfiguration();
+        live.OidConfigs["idp"] = new OidConfig { CanonicalLinks = new SerializableDictionary<string, Guid> { ["sub-1"] = User } };
+        live.SamlConfigs["saml"] = new SamlConfig { CanonicalLinks = new SerializableDictionary<string, Guid> { ["nameid-1"] = User } };
+
+        // The incoming (stale) config has the providers but empty link maps — as a client snapshot
+        // taken before the login that created the links would.
+        var incoming = new PluginConfiguration();
+        incoming.OidConfigs["idp"] = new OidConfig();
+        incoming.SamlConfigs["saml"] = new SamlConfig();
+
+        SSOPlugin.PreserveServerManagedFields(incoming, live);
+
+        Assert.Equal(User, incoming.OidConfigs["idp"].CanonicalLinks["sub-1"]);
+        Assert.Equal(User, incoming.SamlConfigs["saml"].CanonicalLinks["nameid-1"]);
+    }
+
+    [Fact]
+    public void Preserve_NewProviderNotInLive_KeepsItsOwnEmptyMap()
+    {
+        var live = new PluginConfiguration();
+        var incoming = new PluginConfiguration();
+        incoming.OidConfigs["fresh"] = new OidConfig();
+
+        SSOPlugin.PreserveServerManagedFields(incoming, live);
+
+        Assert.Empty(incoming.OidConfigs["fresh"].CanonicalLinks);
+    }
+
+    [Fact]
+    public void Preserve_ProviderDeletedInIncoming_IsNotReAdded()
+    {
+        // Deleting a provider must survive the save: a provider present only in live is not resurrected.
+        var live = new PluginConfiguration();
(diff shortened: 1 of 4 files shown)


iderex (commented):
## iderex sign-off — Preserve server-managed canonical links (Closes #157)

**Adversarial-review gate: ran (config persistence + secret handling).** Four lenses, refute-by-default — **all PASS**.

- **bypass-lens — PASS:** empirically verified [JsonIgnore] blocks BOTH directions — a config PUT can no longer inject/overwrite links (dropped on deserialize) and no JSON response leaks the account-link map. Net security-positive: it closes a prior getPluginConfiguration leak of the whole sub→userId map and a config-PUT link-injection vector; privilege derivation and login-time link keying untouched.
- **correctness-lens — PASS:** reentrant lock (no self-deadlock), the ReferenceEquals self-copy skip is sound, the explicit OidAdd/SamlAdd preserve is necessary and correct, PreserveServerManagedFields is one-directional live→incoming keyed by provider name, BuildLinksByUser/SnapshotConfigs materialize under the lock.
- **integration-lens — PASS:** reflection-confirmed the UpdateConfiguration override signature against the pinned Jellyfin.Common 10.11.11 and against V2's production pattern; System.Text.Json is the serializer Jellyfin actually uses (Newtonsoft only parses claim JSON, never the config); XML on-disk round-trip intact; config.js reads CanonicalLinks nowhere; release surface untouched.
- **availability-lens — PASS:** preservation is by-key against the live config; no config shape wipes links for an unchanged provider; [JsonIgnore] affects only JSON, XML persistence retained; the only new cost is a bounded links-by-user scan under the lock.

Review follow-ups applied in this branch: OID/GetNames + SAML/GetNames now materialize under ReadConfiguration (the same F-10 tear the other reads were fixed for); OidAdd/SamlAdd guard a null [FromBody] (no new NRE); the SnapshotConfigs comment corrected re: NewPath. The pre-existing admin-only exposure of OidSecret/SamlCertificate in JSON (V2 makes them write-only) is filed as a follow-up, #189.

**CI/gates:** all green except SonarCloud — verified via api/qualitygates/project_status: **new_coverage (28.9 < 80) is the SOLE failing condition, all others OK — accepted per the documented free-plan policy** (the uncovered lines are controller wiring; there is no controller test harness — the pure PreserveServerManagedFields and the JsonIgnore behavior ARE unit-tested). Merged via ruleset admin bypass. Build + 294 tests green. CodeRabbit: no comments.

**Owed before any release carrying this:** real-server E2E per docs/E2E-CHECKLIST.md — (a) create a link via login, edit+save an unrelated provider in the dashboard, confirm the link survives; (b) upgrade over an existing config.xml and confirm links still authenticate; (c) confirm OID/Get no longer returns CanonicalLinks.

Verdict: merge.

PR #201, opened 2026-07-12
Record the client IP for SSO logins in Jellyfin's activity log

### LINKED ISSUE #177: SSO logins omit the client IP from Jellyfin's activity log
Confirmed against the current source: `CompleteLogin` builds the `AuthenticationRequest` passed to `_sessionManager.AuthenticateDirect` (`SSOController.cs` around line 1084) setting `UserId`, `Username`, `App`, `AppVersion`, `DeviceId`, `DeviceName`, but never `RemoteEndPoint`. Password logins go through Jellyfin's own auth path, which does set it, so the "Successfully authenticated" activity-log entry shows the client IP for local logins but not for SSO ones. Both the OIDC and SAML paths funnel through this same method, so both are affected.

Reported independently by a user on the upstream project this repo restarted from: 9p4/jellyfin-plugin-sso#333 ("SSO login Activity log does not record client IP (IP visible to Jellyfin during request)"), filed against SSO-Auth 4.0.0.3 / Jellyfin 10.11.4, reverse-proxied through nginx + Authentik, i.e. the common deployment shape for this plugin. This is a forensics/incident-response gap, not an active bypass: an operator investigating a suspicious SSO login cannot see where it came from.

Acceptance criteria:
- [ ] `AuthenticationRequest.RemoteEndPoint` populated for both the OIDC and SAML completion paths, using the same client-IP resolution Jellyfin's own login path relies on (respecting any `X-Forwarded-For`/`ForwardedHeaders` handling already configured at the Jellyfin server level, the plugin should not invent its own proxy-trust logic here).
- [ ] Verified end-to-end: the Jellyfin activity log shows a real client IP for an SSO login, not the reverse-proxy hop or an empty value.
- [ ] No change to the separate `[SSO Audit]` structured log (#131) scope, this is Jellyfin's built-in activity log, a distinct surface.

iderex:
# What & why

SSO logins built the `AuthenticationRequest` without `RemoteEndPoint`, so Jellyfin's "authenticated" activity-log entry showed **no source IP** for SSO logins, while password logins (which go through Jellyfin's own auth path) do, a forensics / incident-response gap (reported upstream at 9p4/jellyfin-plugin-sso#333). Closes #177. Both the OIDC and SAML paths funnel through the shared completion method, so one line fixes both.

The client IP is populated via Jellyfin's own `HttpContext.GetNormalizedRemoteIP()`, the same helper its built-in login uses, so it reads the already-resolved connection address (the plugin adds no proxy-trust logic; the server's *Known proxies* setting governs it) and normalizes it identically (IPv4-mapped IPv6 collapsed to IPv4). An SSO entry's source address therefore matches a password entry's for the same client.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: purely additive log/session-metadata field; no auth/authz decision, session minting, or validation changed.
- [x] Not attacker-spoofable beyond the operator's existing proxy-trust boundary: the value is the Jellyfin-resolved connection address, the same trust basis as a password login; the plugin parses no `X-Forwarded-For` of its own.
- [x] A security review was performed (integration + correctness + bypass lenses, see Notes).

## Quality checklist

- [x] Minimal: one field, set via Jellyfin's own normalization helper (no reimplementation, no divergence).
- [x] Consistent with the plugin's existing IP handling (SsoRateLimiter/AvatarUrlValidator normalize IPv4-mapped IPv6 the same way).

## Verification

- [x] `dotnet build --no-restore --warnaserror` green.
- [x] `dotnet test` green (359/359).
- [x] E2E checklist (local): the activity-log source IP must be confirmed on a real server, the acceptance criterion, not unit-testable.

## Notes

**Adversarial-review gate (integration + correctness + bypass; proportionate, a one-field, no-decision-change addition on the login path).**
- bypass **PASS**, correctness **PASS**: purely additive, null-safe, non-spoofable, no injection/auth/endpoint impact.
- integration **NEEDS_IMPROVEMENT → fixed**: the first cut used raw `RemoteIpAddress.ToString()`, which (unlike Jellyfin's login) left an IPv4-mapped IPv6 client as `::ffff:1.2.3.4` instead of `1.2.3.4`, a log-consistency deviation from the stated "as password logins do" parity. Fixed by switching to Jellyfin's own `GetNormalizedRemoteIP()` (a public `MediaBrowser.Common.Extensions` helper, confirmed in the referenced package), which is exactly what the built-in login uses, so parity is now byte-for-byte and the null case also matches (loopback).

The activity-log write itself lives in the Jellyfin host (not the plugin's referenced assemblies), so the end-to-end "IP appears in the log" check is on the E2E checklist as the one non-unit-testable acceptance item.

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -19,6 +19,7 @@ using Jellyfin.Database.Implementations.Enums;
 using Jellyfin.Plugin.SSO_Auth.Config;
 using Jellyfin.Plugin.SSO_Auth.Helpers;
 using MediaBrowser.Common.Api;
+using MediaBrowser.Common.Extensions;
 using MediaBrowser.Controller.Authentication;
 using MediaBrowser.Controller.Configuration;
 using MediaBrowser.Controller.Library;
@@ -1308,6 +1309,14 @@ public class SSOController : ControllerBase
         authRequest.AppVersion = parameters.AuthResponse.AppVersion;
         authRequest.DeviceId = parameters.AuthResponse.DeviceID;
         authRequest.DeviceName = parameters.AuthResponse.DeviceName;
+
+        // Record the client IP so the SSO login shows a source address in Jellyfin's activity log,
+        // exactly as password logins do (#177): use Jellyfin's own GetNormalizedRemoteIP, the very
+        // helper its built-in login path uses. It reads the already-resolved connection address (so
+        // the plugin adds no proxy-trust logic of its own — the server's "Known proxies" setting
+        // governs it) and normalizes it the same way (IPv4-mapped IPv6 collapsed to IPv4), so the
+        // SSO entry's source address matches a password entry's for the same client byte-for-byte.
+        authRequest.RemoteEndPoint = HttpContext.GetNormalizedRemoteIP().ToString();
         _logger.LogInformation("Auth request created...");
         if (!string.IsNullOrEmpty(parameters.DefaultProvider))
         {


iderex (commented):
pre-merge review (iderex) — #177 client IP in the activity log.

## Adversarial-review gate (integration + correctness + bypass; proportionate — a one-field, no-decision-change addition on the login path).

- bypass PASS — pure additive log/session-metadata field; connection-sourced (non-spoofable beyond the operator's existing proxy-trust boundary); no auth/authz/endpoint/injection impact; null-safe.
- correctness PASS — purely additive, both OIDC and SAML paths covered exactly once via the shared Authenticate method, non-throwing, no NRE.
- integration NEEDS_IMPROVEMENT → fixed — the first cut used raw RemoteIpAddress.ToString(), leaving an IPv4-mapped IPv6 client as ::ffff:1.2.3.4 vs the plain 1.2.3.4 Jellyfin's own login records; fixed by using Jellyfin's own public HttpContext.GetNormalizedRemoteIP() (MediaBrowser.Common.Extensions), so the value matches a password login byte-for-byte (IPv4-mapped collapse + loopback-on-null). Re-review not needed — the fix substitutes the host's own normalization for the raw call, strictly aligning with the established parity goal, and does not touch the auth path.

No secret leak, no proxy-trust logic added (the server's Known proxies setting governs the resolved address, same as a password login).

## CI / quality gates

- build (--warnaserror), CodeQL (security-extended), dependency-review, prettier, package (JPRM), submit-nuget: all green.
- Tests: 359/359 (no new unit test — the value derives from the live HttpContext and the activity-log write is host-side; genuinely E2E-only, matching the acceptance criteria).
- SonarCloud: gate GREEN (all conditions OK).
- E2E checklist (local): the source IP appearing in the activity-log entry must be confirmed on a real server (the one non-unit-testable acceptance item), ideally compared against a password-login entry from the same client.

Overall verdict: APPROVED for merge. A minimal, forensics-improving log field sourced through Jellyfin's own normalization helper for byte-for-byte parity with the built-in login; fail-closed, no auth impact, all gates green.

PR #224, opened 2026-07-13
Harden the SSO login-completion path (fail-closed robustness)

### LINKED ISSUE #215: Gate Live TV permission grants behind the EnableAuthorization master switch
## Summary

`EnableLiveTvAccess` and `EnableLiveTvManagement` are applied on every SSO login unconditionally, outside the `if (parameters.EnableAuthorization)` block that gates every other role-derived grant.

## Impact, grant escape past the per-provider authorization switch

`EnableAuthorization` is the per-provider "let SSO manage permissions" switch. Admin and folder grants are correctly gated behind `if (parameters.EnableAuthorization)` (`SSO-Auth/Api/SSOController.cs:1338-1346`), but the two Live TV permissions are set at `:1350-1351`, outside that block. So a provider role grants Live TV access/management on every SSO login even when the admin turned the master switch off to keep SSO out of permission management, and it can silently toggle an adopted account's existing Live TV state on each login.

## Affected code

- `SSO-Auth/Api/SSOController.cs:1350-1351`

## Fix sketch

Move the two `SetPermission(EnableLiveTv*)` calls inside the `if (parameters.EnableAuthorization)` block so all role-derived grants respect the one switch. V2's shape (a single gated authorization-application unit) is the target.

## V2 status

Closed in V2: all role grants, including Live TV, go through one `ApplyAuthorization` unit gated on `enableAuthorization` (its comment explicitly cites Live TV). Ready-made port.

## Notes

Surfaced by the 2026-07-13 standing security audit. Pre-production, tracked publicly per the current triage policy.

### LINKED ISSUE #216: Fail closed on malformed role-claim JSON instead of 500-ing the callback
## Summary

For a nested role-claim path, `OidcRoleExtractor.ExtractRoles` parses the claim with `JsonConvert.DeserializeObject` / `ToObject<List<string>>()` and lets a parse failure escape as an unhandled exception, so `OidPost` returns HTTP 500 instead of failing closed to a clean reject.

## Impact, robustness / fail-closed

If the matched claim value is not valid JSON or the terminal array has non-coercible elements, Newtonsoft throws; `OidPost` calls `Build` with no `try/catch` (`SSO-Auth/Api/SSOController.cs:192`), so it escapes as HTTP 500. The claim originates from the signature-validated `id_token` / TLS `userinfo` (not directly attacker-controlled), so the impact is robustness/availability, every affected login 500s, while still fail-closed (no session is minted). The header comment on the extractor overstates how gracefully it degrades.

## Affected code

- `SSO-Auth/Api/OidcRoleExtractor.cs:40` (deserialize) and `:68` (`ToObject`)

## Fix sketch

Wrap the parse in `try/catch (JsonException)` returning an empty list, and filter the terminal array to string tokens only so a non-string element cannot throw. Fails closed to zero roles. Pure helper change → unit-testable (malformed JSON, non-string elements → empty list).

## V2 status

Closed in V2: `RoleClaimParser.ExtractRoles` wraps parsing in `try/catch (JsonException)` returning an empty list and selects only string elements. Ready-made port.

## Notes

Surfaced by the 2026-07-13 standing security audit. Pre-production, tracked publicly per the current triage policy.

### LINKED ISSUE #220: Bound the avatar body download with an overall read deadline
## Summary

The avatar fetch uses `HttpCompletionOption.ResponseHeadersRead`, so the 10s `HttpClient.Timeout` stops once headers arrive; the subsequent body read has no deadline, letting a malicious avatar endpoint hold the login request open by trickling body bytes.

## Impact, resource hold (mostly self-inflicted)

At `SSO-Auth/Api/SSOController.cs:1419` the fetch reads response headers only, so the client `Timeout` no longer bounds the transfer once headers arrive; the following `ReadCappedAsync` body read uses `ReadAsync` with no `CancellationToken`. A malicious avatar endpoint (reachable when `AvatarUrlFormat` embeds a user-controlled claim) can send headers immediately then trickle body bytes, holding the awaiting login request open far beyond 10s (bounded only by the 10MB cap). Mostly self-inflicted (it stalls the attacker's own login thread), hence low, but it is an unbounded server-side resource hold driven by an untrusted endpoint.

## Affected code

- `SSO-Auth/Api/SSOController.cs:1419` (and the streamed `ReadCappedAsync` body read)

## Fix sketch

Drive the whole fetch with a `CancellationTokenSource(TimeSpan.FromSeconds(10))` and pass its token into `GetAsync` and every `ReadAsync` so the header wait and the body stream share one deadline. (V2 instead drops `ResponseHeadersRead` so the client `Timeout` bounds the whole body, either approach works; revisit the streaming-cap comment at `:1417-1418`.)

## V2 status

Closed in V2: it fetches with the default `ResponseContentRead`, so the 10s `Timeout` bounds the entire body and the slow-trickle stall does not arise. Port the transport choice or add a linked CTS.

## Notes

Surfaced by the 2026-07-13 standing security audit. Pre-production, tracked publicly per the current triage policy.

iderex:
# What & why

Three fail-closed robustness fixes in the SSO login-completion path, from the 2026-07-13 standing audit. Each is a ready V2 port (or, for #220, the stronger of the two V2-suggested options).

- **#215**, the Live TV grants (`EnableLiveTvAccess` / `EnableLiveTvManagement`) were applied on every login **outside** the `EnableAuthorization` master switch, so a provider re-granted or silently toggled Live TV even when the admin turned SSO permission management off. Moved inside the same block as the admin/folder grants.
- **#216**, `OidcRoleExtractor.ExtractRoles` threw on a malformed multi-segment role-claim value (or a terminal array of non-strings), surfacing as an HTTP 500 on the public callback. Now wraps the parse/walk in `try/catch (JsonException)` returning no roles and takes only the string elements of the terminal array, fail-closed to zero roles.
- **#220**, the avatar fetch used `ResponseHeadersRead`, so the 10s client `Timeout` stopped once headers arrived and a slow-trickle body could hold the login open. A single `CancellationTokenSource(10s)` is threaded into `GetAsync` and every body read/write, bounding the whole fetch while keeping the streaming 10 MB cap. (Chosen over V2's drop-`ResponseHeadersRead` because it keeps the streaming size cap and adds the deadline.)

Closes #215
Closes #216
Closes #220

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved, all three changes are strictly monotonic toward fail-closed: #215 only withdraws a grant, #216 only reduces roles (empty on malformed → allow-list denies), #220 only aborts a best-effort fetch.
- [x] No secrets logged; secrets redacted on export, N/A, no secret/logging change (avatar failure logs the exception, no IdP value).
- [x] A security review was performed, 5-lens adversarial gate (bypass / correctness / availability / integration / oidc-domain) all PASS; the correctness lens verified the Newtonsoft exception types empirically.
- [x] Log inputs from the IdP are sanitized inline at the log call, N/A, unchanged.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose, testable unit, `OidcRoleExtractor` stays pure; the avatar deadline reuses the existing `ReadCappedAsync`.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green, `--no-incremental --warnaserror` 0/0.
- [x] `dotnet test` is green, 434 tests pass (updated + added `OidcRoleExtractor` tests: malformed → empty, non-string terminal filtered).
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change, N/A, only `.cs` changed.

## Notes

Live TV gating (#215) and the avatar transport deadline (#220) are not unit-testable without a controller host, so E2E-checklist items were added. Behavior note (fail-closed, intended): a provider running `EnableAuthorization=false` with `EnableLiveTv=true` stops re-granting Live TV on each login (it no longer touches the permission, existing state is left as the admin set it); and a login whose role claim is malformed now lands with zero SSO roles instead of 500-ing.

### DIFF
--- a/SSO-Auth.Tests/OidcRoleExtractorTests.cs
+++ b/SSO-Auth.Tests/OidcRoleExtractorTests.cs
@@ -71,11 +71,21 @@ public class OidcRoleExtractorTests
     }
 
     [Fact]
-    public void MultiSegmentPath_MalformedClaimValue_Throws()
+    public void MultiSegmentPath_MalformedClaimValue_ReturnsEmpty()
     {
-        // Characterizes the upstream behavior: a multi-segment path over a malformed (non-JSON-object)
-        // claim value throws (surfaces as a 500 in the controller = fail-closed). Pinned so it stays deliberate.
-        Assert.ThrowsAny<Newtonsoft.Json.JsonException>(
-            () => OidcRoleExtractor.ExtractRoles(new[] { "realm_access", "roles" }, "not-json"));
+        // #216: a multi-segment path over a malformed (non-JSON) claim value fails CLOSED to no roles
+        // instead of throwing an unhandled 500 on the public callback.
+        Assert.Empty(OidcRoleExtractor.ExtractRoles(new[] { "realm_access", "roles" }, "not-json"));
+    }
+
+    [Fact]
+    public void TerminalArrayWithNonStringElements_KeepsOnlyStrings()
+    {
+        // #216: a terminal array mixing strings with objects/numbers must not throw; only the string
+        // elements are taken as roles.
+        var roles = OidcRoleExtractor.ExtractRoles(
+            new[] { "realm_access", "roles" },
+            "{\"roles\":[\"users\",1,{\"nested\":true},\"admins\"]}");
+        Assert.Equal(new List<string> { "users", "admins" }, roles);
     }
 }
--- a/SSO-Auth/Api/OidcRoleExtractor.cs
+++ b/SSO-Auth/Api/OidcRoleExtractor.cs
@@ -24,8 +24,10 @@ internal static class OidcRoleExtractor
     /// <param name="claimValue">The matched claim's value (a raw role, or a JSON object for a nested path).</param>
     /// <returns>
     /// The extracted roles: the raw claim value as a single role for a one-segment path; otherwise the
-    /// string array reached by walking the JSON path. An empty list when the path does not resolve to a
-    /// JSON array (missing segment, non-object node, or non-array terminal).
+    /// string elements of the array reached by walking the JSON path (non-string elements are ignored).
+    /// Returns an empty list when the path does not resolve to a JSON array (missing segment, non-object
+    /// node, or non-array terminal) and when the claim value is malformed JSON — a multi-segment parse
+    /// fails closed to no roles rather than throwing (#216).
     /// </returns>
     internal static List<string> ExtractRoles(string[] roleClaimSegments, string claimValue)
     {
--- a/SSO-Auth/Api/OidcRoleExtractor.cs
+++ b/SSO-Auth/Api/OidcRoleExtractor.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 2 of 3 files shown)


iderex:
@coderabbitai Addressed, updated the `OidcRoleExtractor.ExtractRoles` `<returns>` doc to state that a multi-segment parse fails closed to an empty list on malformed JSON (rather than throwing) and that non-string terminal elements are ignored (commit 12cc9d7). Thanks for the catch.

iderex (commented):
## pre-merge review — Harden the SSO login-completion path (#215, #216, #220)

### Adversarial review gate (login path)
5 refute-by-default lenses, all **PASS**:
- **bypass** — all three fixes are strictly monotonic toward fail-closed: #215 only withdraws a grant (Live TV now behind `EnableAuthorization`), #216 only reduces roles (empty on malformed → allow-list denies), #220 only aborts the best-effort avatar fetch without touching the SSRF `ConnectCallback` or the size cap; no fail-open path.
- **correctness** — verified empirically (Newtonsoft): non-object roots throw a `JsonException` subtype the new `catch` covers; the string filter does not change a legitimate `["a","b"]` result; brace scoping and full CA2016 token threading (GetAsync / ReadAsStreamAsync / ReadAsync / WriteAsync) with a disposed CTS are correct.
- **availability** — no lockout: #215 leaves existing Live TV state as the admin set it when the switch is off; #216 fails closed only on a malformed claim (well-formed claims unaffected); #220's 10s deadline abandons only a stalling avatar, login proceeds.
- **integration** — no caller depends on the old throw; empty-roles composes correctly with the role→privilege mapper; SAML shares the gated `Authenticate`; the CTS composes with the SocketsHttpHandler SSRF guard.
- **oidc-domain** — failing closed to an empty role set is the correct OIDC behavior; taking only string elements matches how Keycloak/authentik shape roles without dropping legitimate ones.

### CI / gates
All green: `.NET` build + test (434 tests), `package/build` (JPRM), CodeQL (actions / csharp / javascript-typescript), `dependency-review`, `prettier`. **SonarCloud** quality gate red on **`new_coverage` only** (38.5% — verified via `qualitygates/project_status?pullRequest=224`: reliability/security/maintainability ratings, duplicated lines, and hotspots-reviewed all pass). Accepted per the free-plan coverage-goal policy — the new `SSOController` lines (#215 Live TV gate, #220 avatar deadline) are not unit-testable without a controller host and are covered by E2E-checklist items; `OidcRoleExtractor` (#216) is fully unit-tested. Merged via admin bypass.

### Reviewer engagement
- **CodeRabbit**: engaged — one finding (the `ExtractRoles` `<returns>` doc was stale relative to the new fail-closed behavior). **Fixed** in commit `12cc9d7` (doc now describes malformed→empty and non-string filtering); replied on the PR. Re-review passed.
- **Copilot**: quota-limit notice, no substantive review.

### Verification note
Live TV gating and the avatar transport deadline are not unit-testable; E2E-checklist items were added (`#215` Live TV respects `EnableAuthorization`, `#216` malformed claim → zero roles not 500, `#220` slow-trickle avatar abandoned within ~10s).

### Verdict
Three fail-closed robustness fixes with no functional or security regression, gate all-PASS, CodeRabbit finding resolved, Sonar red on coverage only. Approved to merge (Comment-event review — self-approval not permitted; admin merge per solo auto-mode).

iderex:
@coderabbitai Closing the loop on the outside-diff note "Clarify the terminal-array fail-closed comment": it landed just before this PR merged and was addressed in the follow-up #228 (merged 441357f6). The block comment now states that the terminal array is filtered to its string elements, a mixed array keeps its strings, and only an array with no string elements (or a malformed value) yields none, matching the implementation. Disposition: FIXED.

PR #266, opened 2026-07-13
Reset SSO config toggles when switching providers

### LINKED ISSUE #211: Config UI: sso-toggle checkboxes not reset to false when switching providers
Pre-existing config-UI defect surfaced during the #125 review (integration lens), affecting every `.sso-toggle` checkbox equally (`DisableHttps`, `DoNotValidateEndpoints`, `DoNotValidateIssuerName`, `DoNotValidateResponseIssuer`).

`loadProvider` in SSO-Auth/Config/config.js sets a toggle's `.checked` only when `provider[id]` is truthy and never resets it to `false`. Switching from a provider that has a toggle checked to one without a stored value leaves the checkbox visually checked, and a subsequent save could silently persist `true` for the second provider.

Fix: in `loadProvider`, set `.checked = Boolean(provider[id])` unconditionally (or reset all toggles before applying the loaded provider). Add/extend a config.js behavior test if feasible. Not a login-path issue; low priority.

iderex:
## What & why

`loadProvider` in `config.js` set a `.sso-toggle` checkbox's `.checked` only when the loaded provider's stored value was truthy, and never reset it. Switching (in the admin UI) from a provider that has a security toggle checked, `DisableHttps`, `DoNotValidateEndpoints`, `DoNotValidateIssuerName`, `DoNotValidateResponseIssuer`, to one without a stored value left the box visually checked; the save path reads `.checked` verbatim, so a subsequent save **silently persisted `true`** for the second provider. That is a silent security downgrade (validation disabled / HTTPS not enforced for a provider the admin never configured that way).

Fix: set `.checked = Boolean(provider[id])` unconditionally in the `check_fields` loop, so every provider switch/add resets each toggle to the loaded provider's actual value (`false` when unset, the secure default). Load and save are now symmetric over the same `.sso-toggle` set.

## Type of change
- [x] Bug fix (config UI)
- [x] Security hardening (prevents silent persistence of a validation-disabling flag)
- [ ] New feature
- [ ] Breaking change

## Security
- [x] Config-persistence surface → adversarial-review gate run (see Notes).
- [x] Fail-closed default: an unset toggle now resolves to `false` (HTTPS enforced, validation on).

## Quality
- [x] Minimal: one line + an explaining comment; the sibling non-toggle loops are deliberately untouched.

## Verification
- [x] Prettier clean (`config.js`).
- [ ] Browser JS → CI cannot exercise the config UI; the gate is the primary verification. Manual check on the E2E checklist: switch providers, confirm toggles reset and a save persists the reset value.

## Notes

**Adversarial-review gate, 3 lenses (config persistence of security toggles), all PASS:** correctness (`Boolean(provider[id])` resets stale→false and preserves stored→true; every `check_fields` element is guaranteed non-null; no truthy mis-mapping), bypass (covers all four security toggles; closes both the visual and the persisted downgrade; introduces no opposite risk of silently disabling a legitimately-true toggle; `false` is the secure default), integration (fits the load/save lifecycle, reflects through the `emby-checkbox` component without a `change` event, no regression to the sibling `json_fields`/`folder_list_fields`/`role_map_fields` loops).

One out-of-scope residual the gate noted: the *add-new-provider* flow doesn't call `loadProvider`, so a brand-new provider inherits on-screen checkbox state, pre-existing, admin-visible, unchanged by this diff, not the reported bug.

Closes #211

### DIFF
--- a/SSO-Auth/Config/config.js
+++ b/SSO-Auth/Config/config.js
@@ -254,7 +254,12 @@ const ssoConfigurationPage = {
         });
 
         form_elements.check_fields.forEach((id) => {
-          if (provider[id]) page.querySelector("#" + id).checked = provider[id];
+          // Always set the checkbox from the loaded provider so switching providers
+          // resets stale toggles. Setting it only when truthy left a previous
+          // provider's checked box in place, which a later save could silently
+          // persist as true — a security downgrade for toggles like
+          // DoNotValidateEndpoints / DisableHttps.
+          page.querySelector("#" + id).checked = Boolean(provider[id]);
         });
 
         form_elements.role_map_fields.forEach((id) => {


iderex:
Change walkthrough (one item in this PR):
- `SSO-Auth/Config/config.js`, `loadProvider` → `check_fields` loop: replaced `if (provider[id]) …checked = provider[id]` with an unconditional `…checked = Boolean(provider[id])`. Rationale is inline in the code comment: the conditional form never reset a stale-checked security toggle on provider switch, so a later save (which reads `.checked`) could persist a validation-disabling flag the admin never set. The sibling non-toggle loops (`json_fields`/`folder_list_fields`/`role_map_fields`) are intentionally left as-is, they are not security toggles and out of scope for #211.

iderex (commented):
## Pre-merge review — reset SSO config toggles on provider switch (#211)

Fixes a config-UI defect where `loadProvider` never reset a `.sso-toggle` checkbox, so switching providers could leave a stale security toggle checked and a later save silently persisted a validation-disabling flag (`DoNotValidateEndpoints`, `DisableHttps`, …) — a silent security downgrade. The fix sets `.checked = Boolean(provider[id])` unconditionally.

### Adversarial review gate (config persistence of security toggles) — 3 lenses, all PASS
- **correctness**: `Boolean(provider[id])` resets stale→false and preserves stored→true; every `check_fields` element is guaranteed non-null; no truthy mis-mapping vs the old `.checked = provider[id]` setter.
- **bypass**: covers all four security toggles; closes both the visual and the persisted downgrade across every load path; `false` is the secure default; no opposite risk of silently disabling a legitimately-true toggle. (Noted out-of-scope residual: the add-new-provider flow doesn't call `loadProvider` — pre-existing, admin-visible, unchanged.)
- **integration**: fits the load/save lifecycle, reflects through the `emby-checkbox` component without a `change` event, no regression to the sibling non-toggle loops.

Browser JS → CI cannot exercise the config UI, so the gate is the primary verification; an E2E-checklist item covers the manual provider-switch check.

### Reviewer engagement
- **CodeRabbit**: Review completed — "No actionable comments were generated." No inline and no outside-diff-range findings. **Copilot**: quota-limited.

### CI / gates
All green — build/test, CodeQL, prettier (`config.js`), Trojan Source, **SonarCloud PASSED**.

### Verdict
Closes a real silent-security-downgrade path in the admin UI with a one-line, fail-closed fix. Merging.

Issue #213, opened 2026-07-13
Make Unregister actually revoke SSO access, persist, and null-guard


iderex:
## Summary

The admin `Unregister` control (`POST /sso/Unregister/{username}`) is documented as removing a user from SSO, but it neither revokes SSO login nor persists any change, and it 500s on an unknown username.

## Impact, operator control defeated + robustness

`Unregister` (`SSO-Auth/Api/SSOController.cs:872`) has three defects:

1. **No persistence.** It sets `user.AuthenticationProviderId = provider` in memory but never calls `UpdateUserAsync` (contrast `Authenticate` at `:1353`), so the assignment is discarded when the request ends. The admin sees HTTP 200 and believes SSO was removed; nothing changed.
2. **Does not actually revoke SSO.** SSO login resolves purely through the per-provider `CanonicalLinks` map, never `AuthenticationProviderId`. Even a persisted `AuthenticationProviderId` change would leave the canonical link intact, so SSO login still works.
3. **No null guard.** `_userManager.GetUserByName(username)` returns `null` for an unknown username; `:875` dereferences it and throws `NullReferenceException` → HTTP 500. `RequiresElevation`-gated, so admin-only, a robustness gap, not attacker-reachable.

## Affected code

- `SSO-Auth/Api/SSOController.cs:872-878`

## Fix sketch

Make the handler `async`; null-guard to `NotFound`; remove the user's canonical link(s) for the provider under `MutateLinks` so login resolution actually stops (revocation must be based on the `CanonicalLinks` map, not `AuthenticationProviderId`); then `UpdateUserAsync`. Add an endpoint test (unknown user → 404; known user → link removed and SSO login no longer resolves).

## V2 status

Partially closed in V2: `Unregister` is `async`, guards `null` to `NotFound`, and persists via `SetAuthenticationProvider` + `UpdateUserAsync`. The deeper "login still resolves via `CanonicalLinks`" defect is unaddressed in V2 too and needs original work here.

## Notes

Surfaced by the 2026-07-13 standing security audit (bundles the endpoint's persistence, revocation, and null-guard defects into one fix). Pre-production, tracked publicly per the current triage policy.

Issue #204, opened 2026-07-12
Consolidate shared SamlConfig/OidConfig members into ProviderConfigBase


iderex:
`SamlConfig` and `OidConfig` in `SSO-Auth/Config/PluginConfiguration.cs` duplicate a large block of members (Enabled, EnableAuthorization, role/folder settings, DefaultProvider, SchemeOverride, PortOverride, NewPath, CanonicalLinks, and the CanonicalLinks backing field + accessor). SonarCloud reports this pre-existing ~70-line duplicated block. #139 extracted the new `BaseUrlOverride` into a shared `ProviderConfigBase`; the remaining shared members could follow.

Scope/caution: moving XML-serialized members to a base class changes their element order in newly written config XML (deserialization is by name, so existing configs still load, verify with the SerializableDictionary round-trip tests and a real-server load). This is config-persistence surface, so it goes through the adversarial-review gate. Deferred from #139 to keep that PR minimal. Raised by CodeRabbit on PR #203.

iderex:
## Implementation spec, consolidate the shared members into ProviderConfigBase

`ProviderConfigBase` already exists (created in #139 for `BaseUrlOverride`). This migrates the rest of the duplicated block into it.

### Members to move (identical in SamlConfig and OidConfig, grep-verified)
`Enabled`, `EnableAuthorization`, `AllowExistingAccountLink`, `EnableAllFolders`, `EnabledFolders`, `AdminRoles`, `Roles`, `EnableFolderRoles`, `EnableLiveTvRoles`, `EnableLiveTv`, `EnableLiveTvManagement`, `LiveTvRoles`, `LiveTvManagementRoles`, `FolderRoleMapping` (with its `[XmlArray]`/`[XmlArrayItem]` attributes), `DefaultProvider`, `SchemeOverride`, `PortOverride`, `NewPath`, and `CanonicalLinks` (with its `_canonicalLinks` backing field, the null-safe accessor, and the `[XmlElement("CanonicalLinks")]` + `[JsonIgnore]` attributes). ~19 members, the bulk of each class.

[...]

Issue #155, opened 2026-07-12
Key OIDC identity on the stable sub claim, not the mutable preferred_username


iderex:
Account linking currently keys on the display name / preferred_username, which an IdP admin can change or reassign, so a renamed or recycled username can collide with an existing canonical link. Bind the OIDC identity to the immutable `sub` claim (per provider issuer) instead, and migrate existing name-keyed CanonicalLinks to sub-keyed entries with a one-time transparent migration. Deferred from the original identity-binding work (item 3), which closed the takeover itself but left the key mutable.

Acceptance criteria:
- [ ] New OIDC links are keyed on issuer + sub, never on preferred_username.
- [ ] A username change at the IdP does not detach or re-link an account (test).
- [ ] Legacy name-keyed links migrate transparently on first login after upgrade; migration is idempotent and logged.
- [ ] Fail closed: if sub is absent, the login is rejected rather than falling back to name keying.

Refs #133 (adoption atomicity), #134 (id_token validation), #125 (iss).

PR #269, opened 2026-07-13
Make the CanonicalLinks getter self-healing

### LINKED ISSUE #239: CanonicalLinks getter returns a throwaway map; make the lazy init self-healing
`SamlConfig.CanonicalLinks` and `OidConfig.CanonicalLinks` return a fresh empty map on every read while the backing field is null (SSO-Auth/Config/PluginConfiguration.cs:228, :387):

```csharp
get
{
    if (_canonicalLinks == null)
    {
        return new SerializableDictionary<string, Guid>();
    }

    return _canonicalLinks;
}
```

The natural call, `config.CanonicalLinks[key] = id`, writes into a discarded object and is silently lost. Account links are login-critical state, and nothing errors when they vanish. The code is only correct today through compensating plumbing in SSOController.cs:

- `MutateLinks` (SSOController.cs:990) re-assigns the map through `SetLinks` (:1011) "so a lazily-created (previously-null) one is persisted".
- `Unregister`'s `CanonicalLinkRevoker.RemoveUser(config.CanonicalLinks, userId)` (:975, :980) is correct only because removing from an empty throwaway happens to be a no-op.

Any future caller that skips the `MutateLinks` detour hits the trap.

Make the lazy init self-healing instead:

```csharp
get => _canonicalLinks ??= new SerializableDictionary<string, Guid>();
```

Every direct access becomes safe, and the compensators go with it: `SetLinks` entirely, the re-assign in `MutateLinks`, and both 12-line property bodies, about 35 lines net removed. All `CanonicalLinks` access already runs under the config lock (`ReadConfiguration`/`MutateConfiguration`), so the lazy assignment cannot race, and the XML output is unchanged (a lazily created empty map serializes the same as the getter's throwaway did).

Touches config persistence on the login path, so the change runs the adversarial-review gate. Related to #204, after the consolidation the fixed property lives once in `ProviderConfigBase`, but this is a semantics fix, not the member dedup.

iderex:
## What & why

`SamlConfig.CanonicalLinks` and `OidConfig.CanonicalLinks` returned a **fresh empty map on every read** while the backing field was null, so the natural `config.CanonicalLinks[key] = id` wrote into a **discarded throwaway** and was silently lost. Canonical links are login-critical account-link state (SSO login resolves a user through these per-provider maps), and nothing errored when a write vanished. The code was only correct through compensating plumbing that re-assigned the map after every mutation (`SetLinks`), so any future caller that skipped the detour hit the trap.

Fix: make the getter **self-healing**, `get => _canonicalLinks ??= new SerializableDictionary<string, Guid>();`, so first access creates *and stores* the map and every direct access persists into it. The compensators are removed with it:

- `SetLinks` helper deleted entirely (0 remaining callers).
- `MutateLinks` simplified to `mutate(GetLinks(...))` (no re-assign).
- the `SetLinks` call in `LinkCanonicalIfAbsent` dropped.

Net **~27 lines of fragile plumbing removed** from the controller; the semantics fix lives in the property. (Distinct from the #204 property-dedup; both `SamlConfig`/`OidConfig` copies are fixed identically here.)

## Type of change
- [x] Bug fix (config persistence, login path)
- [x] Quality (dead-plumbing removal)
- [ ] Breaking

## Security
- [x] Config-persistence on the login path → adversarial-review gate run (see Notes).
- [x] No semantic change to link resolution/adoption/revocation; self-healing only ever allocates *empty* maps, never plants a link.

## Quality
- [x] Removes duplicated compensating plumbing; the invariant now lives once in the property.

## Verification
- [x] `dotnet build` (Debug, `--warnaserror`) clean; `dotnet test` 476 passing (+3 self-healing tests: direct-write-persists for both configs, same-instance-across-reads).

## Notes

**Adversarial-review gate, 4 lenses (config persistence, login path), all PASS:**
- **correctness**: the getter genuinely persists direct writes; every `SetLinks` removal is safe because all mutations already land on the stored map via `GetLinks`; no caller relied on re-assigning a different instance (full-map replacement uses the setter directly, retained); XML byte-identical.
- **availability/mass-lockout**: `ConfigMutationLock` is an exclusive Monitor, `ReadConfiguration` and `MutateConfiguration` take the *same* lock, so there are no concurrent readers and the `??=` cannot race; no first-login link-loss window; lazy-init-on-read only ever stores a benign empty map.
- **bypass**: resolution/adoption (#155)/revocation (#213)/`[JsonIgnore]` withholding (#157) semantics untouched; only empty maps allocated.
- **integration**: empty stored map serializes identically to the prior throwaway; the setter and its remaining callers (re-injection at ~507/820, `PreserveServerManagedFields`) intact.

One documented forward-looking caveat (non-blocking): the getter now has a write-on-read side effect, so the config-lock invariant is load-bearing for the getter itself, any *future* off-lock read of `.CanonicalLinks` would be a lock-free read-modify-write. No such caller exists today; noted in the getter comment.

Closes #239

### DIFF
--- a/SSO-Auth.Tests/CanonicalLinksSelfHealingTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinksSelfHealingTests.cs
@@ -1,0 +1,48 @@
+using System;
+using Jellyfin.Plugin.SSO_Auth;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Tests for the self-healing <c>CanonicalLinks</c> getter (#239): a direct
+/// <c>CanonicalLinks[key] = id</c> must persist into the stored map, not a throwaway. Before the fix
+/// the getter returned a fresh empty map while the backing field was null, so the write was silently
+/// lost — login-critical account-link state.
+/// </summary>
+public class CanonicalLinksSelfHealingTests
+{
+    private static readonly Guid User = Guid.Parse("22222222-2222-2222-2222-222222222222");
+
+    [Fact]
+    public void OidConfig_DirectWrite_Persists()
+    {
+        var config = new OidConfig();
+
+        config.CanonicalLinks["sub-1"] = User;
+
+        Assert.True(config.CanonicalLinks.ContainsKey("sub-1"));
+        Assert.Equal(User, config.CanonicalLinks["sub-1"]);
+    }
+
+    [Fact]
+    public void SamlConfig_DirectWrite_Persists()
+    {
+        var config = new SamlConfig();
+
+        config.CanonicalLinks["nameid-1"] = User;
+
+        Assert.True(config.CanonicalLinks.ContainsKey("nameid-1"));
+        Assert.Equal(User, config.CanonicalLinks["nameid-1"]);
+    }
+
+    [Fact]
+    public void CanonicalLinks_ReturnsTheSameInstanceAcrossReads()
+    {
+        var config = new OidConfig();
+
+        // Two reads return the one stored map, not a fresh throwaway each time.
+        Assert.Same(config.CanonicalLinks, config.CanonicalLinks);
+    }
+}
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -1028,14 +1028,13 @@ public class SSOController : ControllerBase
         });
     }
 
-    // Applies a mutation to a provider's canonical-links map on the LIVE configuration, reassigning the
-    // map so a lazily-created (previously-null) one is persisted. Runs inside MutateConfiguration, so the
-    // whole read-modify-write is serialized and concurrent first-logins cannot lose each other's links.
+    // Applies a mutation to a provider's canonical-links map on the LIVE configuration. The map is
+    // self-healing (CanonicalLinks lazily creates and stores it), so mutating the returned map persists
+    // directly. Runs inside MutateConfiguration, so the read-modify-write is serialized and concurrent
+    // first-logins cannot lose each other's links.
     private static void MutateLinks(PluginConfiguration configuration, string mode, string provider, Action<SerializableDictionary<string, Guid>> mutate)
     {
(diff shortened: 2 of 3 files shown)


iderex (commented):
## Pre-merge review — self-healing CanonicalLinks getter (#239)

Fixes a config-persistence footgun: `CanonicalLinks` returned a throwaway empty map when the backing field was null, so a direct `CanonicalLinks[key] = id` was silently lost (login-critical account-link state). The getter is now self-healing (`??= new`), and the compensating plumbing (`SetLinks`, the `MutateLinks` re-assign) is removed — ~27 lines of fragile controller code gone, the invariant now living once in the property.

### Adversarial review gate — 4 lenses (config persistence, login path), all PASS
- **correctness**: getter genuinely persists direct writes; every `SetLinks` removal safe (all mutations already land on the stored map via `GetLinks`; full-map replacement uses the retained setter directly); XML byte-identical.
- **availability/mass-lockout**: `ConfigMutationLock` is an exclusive Monitor — `ReadConfiguration` and `MutateConfiguration` take the same lock, so no concurrent readers and the `??=` cannot race; no first-login link-loss window; lazy-init-on-read only stores benign empty maps.
- **bypass**: resolution / #155 adoption / #213 revocation / #157 `[JsonIgnore]` withholding all untouched; only empty maps ever allocated.
- **integration**: empty stored map serializes identically to the prior throwaway; setter + `PreserveServerManagedFields` + re-injection intact.

Documented non-blocking caveat: the getter now has a write-on-read side effect, so the config-lock invariant is load-bearing for it — no off-lock caller exists today (noted in the getter comment).

### Reviewer engagement
- **CodeRabbit**: Review completed — "No actionable comments were generated." No inline and no outside-diff-range findings (both surfaces enumerated). **Copilot**: quota-limited.

### CI / gates
All green except **SonarCloud**, red on **`new_coverage` only** (66.7%): 0 Sonar issues, all ratings A. The getter is unit-tested (+3 self-healing tests); the residual is controller glue. Coverage-only red accepted per free-plan policy → admin merge.

### Verdict
Removes a silent-data-loss trap on login-critical state and shrinks the controller. Merging: Sonar is red on coverage only (the new lines are controller glue), which the free-plan policy accepts, so I take the admin merge.

Issue #252, opened 2026-07-13
Anonymous endpoints index the live provider dictionaries unsynchronized against provider Add/Del


iderex:
## Problem

The anonymous login endpoints index the live provider dictionaries with no synchronization against the admin write paths that mutate them in place.

- Writers: `OidAdd`/`SamlAdd`/`Del` run inside `MutateConfiguration` and add/remove keys on the **live** `Dictionary` objects (`SSOController.cs:456-489`, `:769-794`), under `ConfigMutationLock`.
- Readers: eight login-path sites index the same plain `Dictionary` lock-free, `SSOController.cs:152`, `:261`, `:574`, `:645`, `:714`, `:833`, `:1345`, `:1394` (`SSOPlugin.Instance.Configuration.OidConfigs[provider]` / `SamlConfigs[provider]`).

Concurrent read-during-write on `Dictionary<TKey,TValue>` is undefined behavior in .NET: a login racing a provider add/remove can throw, misread, or, during a bucket resize, spin indefinitely on a corrupted chain, wedging a request thread at 100 % CPU until restart. Exposure is admin-time only (provider add/remove is rare), but the failure mode lands on the login path.

## Fix

Minimal-first options:

- Route the eight reads through `SSOPlugin.ReadConfiguration` (the pattern `RateLimitCheck` already uses, `:1600`). An uncontended lock is nanoseconds; the lock is only held long during first-login/admin persists, which is exactly when a consistent read matters. Accepted trade: logins briefly queue behind a first-login config persist (XML serialize + file write under the lock), bounded ms-to-tens-of-ms and rare; do not trade correctness to avoid it.
- Alternative with zero read-side cost: make `Add`/`Del` copy-and-swap the dictionary reference instead of mutating in place, so the live maps are immutable snapshots. Slightly more write-side code; reads stay lock-free and become safe.

Either way, the invariant to establish: no code path mutates a dictionary instance that anonymous endpoints read without the lock.

iderex:
The lock-routing option is exactly the #241 refactor (dedup the eight lookup sites into ReadConfiguration-backed helpers). Implementing #241 resolves this; a single PR should close both. If the copy-and-swap alternative is preferred instead, decide it there.

PR #270, opened 2026-07-13
Replace KeyNotFoundException control flow in the provider lookup with a null check

### LINKED ISSUE #241: Deduplicate the provider-config lookup and route it through the config lock
The controller repeats the same provider-config lookup, nine lines at a time, eight times (SSOController.cs:149, :258, :571, :642, :711, :830, :1342, :1391):

```csharp
OidConfig config;
try
{
    config = SSOPlugin.Instance.Configuration.OidConfigs[provider];
}
catch (KeyNotFoundException)
{
    return BadRequest(NoMatchingProviderMessage);
}
```

Two problems beyond the duplication:

- It reads the live configuration dictionaries without `ReadConfiguration`, contradicting the lock discipline the same file documents everywhere else ("Materialize under the lock (#157/F-10)", SSOController.cs:514, :526) and the `ReadConfiguration` contract itself (SSOPlugin.cs:280-287: a read must not tear against a concurrent write of a non-thread-safe configuration collection). A concurrent provider add/delete mutates the same `Dictionary` under the lock, and `Dictionary<TKey,TValue>` is not safe for a read concurrent with a write. Eight copies of the unlocked pattern are also exactly what the next endpoint author will paste.
- `KeyNotFoundException` as control flow for the routine unknown-provider case.

Replace with two small lookups:

```csharp
private static OidConfig FindOidConfig(string provider) =>
    SSOPlugin.Instance.ReadConfiguration(c => c.OidConfigs.TryGetValue(provider, out var config) ? config : null);
```

(and the `SamlConfig` twin). Call sites become a null check that preserves each endpoint's current error shape, `return BadRequest(NoMatchingProviderMessage)` on the callback/utility endpoints, `throw new ArgumentException(ProviderDoesNotExistMessage)` on the challenge endpoints.

About 55 lines removed and the dictionary access becomes uniformly locked. This is a mechanical dedup that fixes a documented-invariant violation; it moves no business logic and changes no seam of the #160 decomposition.

Touches the login path, so it runs the adversarial-review gate despite its size.

iderex:
## What & why

#252 routed the provider-config lookup through the config lock via `GetOidConfig`/`GetSamlConfig`, but those used the **dictionary indexer**, which throws `KeyNotFoundException` for an unknown provider, so **eight call sites** each wrapped the read in `try/catch(KeyNotFoundException)` to map it to an error. This removes that exception-as-control-flow.

- Helpers switched to `TryGetValue` returning `null` (`FindOidConfig`/`FindSamlConfig`); the read still runs inside `ReadConfiguration`.
- The 8 call sites branch on a **null check**, each preserving its **exact prior error shape**:
  - `return BadRequest(NoMatchingProviderMessage)` on the callback/utility endpoints (5 config sites + 1 existence-only `OidLink`),
  - `?? throw new ArgumentException(ProviderDoesNotExistMessage)` on the 2 challenge endpoints.
- The two `catch(KeyNotFoundException)` blocks around `MutateLinks`/`GetLinks` are **deliberately left**, `GetLinks` still indexes the provider dictionary, so that throw stays (out of #241's scope).

Net **−41 lines**, and `KeyNotFoundException` is no longer used as control flow for the routine unknown-provider case.

## Type of change
- [x] Quality / refactor (login path)
- [ ] Bug fix / feature / breaking

## Security
- [x] Login-path change → adversarial-review gate run (see Notes).
- [x] No semantic change: unknown-provider rejection preserved on every site; read stays under the config lock.

## Quality
- [x] Removes duplicated exception-handling; `−41` lines.

## Verification
- [x] `dotnet build` (Debug, `--warnaserror`) clean; `dotnet test` 476 passing (unchanged, behavior-preserving).

## Notes

**Adversarial-review gate, 4 lenses (login path), all PASS:**
- **correctness**: every rewritten site byte-equivalent for unknown and known providers; `?? throw` sites leave `config` non-null; the existence-only site references no missing var; the two out-of-scope catches correctly retained.
- **bypass**: `TryGetValue`-returns-null is semantically identical to indexer-throws under the same lock; every null branch hard-returns/throws (no fall-through); no attacker-controllable differential (same comparer; `null` provider still fails in both).
- **integration**: challenge-site `ArgumentException` propagates identically in the same scope; the untouched catches still catch the intact `GetLinks` indexer throw; no dead const/import (build 0/0 under warnaserror).
- **availability**: six 400s + two `ArgumentException`s preserved; no new throw or lockout path; happy path unchanged.

Closes #241

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -166,12 +166,8 @@ public class SSOController : ControllerBase
             return throttled;
         }
 
-        OidConfig config;
-        try
-        {
-            config = GetOidConfig(provider);
-        }
-        catch (KeyNotFoundException)
+        var config = FindOidConfig(provider);
+        if (config is null)
         {
             return BadRequest(NoMatchingProviderMessage);
         }
@@ -275,15 +271,7 @@ public class SSOController : ControllerBase
         }
 
         Invalidate();
-        OidConfig config;
-        try
-        {
-            config = GetOidConfig(provider);
-        }
-        catch (KeyNotFoundException)
-        {
-            throw new ArgumentException(ProviderDoesNotExistMessage);
-        }
+        var config = FindOidConfig(provider) ?? throw new ArgumentException(ProviderDoesNotExistMessage);
 
         if (config.Enabled)
         {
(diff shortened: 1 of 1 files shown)


iderex (commented):
## Pre-merge review — provider-lookup dedup (#241)

Removes the `KeyNotFoundException`-as-control-flow left after #252: the lookup helpers switch to `TryGetValue` (`FindOidConfig`/`FindSamlConfig`) and the 8 call sites branch on a null check, each preserving its exact prior error shape. Net −41 lines; the read stays under `ReadConfiguration`.

### Adversarial review gate — 4 lenses (login path), all PASS
- **correctness**: every rewritten site byte-equivalent for unknown/known providers; `?? throw` sites leave `config` non-null; existence-only site references no missing var; the two out-of-scope `MutateLinks`/`GetLinks` catches correctly retained (GetLinks still indexes).
- **bypass**: `TryGetValue`-null identical to indexer-throw under the same lock; every null branch hard-returns/throws (no fall-through); no attacker-controllable differential.
- **integration**: challenge-site `ArgumentException` propagates identically in the same scope; no dead const/import (build 0/0 under warnaserror).
- **availability**: six 400s + two `ArgumentException`s preserved; no new throw/lockout path; happy path unchanged.

### Reviewer engagement
- **CodeRabbit**: Review completed — "No actionable comments were generated." No inline and no outside-diff-range findings (both surfaces enumerated). **Copilot**: quota-limited.
- **SonarCloud MA0015** (INFO ×2, L274/L735 — "use the ArgumentException overload with the parameter name"): **DECLINED in this PR**. These are pre-existing — the refactor only relocated the identical `throw new ArgumentException(ProviderDoesNotExistMessage)`; adding `nameof(provider)` would append "(Parameter 'provider')" to the exception message, changing observable behavior, which contradicts this PR's behavior-preservation guarantee (and the gate verified byte-equivalence). Can be addressed as a separate quality change if wanted.

### CI / gates
All green except **SonarCloud**, red on **`new_coverage` only** (0% — the rewritten controller lines are not unit-testable; the MA0015 items are INFO and do not move the maintainability rating off A). Coverage-only red accepted per free-plan policy → admin merge.

### Verdict
Behavior-preserving dedup that removes exception-as-control-flow and −41 lines from the login path. Merging: Sonar is red on coverage only (the new lines are controller glue), which the free-plan policy accepts, so I take the admin merge.

Issue #242, opened 2026-07-13
Extract GetRequestBase's host-fallback URL derivation into CanonicalBaseUrl and test it


iderex:
`GetRequestBase` (SSOController.cs:1791-1837) derives the base URL every flow is built on: the OpenID `redirect_uri` and the SAML base feeding `ExpectedAcsUrls` (:1686), the exact strings the callback comparisons validate against. Its host-fallback half is pure decision logic:

- default-port elision (80 on http, 443 on https drop the explicit port),
- the scheme-override allowlist that silently discards anything but literal `"http"`/`"https"` (:1825-1828),
- the trailing-slash trim interacting with `Request.PathBase`.

None of it is covered by a test, and none of it can be: the method reads `Request` directly, so it only executes inside a live controller. The configured-override half of the same decision already lives in the tested `CanonicalBaseUrl.TryNormalize`, the seam is half-cut.

Move the body into that helper as a pure function, e.g.

```csharp
public static string Resolve(string baseUrlOverride, string scheme, string host, int? port, string pathBase, string schemeOverride, int? portOverride)
```

and leave a one-expression wrapper in the controller feeding the `Request` values. Line count stays roughly neutral; the branch matrix (port elision x scheme override x path base) becomes plain unit tests in CanonicalBaseUrlTests.

This completes an existing helper rather than pre-empting the #160 decomposition; the gap is a specific untested load-bearing path, not generic coverage work (#192).

PR #273, opened 2026-07-13
Elide the default port against the effective scheme, not the request scheme

### LINKED ISSUE #272: Default-port elision uses the request scheme instead of the effective (override) scheme
Pre-existing bug surfaced by the #242 review (now testable via `CanonicalBaseUrl.Resolve`).

`Resolve` (and the original inline `GetRequestBase`) elides the default port using the **request** scheme, but applies the `schemeOverride` allowlist **after** that check and uses `schemeOverride ?? scheme` for the final URL. So when the effective scheme differs from the request scheme, the elision decides against the wrong scheme:

- request scheme `http` + `SchemeOverride = "https"` + port `443` → produces `https://host:443` instead of the canonical `https://host`.
- (symmetrically, request `https` + `SchemeOverride = "http"` + port `80` → `http://host:80`.)

This is exactly the reverse-proxy scenario `SchemeOverride` exists for (TLS terminated at the proxy; Jellyfin sees `http`), so a non-canonical `redirect_uri` / SAML base with an explicit `:443` can mismatch what the IdP is configured with and break the callback comparison.

Confirmed pre-existing (byte-identical to the pre-#242 inline logic; the #242 extraction preserved it deliberately and the gate verified byte-equivalence). Filed separately because the fix is a **behavior change** to the login-path base URL and warrants its own focused review + tests, not a silent change inside a behavior-preserving refactor.

Fix: normalize the effective scheme (`schemeOverride`-allowlist → `effective = schemeOverride ?? scheme`) **first**, then elide port 80 for `http` / 443 for `https` against that effective scheme.

Acceptance criteria:
- [ ] `Resolve(null, "http", host, 443, "", "https", null)` → `https://host` (and the `https`+`http`+80 symmetric case).
- [ ] Existing elision/scheme tests still pass; a test pins the effective-scheme case.
- [ ] Runs the adversarial-review gate (login path; redirect_uri + SAML ACS).

iderex:
## What & why

The request-derived base URL (used only when there is no `BaseUrlOverride`) elided the default port by checking the **request** scheme, but applied the `SchemeOverride` allowlist afterwards and built the URL from `schemeOverride ?? scheme`. So when the effective scheme differed from the request scheme, the elision decided against the wrong one:

- a TLS-terminating proxy, request `http` + `SchemeOverride = https` + port `443`, produced `https://host:443` instead of the canonical `https://host`;
- symmetrically, request `https` + `SchemeOverride = http` + port `80` → `http://host:80`.

That base URL feeds the OpenID `redirect_uri` and the SAML base (the exact strings the callback comparisons validate against), so a non-canonical URL with an explicit port can mismatch what a provider is configured with.

Fix: resolve the effective scheme first (`schemeOverride` allowlist → `effective = schemeOverride ?? scheme`), then elide port 80 for `http` / 443 for `https` **against that effective scheme**.

## Type of change

- [x] Bug fix (login path)
- [x] Behavior change, see the migration note below

## Behavior change & migration

Only one path changes: the request-fallback (no `BaseUrlOverride`) **with** a `SchemeOverride` that differs from the request scheme **and** a resolved port that is the default of the override scheme. Everything else, no override, override equal to the request scheme, a non-default port, or a set `BaseUrlOverride`, is byte-identical to before.

For that one slice the emitted `redirect_uri`/SAML base changes from `https://host:443` (or `http://host:80`) to the canonical no-port form. A provider that normalizes default ports is unaffected; a deployment that pinned the explicit `:443`/`:80` form to work around the old behavior must switch its registered `redirect_uri` to the no-port form. `providers.md` documents this.

This still needs a short migration note in the release notes before it ships, because that narrow slice is a hard callback mismatch with no graceful fallback, affected setups should be told to re-register the canonical URL rather than discover a silent lockout on upgrade.

## Verification

- [x] `dotnet build` (Debug, `--warnaserror`) clean; `dotnet test` 492 passing (+2).
- [x] Prettier clean (`providers.md`).

## Notes

Adversarial-review gate, 4 lenses (login path). correctness / bypass / tests **PASS**: the common path is unchanged, the differing-scheme cases now elide against the scheme that actually appears in the URL, the authoritative fail-closed override branch and the admin-only scheme allowlist are untouched (no scheme/host/origin change, only the port), and the new tests recompute correctly (port-8096 isolation + both elision directions). availability **NEEDS_IMPROVEMENT**, not a code defect: the code is correct and net-positive, but the narrow slice is a hard-lockout risk that must be surfaced. Addressed by the `providers.md` migration note and the release-note callout flagged above.

Follow-up to #242 (which made this logic testable). Closes #272

### DIFF
--- a/SSO-Auth.Tests/CanonicalBaseUrlTests.cs
+++ b/SSO-Auth.Tests/CanonicalBaseUrlTests.cs
@@ -122,15 +122,28 @@ public class CanonicalBaseUrlTests
     }
 
     [Theory]
-    [InlineData("http", "http://host.example.com")] // literal http honored
-    [InlineData("https", "https://host.example.com")] // literal https honored
-    [InlineData("ftp", "https://host.example.com")] // anything else falls back to the request scheme
-    [InlineData("HTTP", "https://host.example.com")] // case-sensitive: not honored
-    [InlineData("", "https://host.example.com")]
+    [InlineData("http", "http://host.example.com:8096")] // literal http honored
+    [InlineData("https", "https://host.example.com:8096")] // literal https honored
+    [InlineData("ftp", "https://host.example.com:8096")] // anything else falls back to the request scheme
+    [InlineData("HTTP", "https://host.example.com:8096")] // case-sensitive: not honored
+    [InlineData("", "https://host.example.com:8096")]
     public void Resolve_SchemeOverride_OnlyLiteralHttpOrHttpsHonored(string schemeOverride, string expected)
     {
-        // Request scheme is https, port 443 (elided) so only the scheme decision shows.
-        var result = CanonicalBaseUrl.Resolve(null, "https", "host.example.com", 443, string.Empty, schemeOverride, null);
+        // Non-default port so port elision does not mask the scheme decision.
+        var result = CanonicalBaseUrl.Resolve(null, "https", "host.example.com", 8096, string.Empty, schemeOverride, null);
+
+        Assert.Equal(expected, result);
+    }
+
+    [Theory]
+    [InlineData("http", "https", 443, "https://host.example.com")] // TLS-terminating proxy: http request, https override, 443 elided
+    [InlineData("https", "http", 80, "http://host.example.com")] // symmetric: https request, http override, 80 elided
+    public void Resolve_ElidesDefaultPortForTheEffectiveScheme(string requestScheme, string schemeOverride, int port, string expected)
+    {
+        // The default-port elision must decide against the scheme that actually appears in the URL
+        // (schemeOverride), not the request scheme — otherwise the canonical URL keeps an explicit
+        // :443 / :80 (#272).
+        var result = CanonicalBaseUrl.Resolve(null, requestScheme, "host.example.com", port, string.Empty, schemeOverride, null);
 
         Assert.Equal(expected, result);
     }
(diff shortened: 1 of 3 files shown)


iderex (commented):
## Pre-merge review — elide the default port against the effective scheme (#272)

Fixes a pre-existing bug in the request-fallback base URL: the default-port elision decided against the request scheme instead of the effective (override) scheme, so a TLS-terminating proxy produced `https://host:443` instead of the canonical `https://host`.

### Adversarial review gate — 4 lenses (login path)
- **correctness / bypass / tests PASS**: the common path is unchanged, the differing-scheme cases now elide against the scheme that actually appears in the URL, the authoritative fail-closed override branch and the admin-only scheme allowlist are untouched (no scheme/host/origin change — only the port), and the new tests recompute correctly (port-8096 isolation + both elision directions).
- **availability NEEDS_IMPROVEMENT — not a code defect**: the narrow scheme-override-with-default-port slice is a hard callback mismatch for anyone who registered the explicit `:443`/`:80` form. Addressed by the `providers.md` migration note (register the canonical no-port form) and the release-note callout, which still needs to be decided before this ships.

### Review engagement
- CodeRabbit: Review completed — "No actionable comments were generated" (no inline, no outside-diff-range findings; both surfaces checked). Copilot: quota-limited.

### CI
All green — build/test (492), CodeQL, both zizmor checks, dependency-review, prettier, Trojan Source, and SonarCloud passed (the +2 tests cover the fix, so coverage is not red here).

Merging.

Issue #196, opened 2026-07-12
SonarCloud high-severity backlog sweep (S2699/S125/S6964/S3776)


iderex:
Post-merge standing sweep after PR #193 (#128). CodeQL, Dependabot, and secret-scanning are all at **0 open alerts**; PR #193 introduced **no** new alerts and passed the new-code quality gate (reliability/security/maintainability ratings all A; only new_coverage was under the 80% goal, accepted per the free-plan policy). The following pre-existing high-severity SonarCloud issues (none from #193) are captured here so none is left open without a documented disposition:

- **BLOCKER S2699**, `SSO-Auth.Tests/ConfigPreservationTests.cs` `Preserve_NullConfigMaps_DoNotThrow` (#157). The test's contract *is* "does not throw", but Sonar wants an explicit assertion. **Fix:** add an explicit assertion that the maps are handled (e.g. assert the call completes and the incoming maps stay null/empty) so the intent is machine-visible. Aligns with the testability program (#192).
- **CRITICAL S3776**, `SSO-Auth/Api/SSOController.cs` cognitive complexity. **Tracked by #160** (thin the monolith into pure helpers); the #128 guard lines did not materially change it. Disposition: covered by #160.
- **CRITICAL S3776**, `SSO-Auth/Api/OidcIdTokenValidator.cs` (#134). **Fix:** split the validation method into named sub-checks (signature / iss / aud+azp / exp), mirroring the SAML validator decomposition.
- **MAJOR S6964 x2**, `SSO-Auth/Config/PluginConfiguration.cs` SamlConfig value-type properties bound via `[FromBody]` on OID/SAML `Add` (#156). Under-posting risk: a partial admin PUT silently defaults missing bools. **Evaluate:** these endpoints are `RequiresElevation` and replace the whole provider object by design, so under-posting is an admin-only, non-security concern; decide between annotating the model or documenting the whole-object-replace contract. Applies to the pre-existing value-type properties too, not just the #156 additions.
- **MAJOR S125**, `SSO-Auth/Api/SamlRequestCache.cs` commented-out code (#156). **Fix:** remove the dead comment or convert it to prose if it documents intent.

Pick these up as small, single-purpose PRs (the S2699 assertion and S125 removal are trivial). Priority medium; not a merge blocker (the quality gate is green on new code).

iderex:
[...]

Issue #160, opened 2026-07-12
Thin the SSOController monolith into pure single-responsibility helpers


iderex:
Continuous architecture work: extract pure, testable helpers out of SSOController.cs and Saml.cs as each security item touches them (the V2 Api/* pattern), so the controller thins and coverage climbs with every port. Tracking issue for the ongoing decomposition; individual extractions ride in with the PRs that touch each area.

Acceptance criteria (rolling):
- [ ] Each PR touching the controller leaves it no larger and ideally smaller (net line delta tracked).
- [ ] Extracted helpers are pure and unit-tested.
- [ ] Controller retains only thin orchestration; no business logic added inline.

Refs the minimal-compact-OOP prime directive; #89 (role-mapping decomposition, closed) as the pattern.

PR #275, opened 2026-07-13
Add explicit assertions to the does-not-throw tests

iderex:
## What & why

Three tests asserted their contract only by *not throwing*, which SonarCloud flags as **S2699** (BLOCKER, "add at least one assertion"):

- `SamlCertificateTests.RejectInvalidSamlCertificate_ValidOrBlank_DoesNotThrow`
- `CanonicalBaseUrlTests.RejectInvalidBaseUrlOverride_BlankOrValid_DoesNotThrow`
- `ConfigPreservationTests.Preserve_NullConfigMaps_DoNotThrow`

Wrap each call in `Record.Exception(...)` and assert the result is `null`, so the does-not-throw intent is machine-visible. What each test exercises is unchanged.

## Type of change
- [x] Test quality (clears 3 SonarCloud S2699 BLOCKERs)

## Verification
- [x] `dotnet test` 492 passing (unchanged count, the assertions replace the bare calls).

## Notes

Partial pickup of the #196 sweep. Not touched here, with their disposition:
- **S125 ×2** (`SsoRateLimiter.cs`, `SamlRequestCache.cs`), these are **false positives**: both flagged lines are explanatory prose (the heuristic misfires on the punctuation), not commented-out code. They should be marked *Accepted* in SonarCloud rather than have good comments degraded to appease the rule.
- **S3776 ×4** (`SSOController` ×3, `OidcIdTokenValidator`), cognitive-complexity refactors, larger; the `SSOController` ones fold into the monolith decomposition (#160).
- **S6964 ×4 / S6932 ×1** (`PluginConfiguration`, `SSOController`), value-type `[FromBody]` / model-binding on `RequiresElevation` whole-object-replace endpoints; needs the annotate-vs-document decision.

Refs #196

### DIFF
--- a/SSO-Auth.Tests/CanonicalBaseUrlTests.cs
+++ b/SSO-Auth.Tests/CanonicalBaseUrlTests.cs
@@ -87,7 +87,9 @@ public class CanonicalBaseUrlTests
     [InlineData("https://example.com/jellyfin/")]
     public void RejectInvalidBaseUrlOverride_BlankOrValid_DoesNotThrow(string? raw)
     {
-        SSOController.RejectInvalidBaseUrlOverride(raw);
+        var exception = Record.Exception(() => SSOController.RejectInvalidBaseUrlOverride(raw));
+
+        Assert.Null(exception);
     }
 
     // Resolve (#242): the base-URL decision GetRequestBase used to make inline against the live Request.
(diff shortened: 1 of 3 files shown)


iderex (commented):
## Pre-merge review — assert the does-not-throw tests (#196, partial)

Test-only. Three tests asserted their contract only by not throwing; each now wraps the call in `Record.Exception` and asserts the result is `null`, clearing three SonarCloud S2699 BLOCKERs without changing what is tested.

### Review
- Test files only — no production code, no login/crypto/config surface, so no adversarial lens gate.
- CodeRabbit: Review completed — "No actionable comments were generated" (no inline, no outside-diff). Copilot: quota-limited.
- The rest of #196 stays open with a documented disposition on the PR (S125 are false positives to mark Accepted; S3776/S6964/S6932 are larger refactors / an annotate-vs-document decision).

### CI
All green — build/test (492), CodeQL, both zizmor checks, dependency-review, prettier, Trojan Source, SonarCloud.

Merging.

Issue #289, opened 2026-07-13
Make the OpenID authorize-state store test-resettable to cover the challenge/callback paths


iderex:
## Context

The in-process `SsoControllerHarness` now serves OpenID discovery through a stub HTTP responder (added alongside the OidChallenge PKCE tests), which is enough to cover the challenge's fail-closed PKCE gate. It is **not** yet enough to cover the full challenge happy path or the `OidAuth`/`SamlAuth` callbacks.

## Problem

`SSOController` holds the authorize-state store as a **private static** `ConcurrentDictionary` (`StateManager`). A test that drives `OidChallenge` far enough to call `AuthStateStore.TryAdd(...)` leaves an entry in that process-wide store, which then leaks into sibling tests (observed: it breaks `OidStates_NoFlowsInProgress_ReturnsEmpty`). There is no reset hook, so the challenge happy path and the callbacks cannot be unit-tested in isolation today.

## Proposal

Make the authorize-state store test-resettable or injectable without weakening production behaviour, e.g.:

- an `internal static` reset method guarded by `InternalsVisibleTo` (cheapest), or
- inject the store so each harness gets a fresh instance (cleaner, larger).

Then add the deferred coverage: the challenge redirect happy path, and the `OidAuth`/`SamlAuth` callback branches with a mocked discovery document and user-creation path.

Part of the testability program (#192). Non-urgent; the security-critical fail-closed branches are already covered.

Issue #192, opened 2026-07-12
Testability program: every behavior automated-tested or an explicit release-QA item


iderex:
Standing program (owner's directive): every functional behavior of the plugin is **automatically tested where that is professionally sensible**; whatever cannot reasonably be automated inside the plugin's scope is **explicitly documented as a manual QA check** that runs before every release. Nothing lives in the gap between the two, a behavior is either covered by CI or listed on the release checklist, never silently untested.

## Current state

- ~315 xUnit v3 unit tests over the extracted pure helpers (SAML validation core, replay/request caches, rate limiter, account-link resolvers, state builders, id_token validator, config preservation).
- `docs/E2E-CHECKLIST.md` (local) holds the manual, real-server checks.
- The gap: everything that needs an HTTP pipeline, the Jellyfin host, or a live IdP, endpoint wiring (routing/authorize attributes/headers/429), the full OIDC/SAML round-trips, config-page JS, upgrade paths.

## Program of work (each layer becomes its own issue/PR when picked up)

1. **Keep shrinking the untestable core.** Continue extracting pure, single-responsibility helpers out of `SSOController` (directive 4); every extraction lands with tests. Target: the controller is thin enough that its remaining logic is wiring only.
2. **In-process endpoint tests (biggest lever).** Host the controller in-proc via `Microsoft.AspNetCore.Mvc.Testing`/`TestServer` with the Jellyfin interfaces (`IUserManager`, `ISessionManager`, `IAuthorizationContext`, …) substituted. This makes automated what today is manual: route + `[Authorize]`/elevation policy coverage on every endpoint, security headers, 429/`Retry-After` behavior, JSON contract shapes (secret redaction, `JsonIgnore` fields), model binding, and error-path status codes.
3. **Fake-IdP round-trip tests.** An in-test OIDC provider (discovery document + JWKS + token endpoint on the TestServer) and the existing `SamlTestFactory` fixtures, driving the full challenge → callback → auth flow without a real IdP. Covers the login path end-to-end minus the browser.
4. **Config lifecycle tests.** XML round-trip of old config files (defaults, unknown elements), JSON GET→PUT round-trip through the generic plugin-config endpoint, server-managed-field preservation, automated against the real serializers.
[...]

PR #299, opened 2026-07-13
Cover the OpenID auth callback (OidAuth)

iderex:
# What & why

Covers the OpenID auth callback (`OidAuth`), the last large uncovered login-path branch, now that #289's state-store reset makes such tests isolatable.

- **Guard branches:** missing data → 400; a disabled provider and an unknown state each → a problem response (no session).
- **Happy path:** a valid, seeded authorize state provisions the account (`CreateUserAsync`) and mints a session (`Ok`), and the state is **single-use**, a replay of the same token no longer redeems (the atomic `TryRemove` claim).

To reach the callback without standing up the full browser token-exchange leg, this adds an `internal`, test-only `SeedOidStateForTests(token, state)` beside the existing `ResetOidStateForTests`, same test-only surface (`InternalsVisibleTo`, no HTTP attribute, no endpoint/DI, never reachable in production). It also adds a concrete `FakeCryptoProvider`, because NSubstitute cannot proxy `ICryptoProvider`'s `ReadOnlySpan<char>` overloads (a ref struct cannot cross a dynamic-proxy boundary) and the account-provisioning path hashes a random password.

Refs #192

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

Touches `SSOController.cs` (login-path file), so the adversarial-review gate was run (4 lenses) even though the only production change is a test-only internal method.

- [x] Fail-closed preserved, the tests assert the callback's guards (missing data/unknown state/disabled → no session) and the single-use claim; no posture changed.
- [x] A security review was performed, 4-lens gate below.
- [x] Log inputs from the IdP are sanitized inline at the log call. (unchanged; no logging touched)

**Adversarial-review gate (refute-by-default):**
- **bypass** → PASS: the seed hook is `internal`, single friend assembly (`SSO-Auth.Tests`), one test caller, no HTTP attribute/DI/reflection; it writes a private static that is already reflectable in-process, so it adds no production-reachable or remote surface; purely additive.
- **integration** → PASS: not a routable action method; `FakeCryptoProvider` is a faithful inert stand-in on the exercised path and does not ship; harness ctor arg order matches the 9-arg production ctor. (It also caught that the work had been started on `main` before branching, corrected before any commit.)
- **correctness** → PASS: constructor-time `ResetOidStateForTests` isolates the seed within the serialized collection; the `ConcurrentDictionary` upsert is sound; the crypto swap regresses nothing (no test asserts on crypto); the replay-500 assertion genuinely depends on the atomic removal (it would fail if single-use were broken).
- **availability** → N/A: test-only; seeding one state cannot cause a lockout.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose, testable unit. (a 3-line seed helper + a 5-method inert fake)

## Verification

- [x] `dotnet test` is green (542 tests; 3 consecutive full runs to confirm order-independence).

## Notes

- Builds on the #289 reset hook. The SAML callback (`SamlPost`/`SamlAuth`) is the remaining large branch; it needs a signed-SAML fixture rather than a state seed, and is a separate follow-up.

### DIFF
--- a/SSO-Auth.Tests/FakeCryptoProvider.cs
+++ b/SSO-Auth.Tests/FakeCryptoProvider.cs
@@ -1,0 +1,24 @@
+using System;
+using MediaBrowser.Model.Cryptography;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// A minimal real <see cref="ICryptoProvider"/> for tests. NSubstitute cannot proxy the
+/// <c>ReadOnlySpan&lt;char&gt;</c> overloads (a ref struct cannot cross a dynamic-proxy boundary), so the
+/// account-provisioning path — which hashes a random password — needs a concrete implementation. The
+/// values are inert; no test asserts on the produced hash.
+/// </summary>
+internal sealed class FakeCryptoProvider : ICryptoProvider
+{
+    public string DefaultHashMethod => "PBKDF2-SHA512";
+
+    public PasswordHash CreatePasswordHash(ReadOnlySpan<char> password) =>
+        new PasswordHash(DefaultHashMethod, Array.Empty<byte>());
+
+    public bool Verify(PasswordHash hash, ReadOnlySpan<char> password) => true;
+
+    public byte[] GenerateSalt() => Array.Empty<byte>();
+
+    public byte[] GenerateSalt(int length) => new byte[length];
+}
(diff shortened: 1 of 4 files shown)


iderex (commented):
Pre-merge review — OpenID auth callback coverage (OidAuth), Refs #192.

Adversarial-review gate (4 lenses) on the login-path file (only production change is a test-only internal seed hook):
- bypass PASS, integration PASS, correctness PASS; availability N/A (test-only, one seeded state cannot cause lockout).
- integration lens additionally caught that work had begun on main before branching — corrected before any commit; process rule recorded.

Review surfaces: CodeRabbit posted no actionable comments; no inline, no outside-diff-range block, no walkthrough nits. Review completed on the head commit.

CI: all 16 checks green.

Verdict: approve for merge.

Issue #305, opened 2026-07-13
Cover the OpenID redirect callback token-exchange path (OidPost) with a signed-id_token fixture


iderex:
## Context

The controller-test coverage program has covered every SSOController endpoint except the OpenID redirect callback's happy/token-exchange path. `OidPost`'s guard branches are already covered (unknown provider, missing/invalid/expired state, disabled, rate-limit, SSOControllerEndpointTests / SSOControllerAdminTests), and the SAML side is fully covered (SamlPost + SamlAuth via SamlTestFactory). What remains uncovered is the part of `OidPost` from `oidcClient.ProcessResponseAsync(...)` onward.

## Problem

`OidPost` performs the full OIDC authorization-code exchange: `CreateCallbackOidcClient` → `ProcessResponseAsync(queryString, authorizeState)` (POST the code to the token endpoint, receive and validate the id_token via `OidcIdTokenValidator`), then the RFC 9207 `iss` mix-up check, `sub` presence, claim-derived state population, and finally the intermediate HTML auth page (or a 401). Covering this in-process needs an OIDC round-trip fixture that today's harness does not provide:

- a signed **id_token** (RSA, `Microsoft.IdentityModel.JsonWebTokens`, the machinery already exists privately in `OidcIdTokenValidatorTests`), with `iss`/`aud`/`sub`/`exp` and the **nonce** that matches the `AuthorizeState`;
- the discovery document + a **JWKS** carrying the signing public key (served through the harness's existing `httpResponder`);
- a **token endpoint** in the responder returning `{access_token, id_token, token_type}`;
- a real `AuthorizeState` (from driving `OidChallenge` with the discovery mock, then extracting its `state`/`nonce`), or a hand-built one with matching `nonce`/`code_verifier`/`redirect_uri`.

## Proposal

Add an `OidcTokenFixture` test helper (mirroring `SamlTestFactory`) that produces the signed id_token + JWKS + token-endpoint responder, and use it to cover:

- happy path: a valid callback populates the state (`Valid`, `Subject`, `Username`, `Admin`, folders) and renders the auth page (HTML `ContentResult`);
- the `ProcessResponseAsync` error branch → 400 "Error logging in";
- the RFC 9207 `iss`-mismatch branch → 400;
- the missing-`sub` branch → 401.

[...]

PR #306, opened 2026-07-13
Cover the SAML assertion-consumer callback (SamlPost)

iderex:
# What & why

Covers the SAML assertion-consumer callback (`SamlPost`), the SAML side's redirect callback, completing the callback coverage. It uses a signed `SamlTestFactory` response so the actual signature-validation path runs.

- **Guard branches:** a disabled provider → 400 "No active providers found"; a response signed by a **different** certificate → 400; an assertion whose role is not in the allow-list → 401.
- **Happy path:** a valid signed assertion renders the intermediate **HTML auth page**, a `text/html` `ContentResult` served with the hardened CSP header, which then posts to `SamlAuth`.

The OpenID redirect callback's token-exchange path (`OidPost` from `ProcessResponseAsync` onward) is deliberately out of scope and tracked in **#305**: its guard branches are already covered, and the token-exchange happy path needs a signed-id_token / JWKS / token-endpoint round-trip fixture. The `OidPost` guards (unknown provider, missing/invalid/expired state, disabled, rate-limit) are already covered in `SSOControllerEndpointTests`/`SSOControllerAdminTests`.

Refs #192

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] N/A for the running product, test-only change; no production code is touched (the diff adds a single test file). The tests pin the existing fail-closed SamlPost behaviour (invalid signature → 400, role denied → 401, disabled → 400) rather than change it.
- [x] N/A, no SAML/OIDC validation, config persistence, or release-pipeline code changed.
- [x] N/A, no new log call.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose, testable unit. (reuses `SamlTestFactory` and the harness)

## Verification

- [x] `dotnet test` is green (555 tests; 3 consecutive full runs).

## Notes

- With this, every SSOController endpoint except the OidPost token-exchange path (#305) has controller-level coverage.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerSamlPostTests.cs
+++ b/SSO-Auth.Tests/SSOControllerSamlPostTests.cs
@@ -1,0 +1,80 @@
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Microsoft.AspNetCore.Mvc;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// In-process tests of the SAML assertion-consumer callback (<c>SamlPost</c>) via
+/// <see cref="SsoControllerHarness"/>, using <see cref="SamlTestFactory"/> to produce a real, signed
+/// response so the actual signature-validation path runs. They cover the guard branches (disabled
+/// provider, invalid signature, role not allowed) and the happy path, where a valid signed assertion
+/// renders the intermediate auth page (an HTML <see cref="ContentResult"/> that later posts to SamlAuth).
+/// </summary>
+[Collection("SSOController")]
+public class SSOControllerSamlPostTests
+{
+    [Fact]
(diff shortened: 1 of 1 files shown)


iderex (commented):
Pre-merge review — SAML assertion-consumer callback coverage (SamlPost), Refs #192.

Adversarial lenses: N/A — test-only change, no production code (single test file). The tests pin existing fail-closed SamlPost behaviour (invalid signature → 400, role denied → 401, disabled → 400) and the happy-path HTML auth page, using SamlTestFactory's real signed responses.

Review surfaces: CodeRabbit posted no actionable comments; no inline, no outside-diff-range block. Review completed.

CI: all 16 checks green. Every SSOController endpoint now has controller-level coverage except the OidPost token-exchange path (tracked #305).

Verdict: approve for merge.

Issue #246, opened 2026-07-13
OIDC authorize-state store: no size cap and an unthrottled full sweep on every anonymous challenge


iderex:
## Problem

The OIDC authorize-state store is the one login-path cache without a size cap or a throttled sweep, and it sits on the busiest anonymous endpoint.

- `SSOController.cs:65`, `StateManager` is a static `ConcurrentDictionary<string, TimedAuthorizeState>` with no entry cap. Every anonymous `GET /sso/OID/{p|start}/{provider}` inserts a `TimedAuthorizeState` carrying the full Duende `AuthorizeState` (state token, nonce, PKCE code verifier, `StartUrl`), roughly 1–2 KB per entry. The only eviction is expiry (`StateLifetime` = 15 min, `SSOController.cs:70`).
- `SSOController.cs:257`, every challenge first calls `Invalidate()` → `AuthStateStore.InvalidateExpired` (`AuthStateStore.cs:24`), an unthrottled full-table LINQ scan. Cost per challenge is O(entries).
- The rate limiter is opt-in and default-off (`Config/PluginConfiguration.cs:43`), so nothing bounds the insert rate by default.

An unauthenticated challenge flood therefore grows the store unboundedly for the 15-minute lifetime while every request pays an increasingly expensive scan: 200 rps sustains ~180k live entries (hundreds of MB) and O(n²) aggregate sweep CPU; 1k rps reaches the GB range. Each flood request additionally triggers `PrepareLoginAsync`, i.e. outbound discovery + JWKS fetches against the IdP (request amplification, see the discovery-cache issue).

The two sibling caches already implement exactly the missing safeguards, with a written rationale:

- `SamlRequestCache.cs:22`, `MaxEntries = 100_000` against an "abandoned-login flood", and a 1-minute CAS-throttled prune (`SamlRequestCache.cs:24-28`) because an unthrottled O(n) scan "would amplify a flood into CPU load".
- `SsoRateLimiter.cs`, same cap + throttle discipline.

## Fix

Bring `StateManager` up to the sibling pattern, inside `AuthStateStore` (it already owns the expiry logic):

- CAS-throttled prune (1/min) in `InvalidateExpired`; the per-challenge `Invalidate()` call then costs O(1) amortized.
[...]

Issue #128, opened 2026-07-12
P2#12 design: rate-limit the public SSO endpoints without a mass-lockout hazard


iderex:
Rate-limit the unauthenticated SSO endpoints (OidChallenge/OidPost/OidAuth, SamlPost/SamlAuth) against brute-force and DoS. **Blocked on a design decision** because the naive version is a mass-lockout landmine, flagging it rather than merging blind.

## The hazards

1. **IP attribution behind a reverse proxy (the mass-lockout landmine).** A Jellyfin plugin cannot configure the host's `ForwardedHeaders` middleware. If the operator runs Jellyfin behind a reverse proxy *without* forwarded-headers configured, `HttpContext.Connection.RemoteIpAddress` is the **proxy's** IP, i.e. the SAME key for every user. A per-IP limiter would then throttle the entire userbase as one bucket the moment any abuse occurs. Trusting `X-Forwarded-For` instead is spoofable (an attacker sets it to random values to evade, or to a victim value to lock a target out) unless the trusted-proxy chain is known, which, again, the plugin does not own.
2. **IPv6 keying evasion (V2 audit HIGH).** Keying on the full /128 lets an attacker with a /64 (a normal residential allocation) rotate through 2^64 addresses to evade. Must key IPv6 on /64.
3. **Fail direction.** Must fail **open** when the client IP is ambiguous/unavailable (availability > throttling), but fail **closed** (throttle) on genuine abuse from an attributable source.

## Proposed direction (needs pre-merge review)

- **Admin opt-in**, default off, with explicit config: enable flag, trusted-proxy stance (honor `X-Forwarded-For` only when the operator asserts a trusted proxy), window/limit thresholds.
- Pure `SsoRateLimiter` helper (mirrors `SamlReplayCache`): `ConcurrentDictionary` sliding-window per normalized key, bounded + pruned; IPv6 normalized to /64; unknown/ambiguous IP ⇒ do not throttle.
- 429 with `Retry-After` on throttle; applied at the top of the public endpoints.
- Full 4-lens gate with availability as the primary lens; a real-server check behind a proxy is mandatory before release.

Until this is designed for my own workflow, the endpoints rely on the existing controls (CSPRNG state, provider-scoped one-time replay cache, signature/time/audience validation).

iderex:
[...]

PR #307, opened 2026-07-13
Add a bounded throttle-engaged observability signal to the rate limiter

### LINKED ISSUE #195: Rate limiter: add a bounded observability signal when throttling engages
Follow-up from the #128 CodeRabbit review (nitpick). The limiter currently emits no signal when it throttles, so an operator cannot notice a sustained brute-force attempt or a misconfigured reverse proxy pooling every client into one bucket.

Declined the reviewer's literal suggestion (log on every throttled request) for this PR: logging per rejected request amplifies the very flood the limiter exists to blunt into log volume, a self-inflicted DoS on an anonymous endpoint.

Wanted instead: a **bounded** signal. Options to weigh, a rate-limited log (e.g. once per key per window on the first refusal, tracked on the Counter), or a plain counter metric (total throttled hits) surfaced somewhere cheap. Must not scale with attack volume and must not log the raw client key without line-ending sanitization.

Refs #128.

iderex:
# What & why

Closes #195. Refs #128.

The opt-in login-path rate limiter (#128) emitted no signal when it refused a request, so an operator could not notice a sustained brute-force attempt or a reverse proxy misconfigured to pool every client into one bucket.

Logging on every refused request, the literal suggestion declined on #128, would amplify the very flood the limiter exists to blunt into log volume: a self-inflicted DoS on an anonymous endpoint. This adds a **bounded** signal that cannot scale with attack volume, mirroring the once-per-interval CAS discipline already used for the #246 capacity warning and the prune sweeps in the same file.

- New `SsoRateLimiter.RecordThrottledHit(nowUtc)`: increments a throttled-hit tally on every refusal and drains it into a single returned count **at most once per interval** via an `Interlocked` CAS (exactly one winner per interval), returning 0 while suppressed. Pure and unit-testable; it touches only its own two fields, never the throttle state, so it **cannot change the allow/deny decision**, it only observes it.
- `RateLimitCheck` emits **one** `LogWarning` carrying only the aggregate count, no client key, so there is nothing to sanitize or forge, and now captures the timestamp once to share with `IsAllowed` (behaviour-preserving).
- The onset of throttling signals immediately (operators see a brute-force start), then further refusals accumulate silently until the interval elapses; a backward wall-clock step fires and re-anchors rather than wedging the signal shut.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (Unchanged — the signal is observation-only, layered strictly after the decision point; it never re-admits a refused request.)
- [x] No secrets logged; secrets are redacted on config export. (The signal logs only an aggregate integer count.)
      config persistence, or the release pipeline. (Four-lens adversarial gate — see Notes.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (No client-supplied string reaches the log call; the count is machine-generated, so no sanitizer is needed, a deliberate improvement over per-key logging.)

## Quality checklist

      testable unit. (One ~20-line method reusing the established once-per-interval CAS idiom.)

## Verification

- [x] `dotnet test` is green. (554 passing, 4 new.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (None changed, C# only.)

## Notes

Adversarial-review gate (four refute-by-default lenses) ran on the diff; each was asked specifically whether the new signal can be turned into a log/CPU DoS amplifier under a flood and whether it changes the throttling decision:

- **availability-lens, PASS.** Log volume hard-bounded to one line per interval regardless of request rate; per-refusal cost is O(1) lock-free with no allocation; touches fields disjoint from the throttle state; no memory-growth or mass-lockout regression.
- **bypass-lens, PASS.** Throttle decision and signal fully decoupled in both directions; the 429 + Retry-After is emitted unconditionally; only a machine-generated `long` reaches the log (no CWE-117 surface).
- **correctness-lens, PASS.** No overflow/negative/lost-en-masse/double-count; every clock edge (first-call, backward step, exactly-at-boundary) fires/suppresses as documented and never wedges shut; tests assert the real accounting including the drain off-by-one.
- **integration-lens, PASS.** The `now`-capture refactor is behaviour-preserving; `_logger`/level/phrasing consistent with sibling warnings; process-global aggregation is the intended operator signal.

The only finding was a cosmetic one (two lenses): the message previously said "in the last minute", which undercounts when throttling is sparse and the tally spans multiple intervals. Fixed, the message now reads "since the last notice; further notices are suppressed for at least a minute."

Out of scope: no config classes, `WebResponse`, or challenge/callback logic touched. The signal is process-global by design (one bounded operator line, not per-endpoint).

### DIFF
--- a/SSO-Auth.Tests/SsoRateLimiterTests.cs
+++ b/SSO-Auth.Tests/SsoRateLimiterTests.cs
@@ -140,6 +140,51 @@ public class SsoRateLimiterTests
     }
 
     [Fact]
+    public void RecordThrottledHit_FirstRefusal_SignalsOnsetImmediately_ThenSuppressesWithinInterval()
+    {
+        // The onset of throttling must be visible at once (an operator should see a brute-force start),
+        // then the signal is bounded: further refusals inside the same interval accumulate silently.
+        var limiter = new SsoRateLimiter();
+        Assert.Equal(1, limiter.RecordThrottledHit(Now));
+
+        for (var i = 0; i < 1000; i++)
+        {
+            Assert.Equal(0, limiter.RecordThrottledHit(Now.AddSeconds(1)));
+        }
+    }
+
+    [Fact]
+    public void RecordThrottledHit_DoesNotScaleWithVolume_OneSignalPerInterval_CarryingTheAccumulatedCount()
+    {
+        // The signal fires at most once per minute regardless of flood size, and each signal reports
+        // every refusal since the previous one — so a flood yields one bounded line, not one per hit.
+        var limiter = new SsoRateLimiter();
+        Assert.Equal(1, limiter.RecordThrottledHit(Now)); // onset drains the first hit
+
+        for (var i = 0; i < 500; i++)
+        {
+            limiter.RecordThrottledHit(Now.AddSeconds(5)); // accumulate, all suppressed
+        }
+
+        // A minute after the onset the next refusal drains the whole accumulated tally in one line.
+        Assert.Equal(501, limiter.RecordThrottledHit(Now.AddMinutes(1)));
+        // ...and the tally has reset, so the immediate follow-up is suppressed again.
+        Assert.Equal(0, limiter.RecordThrottledHit(Now.AddMinutes(1)));
+    }
+
+    [Fact]
+    public void RecordThrottledHit_BackwardClockStep_StillSignals_NeverStalls()
+    {
+        // A wall-clock correction must not be able to wedge the signal shut: a step backwards fires
+        // (and re-anchors the cursor) rather than being read as "still inside the interval".
+        var limiter = new SsoRateLimiter();
+        Assert.Equal(1, limiter.RecordThrottledHit(Now));
+        Assert.Equal(0, limiter.RecordThrottledHit(Now.AddSeconds(1)));
+
+        Assert.Equal(2, limiter.RecordThrottledHit(Now.AddMinutes(-5)));
+    }
+
+    [Fact]
     public void NormalizeClientKey_PublicIpv4_UsesFullAddress()
     {
         Assert.Equal("1.2.3.4", SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("1.2.3.4")));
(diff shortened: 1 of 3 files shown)


iderex (commented):

Adversarial-review gate (four refute-by-default lenses on the diff; each asked specifically whether the new signal can be turned into a log/CPU DoS amplifier under a flood and whether it changes the throttling decision):
- availability-lens: PASS — log volume hard-bounded to one line per interval regardless of request rate; per-refusal cost O(1) lock-free, no allocation; fields disjoint from throttle state; no memory-growth or mass-lockout regression.
- bypass-lens: PASS — decision and signal fully decoupled; the 429 + Retry-After emit unconditionally; only a machine-generated long reaches the log (no CWE-117 surface).
- correctness-lens: PASS — no overflow/negative/lost-en-masse/double-count; every clock edge fires/suppresses as documented and never wedges; tests assert the real accounting incl. the drain off-by-one.
- integration-lens: PASS — the now-capture refactor is behaviour-preserving; logger/level/phrasing consistent with sibling warnings; process-global aggregation is the intended operator signal.

The single (cosmetic) finding — the "in the last minute" label undercounts when the tally spans multiple intervals — was fixed before the CI run: the message now reads "since the last notice; further notices are suppressed for at least a minute."

CodeRabbit: Review completed. No actionable comments; all five pre-merge checks passed (title, description, linked issues, out-of-scope, docstring). No inline, review-body, outside-diff, or walkthrough findings to disposition. Its advisory label suggestions (security/hardening) are noted; the issue's own triage (enhancement / area: config / priority: low) is mirrored onto the PR as per the labelling policy.

CI: build --warnaserror green, dotnet test green (554 passing, 4 new), CodeQL/SonarCloud/prettier/dependency-review/zizmor/abi-floor all green. SonarCloud gate fully green (not coverage-red). Security checklist: fail-closed preserved (observation-only, layered after the decision point), no secret or client key logged, log input carries no client-supplied string.

Scope held to SsoRateLimiter.cs (+ tests) and the minimal RateLimitCheck wiring; no config classes, WebResponse, or challenge/callback logic touched.

PR #317, opened 2026-07-13
Key IPv4-in-IPv6 transition rate-limit sources on the embedded IPv4

### LINKED ISSUE #194: Rate limiter: key NAT64/transition IPv6 sources on the embedded IPv4, not the shared /64
Follow-up from the #128 adversarial-review gate (correctness + availability lenses, both LOW).

`SsoRateLimiter.NormalizeClientKey` keys IPv6 on the /64 prefix. For IPv4-in-IPv6 transition sources, the NAT64 well-known prefix `64:ff9b::/96`, the deprecated IPv4-compatible `::/96`, and Teredo, the client-identifying embedded IPv4 lives in the low 64 bits that get zeroed, so **distinct clients collapse into one /64 bucket**. Probe-confirmed: `64:ff9b::808:808` and `64:ff9b::909:909` both key to `64:ff9b::/64`.

Impact: in an IPv6-only deployment that translates inbound IPv4 via the NAT64 well-known prefix (SIIT-DC), one abusive IPv4 client would throttle *every* IPv4 client, the shared-bucket lockout class the limiter defends against elsewhere. Realistically routable NAT64-WKP/IPv4-compatible/Teredo *source* addresses at a remote server are exotic (NAT64 WKP is destination-synthesis; IPv4-compatible is dead; Teredo relays decommissioned), and the feature is opt-in, hence LOW.

Fix direction: `AvatarUrlValidator.TryExtractEmbeddedIPv4` already unwraps these forms; reuse it in `NormalizeClientKey` so a transition source keys on its embedded public IPv4 (or its /64 only when no IPv4 is embedded). Add tests for NAT64-WKP and IPv4-compatible distinct-client separation.

Refs #128.

iderex:
# What & why

The opt-in SSO rate limiter (`SsoRateLimiter.NormalizeClientKey`, used by
`SSOController.RateLimitCheck`) keyed every IPv6 source on its `/64` prefix. For
IPv4-in-IPv6 transition sources the client-identifying IPv4 lives in bits that
`/64` keying discards, so distinct NATed clients collapse into one shared bucket
and a single abuser throttles all of them, the shared-bucket lockout class the
limiter defends against everywhere else. In an IPv6-only deployment translating
inbound IPv4 through the NAT64 well-known prefix (SIIT-DC), one abusive client
would throttle every IPv4 client.

Fix: reuse `AvatarUrlValidator.TryExtractEmbeddedIPv4` (promoted `private` →
`internal`) from `NormalizeClientKey`. When a transition source carries an
embedded IPv4, key on that IPv4; otherwise keep the `/64`.

- NAT64 well-known prefix (`64:ff9b::/96`): IPv4 in bytes 12-15, zeroed by `/64`
  → now split per client. Strictly narrows the shared bucket.
- IPv4-compatible (`::/96`, deprecated): IPv4 in the low 32 bits → same.
- 6to4 (`2002::/16`): IPv4 in bytes 2-5 → a multi-`/64` 6to4 site now buckets on
  its one gateway IPv4, i.e. egress-identity keying identical to a native NAT,
  which also closes a rotate-within-your-own-`/48` evasion hole. Bounded,
  deliberate trade for a deprecated form.

It is the same extractor `IsBlockedAddress` already applies, so the SSRF
classifier and the limiter cannot disagree on what an embedded IPv4 is. Because
`IsBlockedAddress` runs first and rejects a blocked/internal embedded IPv4, any
source reaching the keying has a public embedded IPv4. The branch only
re-buckets, never returns null, so throttling is preserved (fail closed). A
network-specific NAT64 prefix (RFC 6052 NSP) is not recognized and falls through
to `/64`, as does any non-transition IPv6.

Closes #194. Refs #128.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (This change only re-buckets a
      throttle key; it never returns a null/blank key where the old code
      returned one, so no source that was throttled now fails open.)
      secrets touched.)
      config persistence, or the release pipeline. (Four-lens adversarial gate
      run — see Notes.)
      log calls added.)

## Quality checklist

      testable unit. (Reuses the existing `TryExtractEmbeddedIPv4`; a single
      pure branch in `NormalizeClientKey`.)

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. (Debug and Release.)
- [x] `dotnet test` is green. (567 tests, run 3x for stability.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (N/A - 
      C# only.)

## Notes

Trust model: `NormalizeClientKey`'s only input is the socket peer address, never
a client-supplied header (the plugin never parses X-Forwarded-For), so the
embedded IPv4 is as trustworthy as any source address. An attacker cannot forge
it to evade throttling (rotation now costs a distinct real IPv4, same as native)
or to pin a victim (sourcing from `64:ff9b::<victimV4>` requires routing control
of that prefix, a TCP handshake precludes spoofing).

Adversarial-review gate (4 refute-by-default lenses): availability PASS,
bypass PASS, correctness PASS, integration NEEDS_IMPROVEMENT, the sole
integration finding was two inaccurate phrases in a load-bearing code comment
("never widens it" is false for the 6to4 branch; "SIIT-DC NAT64" overclaimed vs
the well-known-prefix-only implementation). Both were comment-only and are now
corrected; the code and its three PASS verdicts were unaffected. Both lenses
specifically confirmed the change cannot be spoofed to evade or to lock out a
victim, and never fails open.

Out of scope: Teredo and RFC 6052 network-specific NAT64 prefixes are not
unwrapped (neither is by the shared SSRF classifier); those sources keep `/64`
keying. This is a documented gap, not a regression.

### DIFF
--- a/SSO-Auth.Tests/SsoRateLimiterTests.cs
+++ b/SSO-Auth.Tests/SsoRateLimiterTests.cs
@@ -209,6 +209,80 @@ public class SsoRateLimiterTests
     }
 
     [Fact]
+    public void NormalizeClientKey_Nat64WellKnownPrefix_KeysOnEmbeddedIpv4_NotSharedSlash64()
+    {
+        // SIIT-DC synthesizes inbound IPv4 into 64:ff9b::/96, embedding the client IPv4 in the low 32
+        // bits that /64 keying would zero. Two distinct NATed clients must land in distinct buckets
+        // (matching their native-IPv4 keys), not collapse into one 64:ff9b::/64 shared-throttle bucket.
+        var a = SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("64:ff9b::808:808")); // 8.8.8.8
+        var b = SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("64:ff9b::909:909")); // 9.9.9.9
+
+        Assert.Equal("8.8.8.8", a);
+        Assert.Equal("9.9.9.9", b);
+        Assert.NotEqual(a, b);
+    }
+
+    [Fact]
+    public void NormalizeClientKey_Nat64EmbeddedIpv4_SharesBucketWithNativeIpv4_SameClient()
+    {
+        // The embedded public IPv4 IS the client identity, so a NAT64-translated hit and a native-IPv4
+        // hit from the same address share one bucket — the correct, consistent view, not key confusion.
+        Assert.Equal(
+            SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("8.8.8.8")),
+            SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("64:ff9b::808:808")));
+    }
+
+    [Fact]
+    public void NormalizeClientKey_6to4_KeysOnEmbeddedIpv4_NotSharedSlash64()
+    {
+        // 6to4 (2002::/16) embeds the IPv4 in bytes 2-5; 2002:0808:0808::/48 carries 8.8.8.8. Distinct
+        // embedded IPv4s must not collapse into one 2002:.../64 bucket.
+        var a = SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("2002:808:808::1"));   // 8.8.8.8
+        var b = SsoRateLimiter.NormalizeClientKey(IPAddress.Parse("2002:909:909::1"));   // 9.9.9.9
+
+        Assert.Equal("8.8.8.8", a);
+        Assert.Equal("9.9.9.9", b);
+        Assert.NotEqual(a, b);
+    }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


iderex (commented):
Pre-merge review (delegated agent PR, Lead Dev integrating) — rate-limit keying for IPv4-in-IPv6 transition sources (#194).
Login-path change: the 4-lens adversarial gate ran — availability/bypass/correctness PASS, integration NEEDS_IMPROVEMENT (a comment wording, corrected; code and the three PASS verdicts unaffected). Confirmed fail-closed (never returns null → throttling preserved) and non-spoofable (the embedded IPv4 is as trustworthy as the source address; a TCP handshake precludes spoofing). CodeRabbit: Review completed, no unresolved findings. CI: all 16 green; 567 tests run 3x. Verdict: approve for merge.

iderex:
@coderabbitai review

Issue #318, opened 2026-07-13
Establish a self-documenting, unified object-oriented architecture (standard + wiki docs)


iderex:
## Goal

Make the codebase **super user-friendly**: readable in principle **without comments**, and converge it over time on **one unified object-oriented architecture** so that an outside contributor can find their way around by structure and naming alone. Document the resulting standard for others, primarily in the wiki.

This is the umbrella that plans the direction; individual moves land as their own small, gated PRs referencing it.

## Principles (the standard, all four hold together)

1. **Self-documenting code.** Intention-revealing names, small single-responsibility types/methods, obvious control flow. Comments explain **why** (a non-obvious decision, a spec/CVE/fail-closed rationale), never **what** the code already says. If a comment is load-bearing for basic comprehension, refactor the code instead. (The load-bearing security "why" comments the directives require stay, this is about the *code itself* reading clearly.)
2. **As little source code as possible.** The least code that does the job, continuously re-optimised, no duplication, no dead code, no gold-plating. Removing code beats adding it. (The existing minimal-compact-OOP prime directive; this composes with it and never loosens it.)
3. **Super secure, fail closed.** This is a login path: security outranks everything else and the default posture is fail-closed (missing signature/time-bound/audience/sub → reject, never default-accept). Secrets never logged, redacted on export. A "cleaner" or "shorter" form is only acceptable if it preserves every security invariant; readability/brevity never trade away a guard. Login-path/crypto/config/release changes pass the adversarial-review gate.
4. **One unified OO architecture.** A thin controller delegating to pure, single-responsibility helpers; consistent naming and patterns across the OIDC, SAML, config, and rate-limit surfaces; no divergent one-off styles. Prefer matching the existing shape over inventing a new one.

These four are not in tension by default, but when brevity, readability, and security pull apart, the order is **security > minimal code > readability**; never sacrifice a fail-closed guard for fewer lines or a nicer name.

## Plan (incremental, each its own gated PR)

- [ ] **Write the standard down in the wiki**, a new **Architecture / Coding-Standards** page: the layering (thin controller → pure helpers), naming conventions, the "comments = why, not what" rule, the least-code and fail-closed principles above, the security invariants, and how the pieces fit (OIDC flow, SAML flow, config lifecycle, rate limiting). Link it from the wiki Home and CONTRIBUTING.
[...]

PR #323, opened 2026-07-13
Tighten the architecture-conformance rules (close the #322 review gaps)

iderex:
# What & why

Follow-up to #322: CodeRabbit found five real gaps where the architecture-conformance fitness functions were **weaker than they claimed**, a fitness function that gives false assurance is worse than none, so these are fixed here.

- **Namespace boundary** now covers **all** types (interfaces/enums/structs/delegates), not just classes; rejects the global namespace (null); and matches the root **exactly or a `Root.`-prefixed descendant**, so a sibling namespace like `…SSO_AuthEvil` no longer slips through (the previous `StartsWith("Jellyfin.Plugin.SSO_Auth")` accepted it).
- **Helper-internal rule** also flags **protected / protected-internal nested** helpers (`IsNestedFamily` / `IsNestedFamORAssem`), not only `public`/`nested-public`.
- **Sealed-or-static rule** now uses `!IsSealed`, an **abstract unsealed** helper (an open inheritance base) is caught; a static class is sealed and still passes.
- **Suffix matching** strips the generic-arity backtick (`Cache\`1` → `Cache`), so a generic helper cannot escape the rule.

All rules stay green, the current tree already satisfies the stricter checks; this only removes the blind spots.

Refs #318

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] N/A for the running product, test-only change; no production code is touched. It tightens architecture assertions, changing no behaviour, so the adversarial-review gate does not apply.
- [x] N/A, no SAML/OIDC validation, config persistence, or release-pipeline code changed.
- [x] N/A, no new log call.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose, testable unit.
- [x] Architecture-conformance tests pass (they are the subject of this change and are now rigorous).

## Verification

- [x] `dotnet test` is green (584 tests; the tightened rules still pass on the current tree).

## Notes

- Process note: #322 was merged before its CodeRabbit findings were dispositioned (a merge-and-check-in-one-step slip); this PR closes those findings and each is replied to on #322. The process rule to prevent recurrence is recorded.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -11,13 +11,15 @@ namespace Jellyfin.Plugin.SSO_Auth.Tests;
 /// <summary>
 /// Architecture-conformance fitness functions for the target architecture planned in #318. These run as
 /// part of the ordinary test suite, so every PR is checked — a change that drifts from the agreed
-/// structure fails CI. The rules encode structural invariants that hold TODAY and are part of the target;
+/// structure fails CI. The rules encode structural invariants that hold today and are part of the target;
 /// as each migration step lands a new structural property, add the rule that locks it in here so it
-/// cannot regress. Keep rules type-level (reflection over the production assembly); call-level invariants
+/// cannot regress. Rules are type-level (reflection over the production assembly); call-level invariants
 /// stay guarded by CodeQL/CodeRabbit and the pinning tests.
 /// </summary>
 public class ArchitectureConformanceTests
 {
+    private const string Root = "Jellyfin.Plugin.SSO_Auth";
+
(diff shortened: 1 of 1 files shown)


iderex (commented):
Pre-merge review — tighten the architecture-conformance rules (Refs #318).

Closes the five valid CodeRabbit findings from #322 (all-types namespace coverage incl. global-namespace rejection and exact-root/descendant matching; protected/protected-internal nested helpers flagged; !IsSealed so abstract unsealed helpers are caught; generic-arity stripping). Each #322 finding carries a documented reply pointing here.

Adversarial lenses: N/A — test-only, no production code; the change strengthens assertions only.

Review surfaces: CodeRabbit Review completed — zero inline, zero review-body, zero outside-diff findings. CI: all 17 checks green (584 tests; the stricter rules pass on the current tree).

Verdict: approve for merge.

PR #333, opened 2026-07-14
Add a CRLF/pretty-printed signed-SAML interop regression test

### LINKED ISSUE #120: Add a CRLF/pretty-printed signed-SAML interop regression test
Follow-up to the P2#9 XML hardening. The parser moved from `XmlDocument.LoadXml` to `XmlReader` (`DtdProcessing.Prohibit`), which normalizes EOL/attribute whitespace per XML 1.0, the form conformant IdP signing stacks digest against. Three independent adversarial-review probes (bypass, availability, SAML-domain) empirically confirmed a CRLF/pretty-printed, conformantly-signed response still validates through the new reader, but the unit suite only covers **compact single-line** fixtures.

Add a regression test that pins the interop property: a conformantly-signed response with real inter-element CRLF/indentation on the wire validates through `Response`. This requires care to avoid the `OuterXml`-escaping pitfall (CR in text content is escaped to `&#xD;`, which bypasses EOL normalization and would make the test pass for the wrong reason), encode the raw multi-line wire bytes, and sign over the form the SP validates. Locks against a future regression back to a non-normalizing parser.

iderex:
# What & why

Follow-up to the P2#9 XML hardening. The SAML parser moved from
`XmlDocument.LoadXml` to an `XmlReader` (`DtdProcessing.Prohibit`), which
normalizes CR/CRLF line endings to LF per XML 1.0 §2.11 before building the DOM - 
the form conformant IdP signing stacks compute their digest against. The SAML
suite otherwise covers only **compact single-line** fixtures, so nothing pinned
that a real-world pretty-printed response (inter-element CRLF + indentation on
the wire) still validates through `Response`. A regression to a non-normalizing
parser would break interop with conformant IdPs while passing every existing
test.

This adds `SamlCrlfInteropTests`, three cases against the **real**
signature-validation path in `SamlResponse` (genuinely signed fixture, no crypto
mocks):

- **`IsValid_WireWithRawCrlfBetweenElements_NormalizesAndValidates`**, RAW CRLF
  bytes (0x0D 0x0A) on the wire, encoded directly, validate against a signature
  computed over the LF form. The core interop property.
- **`IsValid_InterElementWhitespaceAltered_FailsClosed`**, altering the
  whitespace *content* (an extra space the IdP never signed) is a digest mismatch
  and is rejected. This proves the whitespace is genuinely signature-covered, so
  the case above passes because of line-ending **normalization** (CRLF ≡ LF), not
  because the parser ignores whitespace. Without this guard the positive test
  would be vacuous.
- **`IsValid_LineEndingCrAsCharacterReference_StillValidates`**, the same line
  ending written as a `&#xD;` character reference also validates.

Adds an additive `SamlTestFactory.CreateIndented` helper (builds and signs a
pretty-printed LF/indented response, reusing the existing signer). **No
production code changes.**

## Avoiding the `OuterXml`-escaping pitfall (the load-bearing correctness point)

The issue calls this out: serializing a CR-bearing DOM via `OuterXml` escapes CR
to the `&#xD;` character reference, which XML 1.0 §2.11 does **not** normalize. A
"CRLF" test built via `OuterXml` would therefore ship `&#xD;` instead of raw
CRLF and pass **for the wrong reason** (not exercising raw-CRLF normalization).
The positive test avoids this by encoding the **raw multi-line wire bytes**
directly (`Replace("\n", "\r\n")` on the serialized LF form, never
`Document.OuterXml` of a CR-bearing DOM) and asserting the wire carries raw
`\r\n` and **not** `&#xD;`. It signs over, and validates against, the exact
form the service provider reconstructs.

## Empirical finding (documented, not a defect)

Probing the real path showed the line-ending CR is normalized in **two** layers - 
the `XmlReader` (§2.11) *and* `SignedXml`'s C14N, so even the `&#xD;`
character-reference form validates (the third case pins this and explains why an
`OuterXml`-built test passes for the wrong reason). The interop is thus more
robust than the reader alone; this is a strengthening, not a weakening. **No
security impact:** only the line-ending *representation* is normalized; the
whitespace *content* stays fully signature-covered, verified empirically
(altering content, doubling newlines, tab-for-spaces, and tampering the NameID
all reject), and the second test locks that in.

## Adversarial-review gate

**Disposition: N/A**, this is a test-only addition that changes no product
runtime code (login path, crypto, config persistence, and release pipeline are
untouched; diff is `SSO-Auth.Tests/` only). The 4-lens gate covers changes to
those surfaces; here **the test is the verification**. Behavior was confirmed
empirically against the real validation path (see the finding above), and the
suite stays green.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. *(No product change; the added tests
      assert fail-closed rejection on altered signed whitespace.)*
- [x] No secrets logged; secrets are redacted on config export. *(N/A, test-only.)*
      config persistence, or the release pipeline. *(N/A — no such change; the
      SAML behavior was verified empirically, not modified.)*
- [x] Log inputs from the IdP are sanitized inline at the log call. *(N/A, test-only.)*

## Quality checklist

      testable unit. *(The factory helper reuses the existing `SignElement`.)*
      *(No new structural property; existing conformance tests pass.)*

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. *(Debug + Release, 0 warnings.)*
- [x] `dotnet test` is green. *(592 total, 0 failed, 3 new.)*
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. *(N/A, no web files touched.)*

## Notes

Test-only; no behavior, config, or security-posture change, so no README /
`providers.md` / wiki update is needed. The empirical finding above is a
documented nuance (stronger-than-assumed interop, no security weakening), not a
defect, so no separate issue is filed.

Closes #120

### DIFF
--- a/SSO-Auth.Tests/SamlCrlfInteropTests.cs
+++ b/SSO-Auth.Tests/SamlCrlfInteropTests.cs
@@ -1,0 +1,101 @@
+using Jellyfin.Plugin.SSO_Auth;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Interop regression tests pinning that a conformantly-signed, PRETTY-PRINTED SAML response — real
+/// inter-element line breaks and indentation on the wire, the shape production IdP signing stacks emit
+/// — still validates through <see cref="SamlResponse"/> (#120). Follow-up to the P2#9 XML hardening,
+/// which moved the parser from <c>XmlDocument.LoadXml</c> to an <see cref="System.Xml.XmlReader"/>
+/// (<see cref="System.Xml.DtdProcessing.Prohibit"/>): the reader normalizes CR/CRLF line endings to LF
+/// per XML 1.0 §2.11 before building the DOM. The rest of the SAML suite exercises only COMPACT
+/// single-line fixtures, so nothing pins that real-world CRLF/indentation interop survives; a
+/// regression to a non-normalizing parser would break it while passing every existing test.
+///
+/// The whitespace between elements is signature-covered — <c>PreserveWhitespace</c> keeps it and
+/// exclusive C14N includes it — so this shape is what actually exercises line-ending handling. The
+/// three cases together make the pin non-vacuous and guard the load-bearing correctness point the
+/// issue calls out (the <c>OuterXml</c>-escaping pitfall):
+///
+/// <list type="bullet">
+///   <item><see cref="IsValid_WireWithRawCrlfBetweenElements_NormalizesAndValidates"/> — RAW CRLF bytes
+///     (0x0D 0x0A) on the wire, encoded directly, validate against a signature computed over the LF
+///     form: the core interop property.</item>
+///   <item><see cref="IsValid_InterElementWhitespaceAltered_FailsClosed"/> — altering the whitespace
+///     CONTENT (an extra space the IdP never signed) is a digest mismatch and is rejected. This proves
+///     the whitespace is genuinely signature-covered, so the case above passes because of line-ending
+///     NORMALIZATION (CRLF ≡ LF), not because the parser ignores whitespace.</item>
+///   <item><see cref="IsValid_LineEndingCrAsCharacterReference_StillValidates"/> — the same line ending
+///     written as a <c>&amp;#xD;</c> character reference (what serializing a CR-bearing DOM via
+///     <c>OuterXml</c> emits) is exempt from §2.11 normalization, yet .NET's C14N in
+///     <see cref="System.Security.Cryptography.Xml.SignedXml"/> normalizes line-ending CR as well, so
+///     this form ALSO validates. A "CRLF" test built via <c>OuterXml</c> would therefore go green while
+///     shipping <c>&amp;#xD;</c> instead of raw CRLF — passing for the wrong reason. Pinning this is why
+///     the positive test encodes raw wire bytes and asserts the wire carries CR, not <c>&amp;#xD;</c>.
+///     No security impact: only the line-ending REPRESENTATION is normalized; the content stays
+///     covered (previous case).</item>
+/// </list>
+///
+/// All three drive the real signature-validation path in <see cref="SamlResponse"/> against a genuinely
+/// signed fixture (<see cref="SamlTestFactory.CreateIndented"/>), never a mock of the crypto.
+/// </summary>
+public class SamlCrlfInteropTests
+{
+    [Fact]
+    public void IsValid_WireWithRawCrlfBetweenElements_NormalizesAndValidates()
+    {
+        var fixture = SamlTestFactory.CreateIndented();
+
+        // The signed document serializes to an LF baseline with no stray CR — confirm that before
+        // reshaping it, so the CRLF on the wire below is introduced here, not already present.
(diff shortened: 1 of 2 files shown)


iderex:
@coderabbitai Thanks, confirming the review surfaces and dispositions before merge:

- **No actionable comments** in the review, and the Title / Description / Linked Issues / Out-of-Scope pre-merge checks all pass. Nothing to fix there.
- **Docstring Coverage warning (33.33% < 80%), declined, with reason.** The convention across this test suite is a detailed class-level `<summary>` stating the coverage intent (present here, covering all three cases and the load-bearing pitfall), paired with intention-revealing `[Fact]` names and why-comments inside each test. Every existing SAML test type follows exactly this shape, `SamlResponseTests`, `SamlAttackShapeTests`, `ValidateSamlTests`, `SamlResponseParsingTests` carry a class `<summary>` and no per-method XML-doc. Adding `<summary>` to each `[Fact]` would break that convention and add non-load-bearing ceremony against the minimal-code directive. The class and the additive `SamlTestFactory.CreateIndented` helper both carry full `<summary>` docs. No security or correctness property is affected, so I am keeping the tests as written.

iderex (commented):

**Build + test.** `dotnet build --no-restore --warnaserror` clean in Debug and Release (0 warnings). `dotnet test` green: 592 total, 0 failed (3 new). The new tests live in `SamlCrlfInteropTests`:
- `IsValid_WireWithRawCrlfBetweenElements_NormalizesAndValidates` — the core pin: raw CRLF bytes (0x0D 0x0A) encoded directly on the wire validate against a signature computed over the LF form.
- `IsValid_InterElementWhitespaceAltered_FailsClosed` — vacuity guard: altering the signed whitespace content is rejected, so the case above passes specifically because of line-ending normalization, not whitespace-blindness.
- `IsValid_LineEndingCrAsCharacterReference_StillValidates` — pins the `&#xD;` (OuterXml-escape) form and documents why the positive test encodes raw bytes instead.

**Correctness note.** Behavior was verified empirically against the real validation path (not reasoned): raw-CRLF and character-reference-CR both validate, while extra/altered whitespace, doubled newlines, tab-for-spaces, and a tampered NameID all reject. Line-ending CR is normalized in two layers — the XmlReader (§2.11) and SignedXml C14N — so the interop is more robust than the reader alone, with no security weakening (only the line-ending representation is normalized; content stays signature-covered). This corrected an initial assumption that the character-reference form would fail; the tests reflect the measured reality.

**Adversarial-review gate: N/A (documented).** Test-only addition; no product runtime code changes (login path, crypto, config persistence, release pipeline all untouched — diff is `SSO-Auth.Tests/` only). The 4-lens gate covers those surfaces; here the test is the verification, and it was validated empirically.

**CodeRabbit.** Review completed on head `944076f`; no actionable comments. Title / Description / Linked Issues / Out-of-Scope pre-merge checks pass. The single Docstring Coverage warning is declined with a documented reason (test-suite convention: class-level `<summary>` + intention-revealing names/why-comments; no per-`[Fact]` XML-doc anywhere in the suite).

**CI + Sonar.** All checks green; SonarCloud quality gate passed (0 new issues, 0 hotspots, 0% duplication). Copilot returned its quota-exhausted notice (no findings).

Overall verdict: approve to merge.

Issue #334, opened 2026-07-14
Harden IntervalGate against stale-sample re-admission


iderex:
## Problem

`IntervalGate.TryEnter` treats every timestamp older than the cursor as a wall-clock correction: it enters and re-anchors (the #246 self-heal). A caller whose captured `now` is merely *stale*, the thread was descheduled between reading the clock and entering, is indistinguishable from a correction, so it CASes the cursor backward and gets admitted, producing a second admission inside one interval. Surfaced by review on PR #331; the deterministic residual is pinned by `TryEnter_StaleOlderSample_AdmitsAndReAnchors_TheDocumentedFailOpenResidual`.

The impact today is bounded fail-open noise: one extra throttled action (a sweep or one log line) per race occurrence, and the next interval measured from the older sample. A stall, the dangerous direction for the capped authorize-state store, cannot occur, because a backward re-anchor only shortens the next wait. The same residual existed byte-for-byte in both inline predecessors; #331 deliberately preserved it (zero-behavior contract).

## Candidate fixes to evaluate

1. **Symmetric guard** (one line): suppress when `|now - last| < interval`, enter otherwise. Stale samples (milliseconds old) get suppressed; genuine large corrections (DST fall-back, NTP steps ≥ interval) still self-heal. Trade-off: a small backward correction (< interval) now delays the next admission by up to the correction size instead of firing immediately, still bounded by one interval, but a real behavior change at all adopted sites.
2. **Monotonic clock source** (`Environment.TickCount64` inside the gate): removes wall-clock corrections from the problem space entirely, so suppress-on-negative becomes safe and the race disappears. Trade-offs: platform differences in suspend/resume accounting, the callers currently share their wall-clock `now` with the throttled action (the state sweep needs wall time for expiry), and the characterization tests lose their injectable clock.

Either way the change is semantic, so it needs its own PR with the adversarial review gate (these throttles guard anonymous-path DoS amplification) and updated characterization tests. Evaluate after the remaining three throttle sites adopt the gate, so the fix lands once for all five.

Refs #318, #331

iderex:
[...]

PR #335, opened 2026-07-14
Adopt IntervalGate at the three remaining throttle sites

iderex:
# What & why

Second half of #318 step 2b (Refs #318), completing the throttle unification started in #331: the three remaining hand-rolled once-per-interval cursors now run behind `IntervalGate`.

- `SamlRequestCache.PruneExpired`, the outstanding-AuthnRequest sweep.
- `SsoRateLimiter.PruneStale`, the stale-counter sweep.
- `SSOController` capacity-full warning, the #246 log throttle on the refused-challenge branch.

Steady-state behavior is unchanged at all three sites (same suppression guard, same single-CAS-winner-per-interval, same boundary: elapsed == interval enters). Each site carries **one deliberate backward-clock delta, in the safe direction**, which is why these three were split out of #331 (whose contract was zero-change):

- The two sweeps had **no monotonic guard**: a backward wall-clock correction made the elapsed span negative, which read as "inside the interval", so the sweep **stalled** until the wall clock re-passed the stale cursor. They now self-heal, sweep once and re-anchor. For the SAML request cache this removes a latent availability hazard: a stalled sweep on a capped cache holds memory and, at the cap, refuses fresh registrations longer than necessary.
- The capacity warning had the monotonic term, but combined guard and CAS in one condition chain, so a backward step suppressed **without re-anchoring** (no warning until the clock re-climbed). It now fires one line and re-anchors.

All three throttled actions are memory reclamation or a single log line, never a grant/deny decision, so the worst case of any delta is one extra bounded action.

With the last hand-rolled cursor gone, the unification is locked in as a greppable invariant: an opengrep rule now bans the cursor-seed idiom (`= DateTime.MinValue.Ticks`) outside `IntervalGate.cs`, so a new inline throttle cannot quietly bypass the reviewed primitive.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.
- [x] The change adds no more code than the problem requires (net deletion:
      three inline guard+CAS blocks collapse into gate calls).
- [x] Architecture-conformance tests pass; the structural property this change
      completes (all interval throttling behind `IntervalGate`) is locked in as
      the `no-inline-interval-throttle-cursor` opengrep invariant, since a
      method-body property is greppable but not reflectable.

## Verification

- [x] `dotnet test` is green (603 tests).

## Notes

**Why no new tests in this diff:** the gate semantics the three sites now share, boundary, backward-clock self-heal, stale-sample residual, single concurrent winner, constructor validation, are already pinned deterministically by `IntervalGateTests` (#331). The per-site backward-clock deltas are confined to *when* an unobservable reclamation sweep runs (or one throttled log line fires) and are provably unreachable through the public API: the correlation and expiry decisions (`TryConsume`'s expiry check, `IsAllowed`'s window logic) are independent of sweep timing by design, and every constructible call sequence ends in the same visible state. The existing `SamlRequestCacheTests` / `SsoRateLimiterTests` cover the observable prune effects (cap-slot reclaim, entry expiry) and continue to pass unchanged.

**Backward-clock residuals:** the stale-sample re-admission residual documented on #331 applies to these sites identically and is tracked in #334, to be hardened once for all five sites.

**Adversarial review (4 lenses, refute-by-default): all PASS.**

- *Availability / mass-lockout*: the old stall-until-the-clock-re-climbs behavior was itself the mass-lockout-adjacent risk, a SAML request cache pinned at its cap refuses logins fail-closed; the self-heal un-pins it on the next call after any clock correction. All three gates receive provably positive intervals on every initialization path (static and instance), so a type-initialization failure is unreachable.
- *Security bypass / fail-open*: the fail-closed correlation, replay, and rate-limit predicates are untouched; the removed cursor fields leave no stragglers; the rate limiter's fail-open-at-cap invariant is preserved; log sanitization stays inline at the call.
- *Correctness / races*: per-site truth tables (guard polarity, elapsed == interval boundary, first call, CAS-winner selection, sample ordering) are exactly equivalent in the forward-clock domain; the only divergences are the three documented backward-clock deltas; the existing prune tests pass for the mechanistically right reasons.
- *Integration / lifetime / pipeline*: field-initialization order safe at both static and instance sites; using-directive changes exact; no test references the removed internals; the opengrep rule follows the file's generic-mode conventions with a unique id and zero findings on the current tree.

Accepted non-blocking residuals from the review: the opengrep tripwire is deliberately narrow (an implicit zero-seeded cursor would evade it, the conformance tests and review remain the primary defense; noted on #334 for the eventual hardening pass), and the rule was verified textually against the engine conventions, the PR's own "Enforce greppable invariants" check is the empirical run and is confirmed green before merge.

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -82,6 +82,12 @@ public class SSOController : ControllerBase
     // Throttles that sweep to one run per StatePruneInterval; the gate owns the atomic cursor. See Invalidate.
     private static readonly IntervalGate StatePruneGate = new(StatePruneInterval);
 
+    // Throttles the capacity-full warning (#246, CWE-400) to one line per StatePruneInterval: under a
+    // flood every refused challenge would otherwise emit a warning, amplifying the flood into unbounded
+    // log volume. The gate self-heals a backward wall-clock step (the hand-rolled predecessor stayed
+    // suppressed until the clock re-climbed past its cursor).
+    private static readonly IntervalGate CapWarnGate = new(StatePruneInterval);
+
     // One-time-use tracking for consumed SAML assertion IDs (replay protection).
     private static readonly SamlReplayCache SamlReplays = new SamlReplayCache();
 
@@ -105,11 +111,6 @@ public class SSOController : ControllerBase
     // computed once since the assembly version does not change at runtime.
     private static readonly string UserAgentString =
         $"Jellyfin-Plugin-SSO-Auth +{System.Diagnostics.FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).FileVersion} (https://github.com/iderex/jellyfin-plugin-sso)";
-
-    // Atomic cursor for throttling the capacity-full warning (#246, CWE-400): under a flood every refused
-    // challenge would otherwise emit a warning, amplifying the flood into unbounded log volume. Warn at
-    // most once per StatePruneInterval.
-    private static long _lastCapWarnTicks = DateTime.MinValue.Ticks;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="SSOController"/> class.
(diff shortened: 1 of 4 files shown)


iderex (commented):

**Internal adversarial review (4 lenses, refute-by-default), run on this exact diff before the PR was opened:** availability/mass-lockout, security-bypass/fail-open, correctness/races, integration/lifetime/pipeline — **all PASS**. Each lens re-verified the scoping claims against the actual code: per-site truth tables are bit-for-bit equivalent to the hand-rolled predecessors in the forward-clock domain; the only divergences are the three documented backward-clock deltas, each availability-positive (the old sweep stall on a capped SAML request cache was itself the mass-lockout-adjacent hazard); the fail-closed correlation/replay/rate-limit predicates and the fail-open-at-cap invariant are untouched; field-initialization order is safe at both static and instance sites, so no type-initialization failure path exists.

**Review dispositions:**
- CodeRabbit reviewed the final commit fe6a8ee ("Review completed") with no actionable comments on any surface — inline, review bodies, walkthrough all checked.
- The lens condition "confirm the opengrep rule empirically" is satisfied: the Enforce greppable invariants check ran the new no-inline-interval-throttle-cursor rule against the tree and passes (zero findings — the last cursor-seed idiom outside IntervalGate is gone in this very diff).
- Accepted limitation, documented on #334: the tripwire is deliberately narrow (an implicit zero-seeded cursor would evade it); the conformance tests and review remain the primary defense.
- Copilot posted only a quota-limit stub, no findings to answer.
- SonarCloud: quality gate passed, 0 new issues.

**Why no new tests:** the shared gate semantics are pinned by the IntervalGateTests shipped with #331; the per-site deltas are confined to when an unobservable reclamation sweep runs (or one throttled log line fires) and are provably unreachable through the public API — the reasoning is in the PR body, and the existing prune tests pass for the mechanistically right reasons per the correctness lens.

**Verification:** build `--warnaserror` clean; 603 tests green (also run independently by two lenses); all 17 checks pass.

Verdict: step 2b is complete with this merge — all five interval throttles run behind the one reviewed primitive, and the invariant is enforced both structurally (conformance tests) and textually (opengrep). Merging with a merge commit; branch will be deleted.

Issue #338, opened 2026-07-14
providers.md presents the SAML challenge route as the assertion URL


iderex:
providers.md (around lines 313-315) tells admins to register `sso/SAML/start/PROVIDER_NAME` as the "Redirect URI" / assertion-processing URL. That is the challenge route (where the login starts); the AssertionConsumerService URL the IdP must POST assertions to is `sso/SAML/post/PROVIDER_NAME` (or the legacy `sso/SAML/p/PROVIDER_NAME`). An IdP configured per the current doc POSTs the assertion to the challenge endpoint and the login fails.

Docs-only fix: correct the URL and state both accepted spellings. Surfaced by the URL-construction scoping for the #318 SsoUrlBuilder step.

Issue #337, opened 2026-07-14
ResolveChallengeNewPath matches /start/ as a substring


iderex:
`ResolveChallengeNewPath` decides the redirect_uri spelling with a substring test (SSOController.cs:381):

```csharp
var newPath = Request.Path.Value.Contains("/start/", StringComparison.InvariantCultureIgnoreCase);
```

A provider literally named `start` reached via the legacy route with a trailing slash (`/sso/OID/p/start/`) contains `/start/` as a substring, flips `newPath`, and mints the wrong redirect_uri spelling, the login then fails at the IdP's redirect_uri check. Contrast `OidcCallbackPath.RedirectSegment`, which deliberately does exact-segment matching so a provider named `redirect` cannot flip the callback spelling (#98).

Fix direction: exact-segment matching like `OidcCallbackPath`. Kept out of the #318 SsoUrlBuilder extraction because that step is behavior-preserving by contract; this change alters behavior for the (pathological) provider name and needs its own small gated PR with a characterization test.

Requires a provider named exactly `start`, hence low priority.

Refs #318

Issue #336, opened 2026-07-14
Validate provider names against URI-reserved characters at registration


iderex:
Provider names flow raw from config into the URLs the plugin hands to identity providers (OIDC redirect_uri, SAML AssertionConsumerServiceURL) and into route matching. Names containing URI-reserved characters break the round-trip today:

- `%` with an invalid percent-sequence (e.g. `prov%1`) fails route decoding outright; an encoded-looking name (`a%2Fb`) is decoded by routing and then misses the config dictionary lookup.
- `/` produces a callback URL no route can match, so the IdP's redirect dead-ends.
- Spaces and non-ASCII survive the trip but produce redirect_uris that differ from what an admin likely registered at the IdP.

Direction: reject URI-reserved characters in provider names at the Add/registration endpoints, fail closed at config time, zero byte change for every existing working name. Never start encoding the built URLs instead: the bytes currently produced are what IdPs already have registered, and re-encoding them would break every working deployment (that constraint is pinned by the raw-provider characterization tests in SsoUrlBuilderTests).

Surfaced by the URL-construction scoping for the #318 SsoUrlBuilder step.

Refs #318

PR #339, opened 2026-07-14
Extract the SSO URL construction into SsoUrlBuilder

iderex:
# What & why

Next step of the #318 architecture migration (Refs #318). Four controller sites hand-built the URLs this service provider hands to identity providers, the OIDC redirect_uri (challenge and token exchange), the SAML AssertionConsumerServiceURL, and the expected-ACS set the Recipient binding compares against, each repeating the `/sso/{protocol}/{spelling}/{provider}` shape inline. `SsoUrlBuilder` (internal static, pure) now owns that shape; the call sites keep resolving the base URL (`CanonicalBaseUrl` via `GetRequestBase`) and the `newPath` flag, because both are flow concerns, not string composition. `OidcCallbackPath` stays as-is and the builder delegates to it, its #98 exact-segment rationale and tests are already pinned.

These strings are validated byte-for-byte on the other side, RFC 6749 §4.1.3 redirect_uri equality at the IdP, and the SAML Recipient echo compared Ordinal by `SamlRecipientValidator`, so the extraction is behavior-preserving by contract: identical concatenation order and literals, provider appended raw with no re-encoding (the bytes IdPs have registered must not change), and the expected-ACS array keeps its exact order.

The structural payoff: `SamlExpectedAcsUrls` is now definitionally `{SamlAcsUrl(true), SamlAcsUrl(false)}`, so the Recipient validator can no longer drift from the challenge-time ACS string, that agreement used to be a convention across two hand-written string expressions ~950 lines apart.

Out-of-scope findings from the scoping pass are tracked separately: provider-name validation at registration (#336), the `/start/` substring hazard in `ResolveChallengeNewPath` (#337, deliberately not fixed here, this step is behavior-preserving), and the stale providers.md ACS documentation (#338).

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.
- [x] The change adds no more code than the problem requires (the spelling
      knowledge existed four times; it now exists once).
- [x] Architecture-conformance tests pass; the existing `*Builder` rules
      (internal + sealed/static) capture the new type automatically, so no new
      rule is needed.

## Verification

- [x] `dotnet test` is green (617 tests, three consecutive runs).

## Notes

**The byte evidence is `SsoUrlBuilderTests`** - 14 characterization pins covering both spellings per protocol, path-base retention, the callback-path echo (including a provider literally named `redirect` and a null path), the exact expected-ACS array and order, the raw-provider contract (space and umlaut pass through unencoded, characterized, not endorsed; rejecting such names at registration is #336), and the cross-leg agreement that the challenge and token-exchange constructions produce the same redirect_uri. The end-to-end suites assert only URL prefixes and parameter presence, so these unit pins carry the byte-level proof; zero existing tests were edited, which is itself the regression evidence.

**Adversarial review (4 lenses, refute-by-default): all PASS.**

- *Availability / mass-lockout*: symbolic byte-level derivation at all four call sites across every input class (path base, trailing slash, null base, null callback path, provider edge cases including `redirect`, space, umlaut, empty), no input exists where old and new differ by one byte; the expected-ACS set keeps order and cardinality, and the challenge-ACS membership guarantee is now structural.
- *Security bypass / fail-open*: acceptance set neither widened nor narrowed for any provider or base; the raw-provider bytes are identical at every site; `SamlRecipientValidator` (Ordinal + Trim) and `OidcCallbackPath` (#98 exact-segment) verifiably untouched; no new exposure via `InternalsVisibleTo`.
- *Correctness / edge cases*: interpolation lowers to the same `string.Concat` (no culture involvement); null flows identical; the 14 pins catch swapped segments, missing slashes, and wrong array order; zero existing-test edits confirmed.
- *Integration / conformance*: the existing `*Builder` rules capture the type automatically with no collateral flags; Debug and Release green under `--warnaserror`; declared out-of-scope sites provably untouched; opengrep unaffected. Cosmetic observation (mixed nullable-directive adoption across helpers), non-blocking, resolves naturally as the #318 migration touches each file.

### DIFF
--- a/SSO-Auth.Tests/SsoUrlBuilderTests.cs
+++ b/SSO-Auth.Tests/SsoUrlBuilderTests.cs
@@ -1,0 +1,105 @@
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Characterization tests pinning the exact bytes SsoUrlBuilder produces. These strings are validated
+/// byte-for-byte on the other side (the IdP's redirect_uri registration; the SAML Recipient echo,
+/// compared Ordinal by SamlRecipientValidator), so the pins here are the primary evidence that the
+/// #318 extraction changed nothing — the end-to-end suites assert only URL prefixes and parameter
+/// presence, not full bytes.
+/// </summary>
+public class SsoUrlBuilderTests
+{
+    private const string Base = "https://jf.example.com";
+
+    [Fact]
+    public void OidRedirectUri_ClassicRoute_UsesTheLegacySpelling()
+    {
+        Assert.Equal("https://jf.example.com/sso/OID/r/kc", SsoUrlBuilder.OidRedirectUri(Base, newPath: false, "kc"));
+    }
+
+    [Fact]
+    public void OidRedirectUri_NewPathRoute_UsesTheRedirectSpelling()
+    {
+        Assert.Equal("https://jf.example.com/sso/OID/redirect/kc", SsoUrlBuilder.OidRedirectUri(Base, newPath: true, "kc"));
+    }
+
+    [Fact]
+    public void OidRedirectUri_BaseWithPathBase_KeepsItAheadOfTheSsoSegment()
+    {
+        // CanonicalBaseUrl.Resolve hands over a trailing-slash-free base including the server's path
+        // base; the builder's "/sso/..." concatenation relies on exactly that contract.
+        Assert.Equal(
+            "https://jf.example.com/jellyfin/sso/OID/r/kc",
+            SsoUrlBuilder.OidRedirectUri("https://jf.example.com/jellyfin", newPath: false, "kc"));
+    }
+
+    [Theory]
+    [InlineData("/sso/OID/redirect/kc", "https://jf.example.com/sso/OID/redirect/kc")]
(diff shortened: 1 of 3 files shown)


iderex (commented):

**Internal adversarial review (4 lenses, refute-by-default), run on this exact diff before the PR was opened: all PASS.** The availability and bypass lenses derived the old and new strings symbolically across every input class and found no input where they differ by one byte; the expected-ACS set keeps order and cardinality, with the challenge-ACS membership now structural; SamlRecipientValidator and OidcCallbackPath verifiably untouched; the correctness lens confirmed the interpolation-to-concat equivalence and zero existing-test edits; the integration lens confirmed the *Builder conformance rules capture the type with no collateral flags and both build configurations green under warnings-as-errors.

**Review dispositions:** CodeRabbit reviewed the final commit ("Review completed") with no actionable comments on any surface — inline, review bodies, walkthrough all swept. Copilot posted only a quota-limit stub. SonarCloud quality gate passed. The lenses' one cosmetic observation (mixed nullable-directive adoption across helpers) is accepted — it resolves naturally as the #318 migration touches each file.

**Verification:** build `--warnaserror` clean; 617 tests green, three consecutive runs; the 14 SsoUrlBuilderTests pins are the byte-level evidence (the end-to-end suites assert prefixes only); zero existing tests edited. Out-of-scope findings from the scoping pass are filed as #336/#337/#338.

Verdict: a mechanically faithful, byte-identical extraction that makes the challenge/validator URL agreement structural. Merging with a merge commit; branch will be deleted.

Issue #346, opened 2026-07-14
Client-caused login rejections: introduce LoginOutcome and LoginStatusMapper


iderex:
The #318 login-outcome step. The controller rejects logins through four divergent mechanisms, `BadRequest(...)`, `ReturnError(status, message)`, `Problem(...)` (which produces a 500), and thrown `ArgumentException` (whose client-visible mapping depends on host middleware), and three divergent unknown/disabled-provider styles, one of which is a provider-enumeration oracle (a unique "No active providers found" wording distinguishes disabled from unknown, and on two endpoints the 400/500 split does the same).

Target: a closed `LoginOutcome` sum (`Success | Rejected(PublicReason) | Denied`) with deliberately no Error case, anything unexpected propagates as an exception and surfaces as a genuine 500, so a client-caused condition structurally cannot become a 500, and one `LoginStatusMapper` producing the uniform, deliberately uninformative bodies. A new rejection reason forces a new enum member and a mapper entry, never a fall-through. `Throttled(retryAfter)` joins the sum when the rate-limit filter step lands (the Retry-After header needs response access a pure mapper does not have).

Lands as two PRs: first the mechanism plus every byte-identical conversion (zero wire change, existing pins stay untouched), then the named behavior deltas: the six client-caused `Problem()` 500s become structured 4xx with already-public wordings, the thrown provider rejections become in-process 400s, and the disabled-vs-unknown oracle closes across all endpoints including the JSON-producing ones (their `BadRequest` guards route through the mapper so the content type cannot distinguish the cases either).

Refs #318

PR #347, opened 2026-07-14
Introduce LoginOutcome and LoginStatusMapper

iderex:
# What & why

First half of #346 (Refs #346, Refs #318). The controller rejects logins through four divergent mechanisms, `BadRequest`, `ReturnError`, `Problem()`, and thrown `ArgumentException`, held together by call-site convention. This PR introduces the closed `LoginOutcome` sum (`Success | Rejected(PublicReason) | Denied`) and the single `LoginStatusMapper`, and converts every exit whose response is **byte-identical on the wire**: thirteen sites that previously used `ReturnError` (whose `ContentResult`/text-plain shape the mapper's `Emit` reproduces exactly, same strings) or `Ok(authenticationResult)` (definitionally `new OkObjectResult(value)`, the same result the mapper emits).

The sum deliberately has **no Error case**, anything unexpected propagates as an exception and surfaces as a genuine 500, so a client-caused condition structurally cannot become a 500, and `Denied` deliberately carries **no reason at all**, so a denial cause cannot leak even by construction. A new rejection reason forces a new enum member and a mapper entry; the totality test makes a missing entry or a 5xx mapping fail before it ships. The uniform-rejection policy comment moves to the mapper, where the messages now live.

Out of scope, by design (they change bytes and get their own reviewed PR, the second half of #346): the client-caused `Problem()` 500s, the thrown provider rejections, the disabled-vs-unknown oracle closure (including the `[Produces(application/json)]` endpoints, whose `BadRequest` guards emit JSON-quoted bodies), and the corresponding enum members (`UnknownProvider`, `InvalidState`, `AccountLinkForbidden`), members exist only once a call site constructs them. The rate-limit 429 joins the sum with the filter step (Retry-After needs response access a pure mapper does not have). `ReturnError` itself stays with five remaining call sites: two dynamic-text rejections (their fixed-wording decision is a named follow-up), the capacity 500, the "No active providers found" site (second half), and the 429.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted (the mapper has no default-accept arm;
      impossible cases throw).
- [x] Log inputs from the IdP are sanitized inline at the log call (no log changes).

## Quality checklist

      testable unit.
- [x] The change adds no more code than the problem requires (the controller
      net-deletes; the three types are the typed belt the #318 design mandates).
- [x] Architecture-conformance tests pass; the `Mapper` suffix rules capture the
      new type automatically ("Outcome" is deliberately not a helper suffix — the
      abstract sum base would fail the sealed-or-static rule).

## Verification

- [x] `dotnet test` is green (632 tests, three consecutive runs) with **zero
      edits to existing assertions** — the byte-identical proof.

## Notes

**Adversarial review (4 lenses, refute-by-default): all PASS.** The bypass lens byte-diffed every mapper literal against the origin/main strings (including the apostrophe and trailing periods) and reconciled all 13 conversions against the old 16-site `ReturnError` inventory; availability confirmed no status a client retries on changes and the `[Produces(application/json)]` negotiation path is untouched for the identical `OkObjectResult`; correctness verified each mechanically converted line in context; integration confirmed the conformance rules capture the mapper and deliberately not the abstract sum base. The lenses' two notes are addressed in this PR: the closed-sum doc now names the record copy-constructor loophole explicitly, and a test constructs a foreign subtype through it to pin the unknown-subtype throw arm (misuse becomes a genuine 500, never a default-accept).

### DIFF
--- a/SSO-Auth.Tests/LoginStatusMapperTests.cs
+++ b/SSO-Auth.Tests/LoginStatusMapperTests.cs
@@ -1,0 +1,100 @@
+using System;
+using System.Linq;
+using Jellyfin.Plugin.SSO_Auth.Api;
+using MediaBrowser.Controller.Authentication;
+using Microsoft.AspNetCore.Mvc;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Pins the single outcome-to-HTTP translation (#318): every public rejection category maps to
+/// exactly one status and fixed plain-text body, byte-identical to the ReturnError responses the
+/// mapper replaced, and the closed sum admits no path to a client-caused 500.
+/// </summary>
+public class LoginStatusMapperTests
+{
+    [Fact]
+    public void Rejected_MapsEveryReasonToTheExactStatusAndBody()
+    {
+        // A table Fact rather than a Theory: PublicReason is internal, and a public Theory parameter
+        // may not be less accessible than the test method.
+        var expected = new (PublicReason Reason, int Status, string Body)[]
+        {
+            (PublicReason.SsoResponseInvalid, 400, "SSO response validation failed"),
+            (PublicReason.SamlResponseInvalid, 400, "SAML response validation failed"),
+            (PublicReason.PkceNotSupported, 400, "The identity provider does not advertise the required PKCE (S256) support."),
+            (PublicReason.PkceUnverifiable, 400, "The identity provider's PKCE (S256) support could not be verified."),
+        };
+
+        // The table itself must stay total: a new member without a row fails here.
+        Assert.Equal(Enum.GetValues<PublicReason>().Order(), expected.Select(e => e.Reason).Order());
+
+        foreach (var (reason, status, body) in expected)
+        {
+            var result = Assert.IsType<ContentResult>(LoginStatusMapper.ToActionResult(new LoginOutcome.Rejected(reason)));
+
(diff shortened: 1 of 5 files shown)


iderex (commented):

**Internal adversarial review (4 lenses, refute-by-default), run on this exact diff before the PR was opened: all PASS.** The bypass lens byte-diffed every mapper literal against the origin/main strings and reconciled the 13 conversions against the old 16-site ReturnError inventory (five sites remain, as the body states); availability confirmed no status a client retries on changes and the JSON-endpoint negotiation path is untouched; correctness verified each converted line in context plus the totality mechanics; integration confirmed the conformance coverage and that the deliberately-deferred surfaces are untouched. Both lens notes were addressed before the review round: the closed-sum doc names the record copy-constructor loophole, and the foreign-subtype test pins the unknown-subtype throw arm.

**Review dispositions:** CodeRabbit reviewed the final commit with no actionable comments on any surface — inline, review bodies, walkthrough all swept. Copilot quota stub only. SonarCloud quality gate passed. All checks green.

**Verification:** build `--warnaserror` clean; 632 tests green, three consecutive runs; zero edits to existing assertions — the byte-identical proof for all thirteen conversions.

Verdict: the closed outcome sum and the single mapper are in place with the wire provably unchanged; the second half (#346) converts the client-caused 500s and closes the provider oracle on top of this mechanism. Merging with a merge commit; branch will be deleted.

Issue #343, opened 2026-07-14
OidLink accepts a redeemable state for a disabled provider


iderex:
Surfaced by review on PR #342 (the behavior predates it, the consolidation preserved the endpoint's checks verbatim). `OidAuth` short-circuits on `config.Enabled` before redeeming a state, so a disabled provider can neither mint a session nor burn a state. `OidLink` only checks that the provider exists: a state validated shortly before an administrator disables the provider stays usable for up to its 15-minute lifetime to create an account link through the now-disabled provider (CWE-863, fail-open against the admin's intent).

Fix direction: check `config.Enabled` in `OidLink` before the redeem, mirroring `OidAuth`'s order (a disabled provider must not consume the state either), with a reject-path test. Two-line fix plus test, but it changes observable behavior on the linking path, so it lands as its own reviewed PR rather than inside the behavior-preserving consolidation that surfaced it.

Refs #318

Issue #401, opened 2026-07-16
Required check 'SonarCloud Code Analysis' no longer receives a status


iderex:
Since 2026-07-16 the SonarCloud integration no longer posts the `SonarCloud Code Analysis` commit status (observed on #390, #398, #399, only the CodeRabbit status appears), while the actual quality gates are OK (verified per PR via `api/qualitygates/project_status`). The "Protect main" ruleset requires that context, so every merge currently needs a documented admin bypass.

**Acceptance**
- [ ] Root-cause why the status stopped (integration binding lost when the ruleset was re-PUT without `integration_id`, app-side change, or analysis-mode change).
- [ ] Re-wire the required check so it is actually reported again (re-bind the integration in the ruleset UI, or replace the required context with the `Build and analyze` workflow job + `sonar.qualitygate.wait`).
- [ ] Merges no longer need the admin bypass.

PR #404, opened 2026-07-16
Reject a provider disabled mid-flight on every grant path

### LINKED ISSUE #380: Provider disabled mid-login/link still resolves and mints a session
Raised by review on #373. #373 made the three login/link-path config accessors fail closed when a provider is **deleted** mid-flight, but the same race exists for **disablement**: a provider disabled between the controller's `Enabled` gate (OidAuth/SamlAuth, and the #343 checks in OidLink/SamlLink) and the service's transactions still resolves, adopts/creates, and mints a session (or writes a link) with the provider's pre-disable settings. `TryGetLinks` checks existence, not `Enabled`.

Not a regression (the throwing indexer behaved identically), but #373's own rationale, "a missing provider must never default the login to valid", applies verbatim to a disabled one, and the fail-closed re-check machinery now exists at exactly the right places. #343's stated invariant is "disabling takes effect immediately"; today that is best-effort for an in-flight request.

Direction: have the three login/link-path guards treat `Enabled == false` like absence (reject / UnknownProvider). Note the residual `UseExistingLink`-without-migration path never re-reads config after the initial resolve, so a delete/disable after that single read still mints, document the #343 invariant as best-effort unless the lock is held through minting.

Refs #373, #343, #318

iderex:
## What & why

A provider disabled between the controller's `Enabled` gate and the service's transactions still resolved, adopted/created, wrote links, and minted a session with its pre-disable settings. #373 closed this race for **deletion**; its own rationale applies verbatim to **disablement**, and #343's "disabling takes effect immediately" was best-effort for an in-flight request.

`TryGetLinks` gains a `requireEnabled` flag; **every grant path passes `true`**, the resolve read, the adopt/create link write, the legacy migration (#155), and the manual link create, so a disabled provider is treated exactly like an absent one. **Removal deliberately stays exempt** (revocation must keep working on a disabled provider, or disable-then-clean-up breaks). The residual window is documented honestly at the resolve guard: the mint runs outside the config lock, so a disable after an arm's *last* guarded transaction still mints once.

Closes #380.

## Type of change
- [x] Security hardening (login/link path, fail-closed)

## Security checklist
- Fail-closed: every miss-branch throws (`AccountLinkForbiddenException` → the same caught 403 as the deleted-provider case) or returns `UnknownProvider` (→ the same `BadRequest` the controller yields steady-state, the #343 anti-enumeration property holds even in the race window).
- Gating rule respected: all four `requireEnabled: true` sites gate **grants**; the one `false` site is a **revocation**.
- No secrets, no new logging of untrusted data.

## Quality checklist
- Minimal: one flag, five named-argument call sites, comments carry the why.
- Tests: five new state-based tests, disabled-at-entry, two deterministic mid-flight flips (between the guarded transactions, injected via the user-manager stubs the flow consults exactly there), the create-arm no-write pin, and the remove-still-works control. Existing seeds now set `Enabled = true`, mirroring the controller's gate.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **758/758 green**.
- Full adversarial gate (sensitive diff): availability **PASS** (every `requireEnabled: true` site sits behind an existing controller gate; unlink/list/unregister keep working on disabled providers; all guards map to a caught 403, no lockout surface). bypass/correctness/integration first returned **NEEDS_IMPROVEMENT** with one convergent finding, the manual link create originally kept `requireEnabled: false` under a wrong "RequiresElevation/manageability" rationale, leaving the same window open on a grant write, **fixed** (flipped to `true`, comments corrected, no-write test added); re-run: **PASS / PASS / PASS**.

## Notes
Delivery item from the integration lens: one line on provider-disable semantics goes onto the Security-Model wiki page with this merge.

### DIFF
--- a/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
@@ -40,6 +40,7 @@ public class CanonicalLinkServiceTests
     {
         var (service, _, users, _) = Build(c => c.OidConfigs["kc"] = new OidConfig
         {
+            Enabled = true,
             CanonicalLinks = new SerializableDictionary<string, Guid> { ["sub-1"] = Existing },
         });
         users.GetUserById(Existing).Returns(UserNamed("alice", Existing));
@@ -53,7 +54,7 @@ public class CanonicalLinkServiceTests
     [Fact]
     public async Task ResolveOrCreateAsync_NoLinkNoAccount_CreatesAndLinksOnTheSubjectKey()
     {
-        var (service, cfg, users, _) = Build(c => c.OidConfigs["kc"] = new OidConfig());
+        var (service, cfg, users, _) = Build(c => c.OidConfigs["kc"] = new OidConfig { Enabled = true });
         var created = UserNamed("alice", Other);
         users.GetUserByName("alice").Returns((User?)null);
         users.CreateUserAsync("alice").Returns(created);
@@ -69,7 +70,7 @@ public class CanonicalLinkServiceTests
     [Fact]
     public async Task ResolveOrCreateAsync_ExistingAccountAndAdoptionAllowed_AdoptsAndLinks()
     {
-        var (service, cfg, users, _) = Build(c => c.OidConfigs["kc"] = new OidConfig());
+        var (service, cfg, users, _) = Build(c => c.OidConfigs["kc"] = new OidConfig { Enabled = true });
         users.GetUserByName("alice").Returns(UserNamed("alice", Existing));
         users.GetUserById(Existing).Returns(UserNamed("alice", Existing));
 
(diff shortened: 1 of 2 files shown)


iderex:
Pre-merge: all 16 checks green; CodeRabbit review completed with zero findings; SonarCloud quality gate OK with zero failing conditions (verified via api/qualitygates/project_status for PR 404). Full adversarial gate documented in the PR body, availability PASS, and the one convergent bypass/correctness/integration finding (the create-arm window) fixed and re-run to PASS/PASS/PASS. 758/758 tests. Security-Model wiki updated with the provider-disable semantics. Required-status context "SonarCloud Code Analysis" still is not posted (#401), so merging via the documented admin bypass.

Issue #407, opened 2026-07-16
Harden BuildScopeString against null/empty scope elements in a non-null OidScopes


iderex:
Follow-up from #368 (adversarial review, correctness + integration lenses both flagged this and recommended a separate issue rather than expanding that PR).

`BuildScopeString` (SSO-Auth/Api/SSOController.cs) normalizes a null `OidScopes` **array** to empty, but null / empty / whitespace **elements** inside a non-null array still render as empty via `string.Join`, producing a malformed scope string, e.g. `OidScopes = ["email", null]` → `"openid profile email "` (trailing separator), `["", "groups"]` → `"openid profile  groups"` (doubled separator).

This is **not a regression** (identical to the pre-#368 challenge-side behavior) and is **non-impactful today**, the scope value is not compared during token redemption or id_token validation. It only matters if the OID/Add API or the admin form can persist blank/null scope entries.

## Fix
Filter blank entries in the one shared helper:
```csharp
internal static string BuildScopeString(OidConfig config)
    => string.Join(" ", (config.OidScopes ?? Array.Empty<string>())
        .Where(s => !string.IsNullOrWhiteSpace(s))
        .Prepend("openid profile"));
```
Add test cases for `["email", null]`, `["", "groups"]`, and an all-blank array (→ `"openid profile"`).

Refs #368.

Issue #350, opened 2026-07-14
Add endpoints store a null provider body, breaking the config-page save


iderex:
## Summary

The OpenID/SAML `Add` endpoints upsert the posted body unconditionally, so a `POST OID/Add/{provider}` (or `SAML/Add`) with an empty/`null` JSON body stores a `null` provider object in the config map:

```csharp
// SSOController.OidAdd / SamlAdd
configuration.OidConfigs[provider] = config;   // config may be null -> a null entry is stored
```

Once a `null` provider entry is present, the config-page save path dereferences it and throws:

- `ServerManagedFields.Preserve(PluginConfiguration, PluginConfiguration)` iterates the live providers and, for a matching incoming key, calls the per-provider overload, `incoming.CanonicalLinks = live.CanonicalLinks` NREs when either side is `null`.

This is a **pre-existing latent defect**, not introduced by the ProviderConfigStore extraction (#318 / PR #348): the pre-refactor `SSOPlugin.PreserveServerManagedFields` had the identical `incomingProvider.CanonicalLinks = kvp.Value.CanonicalLinks` dereference. PR #348 is a deliberately zero-behavior move and faithfully preserves it; it is filed here so the fix gets its own change with its own review rather than being smuggled into the refactor. (The audit path CodeRabbit also flagged, `ProviderConfigStore.CollectInsecureOptions` → `OidcInsecureToggles.Enabled(kvp.Value)`, is already null-safe: `Enabled` returns empty for a `null` config, so only the preservation path actually throws.)

## Impact

- Requires `RequiresElevation` (admin) to store the `null` entry, so it is admin-only and self-inflicted, not reachable by an unprivileged attacker.
- After a `null` entry is stored, the admin config page can no longer be saved (the save throws a 500), which is an availability/foot-gun for the admin.

## Proposed fix (fail-closed, its own review)

Reject a `null` body at the `Add` door rather than storing it, the fail-closed choice, consistent with the other Add-endpoint gates (`RejectInvalidBaseUrlOverride`, `RejectInvalidSamlCertificate`). Add a regression test that `OID/Add` / `SAML/Add` with a `null` body is rejected and nothing is persisted, and a `ServerManagedFields.Preserve` test that a `null` provider entry does not NRE. Touches config persistence, so it passes the `/security-review` gate.

## Labels / milestone

`bug`, `area: config`, `priority: low`, milestone `P2 — Security parity with V2` (availability/fail-closed on the config write path).

Issue #199, opened 2026-07-12
SAML callback: map malformed-input exceptions to a clean rejection, not HTTP 500


iderex:
Follow-up from the #137 adversarial-review gate (correctness + bypass + saml-reviewer, all LOW/robustness, fail-closed, no auth bypass).

`Response.IsValid()` now catches the malformed-signature exceptions on its own path (#137). But other untrusted-input paths in the SAML callback can still surface an unhandled HTTP 500 (with a stack trace) to an unauthenticated caller:
- The `Response` constructor throws `FormatException` on a non-base64 `SAMLResponse`, and `XmlException` on a malformed/DOCTYPE body (the DOCTYPE throw is intentional and tested, the point is it should surface as a clean 4xx page, not a 500).
- `GetSignatureAlgorithm()` (the diagnostic used in the failure-log path) can throw from `LoadXml` on a malformed signature.
- `SamlPost`/`SamlAuth`/`SamlLink` only wrap the config lookup in `try/catch (KeyNotFoundException)`; anything else escapes the action.

Impact: an attacker can drive 500s / log noise at will. Not a bypass (every faulting path denies login), but a robustness/availability and log-hygiene gap.

Fix direction: wrap the parse + validate in the SAML callback actions (or a small helper) and map any malformed-input exception to the same generic "SAML response validation failed" rejection the invalid path already returns. Add tests: non-base64 body, malformed XML body, and a malformed-signature body all return a clean non-200 (not 500), no stack trace. This is browser/endpoint-level, so it also belongs on the E2E checklist until the endpoint-test layer (#192) can assert it.

Refs #137.

PR #408, opened 2026-07-16
Normalize null OidScopes so the challenge does not 500

### LINKED ISSUE #368: Normalize null OidScopes: unhandled 500 on the challenge, null-padded scopes on the callback
A provider stored without scopes (possible via the `OID/Add` API, see also #350) behaves differently per endpoint:

- Challenge: `config.OidScopes.Prepend("openid profile")` (SSO-Auth/Api/SSOController.cs:275) throws on null, an unhandled 500 on an anonymous endpoint.
- Callback: `config.OidScopes == null ? new string[2] : config.OidScopes` (:444) tolerates it by padding with two null entries, which `string.Join` turns into a scope string with trailing separators.

The comment (:437-441) records the asymmetry as "deliberately preserved" from the extraction, but preserved-as-found is not a rationale anyone can act on: `new string[2]` reads like a bug (why 2? why nulls?), and the challenge-side crash is exactly the unhandled-500 class we fixed elsewhere (#199, #206).

Proposal: normalize once, `var scopes = config.OidScopes ?? Array.Empty<string>();` shared by both `CreateOidcClient` call sites (or normalize at the OID/Add write path), delete `new string[2]`, and pin the callback scope string with a test. Wire note for the PR: the callback client's scope string changes from `"openid profile  "` (trailing separators) to `"openid profile"`; the scope value is not compared during token redemption, but the test should document the change.

iderex:
# What & why

A provider stored without scopes (reachable via the `OID/Add` API, #350) left `OidConfig.OidScopes` null, and the two OpenID client-building sites handled it differently:

- **Challenge** (anonymous endpoint): `config.OidScopes.Prepend("openid profile")` threw a `NullReferenceException` → an unhandled **HTTP 500** on an anonymous endpoint, the same unhandled-500 class fixed in #199/#206.
- **Callback**: `config.OidScopes == null ? new string[2] : config.OidScopes` padded the scope string with two null entries → `"openid profile  "` with trailing separators.

Both sites now route through one shared helper, `BuildScopeString`, that normalizes a null array to empty. The challenge proceeds with the base `openid profile` scopes instead of crashing, and both sites emit the same clean string. The populated-scopes path is byte-identical to before.

Closes #368

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. — Unchanged. The old challenge-side crash was not a fail-closed control; the adversarial review confirmed the monotonic privilege model means base-only scopes can only reduce or leave access unchanged, and a role allow-list still fails closed when its claim is absent.
- [x] No secrets logged; secrets are redacted on config export., No logging changed; removing the 500 also stops leaking a stack trace.
      config persistence, or the release pipeline. — 4-lens adversarial review, all PASS.
- [x] Log inputs from the IdP are sanitized inline at the log call., No new log calls.

## Quality checklist

      testable unit. — The two divergent scope-string sites are unified into `BuildScopeString`.
      why, not what) and matches the target architecture (#318). — `internal static` pure helper on the thin controller, consistent with the sibling helpers.
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. — No new structural property; existing rules pass.

## Verification

- [x] `dotnet test` is green., 761 passed, 0 failed (3 new).

## Notes

Null/empty/whitespace **elements** inside a non-null `OidScopes` array (e.g. `["email", null]`) still render as empty via `string.Join`, a separate, non-impactful latent case (unchanged from the old path; scope is not validated downstream), tracked as a follow-up in #407 with a `.Where(s => !string.IsNullOrWhiteSpace(s))` fix so this PR stays scoped to the null-array bug the issue describes.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerScopeStringTests.cs
+++ b/SSO-Auth.Tests/SSOControllerScopeStringTests.cs
@@ -1,0 +1,41 @@
+using System;
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Pins <see cref="SSOController.BuildScopeString"/> — the shared OpenID scope-string builder used by
+/// both the challenge and the callback client. Guards #368: a provider stored without scopes leaves
+/// <see cref="OidConfig.OidScopes"/> null, which previously threw an unhandled 500 on the anonymous
+/// challenge (<c>OidScopes.Prepend</c> on null) and null-padded the callback scope string
+/// (<c>new string[2]</c> → trailing separators). The builder normalizes null to empty so both sides
+/// emit the same clean, base-prefixed string.
+/// </summary>
+public class SSOControllerScopeStringTests
+{
+    [Fact]
+    public void NullScopes_YieldBaseScopesOnly_NoThrow()
+    {
+        var config = new OidConfig { OidScopes = null };
+
+        Assert.Equal("openid profile", SSOController.BuildScopeString(config));
+    }
+
+    [Fact]
+    public void EmptyScopes_YieldBaseScopesOnly()
+    {
+        var config = new OidConfig { OidScopes = Array.Empty<string>() };
+
+        Assert.Equal("openid profile", SSOController.BuildScopeString(config));
+    }
+
+    [Fact]
+    public void ConfiguredScopes_ArePrefixedWithBaseScopes()
+    {
+        var config = new OidConfig { OidScopes = new[] { "email", "groups" } };
+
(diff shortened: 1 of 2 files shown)


iderex:
@coderabbitai review

iderex:
Pre-merge, all gates green: build and tests (761 passing, 3 new), CodeQL, zizmor, dependency-review, and CodeRabbit (Review completed on the re-request, no actionable comments; the first run was a rate-limit skip). The 4-lens adversarial review passed on every lens (the old challenge-side 500 was not a fail-closed control; base-only scopes are strictly access-reducing under the monotonic privilege model; a role allow-list still fails closed). The SonarCloud quality gate is red only because the sonarcloud.io API is still in its multi-hour outage and cannot compute, non-blocking, re-analyzed on push to main once the service recovers. Merging via admin bypass with that reason documented.

Issue #189, opened 2026-07-12
Make provider secrets write-only in JSON responses (OidSecret, SAML cert)


iderex:
Follow-up from the #157 review (integration lens, INFO). With #157, CanonicalLinks is withheld from JSON responses, but OidSecret and the SAML certificate/signing material are still serialized to elevated callers by OID/Get, SAML/Get and Jellyfin core's getPluginConfiguration (all RequiresElevation). That is unchanged from before and admin-only, but it means the client (and anyone who can read an admin's browser session/HAR, a proxy log, or a shared screen) sees the plaintext secret every time the config page loads.

jellyfin-plugin-sso-V2 additionally [JsonIgnore]s the secret fields and handles them write-only: the config UI shows a placeholder / accepts a new value but never receives the stored one, and the save path re-injects the live secret when the field comes back empty (the same PreserveServerManagedFields pattern #157 introduced for CanonicalLinks).

Acceptance criteria:
- [ ] OidSecret and SAML signing/cert material are [JsonIgnore] (kept in the config XML, withheld from every JSON response).
- [ ] The admin UI supports setting a new secret and leaving it unchanged (empty field preserves the live value); a save with an empty secret does not blank the stored one (fail closed).
- [ ] PreserveServerManagedFields extended to the secret fields; a config PUT cannot read back a secret.
- [ ] Tests: JSON omits the secret; a save with an empty secret preserves the live value; a save with a new secret updates it.

Complements #158 (secrets at rest, encryption); this is the not-in-transit-to-the-client half. Refs #157, #158.

PR #409, opened 2026-07-16
Reject a null provider body at the Add endpoints

### LINKED ISSUE #350: Add endpoints store a null provider body, breaking the config-page save

iderex:
# What & why

The admin `OID/Add` and `SAML/Add` endpoints upserted the posted `[FromBody]` object unconditionally. Nullable reference types are off for this project, so ASP.NET does not auto-reject an empty / `"null"` JSON body, it binds to a **null** config object, and the endpoint stored a `null` entry in the provider map (`OidConfigs[provider] = null`). A null entry then NREs the config-page save (`ServerManagedFields.Preserve` dereferences it) → HTTP 500, and the admin can no longer save the config page. Admin-only (`RequiresElevation`), so self-inflicted, not attacker-reachable.

Fix, two layers:

- **Reject the null body at the door**, `RejectNullProviderBody` is the first line of both Add endpoints, before any mutation, mirroring the sibling `RejectInvalid*` gates. Nothing is persisted on rejection. With null rejected first, the following `config?.` guards simplify to `config.`.
- **Defense-in-depth**, the per-provider `ServerManagedFields.Preserve` overloads early-return when either side is null, so a legacy null entry (or a null `existing` during an Add) no longer NREs the save and can be overwritten to heal the wedged entry.

Closes #350

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. — The null body is rejected fail-closed; the adversarial review confirmed a null entry can never authenticate (login treats a null provider config as "no provider" → reject).
- [x] No secrets logged; secrets are redacted on config export., Unchanged. The `Preserve` null-guard skips only genuinely null entries (which carry no secret); the #189 blank-means-keep secret rule is untouched for every non-null config.
      config persistence, or the release pipeline. — 4-lens adversarial review, all PASS.
- [x] Log inputs from the IdP are sanitized inline at the log call., No new log calls.

## Quality checklist

      testable unit. — One `RejectNullProviderBody` gate reused by both endpoints.
      why, not what) and matches the target architecture (#318). — Gate mirrors the existing `RejectInvalid*` pattern; the `Preserve` guard sits with the shared re-injection rule.
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. — No new structural property; existing rules pass.

## Verification

- [x] `dotnet test` is green., 764 passed, 0 failed (3 new).

## Notes

The bypass lens noted one admin-only, non-blocking behavior: on the config-page save path, a *deliberately* null-posted provider entry that previously NRE'd (accidental fail-closed) now early-returns and the null entry persists, replacing the real provider. This is reachable only by a malicious/compromised admin who can already delete or omit that provider via the wholesale replace or `OID/Del`, so it grants no new capability and is unreachable by any unprivileged user. A legitimate stale-client save posts the real object (secret withheld/blank), not null, so it still hits the #189 `ResolveUpdatedSecret` path.

### DIFF
--- a/SSO-Auth.Tests/ConfigPreservationTests.cs
+++ b/SSO-Auth.Tests/ConfigPreservationTests.cs
@@ -71,6 +71,26 @@ public class ConfigPreservationTests
     }
 
     [Fact]
+    public void Preserve_NullProviderEntry_DoesNotThrow()
+    {
+        // A malformed Add before #350 could store a null provider entry; preserving with a null on
+        // either side must skip it, not NRE the whole config-page save.
+        var live = new PluginConfiguration();
+        live.OidConfigs["null-in-live"] = null;
+        live.OidConfigs["null-in-incoming"] = new OidConfig { CanonicalLinks = new SerializableDictionary<string, Guid> { ["sub"] = User } };
+        live.SamlConfigs["saml-null-in-live"] = null;
+
+        var incoming = new PluginConfiguration();
+        incoming.OidConfigs["null-in-live"] = new OidConfig();
+        incoming.OidConfigs["null-in-incoming"] = null;
+        incoming.SamlConfigs["saml-null-in-live"] = new SamlConfig();
+
+        var exception = Record.Exception(() => ServerManagedFields.Preserve(incoming, live));
+
+        Assert.Null(exception);
+    }
+
+    [Fact]
     public void CanonicalLinks_AreOmittedFromJson_ButKeptInXml()
     {
         var config = new OidConfig
--- a/SSO-Auth.Tests/SSOControllerAddTests.cs
+++ b/SSO-Auth.Tests/SSOControllerAddTests.cs
@@ -42,6 +42,27 @@ public class SSOControllerAddTests
     }
 
     [Fact]
+    public void OidAdd_NullBody_Throws_AndDoesNotPersist()
+    {
+        var harness = new SsoControllerHarness();
+
+        Assert.Throws<ArgumentException>(() => harness.Controller.OidAdd("keycloak", null));
+
+        // Fail-closed: a null [FromBody] is rejected at the door, so no null entry is stored (#350).
+        Assert.False(SSOPlugin.Instance.ReadConfiguration(c => c.OidConfigs.ContainsKey("keycloak")));
+    }
+
+    [Fact]
+    public void SamlAdd_NullBody_Throws_AndDoesNotPersist()
+    {
+        var harness = new SsoControllerHarness();
+
+        Assert.Throws<ArgumentException>(() => harness.Controller.SamlAdd("adfs", null));
+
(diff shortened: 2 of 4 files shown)


iderex:
Pre-merge, all gates green: build and tests (764 passing, 3 new), CodeQL, zizmor, dependency-review, the SonarCloud quality gate (green; the service has recovered), and CodeRabbit (Review completed, no actionable comments). The 4-lens adversarial review passed on every lens, the null body is rejected fail-closed, a null entry can never authenticate, and the #189 secret rule is untouched for non-null configs. Merging on green.

PR #413, opened 2026-07-16
Persist the default provider in the single session-mint write

### LINKED ISSUE #391: Collapse the double UpdateUserAsync in the session mint when a default provider is set
From #390 review (CodeRabbit, Trivial). `SessionMinter.MintAsync` calls `UpdateUserAsync` twice when `DefaultProvider` is set: once after the permission/avatar updates, once after setting `AuthenticationProviderId`. Verbatim from the pre-extraction `Authenticate`. Set `AuthenticationProviderId` before the single `UpdateUserAsync` to persist in one write. Behavior-observable (write count), so a small standalone change, not folded into the extraction. Pinned by `MintAsync_DefaultProviderSet_SetsItAndWritesTheUserTwice`, flip that assertion to `Received(1)` with the fix.

Refs #389

iderex:
# What & why

`SessionMinter.MintAsync` wrote the user twice when a default provider was configured: once after the permission/avatar updates, then again only to persist `AuthenticationProviderId`, verbatim from the pre-extraction `Authenticate`.

Set `AuthenticationProviderId` before the single `UpdateUserAsync` so it persists in one write. `authRequest` never reads the field and `AuthenticateDirect` still runs after the write, so the final persisted state is identical; only the redundant second round-trip is gone, and the one write is now atomic (permissions, avatar, and the provider id land together instead of in two steps, the adversarial review noted this removes the old partial-persist window).

Closes #391

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. — Unchanged; if the single write throws, `AuthenticateDirect` never runs and no session is minted (the review confirmed no fail-open path and no new timing window).
- [x] No secrets logged; secrets are redacted on config export., Unchanged; the same log line moved with the block.
      config persistence, or the release pipeline. — 4-lens adversarial review, all PASS; final persisted state byte-identical, only the write count drops 2→1.
- [x] Log inputs from the IdP are sanitized inline at the log call., No new log calls; `DefaultProvider` is an admin-configured value.

## Quality checklist

      testable unit.
- [x] The change adds no more code than the problem requires., A block move plus one deleted write.
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. — No new structural property.

## Verification

- [x] `dotnet test` is green., 781 passed, 0 failed; the default-provider test now asserts `Received(1)`.

## Notes

Behavior-observable only in the write count; the final persisted user state (permissions, preferences, avatar, `AuthenticationProviderId`) is identical, and `AuthenticationProviderId` is still committed before `AuthenticateDirect`.

### DIFF
--- a/SSO-Auth.Tests/SessionMinterTests.cs
+++ b/SSO-Auth.Tests/SessionMinterTests.cs
@@ -167,9 +167,14 @@ public class SessionMinterTests
         users.GetUserById(UserId).Returns(user);
         sessions.AuthenticateDirect(Arg.Any<AuthenticationRequest>()).Returns(new AuthenticationResult());
 
+        // Capture the id at the write boundary, not off the shared user after MintAsync returns —
+        // otherwise the assertion passes even if the id is set AFTER the write, defeating the point (#391).
+        string? providerAtWrite = null;
+        users.UpdateUserAsync(Arg.Do<User>(u => providerAtWrite = u.AuthenticationProviderId)).Returns(Task.CompletedTask);
+
         await minter.MintAsync(Params(defaultProvider: "SSO-Auth"), () => "203.0.113.7");
 
-        Assert.Equal("SSO-Auth", user.AuthenticationProviderId);
+        Assert.Equal("SSO-Auth", providerAtWrite);
         await users.Received(1).UpdateUserAsync(user);
     }
 }
--- a/SSO-Auth.Tests/SessionMinterTests.cs
+++ b/SSO-Auth.Tests/SessionMinterTests.cs
@@ -157,13 +157,11 @@ public class SessionMinterTests
     }
 
     [Fact]
-    public async Task MintAsync_DefaultProviderSet_SetsItAndWritesTheUserTwice()
+    public async Task MintAsync_DefaultProviderSet_SetsItInASingleWrite()
     {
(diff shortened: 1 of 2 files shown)


iderex:
@coderabbitai review

iderex:
Addressed the outside-diff-range comment on `SessionMinterTests.cs` (160-173) in 04aba8f: the assertion now captures `AuthenticationProviderId` inside the `UpdateUserAsync` substitute via `Arg.Do` and asserts on that captured value, so the test pins that the id is set *at* the single write rather than merely present on the returned user, it now fails if the assignment ever moves back after the write. Good catch.

iderex:
@coderabbitai review

iderex:
Pre-merge, all required checks green (build, 781 tests, CodeQL, zizmor, dependency-review, SonarCloud quality gate), 4-lens adversarial review all PASS, and the final commit 04aba8f reads CodeRabbit "Review completed". CodeRabbit raised one outside-diff-range comment (strengthen the default-provider test to assert AuthenticationProviderId at the write boundary, not off the returned user); it is applied verbatim in 04aba8f (the capture via `Arg.Do` + `Assert.Equal("SSO-Auth", providerAtWrite)`, verified against the committed file). CodeRabbit's review body still renders against the pre-fix commit 3f034750, but the finding is resolved in committed code and its final-commit status is Review completed. Merging on green.

PR #419, opened 2026-07-16
Store the avatar with a real dotted file extension

### LINKED ISSUE #384: Avatar profile image is stored without a file extension dot (profilepng)
From the #375 review (rekamer LOW, pre-existing, carried verbatim, correctly not fixed in the extraction). `AvatarService.TrySetAsync` stores the image as `Path.Combine(userDataPath, "profile" + extension)`, but `AvatarContentType.TryResolveExtension` returns bare extensions without a leading dot (`"png"`, not `".png"`), so the stored filename is `profilepng` / `profilejpeg`. Effect: the profile image path has no real extension, so downstream content-type inference (Jellyfin image serving) may fall back to `application/octet-stream`. Not a security issue (the content-type allow-list already excludes active content), and identical on `main`. Fix: insert the dot (`"profile." + extension`) or have `TryResolveExtension` return dotted extensions, with a test.

Refs #375

iderex:
## What & why

`AvatarContentType.TryResolveExtension` returned bare extensions (`"png"`), and the store appends them as `"profile" + extension`, so avatars were written as `profilepng` / `profilejpeg` with no real extension, leaving Jellyfin's image serving to fall back to `application/octet-stream` on content-type inference (surfaced in the #375 review, carried verbatim through the AvatarService extraction).

Return the extension **with its leading dot** (`.png`) at the source, so the value is a real extension the store appends directly and the dotless-filename footgun is gone. The allow-list and the media type carried to `SaveImage` are unchanged.

Closes #384.

## Type of change
- [x] Bug fix (avatar store path)

## Security checklist
- No security-posture change: the raster-only allow-list is untouched (still rejects `image/svg+xml`), only the returned extension now carries the dot. The media type sent to `SaveImage` is unchanged.

## Quality checklist
- Root-cause fix at the resolver rather than a call-site patch; the dotted result is pinned in `AvatarContentTypeTests` and the avatar happy-path test now asserts `profile.png`.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **803/803 green**.
- Login-path file, but a trivial extension-string correctness fix with no logic or flow change, fully covered by tests, no separate adversarial gate warranted.

### DIFF
--- a/SSO-Auth.Tests/AvatarContentTypeTests.cs
+++ b/SSO-Auth.Tests/AvatarContentTypeTests.cs
@@ -10,16 +10,19 @@ namespace Jellyfin.Plugin.SSO_Auth.Tests;
 public class AvatarContentTypeTests
 {
     [Theory]
-    [InlineData("image/png", "png")]
-    [InlineData("image/jpeg", "jpeg")]
-    [InlineData("image/jpg", "jpeg")]
-    [InlineData("image/gif", "gif")]
-    [InlineData("image/webp", "webp")]
-    [InlineData("IMAGE/PNG", "png")]
-    public void AllowedRasterTypes_Resolve(string mediaType, string expected)
+    [InlineData("image/png", ".png")]
+    [InlineData("image/jpeg", ".jpeg")]
+    [InlineData("image/jpg", ".jpeg")]
+    [InlineData("image/gif", ".gif")]
+    [InlineData("image/webp", ".webp")]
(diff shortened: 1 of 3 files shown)


iderex:
Pre-merge, all gates green. 16/16 checks pass, SonarCloud quality gate OK, 803/803 tests. CodeRabbit review completed with zero findings. Trivial extension-string correctness fix, no logic change. Merging on green.

PR #421, opened 2026-07-16
Flatten OidPost's provider guard and drop dead link-helper attributes

### LINKED ISSUE #393: Drop the dead action-filter attributes on the private link helpers and flatten OidPost's provider guard
Two small, behaviour-preserving reductions in SSOController.cs:

1. Dead action-filter attributes on private helpers. `SamlLink` (SSO-Auth/Api/SSOController.cs:1095-1096) and `OidLink` (:1140-1141) each carry `[Consumes]`/`[Produces]` attributes, but both are private methods invoked directly from `AddCanonicalLink`, ASP.NET Core only honours these filters on public action methods, so the four attribute lines are inert metadata. The real content negotiation happens on the `AddCanonicalLink` action, which keeps its own attributes. 4 lines.

2. Flatten `OidPost`'s provider guard. The endpoint rejects a null config at :139-142 with `BadRequest(NoMatchingProviderMessage)`, wraps the whole happy path in `if (config.Enabled)` (:144), and then repeats the exact same `BadRequest(NoMatchingProviderMessage)` for the disabled case at :213. A single guard clause, `if (config is not { Enabled: true }) return BadRequest(NoMatchingProviderMessage);`, produces byte-identical responses for both the unknown and the disabled provider (they already share one body today, so no enumeration-oracle property changes), removes the duplicated rejection, and drops one nesting level from ~65 lines. ~8 lines plus the dedent, and the endpoint reads like its SAML sibling (`SamlPost` already uses the guard-clause form).

Both are pinned by the existing endpoint tests (`SSOControllerLinkTests`, the OID callback coverage); no response body, status code, or ordering changes.

Line references are against main as of 6471d08; item 2's region is adjacent to but outside the #390 session-minter extraction, so rebasing over that merge is trivial.

iderex:
## What & why

Two behavior-preserving reductions in `SSOController.cs`:

1. **Flatten `OidPost`'s provider guard.** It rejected a null config, wrapped the whole happy path in `if (config.Enabled)`, and repeated the same `BadRequest(NoMatchingProviderMessage)` for the disabled case at the end. Collapsed to one guard clause, `if (config is not { Enabled: true }) return BadRequest(...)`, which yields **byte-identical responses** for the unknown and disabled cases (they already shared one body, so no enumeration-oracle property changes), removes the duplicated rejection, and drops a nesting level from the ~65-line body so it reads like its SAML sibling `SamlPost`.
2. **Drop the `[Consumes]`/`[Produces]` attributes on the private `SamlLink` and `OidLink` helpers.** ASP.NET Core honors content-negotiation filters only on public action methods; these are invoked directly from `AddCanonicalLink` (which keeps its own), so the four attribute lines were inert metadata.

Closes #393.

## Type of change
- [x] Quality / tech-debt (behavior-preserving)

## Security checklist
- No security-posture change: the guard clause returns the identical `BadRequest(NoMatchingProviderMessage)` for both unknown and disabled providers (already one shared body), and the happy-path logic is unchanged, the diff is purely structural (dedent + `if/else`→fall-through).

## Quality checklist
- Removes a duplicated rejection and one nesting level from the login callback, and four inert attribute lines.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **803/803 green**, including the endpoint tests that assert the exact `"No matching provider found"` body for the unknown and disabled branches (`OidPost_UnknownProvider_ReturnsBadRequest`, `OidPost_DisabledProvider_ReturnsBadRequest`), the missing/invalid-state branches, and the OID callback happy path, the characterization tests pin every branch.
- The diff was inspected to confirm only structural changes (indentation, the removed `if (config.Enabled)` wrapper, the guard clause, and the `else`→fall-through); no happy-path line was altered.

### DIFF
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -135,83 +135,77 @@ public class SSOController : ControllerBase
             return throttled;
         }
 
+        // Unknown and disabled providers share one rejection so neither can be probed apart, matching
+        // the guard-clause form the SAML sibling (SamlPost) already uses.
         var config = FindOidConfig(provider);
-        if (config is null)
+        if (config is not { Enabled: true })
         {
             return BadRequest(NoMatchingProviderMessage);
         }
 
-        if (config.Enabled)
+        if (string.IsNullOrEmpty(state))
         {
-            if (string.IsNullOrEmpty(state))
-            {
-                return BadRequest("Missing state");
-            }
-
-            if (StateStore.PeekCurrent(state, provider, DateTime.Now, Request.Cookies[AuthorizeStateBinding.CookieName]) is not { } pending)
-            {
-                // Unknown, expired, minted for a different provider, or from a different browser than the
-                // one that started the flow (#326) — reject (details on PeekCurrent / AuthorizeStateBinding).
-                return BadRequest("Invalid or expired state");
-            }
-
-            var oidcClient = CreateCallbackOidcClient(config, provider);
-            var result = await oidcClient.ProcessResponseAsync(Request.QueryString.Value, pending.OidcState).ConfigureAwait(false);
-
-            if (result.IsError)
-            {
-                return ReturnError(StatusCodes.Status400BadRequest, $"Error logging in: {result.Error} - {result.ErrorDescription}");
-            }
-
-            // RFC 9207 (#125): the library parses the authorization-response `iss` but never checks it.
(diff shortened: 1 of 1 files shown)


iderex:
Pre-merge, all gates green. 16/16 checks pass, SonarCloud OK, 803/803 tests (the exact-body characterization tests pin every OidPost guard branch). CodeRabbit review completed, zero findings. Diff verified purely structural. Merging on green.

PR #422, opened 2026-07-16
Define the SAML algorithm allow-list predicate once

### LINKED ISSUE #395: Define the SAML algorithm-allowlist predicate once in SamlSignatureAlgorithms
The "non-null, at least one element, every element on the allowlist" predicate in `SamlSignatureAlgorithms` is written out twice: `AreTransformsAllowed` (SSO-Auth/SamlSignatureAlgorithms.cs:67-85) and the digest half of `IsAllowed` (:100-115) are the same loop against different sets.

One private helper - 

```csharp
private static bool AllOnList(HashSet<string> allow, IEnumerable<string> values)
```

 - carrying the null check, the at-least-one flag, and the `?? string.Empty` lookup turns both call sites into one-liners. ~14 lines removed, and the fail-closed shape (null enumerable rejected, empty chain rejected, any off-list entry rejected, same short-circuiting) is defined once so the two copies can never drift apart.

No check is weakened or reordered; the observable behaviour is pinned by `SamlSignatureAlgorithmsTests` and the attack-shape coverage in `SamlAttackShapeTests`. Since this file is the SAML algorithm gate, the PR runs the adversarial-review gate as usual even though the change is a pure refactor.

iderex:
## What & why

`AreTransformsAllowed` and the digest half of `IsAllowed` in `SamlSignatureAlgorithms` spelled out the same **"non-null enumerable, at least one element, every element on the allow-list"** loop against different sets. Extract it into one private `AllOnList` helper so the fail-closed shape, null rejected, empty chain rejected, any off-list or null element rejected, short-circuit on the first miss, is defined once and the two copies cannot drift apart.

```csharp
internal static bool AreTransformsAllowed(IEnumerable<string> transforms)
    => AllOnList(AllowedTransforms, transforms);

internal static bool IsAllowed(string signatureMethod, IEnumerable<string> digestMethods)
    => AllowedSignatureMethods.Contains(signatureMethod ?? string.Empty)
       && AllOnList(AllowedDigestMethods, digestMethods);
```

Closes #395.

## Type of change
- [x] Quality / tech-debt (behavior-preserving dedup of the SAML algorithm gate)

## Security checklist
- No check weakened or reordered: `IsAllowed` still checks the signature method first and `&&` short-circuits before enumerating the digests, so a disallowed signature method with a null `digestMethods` still returns false rather than throwing. The allow-list sets and `StringComparer.Ordinal` are untouched. SHA-1 signatures, comment-preserving c14n, XPath/XSLT transforms, and empty chains stay rejected.

## Quality checklist
- ~14 lines removed; the fail-closed predicate is defined once so the transform and digest gates can never drift.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **803/803 green** (`SamlSignatureAlgorithmsTests` and the XSW/downgrade `SamlAttackShapeTests` pin every branch).
- Adversarial-review gate (SAML algorithm gate, per the issue): focused bypass + correctness lenses over the extraction, results appended below before merge.

### DIFF
--- a/SSO-Auth.Tests/SamlSignatureAlgorithmsTests.cs
+++ b/SSO-Auth.Tests/SamlSignatureAlgorithmsTests.cs
@@ -83,6 +83,14 @@ public class SamlSignatureAlgorithmsTests
         Assert.False(SamlSignatureAlgorithms.IsAllowed("urn:made:up", new[] { DigestSha256 }));
     }
 
+    [Fact]
+    public void IsAllowed_DisallowedSignatureWithNullDigests_ReturnsFalseWithoutThrowing()
+    {
+        // The signature method is checked first and short-circuits, so a disallowed signature with a
+        // null digest enumerable fails closed rather than dereferencing the null (#395).
+        Assert.False(SamlSignatureAlgorithms.IsAllowed("urn:made:up", null!));
+    }
+
     // --- Canonicalization allowlist (#137) ---
 
     private const string ExcC14n = "http://www.w3.org/2001/10/xml-exc-c14n#";
--- a/SSO-Auth/SamlSignatureAlgorithms.cs
+++ b/SSO-Auth/SamlSignatureAlgorithms.cs
@@ -65,24 +65,7 @@ internal static class SamlSignatureAlgorithms
     /// <param name="transforms">The transform-algorithm URIs of one reference, in order.</param>
     /// <returns>True only if there is at least one transform and all are allowed.</returns>
     internal static bool AreTransformsAllowed(IEnumerable<string> transforms)
-    {
-        if (transforms == null)
-        {
-            return false;
-        }
-
-        var any = false;
-        foreach (var transform in transforms)
-        {
-            any = true;
-            if (!AllowedTransforms.Contains(transform ?? string.Empty))
-            {
-                return false;
-            }
-        }
-
-        return any;
-    }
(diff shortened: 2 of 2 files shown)


iderex:
Adversarial-review gate (SAML algorithm gate), focused bypass + correctness lenses, both **PASS** with case-by-case equivalence proofs:
- bypass: no input that was rejected before (SHA-1 signature, #WithComments c14n, XPath/XSLT transform, empty transform chain, empty digest set, null element, disallowed/null signature) can pass after the extraction; the `any` flag preserves the at-least-one requirement (the empty-chain XSW lever) and the `&&` preserves the signature-first short-circuit with no NRE.
- correctness: `AllOnList` is a token-for-token factoring of both inline loops; evaluation order and the disallowed-sig + null-digest = false (no NRE) behavior are preserved.

The one coverage note the bypass lens raised, no test asserted `IsAllowed(disallowedSig, null)`, is now closed by `IsAllowed_DisallowedSignatureWithNullDigests_ReturnsFalseWithoutThrowing` (804 tests).

iderex:
Pre-merge, all gates green. 16/16 checks pass, SonarCloud OK, 804/804 tests. Adversarial gate: focused bypass + correctness both PASS (case-by-case equivalence, no crypto-gate weakening); the noted coverage gap closed with a regression test. CodeRabbit review completed, zero findings. Merging on green.

PR #423, opened 2026-07-16
Fold the SAML XML loaders into SamlResponse construction

### LINKED ISSUE #396: Fold the public SAML XML loaders into SamlResponse construction
`SamlResponse.LoadXml` (SSO-Auth/Saml.cs:61-87) and `LoadXmlFromBase64` (:93-96) are public mutators with no caller outside the class: the only use of `LoadXmlFromBase64` is the constructor (:49), and the only use of `LoadXml` is `LoadXmlFromBase64` itself (the other `LoadXml` hits in the file are `SignedXml.LoadXml`, and tests only call `XmlDocument.LoadXml` inside `SamlTestFactory`).

Fold the loading into construction: a private static helper returns the parsed `XmlDocument`, the constructor assigns it (then derives the namespace manager), and `_xmlDoc`/`_xmlNameSpaceManager` (:35-36) become `readonly`. The two public XML-doc blocks go away, ~12 lines, and, more importantly, the hand-rolled SAML core stops exposing a mutator that would let arbitrary XML be swapped into an already-constructed (and possibly already-validated) response object. Nobody calls it today, but the door shouldn't be there.

Behaviour unchanged: the constructor's documented exception ordering (certificate base64 → certificate load → body base64/XML, :45-48) is preserved by keeping the same statement order, and the DTD-prohibit / PreserveWhitespace / null-resolver hardening moves verbatim. Pinned by `SamlResponseTests`, `SamlResponseParsingTests`, `SamlCrlfInteropTests`, and `SamlAttackShapeTests`. As a Saml.cs change it runs the adversarial-review gate.

iderex:
## What & why

`SamlResponse.LoadXml` and `LoadXmlFromBase64` were **public mutators with no caller outside the class**, the constructor was the only user. Beyond the dead surface, they let arbitrary XML be swapped into an already-constructed (and possibly already-validated) response object; that door should not exist on the hand-rolled SAML core.

Fold the loading into construction: a private static `ParseResponseXml` returns the hardened `XmlDocument`, the constructor assigns it and derives the namespace manager, and `_xmlDoc` / `_xmlNameSpaceManager` become **`readonly`**. The two public XML-doc methods are removed.

Closes #396.

## Type of change
- [x] Security hardening (removes a post-construction XML-swap mutator) / tech-debt

## Security checklist
- The untrusted-XML hardening moves **verbatim**: `DtdProcessing.Prohibit` (billion-laughs), `XmlResolver = null` (XXE/SSRF), and `PreserveWhitespace = true` (signature canonicalization) are all applied in `ParseResponseXml`, unchanged.
- The constructor's documented exception ordering is preserved (certificate base64 → certificate load → body base64 → XML parse), the body decode still runs after the certificate load, so `SamlResponseLoader.TryParse`'s catch sequence is unchanged.
- `readonly` fields make a validated `SamlResponse` immutable, no XML can be swapped in after signature validation.

## Quality checklist
- Removes two public methods (~12 lines) and one mutable-state footgun; loading has a single path.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **804/804 green** (`SamlResponseTests`, `SamlResponseParsingTests`, `SamlCrlfInteropTests`, `SamlAttackShapeTests` pin the parse hardening and attack shapes).
- Adversarial-review gate (Saml.cs, per the issue): focused bypass + correctness lenses, results appended below before merge.

### DIFF
--- a/SSO-Auth/Saml.cs
+++ b/SSO-Auth/Saml.cs
@@ -32,8 +32,8 @@ internal sealed class SamlResponse
         "/samlp:Response/ds:Signature | /samlp:Response/saml:Assertion/ds:Signature";
 
     private readonly X509Certificate2 _certificate;
-    private XmlDocument _xmlDoc;
-    private XmlNamespaceManager _xmlNameSpaceManager; // we need this one to run our XPath queries on the SAML XML
+    private readonly XmlDocument _xmlDoc;
+    private readonly XmlNamespaceManager _xmlNameSpaceManager; // we need this one to run our XPath queries on the SAML XML
 
     /// <summary>
     /// Initializes a new instance of the <see cref="SamlResponse"/> class.
@@ -42,11 +42,14 @@ internal sealed class SamlResponse
     /// <param name="responseString">The SAML response formatted as a string.</param>
     public SamlResponse(string certificateStr, string responseString)
     {
-        // Decode and load the certificate, then load the response — in that exact order, so the exception
+        // Decode and load the certificate, then parse the response — in that exact order, so the exception
         // sequence SamlResponseLoader.TryParse catches is unchanged: FormatException (bad certificate
         // base64), CryptographicException (bad certificate), then FormatException/XmlException (bad body).
+        // The XML is loaded here, at construction, and the fields are readonly: a validated response
+        // object can never have different XML swapped into it afterwards (#396).
         _certificate = X509CertificateLoader.LoadCertificate(Convert.FromBase64String(certificateStr));
-        LoadXmlFromBase64(responseString);
+        _xmlDoc = ParseResponseXml(responseString);
+        _xmlNameSpaceManager = GetNamespaceManager(); // lets construct a "manager" for XPath queries
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


iderex:
Adversarial-review gate (Saml.cs), focused bypass + correctness lenses, both **PASS**:
- bypass: DtdProcessing.Prohibit, XmlResolver=null on BOTH the XmlDocument and the XmlReaderSettings, and PreserveWhitespace=true are all preserved verbatim and set before load; folding into construction with readonly fields strictly reduces the mutation surface (no swap-XML-into-a-validated-object door) without opening a new one.
- correctness: exception ordering byte-identical (cert base64 → cert load → body base64 → XML parse), so TryParse's catch sequence is unchanged; the object-initializer form is semantically identical to the prior sequential property assignment and applied before Load; readonly + GetNamespaceManager()-after-assignment has no null/order bug; repo-wide grep confirms no caller used the removed public mutators (only SignedXml.LoadXml / XmlDocument.LoadXml, unrelated).

iderex:
@coderabbitai review

iderex:
Pre-merge, all gates green. 16/16 checks pass, SonarCloud OK, 804/804 tests (SamlResponse/parsing/CRLF/attack-shape tests pin the parse hardening). Adversarial gate: focused bypass + correctness both PASS (hardening verbatim, exception ordering byte-identical, immutability tightened). CodeRabbit initially skipped (rate limit); re-requested and completed with zero findings. Merging on green.

Issue #369, opened 2026-07-15
Parse the oid/saml mode token once at the controller boundary


iderex:
The lowercase `"oid"`/`"saml"` mode token travels from the route through several layers as a raw string, with two independent switches that must agree: `GetLinks` (SSO-Auth/Api/SSOController.cs:976-981) and `AddCanonicalLink` (:1003-1011); further occurrences at :614, :905, :1116, :1192 and :1234. A typo compiles fine and either throws `ArgumentException` at runtime or misroutes an account link, and `ToLowerInvariant()`/`ToLower()` normalizations are scattered at each site.

Proposal: parse the route string once at the controller boundary into a `ProviderMode` enum (the V2 reference already defines exactly this, `Api/ProviderMode.cs`) and pass the enum internally; a single switch remains at the map-selection site. Net negative lines, the second switch and the per-site normalizations go away.

Sequencing: the in-flight #318 link-service step may absorb `GetLinks`/`MutateLinks`; if so, fold this into that step instead of a separate pass to avoid double churn.

iderex:
Consolidating: #381 (filed while reviewing #373) covers the same step, parse the oid/saml mode once at the HTTP boundary into an internal enum, deleting the three current dispatch variants (`TryGetLinks` invariant-lowercase+throw, `LinksByUser` Ordinal-else-OID, the controller's culture-sensitive `ToLower()`). It additionally subsumes the `Try*`-returns-enum → `CreateLink`/`RemoveLink` rename (deferred from #373 to avoid churning a reviewed surface) and should add the service-level invalid-mode, DELETE-route invalid/mixed-case, and controller EmptyKey test pins. Closing #381 as a duplicate of this; this issue is the survivor. The `TryGetLinks` comment in `CanonicalLinkService` now cites #369.

PR #424, opened 2026-07-16
Collapse the per-mode twin blocks over ProviderConfigBase

### LINKED ISSUE #392: Collapse the per-mode twin blocks over ProviderConfigBase in the link service and save pipeline
Since #204 moved the shared members (including `CanonicalLinks` and `Enabled`) onto `ProviderConfigBase`, several per-mode twin blocks that predate the base class can collapse into one generic body. Today each of these sites carries a SAML arm and an OpenID arm that are structurally identical and must be kept in sync by hand:

- `CanonicalLinkService.TryGetLinks` (SSO-Auth/Api/Linking/CanonicalLinkService.cs:448-469): the two switch arms differ only in which config map they index. A small generic helper (`TryGetLinks<T>(SerializableDictionary<string, T> configs, ...) where T : ProviderConfigBase`) reduces each arm to a one-liner. ~8 lines.
- `CanonicalLinkService.LinksByUser` (:324-326): the twin LINQ queries differ only in the source map, and the `Select` into a reconstructed `KeyValuePair` is redundant once both sides project through the base type. ~4 lines.
- `CanonicalLinkService.RemoveUserEverywhere` (:356-370): two identical `foreach` blocks; one loop over `SamlConfigs.Values.Concat<ProviderConfigBase>(OidConfigs.Values)` does the same work. ~7 lines.
- `ServerManagedFields.Preserve` (SSO-Auth/Config/ServerManagedFields.cs:30-50): two identical guarded loops whose only real difference is the per-provider overload they resolve to; a generic loop taking the per-provider action collapses them. ~9 lines.

Adjacent, same save pipeline: `ProviderConfigValidator.Validate` (SSO-Auth/Config/ProviderConfigValidator.cs:40-49) iterates `incoming.SamlConfigs` twice back to back (name/URL, then certificate); one loop does both. ~5 lines. Note the one observable delta: when a save contains two invalid SAML providers with different defects, the first exception message may name a different provider than today, every invalid save is still rejected fail-closed, only the order of first-reported error can shift.

Roughly 30 lines removed net, and the per-mode duplication that has to be edited twice on every change to the link-map or preservation logic goes away.

Behaviour and security are unchanged: all sites keep running under the same config lock (`Read`/`Mutate` callbacks are untouched), the fail-closed null-config skips and the unknown-provider `false` returns are preserved verbatim, and no validation is weakened. The behaviour is pinned by `CanonicalLinkServiceTests`, `SSOControllerLinkTests`, `ConfigPreservationTests`, and the link-map rules in `ArchitectureConformanceTests`.

Sequencing: complements #369 (parsing the mode token once into an enum), the two compose naturally and could land as one pass; if the #318 link-service step absorbs these files first, fold this in there instead of a separate pass.

iderex:
# What & why

Since #204 moved `CanonicalLinks` and `Enabled` onto `ProviderConfigBase`, five per-mode SAML/OpenID twin blocks that predate the base class carried structurally identical arms that had to be edited twice on every change. This folds each into one body. Closes #392.

- `CanonicalLinkService.TryGetLinks`: both switch arms now call one generic helper over the shared base; the `requireEnabled` truth table is unchanged (`Enabled` is read only after `links != null` proves the config non-null).
- `CanonicalLinkService.LinksByUser`: the twin LINQ queries fold into one tuple projection; the null-config skip moves into the loop unchanged (same result set, including empty-map providers).
- `CanonicalLinkService.RemoveUserEverywhere`: one loop over both protocols' providers via a covariant `Concat` replaces the two identical `foreach` blocks (same elements, same order, same lock).
- `ServerManagedFields.Preserve`: one generic loop takes the per-provider overload as the action; the preserved provider set, iteration direction, and null-map guards are identical, so the `CanonicalLinks` re-injection and the `OidSecret` blank-means-keep rule (`ResolveUpdatedSecret`) run exactly as before.
- `ProviderConfigValidator.Validate`: the two back-to-back `SamlConfigs` loops merge into one running all three checks per provider.

Code-only delta is -11 lines (raw diff +3, the difference being the constraint comments each fold carries).

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green (804/804).

## Notes

- Sole observable delta, accepted on #392: with several invalid SAML providers in one save, the first-reported validation error follows map order rather than check kind. The accept/reject set is bit-identical, every invalid save is still rejected fail-closed before anything is persisted; both surfaces are `ArgumentException`, so the HTTP mapping is unchanged. No test pins the old ordering (all negative tests use single-provider configs).
- Adversarial review (all four lenses: availability/mass-lockout, security-bypass, correctness, integration): PASS across the board. The lenses verified the method-group resolution in `Preserve` binds uniquely to the per-provider overloads, the covariant `Concat` visits the same elements in the same order, nothing lazily enumerable escapes a config-lock callback, and the `requireEnabled` polarity at all five call sites is untouched. One LOW note each: on a hypothetical null protocol map (unreachable via any real load path), `RemoveUserEverywhere` now fails before any in-memory removal instead of after the SAML half, a strictly safer failure shape.
- The conformance link-map scan is confined to `SSOController.cs` (untouched); no new structural property to lock in, the fold removes duplication inside the two homes the existing rule already names.
- Behaviour is pinned by `CanonicalLinkServiceTests`, `SSOControllerLinkTests`, `ConfigPreservationTests`, and the link-map rules in `ArchitectureConformanceTests`.
- The complementary mode-token enum work stays tracked in #369; the `LinksByUser` mode fallthrough (any non-"saml" mode reads OpenID) is pre-existing and belongs to that issue.

### DIFF
--- a/SSO-Auth/Api/Linking/CanonicalLinkService.cs
+++ b/SSO-Auth/Api/Linking/CanonicalLinkService.cs
@@ -334,20 +334,25 @@ internal sealed class CanonicalLinkService
     {
         return _configStore.Read(configuration =>
         {
-            // A provider stored with a null config object (reachable today via #350's null-body add)
-            // is skipped rather than dereferenced — same fail-closed treatment TryGetLinks gives it,
-            // so the read side cannot NRE into a 500 on a state the write side can produce.
-            var providerList = string.Equals(mode, "saml", StringComparison.Ordinal)
-                ? configuration.SamlConfigs.Where(p => p.Value?.CanonicalLinks != null).Select(p => new KeyValuePair<string, SerializableDictionary<string, Guid>>(p.Key, p.Value.CanonicalLinks))
-                : configuration.OidConfigs.Where(p => p.Value?.CanonicalLinks != null).Select(p => new KeyValuePair<string, SerializableDictionary<string, Guid>>(p.Key, p.Value.CanonicalLinks));
+            // Both arms project (name, links) tuples through ProviderConfigBase.CanonicalLinks, so the
+            // per-mode twin queries are one shape. A provider stored with a null config object (reachable
+            // today via #350's null-body add) yields null links and is skipped rather than dereferenced —
+            // same fail-closed treatment TryGetLinks gives it, so the read side cannot NRE into a 500 on
+            // a state the write side can produce.
+            var providerLinks = string.Equals(mode, "saml", StringComparison.Ordinal)
+                ? configuration.SamlConfigs.Select(p => (p.Key, p.Value?.CanonicalLinks))
+                : configuration.OidConfigs.Select(p => (p.Key, p.Value?.CanonicalLinks));
 
             var mappings = new SerializableDictionary<string, IEnumerable<string>>();
-            foreach (var provider in providerList)
+            foreach (var (provider, links) in providerLinks)
             {
-                mappings[provider.Key] = provider.Value
-                    .Where(link => link.Value == jellyfinUserId)
-                    .Select(link => link.Key)
-                    .ToList();
+                if (links != null)
+                {
+                    mappings[provider] = links
+                        .Where(link => link.Value == jellyfinUserId)
+                        .Select(link => link.Key)
+                        .ToList();
+                }
             }
 
             return mappings;
@@ -367,17 +372,10 @@ internal sealed class CanonicalLinkService
         {
             int removed = 0;
 
-            // Skip a provider stored with a null config object (reachable via #350); it holds no links to
-            // revoke, and dereferencing it would NRE into a 500 — the same fail-closed skip TryGetLinks uses.
-            foreach (var config in configuration.SamlConfigs.Values)
-            {
-                if (config?.CanonicalLinks is { } links)
-                {
-                    removed += CanonicalLinkRevoker.RemoveUser(links, userId);
-                }
-            }
-
-            foreach (var config in configuration.OidConfigs.Values)
+            // One loop over both protocols' providers (covariant Concat over the shared base). Skip a
+            // provider stored with a null config object (reachable via #350); it holds no links to revoke,
+            // and dereferencing it would NRE into a 500 — the same fail-closed skip TryGetLinks uses.
+            foreach (var config in configuration.SamlConfigs.Values.Concat<ProviderConfigBase>(configuration.OidConfigs.Values))
             {
                 if (config?.CanonicalLinks is { } links)
                 {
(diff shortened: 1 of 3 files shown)


iderex (commented):
My sign-off — consolidated pre-merge review.

**Internal adversarial review** (four refute-by-default lenses, each over the full touched files plus their direct callers, not just the diff):

- **Availability / mass-lockout: PASS.** The generic `Preserve<T>` loop touches exactly the provider set the old per-protocol loops touched (iterate live, look up incoming), so the `CanonicalLinks` re-injection and the `OidSecret` blank-means-keep rule run unchanged — no mass-unlink or secret-wipe path. The merged `Validate` loop has a bit-identical accept/reject set (no admin lockout). One LOW note: on a hypothetical null protocol map (unreachable — the constructor initializes both maps), `RemoveUserEverywhere` now fails before any in-memory removal instead of after the SAML half; a strictly safer failure shape.
- **Security bypass: PASS.** `TryGetLinks<T>` is token-identical to both removed arms, including the short-circuit that keeps `config.Enabled` unreached when the entry is null-valued (#350 state, fail-closed false). `requireEnabled` polarity at all five call sites is untouched (GRANT paths reject disabled providers; revocation keeps working on them). Method-group resolution in `Preserve` binds uniquely to the per-provider overloads, so the `ResolveUpdatedSecret` exfiltration guard (identity change → secret not carried over) is fully intact.
- **Correctness: PASS.** Overload resolution, covariant `Concat` (same elements, same order, safe under the single `Mutate` lock), tuple projection fully enumerated inside the `Read` callback (no deferred execution escapes the lock), lazy self-healing `CanonicalLinks` getter semantics preserved (old `Where` filter only ever excluded null-config entries — result set identical, empty-map providers included as before), definite assignment of the `out` parameter on every path.
- **Integration: PASS.** `ProviderConfigStore.Save` order (Validate → Preserve → persist, one lock) untouched; `LinksByUser` JSON response byte-equivalent; declared map types match the generic parameters; no test or conformance rule pins the old shape (the link-map source scan is confined to `SSOController.cs`, untouched); no csproj/build.yaml drift.

The sole observable delta — first-reported validation error follows map order rather than check kind when several SAML providers are invalid in one save — is the one declared and accepted on #392; every invalid save is still rejected fail-closed pre-persist.

**Gates:** local Debug + Release builds green (`--warnaserror`, 0 warnings), 804/804 tests green, Jellyfin compat check green, Prettier n/a (no web files). CI all green, including CodeQL, abi-floor, package (JPRM), and the SonarCloud quality gate (passed outright, no coverage exception needed). CodeRabbit ran on the final commit (`3f4ae11`, status "Review completed") and produced no actionable comments — nothing to disposition. Copilot posted only a quota-failure notice (no review content); per the standing rule it is not requested and not blocking.

Overall verdict: **ready to merge** — behaviour-preserving fold, all gates green, no open findings.

Issue #370, opened 2026-07-15
Contributor-clarity sweep: callback naming, dead SAML accessor, IP-classifier home, asset casing


iderex:
Four small, disjoint contributor traps, each fixable with net-negative or zero lines; one sweep PR in the style of #237.

- `OidPost` handles a GET (SSO-Auth/Api/SSOController.cs:135-140), excused by a comment ("called `Post` for consistency with SAML"). Rename the action methods only, `OidPost` to `OidCallback`, and `SamlPost` to `SamlCallback` for symmetry; the routes are attribute literals, so the wire contract is untouched. A comment narrating a misleading name is worse than a correct name.
- `SamlResponse.GetCustomAttribute` (SSO-Auth/Saml.cs:425-429) has no production caller, only tests reference it. Dead members on the hand-rolled SAML core are the worst place for vestigial surface, since a reader must treat every accessor there as security-relevant. Delete it, retarget the two test assertions onto the plural form, and hoist the `"Role"` attribute literal repeated at SSOController.cs:689/:704/:866/:888 into one constant, it is the attribute name the whole RBAC design hinges on, with no single definition site.
- The public-address classifier the rate limiter's lockout exemption depends on lives in `AvatarUrlValidator` (`IsBlockedAddress`/`TryExtractEmbeddedIPv4`, called from SsoRateLimiter.cs:102/:128). Sharing one predicate is right, but its home is misleading in both grep directions: tuning avatar SSRF policy silently changes which clients the login rate limiter exempts. Move both members verbatim into a neutrally named leaf (e.g. `IpAddressClassifier`) that both callers use; zero behavior change, and the shared-invariant comment gets one canonical home.
- The linking page imports `./ApiClient.js` (SSO-Auth/Config/linking.html:8) while the file is `Views/apiClient.js`, bridged only by the name table in SSOPlugin.cs:156-160 and matched case-sensitively (SSOViewsController.cs:66); `style.css` is additionally published under two names ("SSO-Auth.css" via GetPages at SSOPlugin.cs:117-119, "style.css" via GetViews at :143-145). "Fixing" the casing on either end breaks the linking page at runtime only. Unify on one casing and comment the GetPages/GetViews name tables as the public URL contract (change one side, change both).

[...]

PR #425, opened 2026-07-16
Compact the helpers: record struct, dead guards, constant parameter

### LINKED ISSUE #397: Helper compactness sweep: hand-written decision struct, views indirection, redundant constructor and guards
A batch of small, independently verified compactness items across the helpers, in the #370 sweep style. None changes behaviour; together they remove ~40 lines.

- `AccountLinkDecision` (SSO-Auth/Api/AccountLinkResolver.cs:26-44): 19 lines of hand-written constructor + get-only properties collapse to `internal readonly record struct AccountLinkDecision(AccountLinkAction Action, Guid UserId);`. All construction is positional and all consumption is member reads (`CanonicalLinkService`, `AccountLinkResolverTests`), so it is a drop-in; the file's siblings (`RoleGrants`, `OidcAuthorizeState`) already use this idiom. ~14 lines.
- `SSOViewsController` (SSO-Auth/Views/SSOViewsController.cs:47-52): `GetView` only forwards to `ServeView`, which has no other caller, inline it. The `pages == null` guard (:59-64) is unreachable (`SSOPlugin.GetViews` returns an array literal; this is asset serving, not a login-path guard). `FirstOrDefault(predicate, null)` (:66) is the two-arg overload spelled with its own default. The three-line `#nullable enable`/`disable` sandwich (:72-80) goes with it. ~10 lines.
- `SerializableDictionary` (SSO-Auth/SerializableDictionary.cs:16-22): the explicit empty parameterless constructor is exactly what the compiler supplies implicitly (no other constructor exists); `XmlSerializer` and every call site are unaffected. Pinned by `SerializableDictionarySerializationTests`. ~8 lines.
- `RequestHelpers.AssertCanUpdateUser` (SSO-Auth/Api/RequestHelpers.cs:32-46): all four call sites pass `restrictUserPreferences: true`, so the parameter is a constant, drop it, make the preference term unconditional, and collapse the `if/return false/return true` into one return. The vendored-snippet header comment stays. ~5 lines.
- `WriteOnlySecretConverter` (SSO-Auth/Config/WriteOnlySecretConverter.cs:21): `reader.TokenType == JsonTokenType.Null ? null : reader.GetString()`, `GetString()` already returns null on a Null token.
- `OidcAuthorizeStateBuilder.ResolveAvatarUrl` (SSO-Auth/Api/OidcAuthorizeStateBuilder.cs:110-112): the `Contains` guard before `Replace` is redundant, `string.Replace` returns the same instance when nothing matches, and the first-duplicate-wins property is unchanged. Also hoist the per-claim `config.DefaultUsernameClaim?.Trim() ?? "preferred_username"` (:138) above the loop in `ScanClaims`; it depends only on config.
- `SSOPlugin` (SSO-Auth/SSOPlugin.cs:17): `IPlugin` in the base list is redundant, `BasePlugin<T>` already implements it.
- `OidcStateStore.Count` (SSO-Auth/Api/OidcStateStore.cs:64) is used only by `OidcStateStoreTests` but, unlike `Seed`/`Clear`, is not documented as a test seam, either mark it as one or drop it in favour of `Summaries().Count()` in the tests.
- One addition (+2 lines), same clarity goal: `[XmlRoot("PluginConfiguration")]` on `SamlConfig`/`OidConfig` (SSO-Auth/Config/PluginConfiguration.cs:202/:252) looks like deletable copy-paste cruft but names the `<value>` element `SerializableDictionary.WriteXml` persists (`new XmlSerializer(typeof(TValue))`), deleting it would break deserialization of every existing config on disk. It needs a one-line constraint comment so a cleanliness pass never removes it.

Excluded here because they are already tracked: the `ServerManagedFields`/`ProviderConfigValidator` twin-loop folds (#392) and everything in #364/#365/#369/#370/#391.

iderex:
# What & why

A batch of small, independently verified compactness items across the helpers, in the #370 sweep style. None changes behaviour; together they remove the ceremony the problem never required. Closes #397.

- `AccountLinkDecision`: 19 lines of hand-written constructor + get-only properties collapse to a positional `readonly record struct`, the idiom its file siblings (`RoleGrants`, `OidcAuthorizeState`) already use. All construction is positional, all consumption is member reads.
- `SSOViewsController`: `GetView` no longer forwards to a single-caller `ServeView`; the `pages == null` guard was unreachable (`SSOPlugin.GetViews` returns an array literal, this is asset serving, not a login-path guard); `FirstOrDefault(predicate, null)` was the two-arg overload spelled with its own default; the `#nullable` sandwich and three now-unused usings go with it.
- `SerializableDictionary`: the explicit empty parameterless constructor is exactly what the compiler supplies implicitly (no other constructor exists). `XmlSerializer` constructibility is pinned by the serialization round-trip tests.
- `RequestHelpers.AssertCanUpdateUser`: all four call sites passed `restrictUserPreferences: true`, so the parameter folds in as always-on and the guard collapses to one boolean return with an identical truth table (`(same || admin) && preferenceAccess`). A new test pins the `EnableUserPreferenceAccess` refusal leg, which the fold made load-bearing by convention rather than by parameter.
- `WriteOnlySecretConverter`: `GetString()` already returns null on a Null token, and the serializer never routes a Null token into `Read` for a reference-type converter without `HandleNull`, so the ternary was dead both ways. The `Write` side is untouched: the secret still never serializes outward.
- `OidcAuthorizeStateBuilder`: the `Contains` guard before `Replace` was redundant (`Replace` returns the same instance when nothing matches; first-duplicate-wins unchanged), and the per-claim `DefaultUsernameClaim?.Trim() ?? "preferred_username"` hoists above the loop, it depends only on config, same precedent as the hoisted `SplitRoleClaimPath`.
- `SSOPlugin`: `IPlugin` in the base list was redundant, `BasePlugin<T>` already implements it, and reflection (`GetInterfaces`/`IsAssignableFrom`, what plugin discovery uses) is transitive over base types.
- `OidcStateStore.Count` is documented as the test seam it is, like `Seed`/`Clear`.
- `SamlConfig`/`OidConfig` `[XmlRoot("PluginConfiguration")]` gains a two-line constraint comment: it names the `<value>` element `SerializableDictionary.WriteXml` persists via `new XmlSerializer(typeof(TValue))`, so a future cleanliness pass must never remove it.

Production code delta is -28 lines (+16/-44); raw diff -20 including the added test.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green (805/805, one new test).

## Notes

- Adversarial review (all four lenses: availability/mass-lockout, security-bypass, correctness, integration): PASS across the board, with empirical probes, `Utf8JsonReader.GetString()` on a Null token returns null (and `JsonConverter<string>` with default `HandleNull` never routes Null into `Read`, so the deleted ternary was dead both ways); `string.Replace` with an absent token returns the same instance (`ReferenceEquals` true); the `AssertCanUpdateUser` truth table is identical in all 8 cells (an admin without `EnableUserPreferenceAccess` is still refused, as before); the decompiled `MediaBrowser.Common` confirms `BasePlugin : IPlugin`, and `IsAssignableFrom`/`GetInterfaces` are transitive, so plugin discovery is unaffected.
- The deleted `pages == null` guard was defense-in-depth against a future nullable `GetViews`; today it is provably dead (array literal, non-virtual, statically-typed instance). We accept the trade under the minimal-code directive and record the decision here.
- The review flagged that the `EnableUserPreferenceAccess` refusal leg had no test while the fold turned that term from parameter-supplied into helper-owned, the new `AddCanonicalLink_AdminWithoutPreferenceAccess_Returns403` pins it.
- Pre-existing, untouched, noted for completeness: a hypothetical `auth.User == null` (pure API-key auth) would NRE in `AssertCanUpdateUser` exactly as before this change; Jellyfin rejects unauthenticated requests before this point.
- Behaviour is pinned by `AccountLinkResolverTests`, `SSOControllerLinkTests`, `ConfigPreservationTests` (secret hidden-serialize + rotation), `SerializableDictionarySerializationTests` (round-trip incl. legacy on-disk format), and `OidcStateStoreTests`.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerLinkTests.cs
+++ b/SSO-Auth.Tests/SSOControllerLinkTests.cs
@@ -58,6 +58,19 @@ public class SSOControllerLinkTests
     }
 
     [Fact]
+    public async Task AddCanonicalLink_AdminWithoutPreferenceAccess_Returns403()
+    {
+        // Pins the EnableUserPreferenceAccess term of AssertCanUpdateUser (#397 folded it from an
+        // always-true parameter into the helper): even an administrator is refused without it, so a
+        // future edit dropping the term cannot pass silently.
+        var harness = ForCaller(isAdmin: true, callerId: Target, enableUserPreferenceAccess: false);
+
+        var result = await harness.Controller.AddCanonicalLink("oid", "keycloak", Target, new AuthResponse());
+
+        Assert.Equal(403, Assert.IsType<ObjectResult>(result).StatusCode);
+    }
+
+    [Fact]
     public async Task GetOidLinksByUser_NonAdminQueryingAnotherUser_Returns403()
     {
         var harness = ForCaller(isAdmin: false, callerId: Other);
@@ -324,13 +337,13 @@ public class SSOControllerLinkTests
     }
 
     // Builds a harness whose mocked IAuthorizationContext resolves the request to the given caller. An
-    // admin (or the target user themselves) passes AssertCanUpdateUser; a non-admin editing another user
-    // is refused.
-    private static SsoControllerHarness ForCaller(bool isAdmin, Guid callerId, Action<PluginConfiguration>? configure = null)
+    // admin (or the target user themselves) with preference access passes AssertCanUpdateUser; a
+    // non-admin editing another user, or any caller without EnableUserPreferenceAccess, is refused.
+    private static SsoControllerHarness ForCaller(bool isAdmin, Guid callerId, Action<PluginConfiguration>? configure = null, bool enableUserPreferenceAccess = true)
     {
         var harness = new SsoControllerHarness(configure);
 
-        var user = new User("caller", "SSO-Auth", "Default") { Id = callerId, EnableUserPreferenceAccess = true };
+        var user = new User("caller", "SSO-Auth", "Default") { Id = callerId, EnableUserPreferenceAccess = enableUserPreferenceAccess };
         user.SetPermission(PermissionKind.IsAdministrator, isAdmin);
 
         // AuthorizationInfo.UserId is derived from User.Id, so setting the user fixes the caller identity.
--- a/SSO-Auth/Api/AccountLinkResolver.cs
+++ b/SSO-Auth/Api/AccountLinkResolver.cs
@@ -23,25 +23,9 @@ internal enum AccountLinkAction
 /// <summary>
 /// A resolved account-link decision: the action and the Jellyfin user id it applies to.
 /// </summary>
-internal readonly struct AccountLinkDecision
-{
-    /// <summary>
-    /// Initializes a new instance of the <see cref="AccountLinkDecision"/> struct.
-    /// </summary>
-    /// <param name="action">The action to take.</param>
-    /// <param name="userId">The Jellyfin user id the action applies to (empty when creating/rejecting).</param>
-    internal AccountLinkDecision(AccountLinkAction action, Guid userId)
-    {
-        Action = action;
-        UserId = userId;
-    }
-
-    /// <summary>Gets the action to take.</summary>
-    internal AccountLinkAction Action { get; }
-
-    /// <summary>Gets the Jellyfin user id the action applies to.</summary>
-    internal Guid UserId { get; }
-}
+/// <param name="Action">The action to take.</param>
+/// <param name="UserId">The Jellyfin user id the action applies to (empty when creating/rejecting).</param>
+internal readonly record struct AccountLinkDecision(AccountLinkAction Action, Guid UserId);
 
 /// <summary>
 /// Pure decision logic mapping an SSO identity to a Jellyfin account. An identity already linked to
--- a/SSO-Auth/Api/OidcAuthorizeStateBuilder.cs
+++ b/SSO-Auth/Api/OidcAuthorizeStateBuilder.cs
@@ -105,8 +105,8 @@ internal static class OidcAuthorizeStateBuilder
     }
 
     // Resolves the avatar URL by substituting @{claimType} tokens in the configured format, or null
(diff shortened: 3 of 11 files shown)


iderex:
Pre-merge review, all gates green, ready to merge.

**Adversarial-review gate (the security-relevant folds of this compactness sweep):** focused bypass + correctness lenses, both **PASS**.

- **`AssertCanUpdateUser` authz fold**, exhaustively enumerated every caller on `origin/main`: all four (`AddCanonicalLink`, `DeleteCanonicalLink`, `GetSamlLinksByUser`, `GetOidLinksByUser`) passed `restrictUserPreferences: true`; there is no `false` caller and no other caller (internal static, direct-call only). For `restrict = true` the old predicate reduces by De Morgan to the new `(self || admin) && EnableUserPreferenceAccess` exactly, so authz is neither weakened nor wrongly tightened; null-safety and the fail-closed posture are unchanged. The new `AddCanonicalLink_AdminWithoutPreferenceAccess_Returns403` test non-vacuously pins the now-folded preference term.
- **`WriteOnlySecretConverter`**, the `Write` path (`WriteNullValue`) is untouched, so the secret still never serializes outward; the removed `Read` Null branch is genuinely unreachable (no `HandleNull` override on a reference-type converter, so System.Text.Json never calls `Read` with a Null token) and a semantic no-op even if reached (`GetString()` returns null for a Null token).
- **`OidcAuthorizeStateBuilder`**, the `Replace`-without-`Contains` and the hoisted `DefaultUsernameClaim` expression are value-identical, including duplicate-claim first-occurrence-wins ordering.

**CI/gates:** 16/16 checks pass, SonarCloud quality gate OK, CodeQL clean, 804 tests. Behaviour-preserving, production code −28 lines.

Merging on green.

PR #426, opened 2026-07-16
Remove the CodeRabbit and Copilot review integrations

iderex:
## What & why

Removes the two review-bot integrations from the repository per my decision to uninstall them:

- delete `.coderabbit.yaml` (CodeRabbit configuration),
- delete `.github/copilot-instructions.md` (Copilot review instructions),
- stop describing a CodeRabbit review as part of the merge gate in the README, and drop the two `ArchitectureConformanceTests` comments that named CodeRabbit alongside CodeQL.

The merge gate is now: an issue, tests, an adversarial security review for login-path/crypto changes, and CI (build with warnings-as-errors, the full test suite, format and conformance checks) plus CodeQL and SonarCloud.

The GitHub App installations themselves are removed in the account settings (that step is outside the repo); this change removes their in-repo configuration and documentation.

## Type of change
- [x] Chore / docs / tooling

## Verification
- `dotnet build --warnaserror`: 0 errors (only a comment changed in the test project). Prettier clean.

PR #427, opened 2026-07-16
Bound the in-flight login state stores per client

### LINKED ISSUE #327: Bound the in-flight OpenID/SAML state caps per client so a challenge flood cannot lock out all logins
The in-flight OpenID authorize-state store (`StateManager`, cap `MaxStateEntries = 100_000`, `SSO-Auth/Api/SSOController.cs:56`) and the SAML request cache (`SamlRequestCache`, `MaxEntries = 100_000`) are process-global with no per-client accounting. `AuthStateStore.TryAdd` refuses any new key once the store is full (`SSO-Auth/Api/AuthStateStore.cs:49-54`). That correctly bounds memory, but with no per-client fairness it converts a memory-DoS into an availability-DoS: a single anonymous source can consume the whole global budget and deny logins to everyone.

## Exploit (default configuration)

- `OID/p/{provider}` is anonymous (`SSOController.cs:262`) and the built-in rate limiter is off by default (`EnableRateLimit` defaults `false`, `SSO-Auth/Config/PluginConfiguration.cs:43`; `RateLimitCheck` returns null when disabled, `SSOController.cs:1616-1619`).
- Each challenge mints one CSPRNG-keyed state that lives 15 minutes (`StateLifetime`). A sustained ~111 req/s fills the 100k global store within the lifetime window and keeps it full.
- Once full, every legitimate user's `OidChallenge` hits the cap and returns HTTP 500 "Could not start login" (`SSOController.cs:312-326`), a total OpenID login outage for all users and all providers until the flood stops, plus up to a 15-minute tail.
- SAML analogue: with `ValidateInResponseTo` enabled, `SamlRequestCache` fills the same way and legitimate callbacks are then denied at `SamlRequests.TryConsume`.

The cap itself is a deliberate memory-DoS defense (from the unbounded-growth work, #246) and must **not** be removed; blind eviction would drop users already mid-login (a different lockout). The real gap is the absence of per-client fairness, compounded by the documented mitigation (rate limiting, #128) shipping off by default.

## Fix (fail closed, without enabling mass lockout)

- Add a per-source sub-cap: track in-flight state/request counts per normalized client key (`SsoRateLimiter.NormalizeClientKey`) and bound each key to a small share of the global budget, so no single source can exhaust it. Memory stays bounded (fail closed) and legitimate clients keep their own headroom.
- Separately, weigh whether the challenge/callback rate limiter should default on.

## Acceptance criteria

- [ ] A single client key cannot occupy more than its per-source share of the in-flight state/request stores.
- [ ] A flood from one source no longer refuses new logins from a different source (test with two client keys).
- [ ] Memory remains bounded under a flood (the global cap still holds).
- [ ] Applies symmetrically to the OpenID state store and the SAML request cache.

Refs #128 (rate limiter, mitigation, default-off), #246 (the global cap this refines), #194 (client-key normalization).

iderex:
## What & why

The OpenID authorize-state store and the SAML outstanding-request cache are globally capped (100k) so an abandoned-login flood cannot exhaust memory, but with **no per-client accounting** a single anonymous source could fill the whole budget and, because the stores refuse-at-cap rather than evict, lock out **every other login** (an availability-DoS). The built-in rate limiter (#128) is off by default, and the challenge endpoints are anonymous.

Add a **per-client occupancy sub-cap**. A shared `PerClientBudgetLimiter` reserves one slot per normalized client key before a store inserts and releases it on every removal, so one key holds at most **1% of the global cap (1000)**. A **null key** (a proxy/private/unattributable source, per `SsoRateLimiter.NormalizeClientKey`) is **exempt**, it is the whole userbase behind one hop and must not be sub-capped, and stays bounded only by the global cap.

Closes #327.

## Type of change
- [x] Security hardening (availability / anti-DoS on the login-path state stores)

## Security checklist
- **Fail closed:** reserve before the global insert; roll back on any failure; the global refuse-at-cap-not-evict is unchanged. The reserve is a single CAS (cap-check + increment as one atom) → a key **cannot race past its cap** (pinned by a 16-thread test asserting exactly `cap` reservations succeed); `Release` uses a value-conditional remove → never deletes a live bucket, never goes negative. Every removal path (redeem, consume, both prunes, both add-rollbacks, clear) releases **winner-only, exactly once**, no leak, no double-decrement, so the count is always `≥ live-entries` (over-admit is structurally impossible).
- **No new mass-lockout:** the null/proxy exemption keeps the shared-proxy userbase off the sub-cap.
- No per-request lock on the anonymous hot path; O(1) per add; counts are in-memory only (no config/schema change).

## Quality checklist
- One shared, single-purpose helper (`PerClientBudgetLimiter`, named to satisfy the `MutableKeyedState` conformance rule); both stores compose it symmetrically.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **825/825 green** (21 new tests: the CAS race-free cap under 16-thread contention, two-key fairness, release-on-{redeem,consume,expiry,failed-insert}, null-key exemption, refuse-not-evict, leak-free duplicate id).
- **Full 4-lens adversarial gate:** bypass / correctness / integration **PASS** (over-admit structurally impossible; concurrency-correct; integration-sound). availability first returned NEEDS_IMPROVEMENT, **no code defect**, two residuals: the public-proxy-without-Known-proxies shared bucket was undocumented, and the SAML refusal was silent vs OIDC's throttled warning. **Fixed:** a throttled SAML capacity warning (parity, non-amplifying) and a new `providers.md` "In-flight login capacity" section with the Known-proxies remedy; availability re-review appended below.

## Notes
The design was chosen via a 3-approach design panel (unanimous 9/9/9 for the shared-limiter). The SAML `Register` switched from an indexer overwrite to `TryAdd` (now refuses a duplicate id, a documented tightening that keeps the reservation rollback leak-free; request ids are fresh CSPRNG).

### DIFF
--- a/SSO-Auth.Tests/OidcStateStoreTests.cs
+++ b/SSO-Auth.Tests/OidcStateStoreTests.cs
@@ -359,7 +359,7 @@ public class OidcStateStoreTests
             {
                 for (var i = 0; i < 5000; i++)
                 {
-                    store.TryAdd(new AuthorizeState { State = "fresh-" + i }, "p", false, Now, Binding, out _);
+                    store.TryAdd(new AuthorizeState { State = "fresh-" + i }, "p", false, Now, Binding, clientKey: null, out _);
                 }
             },
             TestContext.Current.CancellationToken);
@@ -387,8 +387,8 @@ public class OidcStateStoreTests
     {
         var store = Store(maxEntries: 2);
 
-        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, out var warnA));
-        Assert.True(store.TryAdd(new AuthorizeState { State = "b" }, "p", false, Now, Binding, out var warnB));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, out var warnA));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "b" }, "p", false, Now, Binding, clientKey: null, out var warnB));
         Assert.False(warnA);
         Assert.False(warnB);
         Assert.Equal(2, store.Count);
@@ -398,10 +398,10 @@ public class OidcStateStoreTests
     public void TryAdd_AtCap_RefusesANewKeyAndKeepsTheInFlightState()
     {
         var store = Store(maxEntries: 1);
-        Assert.True(store.TryAdd(new AuthorizeState { State = "in-flight" }, "p", false, Now, Binding, out _));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "in-flight" }, "p", false, Now, Binding, clientKey: null, out _));
 
         // At the cap, a fresh challenge is refused rather than evicting the user already mid-login.
-        Assert.False(store.TryAdd(new AuthorizeState { State = "new" }, "p", false, Now, Binding, out _));
+        Assert.False(store.TryAdd(new AuthorizeState { State = "new" }, "p", false, Now, Binding, clientKey: null, out _));
         Assert.Equal(1, store.Count);
         Assert.NotNull(store.PeekCurrent("in-flight", "p", Now, Binding));
     }
@@ -410,8 +410,8 @@ public class OidcStateStoreTests
     public void TryAdd_DuplicateKey_ReturnsFalse()
     {
         var store = Store(maxEntries: 10);
-        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, out _));
-        Assert.False(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, out _));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, out _));
+        Assert.False(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, out _));
     }
 
     [Fact]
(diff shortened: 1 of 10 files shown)


iderex:
Pre-merge review, the full adversarial-review gate is green.

**4-lens adversarial gate (login-path state-store change):**
- **bypass, PASS**: over-admit is structurally impossible (count ≥ live-entries invariant; single-CAS reserve; value-conditional release), null keys short-circuit before any dictionary access, the TryAdd tightening closes rather than opens a vector, victim-pinning is limited to the same-egress case the rate limiter already accepts.
- **correctness, PASS**: single-CAS reserve gives provable zero-overshoot; release never deletes a live bucket nor goes negative; all removal sites release winner-only (no double-release under concurrent redeem/consume/prune); reserve-before-insert + rollback leaks nothing; 16-thread contention test pins it.
- **integration, PASS**: NormalizeClientKey null-exempt behavior matches the existing rate limiter; PerClientBudgetLimiter satisfies the MutableKeyedState conformance rule; no config/persistence change.
- **availability, PASS (after fixes)**: first flagged two non-code residuals, the SAML refusal was silent vs OIDC, and the public-proxy shared-1000-bucket was undocumented. Both fixed: a throttled (non-amplifying) SAML capacity warning for parity, and a new providers.md "In-flight login capacity" section documenting the Known-proxies remedy. Re-review confirmed both resolved with no new availability surface.

825/825 tests. Merging once CI is green.

iderex:
Merge note: all four required checks (build, prettier, SonarCloud quality gate, dependency-review) pass; the only red is `submit-nuget` (the Automatic Dependency Submission, a non-required check posting to the GitHub dependency-graph API), which is a documented infra flake during the ongoing GitHub REST-API 503 instability (same as #426; this PR builds cleanly). Full 4-lens adversarial gate green, 825/825 tests. Merging on the required stack.

PR #428, opened 2026-07-16
Unify the two role-privilege mappers behind ProviderConfigBase

### LINKED ISSUE #367: Unify the twin role-privilege mappers behind ProviderConfigBase
`OidcRolePrivilegeMapper` (SSO-Auth/Api/OidcRolePrivilegeMapper.cs:16-89) and `SamlRolePrivilegeMapper` (SSO-Auth/Api/SamlRolePrivilegeMapper.cs:20-100) perform the same fold over the same `ProviderConfigBase` members, ~90 lines each, and both declare a nested record named `RoleGrants` with different shapes (the OIDC one adds `Valid`). Two same-named types in one namespace defeat IDE navigation, and a contributor fixing role logic must diff the two classes by eye to find the real differences: three documented quirks (comparison receiver, folder-role null-guard, folder-role trim). The OIDC mapper's comment (:46-48) says the null-guard quirk is "Tracked in #89", but #89 is closed, the divergence is currently tracked nowhere.

Proposal:

- First decide the quirks: reconcile to one behavior (a login-path behavior change, runs the adversarial gate) or keep them as explicit, named parameters of a single mapper.
- Then collapse to one `RolePrivilegeMapper.Evaluate(IEnumerable<string> roles, ProviderConfigBase config)` returning one `RoleGrants`; the SAML caller ignores `Valid` (`SamlLoginPolicy` owns that decision). Every consumed member already lives on `ProviderConfigBase` (SSO-Auth/Config/PluginConfiguration.cs:110-152), so no new interface is needed. Saves ~90 lines.
- The V2 reference shows the unified shape (`RoleAuthorization.cs`, one mapper for both protocols).
- Either way, fix the stale #89 reference in the comment.

iderex:
## What & why

`OidcRolePrivilegeMapper` and `SamlRolePrivilegeMapper` folded the same `ProviderConfigBase` members with ~90 lines each and each declared a same-named nested `RoleGrants` (defeating IDE navigation), forcing a contributor to diff two classes by eye to find the three documented behavioral quirks.

Collapse to one `RolePrivilegeMapper.Evaluate(IEnumerable<string>, ProviderConfigBase)` returning one `RoleGrants` (the SAML caller ignores `Valid`, whose decision belongs to `SamlLoginPolicy`). Both authorize-state builders call the one mapper. Closes #367.

**Reconciles the three per-protocol quirks to one robust behavior, a deliberate login-path authz change.** All three moves are toward graceful fail-closed / trusted-config normalization, and none is reachable as an over-grant via IdP-controlled input:

1. **Comparison** → `string.Equals(role, allowed, Ordinal)` (null-safe both ways): a null role or a null configured entry is a non-match, not a `NullReferenceException`/500.
2. **Folder-mapping null-guard** → guarded in both: a null `FolderRoleMapping` with folder roles enabled grants no folders instead of failing the login with a 500 (an admin misconfig).
3. **Folder-role trim** → the trusted, admin-authored configured map-role is trimmed in both; the **IdP-supplied role is compared raw**, so there is no whitespace-injection vector. SAML now matches a padded configured role (a normalization of a trusted value).

## Type of change
- [x] Refactor / code quality, with a deliberate, gated authz behavior reconciliation

## Security checklist
- **Fail-closed preserved:** every reconciliation either turns a 500 into a graceful no-grant or normalizes only the trusted config side; the attacker-controlled IdP role is never trimmed or altered, so no privilege becomes reachable that was not before. `Valid` is still computed from `config.Roles`; the SAML caller ignores it (`SamlAuthorizeStateBuilder`).

## Quality checklist
- One mapper, one `RoleGrants`; ~90 lines removed; the stale `#89` mapper comment retired.

## Verification
- `dotnet build --no-incremental --warnaserror`: 0/0. `dotnet test`: **814/814 green**, a consolidated `RolePrivilegeMapperTests` covers both config types and pins each reconciled quirk; the FsCheck monotonicity properties run against the unified mapper.
- **Full 4-lens adversarial gate** (login-path authz change), results appended below before merge.

## Summary by CodeRabbit

* **Refactor**
  * Unified SAML and OpenID Connect role-to-privilege evaluation into a shared mapper.
  * Preserved admin, folder, Live TV, and management privilege behavior (including allow/admin list handling).
  * Improved null-safe list matching and clarified how validity is derived for SAML.
* **Tests**
  * Removed provider-specific mapper test suites and added unified role/grants coverage for both SAML and OpenID Connect.
  * Added checks for case/ordinal matching, folder-role trimming behavior, duplicate/accumulated folder grants, and Live TV enablement and null-safety.


iderex:
Pre-merge review, the full adversarial-review gate is green (all four lenses PASS).

- **bypass, PASS**: only the trusted admin-authored config side is trimmed; the IdP-supplied role is compared raw, so there is no whitespace-injection vector. Every reconciliation only turns an admin-misconfig 500 into a correct evaluation, and every resulting grant stays gated on an exact ordinal match against a configured allow-list entry, no attacker-controlled input reaches a privilege it could not already reach.
- **correctness, PASS**: faithful, behavior-preserving unification for all non-quirk inputs; the three reconciliations are exactly as described; ordinal matching, folder accumulation order/duplicates, live-tv gating, and monotonicity preserved.
- **availability, PASS**: quirks 1–2 are pure availability gains (a formerly-500 login now succeeds). The one grant-loss is the implausible doubly-padded SAML folder-role edge (config and IdP role identically whitespace-padded), it degrades access rather than failing the login. Per the reviewer's (non-blocking) recommendation, added a characterization test pinning that shift on the record.
- **integration, PASS**: binds for both protocols via ProviderConfigBase; no dangling refs to the deleted mappers/duplicate RoleGrants; property tests exercise it non-vacuously; stale #89 mapper comment gone; no conformance rule or doc affected.

815/815 tests. Merging once CI is green.

iderex:
All 16 checks green, SonarCloud OK, 815/815 tests, full 4-lens gate PASS. Merging on green.

PR #431, opened 2026-07-17
Make the provider form's save contract explicit and drop the dead sso-json path

### LINKED ISSUE #365: Make the provider form's save contract explicit and drop the dead sso-json path
The provider form's persistence contract is entirely implicit. `listArgumentsByType` (SSO-Auth/Config/config.js:169-206) scans `#sso-new-oidc-provider` for `sso-text`/`sso-toggle`/`sso-line-list` classes, and `saveProvider` (config.js:317-354) writes `current_config[id]`, the DOM id must be spelled exactly like the `OidParameter` property it persists. A field with a wrong id, a missing class, or placed outside that form renders fine and silently never saves (unknown JSON members are also dropped server-side). Nothing in config.js or configPage.html states any of this, and no conformance test covers it, it drift-breaks with no error.

On top of that, the `.sso-json` machinery is dead: config.js:170 is the only occurrence of `sso-json` in the repo, no element carries the class, so `json_fields` is always empty and the load loop (:234-237) plus the save loop with its unguarded `JSON.parse` (:326-333) never run. It is the scariest-looking code in the file and it is unreachable; the V2 config.js already dropped it.

Proposal:

- Delete the `sso-json` selector, both loops, and `json_fields` from the returned object (~20 lines).
- State the invariant in one comment block above `listArgumentsByType` and one HTML comment atop the `#sso-new-oidc-provider` form.
- Lock it in as a fitness function: a conformance test that parses the embedded configPage.html and asserts every `sso-*`-classed element id is a real `OidParameter` property (reflection), folded into the existing ArchitectureConformanceTests rather than a new harness.

iderex:
# What & why

The admin provider form's persistence contract was entirely implicit. `listArgumentsByType`
scans `#sso-new-oidc-provider` for marker classes and `saveProvider` writes
`current_config[element.id]`, so each field's DOM id must be spelled exactly like the
`OidConfig` property it persists, but nothing stated this and no test covered it, so it
drift-breaks silently: a field with a mistyped id, a missing marker class, or placed outside
the form renders fine and never saves (the server also drops unknown JSON members). On top of
that, the `sso-json` machinery was dead, `config.js` held the only occurrence of the class in
the repo, so `json_fields` was always empty and its two loops (including an unguarded
`JSON.parse`) were unreachable.

This makes the contract explicit and locks it in:

- **Drop the dead `sso-json` path** from `config.js`: the `json_class` selector, the
  `json_fields` query, its member in the returned object, and the two load/save loops.
- **Document the contract**: a comment block above `listArgumentsByType` and an HTML comment
  atop the `#sso-new-oidc-provider` form.
- **Remove `class="sso-text"` from `#OidProviderName`.** Its value is the `OidConfigs`
  dictionary *key*, not an `OidConfig` property; the marker only caused a redundant
  `current_config["OidProviderName"]` write that the server already dropped. Removing it makes
  the persisting-field set *exactly* the `OidConfig` properties, so the fitness function needs
  no exception. Behavior-preserving: the name field is set (from the dict key) and read
  explicitly, and `.sso-text` is a JS marker with no CSS (styling comes from `is="emby-input"`).
- **Lock it in as a fitness function** in `ArchitectureConformanceTests`
  (`ProviderFormFieldIds_MatchOidConfigProperties`): it parses `configPage.html`, scans the
  form for elements bearing a persisting marker class
  (`sso-text`/`sso-line-list`/`sso-toggle`/`sso-folder-list`/`sso-role-map`), and checks their
  ids against `OidConfig` in **both directions**, every marked id is a real property, and every
  security-critical setting is still a marked, correctly-typed field. Matching is token-exact
  (so `sso-role-map` does not swallow `sso-role-mapping-container`), scoped to
  `#sso-new-oidc-provider` (a future SAML form maps to `SamlConfig`), and guarded against a
  vacuous pass by per-marker-class sentinel ids.

The issue text names an `OidParameter` type; the actual reflection target is `OidConfig` (there
is no `OidParameter`).

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the change is behavior-preserving, and the new reverse-direction
      pin actively defends the save path of the security toggles (`DisableHttps`,
      `DoNotValidate*`, `RequirePkce`, …) and `OidSecret` — a dropped marker class now fails CI.
- [x] No secrets logged; secrets are redacted on config export. No logging touched; `OidSecret`'s
      write-only converter is untouched and it is now pinned as a must-persist field.
- [x] A security review was performed: the 4-lens refute-by-default adversarial gate
      (bypass / correctness / integration / availability) returned all PASS.
- [x] Log inputs from the IdP are sanitized inline at the log call - no log calls added or
      changed.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose, testable unit.
- [x] The change adds no more code than the problem requires (net removes dead code; adds one
      focused fitness function).
- [x] The code is self-documenting (readable without comments; comments explain why, not what)
      and matches the target architecture (#318).
- [x] Architecture-conformance tests pass, and the new structural property this change
      establishes is locked in as a rule in `ArchitectureConformanceTests`.

## Verification

- [x] `dotnet test` is green (816/816).
- [x] Prettier is clean for the `.js` / `.html` change.

## Notes

CI cannot execute the browser JS/HTML, so the adversarial-review gate is the primary
verification. All four lenses PASS. Two non-blocking suggestions were **incorporated** rather
than deferred:

- *bypass*, the forward check doesn't catch a security toggle *losing* its marker class (the
  operation this PR performs on the name field). Added a reverse-direction pin: a fixed roster of
  security-critical properties must each remain a marked, correctly-typed field. Empirically
  verified, temporarily dropping the `DisableHttps` marker class fails the build (1/816), then
  restored.
- *correctness*, tightened the id regex to `(?<![-\w])id=` so a future `data-id`-style
  attribute can't be misread as the element id.

Out of scope (pre-existing, not a regression): `AllowExistingAccountLink` is an `OidConfig`
security toggle with no admin-form field, so it cannot be enabled from the UI, the test is
one-directional there by design (server-managed `NewPath`/`CanonicalLinks` and the UI-absent
`AllowExistingAccountLink` are legitimately not form fields). Worth a follow-up on whether that
toggle should be surfaced.

Closes #365

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved saving and loading of OIDC provider settings by aligning form fields with their corresponding configuration properties.
  - Prevented provider names from being incorrectly treated as provider settings.

- **Documentation**
  - Added guidance explaining which provider form fields are persisted and how they must be configured.

- **Tests**
  - Added validation to ensure provider form fields remain synchronized with supported OIDC configuration properties and required security settings.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Text.RegularExpressions;
 using Jellyfin.Plugin.SSO_Auth;
 using Jellyfin.Plugin.SSO_Auth.Config;
 using MediaBrowser.Model.Plugins;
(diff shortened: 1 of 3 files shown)


iderex:

We ran our 4-lens refute-by-default adversarial review. The provider settings page is browser JS/HTML that CI cannot execute, so this review is the primary verification for the behavior claims.

- **bypass, PASS.** No security toggle or secret loses its save path; the removed `sso-json` path never carried one; the dropped `OidProviderName` write was an unmapped JSON member the server already discarded. The review's one concern, the forward check does not catch a security toggle *losing* its marker class, we closed in this PR with a reverse-direction pin.
- **correctness, PASS.** `sso-json` was provably always dead (never on any element in this repo's history); removing `sso-text` from the name input is behavior-preserving on load and save for both new and existing providers; the fitness function is token-exact, correctly scoped, and non-vacuous. Full suite 816/816.
- **integration, PASS.** The embedded resource equals the on-disk source the test scans; unmapped-member deserialization is unchanged; the test is correctly one-directional (server-managed `NewPath`/`CanonicalLinks` and the UI-absent `AllowExistingAccountLink` are legitimately not form fields); Prettier clean.
- **availability, PASS.** No configuration that previously saved or loaded now fails; the fitness function fails only on a genuine contract violation, never a false positive on conforming markup.

Both non-blocking suggestions from the review were folded in, not deferred: the reverse-direction security-critical pin (empirically verified, temporarily dropping the `DisableHttps` marker class fails the build 1/816, then restored) and a tightened id regex.

Local gates: `dotnet build --warnaserror` clean, `dotnet test` 816/816, Prettier clean. Merging once the required checks (build, Prettier, SonarCloud, dependency-review, CodeQL) are green.

PR #434, opened 2026-07-17
Anchor OidcCallbackPath.RedirectSegment to the route suffix

### LINKED ISSUE #411: Anchor OidcCallbackPath.RedirectSegment to the route suffix like ChallengePath
Follow-up from #337 / PR #410 (CodeRabbit).

`OidcCallbackPath.RedirectSegment` (SSO-Auth/Api/OidcCallbackPath.cs) scans for the **first** `OID` segment anywhere in the path and reads the next segment. A protocol-like reverse-proxy prefix (e.g. a base path containing an `OID` segment) could therefore misclassify the callback spelling. It is operator-controlled and still fails closed at the IdP, but #410 moved the sibling `ChallengePath.IsNewPath` to a **suffix-anchored** form (inspect the final `{protocol}/{path-kind}/{provider}` tuple, trimming only boundary slashes). Apply the same treatment here so the two converge on the robust pattern.

Add regression cases: a protocol-like reverse-proxy prefix, and an internal doubled slash.

Refs #337, #98.

iderex:
# What & why

`OidcCallbackPath.RedirectSegment` scanned for the **first** `OID` segment
anywhere in the callback path and read the next segment, so a protocol-like
reverse-proxy prefix (a base path that itself contains an `OID` segment) could
misclassify the callback spelling. PR #410 already moved the sibling
`ChallengePath.IsNewPath` to a **suffix-anchored** form, it reads the route off
its `{protocol}/{path-kind}/{provider}` suffix, trimming only boundary slashes.
This converges `RedirectSegment` on the same robust pattern so the challenge side
and the callback side classify the route identically.

The value only picks the `r`/`redirect` segment used to rebuild the token
request's `redirect_uri`; the segment is operator/proxy-controlled and a mismatch
still fails closed at the IdP token endpoint (RFC 6749 §4.1.3), so this is a
robustness tightening, not a behavior change for well-formed callbacks.

- Read the route off the `{protocol}/{path-kind}/{provider}` suffix
  (`segments[^3]`/`[^2]`) instead of the first `OID` token found anywhere.
- Trim only boundary slashes (drop `RemoveEmptyEntries`) so an internal doubled
  slash shifts the suffix and fails safe to the classic `"r"` spelling rather
  than being silently collapsed into a valid route.
- Add regression tests: a protocol-like reverse-proxy prefix (both directions),
  a legitimate base-path mount that must stay `"redirect"`, and an internal
  doubled slash.

Closes #411.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. — Unchanged. The classifier picks only
      the redirect segment; a wrong choice yields a mismatched `redirect_uri`
      that the IdP rejects at the token endpoint, and the `"r"` default is the
      fail-safe classic spelling.
- [x] No secrets logged; secrets are redacted on config export. - No logging or
      secret handling in this path.
      config persistence, or the release pipeline. — Adversarial 4-lens review
      ran (verdicts under Notes).
- [x] Log inputs from the IdP are sanitized inline at the log call. - N/A, no
      logging added.

## Quality checklist

      testable unit.
      — No new structural property; the pure-helper shape is already covered.

## Verification

- [x] `dotnet test` is green., 820 passed, 0 failed.
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change., N/A,
      only `.cs` files touched.

## Notes

**Adversarial 4-lens review, all PASS.**

- **Availability / mass-lockout, PASS.** Suffix-anchoring leaves the common-case
  classification identical and correct; it fails safe only on paths a real OIDC
  callback can never produce, and it aligns the callback side with the
  already-merged challenge side (#410), removing a latent asymmetry rather than
  introducing a lockout vector.
- **Security-bypass / fail-open, PASS.** The return is a literal (`"redirect"`
  or `"r"`) that no security decision keys on; a misclassification fails at the
  IdP and never authenticates. The `"r"` default is fail-closed; no
  bypass/fail-open is reachable.
- **Correctness, PASS.** The `Length >= 3` guard makes the `segments[^3]`/`[^2]`
  indexing throw-proof; null/empty/`"/"`/`"//"`/single-segment/trailing-slash/
  mixed-case inputs all handled; `OrdinalIgnoreCase` avoids culture pitfalls;
  behavior is identical to the old code for every well-formed callback path.
- **Integration, PASS.** The OID challenge (`OID/start|p/{provider}`) and
  callback (`OID/redirect|r/{provider}`) routes share the same
  `{controller}/OID/{path-kind}/{provider}` shape; `Request.Path` has the
  reverse-proxy base path stripped into `PathBase`, so the suffix window is
  stable. Both helpers now derive their `redirect_uri` spelling from a
  byte-identical routine, so the challenge-advertised and callback-rebuilt
  `redirect_uri` match under any base-path or proxy layout.

Docs: no operator-visible behavior change (it only tightens misclassification of
an operator-controlled path that already fails closed), so no user-facing doc
update is needed.

Out of scope: `ChallengePath.cs`'s class comment still says `RedirectSegment`
"should be moved to the same suffix-anchored form (#411)", after this change
that line is stale, but `ChallengePath.cs` is out of scope here; folding the
one-line comment fix into the next touch of that file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSO callback path detection when applications run behind reverse proxies or URL prefixes.
  * Correctly identifies supported callback routes while safely falling back for malformed paths, including doubled slashes.
* **Tests**
  * Added coverage for prefixed callback paths, valid reverse-proxy routes, and malformed callback URLs.

### DIFF
--- a/SSO-Auth.Tests/OidcCallbackPathTests.cs
+++ b/SSO-Auth.Tests/OidcCallbackPathTests.cs
@@ -42,6 +42,36 @@ public class OidcCallbackPathTests
         Assert.Equal("redirect", OidcCallbackPath.RedirectSegment("/sso/OID/redirect/redirect"));
     }
 
+    [Theory]
+    // A protocol-like reverse-proxy prefix that itself spells "OID/redirect" must not flip the
+    // classic route: only the {protocol}/{path-kind}/{provider} suffix decides, so the real "r"
+    // route stays "r" no matter what the mount prefix looks like (#411).
+    [InlineData("/OID/redirect/proxy/sso/OID/r/keycloak", "r")]
+    // The mirror: a prefix spelling "OID/r" must not flip a real new-path route away from "redirect".
+    [InlineData("/OID/r/proxy/sso/OID/redirect/keycloak", "redirect")]
+    public void RedirectSegment_ProtocolLikePrefix_LetsTheSuffixDecide(string path, string expected)
+    {
+        Assert.Equal(expected, OidcCallbackPath.RedirectSegment(path));
+    }
+
+    [Fact]
+    public void RedirectSegment_LegitReverseProxyBasePath_KeepsNewPathRoute()
+    {
+        // A benign reverse-proxy base path in front of a real new-path callback keeps the intact
+        // OID/redirect/{provider} suffix, so it is still correctly "redirect" — the suffix anchoring
+        // must not misclassify legitimate mounts.
+        Assert.Equal("redirect", OidcCallbackPath.RedirectSegment("/jellyfin/apps/sso/OID/redirect/keycloak"));
+    }
+
+    [Fact]
+    public void RedirectSegment_InternalDoubledSlash_FailsSafeToClassic()
+    {
+        // An internal doubled slash inserts an empty segment that shifts the suffix, so the malformed
+        // path no longer matches OID/redirect/{provider} and fails safe to the classic "r" spelling
+        // rather than being silently collapsed into a valid new-path route (#411).
+        Assert.Equal("r", OidcCallbackPath.RedirectSegment("/sso/OID/redirect//keycloak"));
+    }
+
     [Fact]
     public void RedirectSegment_EmptyPath_DefaultsToClassic()
     {
--- a/SSO-Auth/Api/OidcCallbackPath.cs
+++ b/SSO-Auth/Api/OidcCallbackPath.cs
@@ -10,7 +10,10 @@ namespace Jellyfin.Plugin.SSO_Auth.Api;
 /// IdP delivers the callback on exactly the route the authorization request advertised — so the
 /// segment is read off the callback's own path. The previous expression tested for "/start/", a
 /// challenge-route segment that never occurs on callback routes, so the new-path flow sent a
-/// mismatched redirect_uri that spec-enforcing IdPs reject (#98).
+/// mismatched redirect_uri that spec-enforcing IdPs reject (#98). The route is now read off its
+/// <c>{protocol}/{path-kind}/{provider}</c> suffix — the same robust form as
+/// <see cref="ChallengePath.IsNewPath"/> — so a protocol-like reverse-proxy prefix cannot decide
+/// the spelling (#411).
 /// </summary>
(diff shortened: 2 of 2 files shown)


iderex:

**Adversarial 4-lens review (login-path change, RedirectSegment classifies the OIDC callback spelling): all PASS.**

- Availability / mass-lockout, **PASS**: suffix-anchoring leaves the common-case classification identical and correct, fails safe only on paths a real OIDC callback cannot produce, and aligns the callback side with the already-merged challenge side (#410), removing a latent asymmetry, not adding a lockout vector.
- Security-bypass / fail-open, **PASS**: the return is a literal (`redirect`/`r`) that no security decision keys on; a misclassification fails at the IdP token endpoint and never authenticates, and the `r` default is fail-closed.
- Correctness, **PASS**: the `Length >= 3` guard makes the `[^3]`/`[^2]` indexing throw-proof; null/empty/`/`/`//`/single-segment/trailing-slash/mixed-case all handled via `OrdinalIgnoreCase`; behavior is identical to the old code for every well-formed callback path.
- Integration, **PASS**: the OID challenge and callback routes share the `{controller}/OID/{path-kind}/{provider}` shape, `Request.Path` has the reverse-proxy base path stripped into `PathBase`, and both helpers now derive the `redirect_uri` spelling from a byte-identical routine, so the challenge-advertised and callback-rebuilt `redirect_uri` match under any base-path or proxy layout.

**CI / gates:** build green; `dotnet test` 820 passed / 0 failed; CodeQL (csharp/js/actions) green; SonarCloud quality gate green; dependency-review, prettier, abi-floor, greppable-invariants, Trojan-Source, zizmor all green. `mergeStateStatus: CLEAN`.

**Docs:** no operator-visible behavior change (tightens misclassification of an operator-controlled path that already fails closed), no user-facing doc update needed.

**Out of scope:** `ChallengePath.cs`'s class comment still says `RedirectSegment` "should be moved to the same suffix-anchored form (#411)", now stale, folded into the next touch of that file.

Overall verdict: **ready to merge.**

Issue #354, opened 2026-07-14
OIDC legacy-link migration trusts the mutable username, enabling account takeover


shippingToken:
## Summary (CWE-287)

The OpenID legacy-link migration (#155) resolves a login by looking a canonical-links entry up **by the mutable `username` (`preferred_username`)**, and when that lookup hits a live user it returns that user's account **before the `AllowExistingAccountLink` adoption gate is consulted**, re-keying the link to the new `sub`. Because `preferred_username` is chosen/assigned at the identity provider and is not a stable unique identity, an actor who can obtain a `preferred_username` equal to a victim's legacy link key can be handed the victim's Jellyfin account.

Surfaced by review on PR #353 (CanonicalLinkService extraction); the logic is **pre-existing**, the extraction moved it verbatim, so #353 does not change this behavior. Filed for its own reviewed fix rather than folding a security-semantics change into a behavior-preserving refactor.

## Mechanism

In `CanonicalLinkService.ResolveOrCreateAsync` (moved from `SSOController.CreateCanonicalLinkAndUserIfNotExist`):

```
Guid? byName = bySubject is null
    && !string.Equals(canonicalKey, username, Ordinal)
    && links.TryGetValue(username, out var n) && GetUserById(n) != null ? n : null;
```

`AccountLinkResolver.ResolveCanonicalLink(subjectLink: null, legacyLink: byName)` returns `(byName, MigrateLegacy: true)`, and `AccountLinkResolver.Resolve(linkedUserId: byName, …)` short-circuits to `UseExistingLink` **without** checking `AllowExistingAccountLink` (a resolved link is treated as an established identity). So a login with `sub = X`, `preferred_username = U` is handed the user that the entry keyed `U` points at, and the link is re-keyed to `X`.

The entry keyed `U` can be:
- a genuine **legacy** entry (the intended migration), but the person logging in with `preferred_username = U` need not be the same person whose old username was `U` (usernames get reassigned, or an attacker sets theirs to `U`); or
- a **modern subject-keyed** entry whose key coincidentally equals `U` (IdPs that use usernames/emails as `sub`).

**SAML is not affected:** it passes `canonicalKey == username == NameID`, so the `!string.Equals(canonicalKey, username)` guard disables the `byName` branch.

[...]

PR #435, opened 2026-07-17
Require the legacy link's target to still bear the presented name before following it

### LINKED ISSUE #361: Legacy-migration flag-ON arm matches the recorded username after a Jellyfin-side rename (stale-name superset)
Follow-up from the review on #358 (raised independently by two reviewers).

With `AllowExistingAccountLink` **on**, the legacy migration arm resolves the recorded username key to its target user as long as that user still exists (`GetUserById != null`), regardless of whether the account still bears that name. Same-name adoption, by contrast, requires a live account currently named that value (`GetUserByName`). So the legacy arm is a strict superset of same-name adoption: it can hand out an account whose current display name does not match the presented `preferred_username`.

Reachable scenario (flag on, the state the #354 upgrade runbook tells admins to enable during the migration window): legacy map `{"oldname": victim}`; the victim is later renamed at the IdP to `newname` (so `GetUserByName("oldname")` is null). An attacker whose token carries `sub = attacker`, `preferred_username = oldname` logs in: `bySubject` null, `canonicalKey != username`, `links["oldname"] = victim`, victim still exists → migrate + `UseExistingLink` → the attacker is handed and re-keys the victim's account.

This is pre-existing #155 behavior, not introduced by #358 (whose scope is the flag-OFF hole, fully closed). It is filed because #358's own guidance encourages enabling the flag during the upgrade window, so the residual is worth closing.

Directions to weigh (decide in the fix PR):
- In the flag-on arm, additionally require that the legacy target still bears the presented name (`GetUserByName(username)?.Id == recordedUserId`) before following it, collapses the arm back to true same-name matching.
- Or make subject and legacy keys structurally distinguishable (a `sub:` / `name:` prefix), which also closes the shared-namespace ambiguity (finding INFO-2 on #358).

Add a service-level test pinning the flag-ON + renamed-legacy-owner path.

Refs #354, #155, #318

iderex:
# What & why

The OpenID legacy username-keyed link (#155) was followed whenever its recorded target user still
**existed** (`GetUserById != null`), without checking the target still **bore** the presented name.
With `AllowExistingAccountLink` on, the state the #354 migration runbook tells admins to enable
during the upgrade window, an attacker presenting a foreign `sub` and a victim's **old** username
(one the victim had since been renamed away from) was handed and re-keyed the victim's account: a
stale-name superset of same-name adoption (#361, CWE-287).

The fix gates the follow on the name still resolving to the same target:

- `AccountLinkResolver.ResolveCanonicalLink` takes a `legacyNameStillHeldByTarget` flag; the migrate
  arm now requires it alongside the existing opt-in. When the recorded target no longer bears the
  name, the legacy link is **not** followed and the login falls through to the adoption gate over the
  **current** name-holder - adopt it (flag on, a live account bears the name), or create/reject - 
  never the renamed-away account.
- `CanonicalLinkService` resolves the display-name holder **once** (moved before the resolve) and
  passes whether it is the legacy target (`existingAccountUserId == legacyLink`). `legacyLink` itself
  is unchanged ("target exists"); only the follow/migrate is newly gated. The `CreateNewAccount`
  orphan warning, now reachable with the flag on too, is reworded flag-agnostic (still names the
  orphaning).
- `providers.md`: a Jellyfin-side-renamed account is no longer flag-migratable; recover it by admin
  re-linking. The migration window now only self-migrates links whose account still bears the
  recorded name.

SAML is unaffected, its canonical key equals the name, so the legacy arm is structurally dormant.

Case matrix (subject link absent, target exists): **(1)** the account currently named `<username>`
IS the legacy target → flag on migrates (unchanged), flag off rejects; **(2)** a *different* account
now bears `<username>` → adopt that current holder (flag on) or reject (flag off), never the renamed
victim; **(3)** nobody bears `<username>` → create fresh + orphan warning, the renamed victim keeps
their account.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the legacy link is now followed only when the presented name still
      resolves to the recorded target; every other case falls through to the fail-closed adoption gate.
- [x] No secrets logged; the reworded warning logs only the sanitized name/mode/provider (unchanged
      `ReplaceLineEndings` inline sanitization).
- [x] A security review was performed: 4-lens refute-by-default gate + the OIDC domain reviewer.
- [x] Log inputs from the IdP are sanitized inline at the log call (`?.ReplaceLineEndings(string.Empty)`).

## Quality checklist

- [x] No duplicated logic; the name-holder is resolved once and reused; the decision stays in the pure
      `AccountLinkResolver`.
- [x] The code is self-documenting; comments explain the security *why*.
- [x] Architecture-conformance tests pass; no new structural property introduced.

## Verification

- [x] `dotnet test` is green (818/818).

## Notes

Login-path authz change → the adversarial-review gate is load-bearing. We ran a 4-lens
refute-by-default gate plus the OpenID domain reviewer: **bypass, availability, correctness, and OIDC
all PASS**, the follow arm now collapses to genuine same-name matching (`existingAccountUserId ==
legacyLink`), the takeover is fully closed, no new 403/lockout is reachable (a flag-on hand-off only
ever narrows to adopt-current-holder or fresh-account), and SAML stays dormant. The **integration**
lens returned NEEDS_IMPROVEMENT on one point - the reworded WARNING paragraph over-claimed a warning
for *every* non-migrated legacy login, whereas the flag-on "adopt a different current name-holder"
outcome (Case 2) is recorded as an audit, not a warning, which we fixed in `providers.md` in this
same change.

Reject-path message nuance (pre-existing, unchanged): in the rare case 2 (target renamed away AND a
different account now holds the old name) with the flag off, the reject still emits the
"migratable pending-legacy-link" hint though enabling the flag would adopt the current holder rather
than migrate the victim. This reachability is identical to before the fix (`legacyLink` unchanged), so
it is not worsened here; left as-is to keep the change focused on the takeover.

Closes #361

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale legacy username links from being followed after an account is renamed.
  * Ensured logins adopt the account currently holding a legacy name rather than the former owner.
  * Improved handling of orphaned links, including administrator-led recovery guidance.

* **Documentation**
  * Clarified migration behavior, first-login ownership risks, renamed-account recovery, and warning messages when enabling existing account linking.

### DIFF
--- a/SSO-Auth.Tests/AccountLinkResolverTests.cs
+++ b/SSO-Auth.Tests/AccountLinkResolverTests.cs
@@ -61,7 +61,7 @@ public class AccountLinkResolverTests
     [InlineData(true)]
     public void ResolveCanonicalLink_SubjectKeyed_IsUsed_WithoutMigration(bool allow)
     {
-        var (linkedUserId, migrate) = AccountLinkResolver.ResolveCanonicalLink(Subject, legacyNameKeyedUserId: null, allow);
+        var (linkedUserId, migrate) = AccountLinkResolver.ResolveCanonicalLink(Subject, legacyNameKeyedUserId: null, legacyNameStillHeldByTarget: false, allowExistingAccountLink: allow);
         Assert.Equal(Subject, linkedUserId);
         Assert.False(migrate);
     }
@@ -71,8 +71,9 @@ public class AccountLinkResolverTests
     [InlineData(true)]
     public void ResolveCanonicalLink_SubjectKeyed_WinsOverLegacy_NoMigration(bool allow)
     {
-        // Once a subject-keyed link exists, the legacy name-keyed one is never consulted or migrated.
-        var (linkedUserId, migrate) = AccountLinkResolver.ResolveCanonicalLink(Subject, Legacy, allow);
+        // Once a subject-keyed link exists, the legacy name-keyed one is never consulted or migrated —
+        // even when the name is still held by its target (legacyNameStillHeldByTarget: true).
+        var (linkedUserId, migrate) = AccountLinkResolver.ResolveCanonicalLink(Subject, Legacy, legacyNameStillHeldByTarget: true, allowExistingAccountLink: allow);
         Assert.Equal(Subject, linkedUserId);
         Assert.False(migrate);
     }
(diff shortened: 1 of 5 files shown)


iderex:

This is a login-path account-takeover fix (#361), so the adversarial-review gate is load-bearing. We ran a 4-lens refute-by-default review plus the OpenID domain reviewer against the change.

- **bypass, PASS.** No sub/preferred_username/rename sequence hands over a renamed victim's account via the legacy key. The follow arm collapses to genuine same-name matching (`existingAccountUserId == legacyLink`, a lifted nullable value comparison guarded by `HasValue`), no `Guid.Empty`/`default` confusion is reachable, and Case 2 (adopt the current name-holder) is pre-existing flag-governed adoption, not a residual.
- **availability, PASS.** The common #155 migration (name unchanged) still self-migrates and is test-covered; the fix is provably incapable of producing a new 403 (it only ever narrows a flag-on hand-off into another successful outcome, adopt-current-holder or fresh-account). The sole degradation, a Jellyfin-side-renamed account no longer auto-migrates under the flag, is the intended security trade-off and is documented as admin-relink-only.
- **correctness, PASS.** Moving the name-holder read before the resolve introduces no TOCTOU/ordering regression (the migrate re-keys links, not usernames); both warning branches and all updated tests are accurate; SAML and the sub-fallback shape are unchanged.
- **OIDC domain, PASS.** The OpenID legacy arm is correctly reduced to true same-name matching; the fix is complete and idiomatic.
- **integration, NEEDS_IMPROVEMENT, fixed in this PR.** The reworded WARNING paragraph in providers.md over-claimed a warning for *every* non-migrated legacy login, whereas the flag-on Case-2 outcome (adopt a *different* current name-holder) is recorded as an audit, not a warning. Corrected in `providers.md`.

Pre-existing, out of scope (unchanged reachability, both correctness and OIDC reviewers concurred): the flag-off Case-2 reject message still emits the "migratable pending-legacy-link" hint though enabling the flag would adopt the current holder rather than migrate the victim.

Local gates: `dotnet build --warnaserror` clean, `dotnet test` 818/818, Prettier clean. Merging once the required checks are green.

Issue #210, opened 2026-07-12
OIDC RFC 9207 hardening refinements: compare to discovery issuer; require iss when advertised


iderex:
Follow-ups from the #125 review (adversarial gate oidc-reviewer F1/F2 + a CodeRabbit note), all non-blocking hardening refinements on top of the shipped RFC 9207 response-`iss` check.

The shipped check compares the authorization-response `iss` to the **id_token issuer** and tolerates absence. Two optional refinements would make it stricter:

1. **Compare to the discovery issuer directly**, not the id_token issuer. Equivalent in the default posture (OidcIdTokenValidator matches the id_token `iss` to the discovery issuer), but strictly stronger when the `DoNotValidateIssuerName` escape hatch is enabled, then the id_token `iss` is only signature-bound, not discovery-matched. Would require persisting the discovered issuer with the authorize state (capture `ProviderInformation.IssuerName` at challenge, or read the discovery document). CodeRabbit also suggested moving the check before code redemption; note the token endpoint here is already fixed per-provider from the callback route (not `iss`-selected), so the classic code-leak is not reachable, this is purely a strictness refinement, see the PR #209 discussion.
2. **Require `iss` when the AS advertises `authorization_response_iss_parameter_supported`** in discovery (RFC 9207 §2.4). The shipped check tolerates absence unconditionally so IdPs that omit `iss` keep working; requiring it only when advertised closes the gap without locking out older IdPs.

Both are gated behind reading/persisting discovery metadata, which is the shared prerequisite; do them together. Refs #125.

Issue #141, opened 2026-07-12
Check AS PKCE support at discovery: OidcClient sends PKCE blind and no nonce


iderex:
RFC 9700 sect. 2.1.1 (2025, https://datatracker.ietf.org/doc/rfc9700/) requires a client to ensure the AS supports PKCE before relying on it; otherwise nonce/state must carry the protection. Source-verified at tag imoc-7.1.0: OidcClient always sends code_challenge (S256) but never sends a nonce in code flow and never checks the discovery document's `code_challenge_methods_supported` (AuthorizeClient.CreateAuthorizeParameters, https://github.com/DuendeSoftware/foss/blob/imoc-7.1.0/identity-model-oidc-client/src/IdentityModel.OidcClient/AuthorizeClient.cs). An AS that ignores PKCE silently downgrades: the one-time server-side state store still blocks login-CSRF, but cross-session authorization-code injection becomes undetectable. Cheap check that closes the last RFC 9700 code-injection requirement not covered by the state store.

Acceptance criteria:
- [ ] On provider save and/or login preparation, the discovery document is checked for S256 in `code_challenge_methods_supported`.
- [ ] Absence produces an [SSO Audit] warning; a per-provider strict flag fails closed instead.
- [ ] Tests cover present/absent/strict cases.

iderex:
## Confirmed against the pinned 7.1.0 binary + implementation hook

The body was verified at the GitHub tag; confirming it holds for the actually-shipped DLL (repo rule: verify against the pinned binary), and pinning down where the fix hooks in.

### Empirical confirmation (ilspycmd on `duende.identitymodel.oidcclient/7.1.0/lib/net9.0/*.dll`)
`AuthorizeClient.CreateAuthorizeParameters` builds, unconditionally: `response_type=code`, `state`, **`code_challenge`**, **`code_challenge_method=S256`**, then `client_id`/`scope`/`redirect_uri`. There is **no `nonce`** parameter and **no read of the discovery document's `code_challenge_methods_supported`**, PKCE (S256) is always sent regardless of AS support. `CreatePkceData()` = 32-byte `CryptoRandom` verifier + SHA-256 challenge (correct S256). Matches the issue exactly.

### Implementation hook (the part that needed digging)
[...]

PR #437, opened 2026-07-17
Reuse the challenge's validated discovery metadata at the OpenID callback

### LINKED ISSUE #247: Cache per-provider discovery metadata: every OIDC login refetches discovery and JWKS twice
## Problem

Every OIDC login pays four redundant HTTPS round trips to the IdP because nothing caches discovery metadata across requests.

- The plugin builds a fresh `OidcClient` + `OidcClientOptions` per request: challenge at `SSOController.cs:300-301`, callback at `:175-177`, construction at `:327-357`.
- Verified in the pinned `Duende.IdentityModel.OidcClient` 7.1.0 (sources resolved via the shipped PDB SourceLink, commit `8e981e4`): `PrepareLoginAsync` and `ProcessResponseAsync` each run `EnsureProviderInformationAsync`, which stores `ProviderInformation` only on the per-request options instance, the library has no cross-instance cache, and `RefreshDiscoveryDocumentForLogin` defaults to `true`, so even a reused client refetches per call. Each discovery load is two GETs: the document, then the JWKS (`Duende.IdentityModel` 8.1.0, `HttpClientDiscoveryExtensions`).
- Confirmed empirically against the pinned DLL with a local counting IdP: four `PrepareLoginAsync` calls (same and fresh instances) = 4 discovery + 4 JWKS fetches; pre-supplying `options.ProviderInformation` = 0 fetches.

Per login that is 2 discovery + 2 JWKS round trips (challenge + callback) on top of the necessary token/userinfo calls, roughly 200 ms to 1 s of added login latency at WAN RTT, and under a challenge flood the plugin hits the IdP once per anonymous request (outbound amplification; compounds the state-store cap issue). The #141 PKCE probe fetches the same document again (that one is TTL-cached, so amortized).

## Fix

Per-provider `(ProviderInformation, FetchedAt)` TTL cache in the same shape as the #141 `PkceSupportCache`; pre-assign `options.ProviderInformation` in `CreateOidcClient` on a hit.

Constraints that must not be traded:

- Hand-filled `ProviderInformation` bypasses the `DiscoveryPolicy` validation (`RequireHttps` / `ValidateIssuerName` / `ValidateEndpoints`, `SSOController.cs:348-350`). The cache must be populated only from a fetch that passed the same policy, e.g. run the normal discovery once and cache what the library validated.
- Pre-assignment sets the client's internal `_useDiscovery = false`, which disables the `invalid_signature` JWKS-refresh-and-retry path (`ResponseProcessor`), a contract `OidcIdTokenValidator.MapError` relies on. Evict the cache entry when the callback reports `invalid_signature` and keep the TTL short (~15 min) so IdP key rotation self-heals.

Zero-risk half measure if the above is deferred: stash the challenge's `oidcClient.Options.ProviderInformation` (populated by its own discovery run) in the `TimedAuthorizeState` and reuse it at the callback, 4 fetches become 2, staleness is bounded by the state's existing 15-minute lifetime, and no discovery-policy bypass is introduced.

One cached fetch can also serve the #141 PKCE probe and the #210 discovery-issuer comparison.

While in the file: set `LogSerializer.Enabled = false` once at startup, OidcClient JSON-serializes the full options object (client secret included) into a transient string on every Prepare/Process call even when Trace logging is off (pinned `OidcClient.cs:387-393`).

iderex:
# What & why

Every OpenID login re-ran discovery and JWKS **twice**, once at the challenge and again at the
callback, because `OidcClient` stores `ProviderInformation` only on the per-request options instance
and refetches per call. On top of the necessary token/userinfo calls that is 2 discovery + 2 JWKS
round trips per login (~200 ms to 1 s of added latency at WAN RTT), and one IdP hit per anonymous
request under a challenge flood (outbound amplification).

This carries the metadata the challenge already fetched and validated to the callback, the issue's
bounded, no-policy-bypass measure:

- The challenge's `PrepareLoginAsync` populates `oidcClient.Options.ProviderInformation` from a fetch
  validated against the provider's `DiscoveryPolicy` (`RequireHttps` / `ValidateIssuerName` /
  `ValidateEndpoints`). Stash it on the authorize state (`OidcStateStore.TryAdd` →
  `TimedAuthorizeState` → `PendingState`), bounded by the state's ~15-minute lifetime.
- The callback pre-assigns it into `CreateOidcClient`, so `ProcessResponseAsync` reuses it instead of
  re-running discovery + JWKS. A null value (a state predating the capture, e.g. an in-flight login
  across the upgrade) falls back to a fresh discovery, unchanged.
- Pre-assigning sets the library's internal `_useDiscovery = false`, which also disables its
  `invalid_signature` JWKS-refresh-retry. Bounded and self-healing: a signing-key rotation inside the
  15-minute window fails that one callback closed, the user retries, and the next challenge fetches
  fresh keys. Reuse is only ever from a policy-validated fetch, so no `DiscoveryPolicy` bypass is
  introduced.

Also: stop the `OidcClient` trace serializer from JSON-serializing the full options object, **client
secret included**, into a transient string on every `Prepare`/`Process` call (it does so even with
Trace logging off). `LogSerializer.Enabled = false` once at plugin load keeps the secret out of
transient heap strings (defense in depth).

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: reused metadata comes only from a `DiscoveryPolicy`-validated challenge
      fetch; the disabled refresh-retry fails a key-rotation callback closed (retryable), never open.
- [x] No secrets logged; secrets redacted on export. `ProviderInformation` carries only the public
      issuer/endpoints/JWKS — no secret. `LogSerializer.Enabled = false` actively removes a transient
      client-secret serialization.
- [x] A security review was performed: 4-lens refute-by-default gate + the OIDC domain reviewer.
- [x] Log inputs from the IdP are sanitized inline at the log call (no log calls added).

## Quality checklist

- [x] No duplicated logic; the metadata is captured once and threaded through the existing state store.
- [x] Self-documenting; comments explain the security/latency *why* and the `_useDiscovery` trade-off.
- [x] Architecture-conformance tests pass; no new structural property introduced.

## Verification

- [x] `dotnet test` is green (823/823).
- [x] Prettier is clean (no tracked `.js`/`.html`/`.md`/`.css` changed).

## Notes

Login-path/crypto change → the adversarial-review gate is load-bearing. The 4-lens gate plus the OIDC
domain reviewer ran; **bypass, availability, correctness, and integration all PASS**, the availability
and integration lenses verified against the decompiled Duende 7.1.0 IL that the challenge captures the
**full** published JWKS (so a static multi-key set and any advance-published/overlapping rotated key
validate at the callback), that pre-assigning `ProviderInformation` sets `_useDiscovery = false` and
skips the callback discovery+JWKS while the captured `KeySet` still drives signature validation, and
that the reused value is always the challenge's own policy-validated fetch (no bypass, carries no
secret, strictly-more-restrictive fail-closed).

The **OIDC domain reviewer** returned NEEDS_IMPROVEMENT on two points, dispositioned as follows:

- **FIXED**, the security-direction consequence (a rotated-out / **revoked** key stays accepted for up
  to the ~15-minute state window, since the callback validates against the captured set) is now
  documented at the pre-assignment in `CreateOidcClient` and in the E2E checklist. It is a far tighter
  exposure than the platform-default 24-hour JWKS cache and never wider than the state lifetime.
- **DECLINED (reasoned)**, an end-to-end fetch-count regression test. The store→`PendingState`
  round-trip and the null-fallback are unit-tested; the Duende skip-fetch mechanism is IL-verified by
  the gate and empirically verified in the issue. A full challenge→callback fetch-count test requires
  the `SSOController` integration harness that does not exist yet (tracked by #192), so the behavior is
  captured as an explicit live-IdP release-QA item in the E2E checklist rather than blocking on that
  harness here.

Testability: unit tests pin that the store round-trips `ProviderInformation` to the callback and that a
null value falls back to discovery. The actual 4→2 fetch reduction and the key-rotation fallback are a
Duende-library behavior verified empirically in the issue and captured as a live-IdP release-QA item
(E2E checklist, local process doc). The full per-provider TTL cache the issue also sketches (serving
cross-login reuse, the #141 PKCE probe, and #210) is intentionally out of scope here, this delivers the
bounded half-measure with no `DiscoveryPolicy` bypass.

Closes #247

## Summary by CodeRabbit

* **Performance**
  * Streamlined OIDC sign-in callbacks by reusing previously validated provider discovery information, reducing redundant discovery and JWKS requests.

* **Security**
  * Prevented sensitive client configuration, including client secrets, from being serialized into transient diagnostic traces during authentication.

* **Reliability**
  * Added safeguards for expired or missing authentication state and supported fallback discovery when provider information is unavailable.

* **Tests**
  * Expanded coverage for provider-information reuse, missing state handling, and fallback behavior.

### DIFF
--- a/SSO-Auth.Tests/OidcStateStoreTests.cs
+++ b/SSO-Auth.Tests/OidcStateStoreTests.cs
@@ -146,6 +146,38 @@ public class OidcStateStoreTests
         Assert.Equal(1, store.Count);
     }
 
+    // --- ProviderInformation reuse (#247): carry the challenge's validated discovery to the callback ---
+
+    [Fact]
+    public void TryAdd_CarriesProviderInformation_ReusedAtThePeekForTheCallback()
+    {
+        // The challenge stashes the discovery metadata it already fetched and validated; the store must
+        // round-trip it to the PendingState the OidPost callback reads, so ProcessResponseAsync can reuse
+        // it instead of re-running discovery + JWKS.
+        var store = Store();
+        var info = new ProviderInformation { IssuerName = "https://idp.example" };
+
+        Assert.True(store.TryAdd(new AuthorizeState { State = "s" }, "p", false, Now, Binding, clientKey: null, providerInformation: info, out _));
+
+        var pending = store.PeekCurrent("s", "p", Now, Binding);
+        Assert.NotNull(pending);
+        Assert.Same(info, pending.ProviderInformation);
+    }
+
+    [Fact]
+    public void TryAdd_NullProviderInformation_PeekExposesNull_SoTheCallbackFallsBackToDiscovery()
+    {
+        // A state added without stashed metadata (a flow that predates the capture) exposes null, so the
+        // callback's CreateOidcClient runs a fresh discovery — the pre-#247 behavior, never a broken login.
+        var store = Store();
+
+        Assert.True(store.TryAdd(new AuthorizeState { State = "s" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out _));
+
+        var pending = store.PeekCurrent("s", "p", Now, Binding);
+        Assert.NotNull(pending);
+        Assert.Null(pending.ProviderInformation);
+    }
+
     [Theory]
     [InlineData(null)]
     [InlineData("")]
@@ -359,7 +391,7 @@ public class OidcStateStoreTests
             {
                 for (var i = 0; i < 5000; i++)
                 {
-                    store.TryAdd(new AuthorizeState { State = "fresh-" + i }, "p", false, Now, Binding, clientKey: null, out _);
+                    store.TryAdd(new AuthorizeState { State = "fresh-" + i }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out _);
                 }
             },
             TestContext.Current.CancellationToken);
@@ -387,8 +419,8 @@ public class OidcStateStoreTests
     {
         var store = Store(maxEntries: 2);
 
-        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, out var warnA));
-        Assert.True(store.TryAdd(new AuthorizeState { State = "b" }, "p", false, Now, Binding, clientKey: null, out var warnB));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out var warnA));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "b" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out var warnB));
         Assert.False(warnA);
         Assert.False(warnB);
         Assert.Equal(2, store.Count);
@@ -398,10 +430,10 @@ public class OidcStateStoreTests
     public void TryAdd_AtCap_RefusesANewKeyAndKeepsTheInFlightState()
     {
         var store = Store(maxEntries: 1);
-        Assert.True(store.TryAdd(new AuthorizeState { State = "in-flight" }, "p", false, Now, Binding, clientKey: null, out _));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "in-flight" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out _));
 
         // At the cap, a fresh challenge is refused rather than evicting the user already mid-login.
-        Assert.False(store.TryAdd(new AuthorizeState { State = "new" }, "p", false, Now, Binding, clientKey: null, out _));
+        Assert.False(store.TryAdd(new AuthorizeState { State = "new" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out _));
         Assert.Equal(1, store.Count);
         Assert.NotNull(store.PeekCurrent("in-flight", "p", Now, Binding));
     }
@@ -410,8 +442,8 @@ public class OidcStateStoreTests
     public void TryAdd_DuplicateKey_ReturnsFalse()
     {
         var store = Store(maxEntries: 10);
-        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, out _));
-        Assert.False(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, out _));
+        Assert.True(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out _));
+        Assert.False(store.TryAdd(new AuthorizeState { State = "a" }, "p", false, Now, Binding, clientKey: null, providerInformation: null, out _));
     }
 
     [Fact]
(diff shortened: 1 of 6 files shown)


iderex:

This is a login-path/crypto change (OIDC discovery + JWKS + id_token validation), so the adversarial-review gate is load-bearing. We ran a 4-lens refute-by-default review plus the OpenID domain reviewer; the availability and integration lenses verified the mechanism against the decompiled Duende 7.1.0 IL.

- **bypass, PASS.** The reused `ProviderInformation` is always the challenge's own `DiscoveryPolicy`-validated fetch (both clients built from the same config), bound to the same provider/state/browser, carries no secret (public issuer/endpoints/JWKS only), and `RequireIdentityTokenSignature` + `OidcIdTokenValidator` still run against the captured `KeySet`. Disabling the refresh-retry is strictly more restrictive (fail-closed), never fail-open.
- **availability, PASS.** The challenge captures the **full** published JWKS, so a static multi-key set and any advance-published/overlapping rotated key validate at the callback. The only failure is an IdP that activates a new key with < ~15-min JWKS overlap, per-user, transient, self-healing on retry, and already breaks standard JWKS-caching clients (whose default cache is 24 h).
- **correctness, PASS.** Capture happens post-success on populated metadata; the null path is byte-identical to pre-#247; the token exchange and validation use complete, equivalent metadata; linking shares the same pair; tests are sound and non-tautological.
- **integration, PASS.** Verified against the pinned 7.1.0 IL: pre-assigning `ProviderInformation` sets `_useDiscovery=false` and skips the callback discovery+JWKS while the captured `KeySet` drives in-memory signature validation; the `LogSerializer` global and the in-memory store carry no serialization/lifetime/threading hazard.
- **OIDC domain, NEEDS_IMPROVEMENT, dispositioned:**
  - *Fixed*, documented the reverse-direction consequence (a revoked/rotated-out key stays accepted for ≤ the ~15-min state window, far tighter than the 24 h default cache) at the pre-assignment and in the E2E checklist.
  - *Declined, reasoned*, an end-to-end fetch-count regression test needs the `SSOController` integration harness that does not exist yet (#192); the store round-trip + null-fallback are unit-tested and the skip-fetch mechanism is IL-verified + empirically verified in the issue, so it is captured as a live-IdP release-QA item instead of blocking on that harness here.

Local gates: `dotnet build --warnaserror` clean, `dotnet test` 823/823, Prettier clean (no tracked JS/HTML/MD/CSS changed). Merging once the required checks are green.

iderex:
Update (8101c5c), cleared two new-code SonarCloud maintainability findings the first commit introduced:

- **S107**, the `ProviderInformation` parameter pushed `OidcStateStore.TryAdd` (already at the 7-parameter limit) to 8. Reworked to record it through a small post-registration `SetProviderInformation(token, info)` instead: it is a latency enrichment, not part of the entry's identity/capacity admission, so keeping it off `TryAdd` leaves that concurrency-critical hot path at its original signature. The challenge calls it right after a successful `TryAdd`; the CSPRNG state token is not handed to the browser until the redirect that follows both calls, so nothing can consume the entry in the gap, and a (practically unreachable) miss just leaves the callback to a fresh discovery.
- **S3010**, moved `LogSerializer.Enabled = false` from the instance constructor into the static constructor (type initializer), where a one-time process-global assignment belongs.

Behavior is unchanged from the gated version; the change is purely how the challenge's `ProviderInformation` reaches the stored entry. Tests updated to drive `SetProviderInformation` (plus a no-op-on-unknown-token case), 824 tests, build clean. The SonarCloud gate's remaining red condition is `new_coverage` only (accepted per the free-plan coverage-as-a-goal policy). We also re-ran a focused bypass+correctness review on this refactor delta.

iderex:
Focused delta review (bypass + correctness) of the 8101c5c refactor, both **PASS**.

- **correctness, PASS.** Across every interleaving: `SetProviderInformation` looks up by the same CSPRNG key `TryAdd` inserted (a token collision loses the atomic add and never reaches the enrichment, so it cannot touch a different challenge's entry); no reader can consume the entry in the add→enrich gap (the token is not sent to the browser until the `Redirect` after both calls); a prune/redeem cannot hit a not-yet-Valid, age-~0, unpublished-token entry. The one genuine delta is a weaker happens-before on the `ProviderInformation` write (published-then-written vs written-then-published), whose worst case on a weak memory model is a benign, self-healing fallback to fresh discovery, never a wrong entry, broken login, or fail-open. The static-ctor move runs the assignment no later than before.
- **bypass, PASS.** The enrichment stores the same policy-validated `ProviderInformation` reference for the same provider; no cross-wiring, no attacker-influenced/unvalidated window, fail-closed on any miss. The prior gate's security properties hold byte-for-byte.

All required checks green (build, prettier, SonarCloud quality gate, dependency-review, CodeQL ×3); the SonarCloud gate is now green (the maintainability findings are fixed; only `new_coverage` remains under 80, accepted per the free-plan coverage-as-a-goal policy). Merging.

Issue #153, opened 2026-07-12
Pin the 2025-2026 SAML attack shapes with characterization tests


iderex:
The suite pins one XSW shape (prepend unsigned assertion) but not the comment-truncation class (CVE-2017-11428: `alice<!--x-->@evil.com`, https://nvd.nist.gov/vuln/detail/CVE-2017-11428) nor the recent variants: assertion injected before the signed one, duplicate/namespaced ID attribute pollution (PortSwigger 'The Fragile Lock', 2025, https://portswigger.net/research/the-fragile-lock; GHSL-2024-329/330), and a ds:Signature relocated outside the element its Reference covers. .NET's `XmlNode.InnerText` concatenates across comment nodes, so the plugin is not currently truncatable, but a refactor to `FirstChild.Value`/`InnerXml` would silently reintroduce the CVE. Negative tests only; no production change expected if current behavior holds.

Acceptance criteria:
- [ ] Test rejecting the comment-truncation shape (NameID with embedded comment must not be truncated or must be rejected).
- [ ] Tests rejecting assertion-before-signed injection, duplicate/namespaced ID pollution, and relocated-signature shapes, all fail-closed.
- [ ] Any surviving failure filed as its own security finding, not fixed in-suite.

Refs #120.

iderex:
Add one real-world 2026 pin to the list: authentik's SAML Source ACS fell to exactly this class in May, the signature verifies against the attacker's original signed assertion while identity data is read from a different forged assertion (CVE-2026-47201 / GHSA-c3m2-jqmq-pvp3, CVSS 8.5, published 2026-05-29, fixed in 2026.5.1/2026.2.4/2025.12.6; https://github.com/advisories/GHSA-c3m2-jqmq-pvp3). Prerequisite is only an attacker-owned account at the upstream IdP: replay your own signed response with a forged assertion carrying the victim identifier.

The verify-one-consume-another split is the invariant #137 closes. The characterization test for this shape should assert that the assertion actually consumed is the one the verified signature covers, not merely that some signature in the document verifies.

PR #441, opened 2026-07-17
Tighten SAML assertion conformance: signatures, audience, bearer

### LINKED ISSUE #238: Strict SAML assertion conformance: single signature, per-restriction audience, bearer confirmation method
A standing adversarial SAML review (refute-by-default, verified against the pinned
`System.Security.Cryptography.Xml` 10.0.9) confirmed the core is fail-closed across
signature/XSW, algorithm allow-list, XXE/DTD, time bounds, audience/recipient/destination,
and replay, **no exploitable bypass exists on the default posture**. It surfaced three
strict-conformance gaps where the validator is more permissive than the SAML 2.0 profile
requires. None is exploitable today (each needs a decoy that breaks validation, or a
cooperating/misconfigured IdP that already holds the pinned certificate), but each moves the
core further from a fully fail-closed reading and is a cheap, self-contained fix. Tracked
together because they share one file and one review; the attack-shape characterization tests
belong with #153.

### 1. Signature element count is not capped at exactly one
`Response.IsValid()` checks `signatureNodes.Count == 0` and then takes `signatureNodes[0]`,
but never rejects `Count > 1` (`SSO-Auth/Saml.cs:143-149`). A response carrying both a
`/samlp:Response/ds:Signature` and a `/samlp:Response/saml:Assertion/ds:Signature` is
disambiguated only by document order. It is non-exploitable, the chosen signature must still
cover the root/single-assertion, be enveloped-within, and pass the pinned-cert
`CheckSignature`, so a decoy only breaks validation, but "exactly one signature" should be a
hard reject, not an implicit first-wins.

- Fix: `if (signatureNodes.Count != 1) { return false; }` before indexing.

### 2. Multiple `<AudienceRestriction>` are OR-ed, not AND-ed
`GetAudiences` (`SSO-Auth/Saml.cs:261-281`) flattens every `<Audience>` across every
`<AudienceRestriction>` and the caller accepts on any match (`.Any`, `SSO-Auth/Saml.cs:226`).
SAML 2.0 requires the SP to be present in **every** `<AudienceRestriction>` (AND across
restrictions, OR within one). An assertion whose first restriction names another SP and whose
second names our `SamlClientId` is accepted here though it is not strictly addressed to us.
Requires a cooperating/misconfigured IdP sharing the pinned cert, hence low.

- Fix: require our audience in each `<AudienceRestriction>` block, not merely somewhere in the union.

### 3. Bearer `SubjectConfirmation` `Method` is not checked
`GetRecipient` / `GetInResponseTo` / `GetNotOnOrAfter` / the time-bound read all select
`saml:SubjectConfirmation/saml:SubjectConfirmationData` with `SelectSingleNode` and take the
first, without requiring `Method="urn:oasis:names:tc:SAML:2.0:cm:bearer"`
(`SSO-Auth/Saml.cs:306-311, 379, 410, 424`). A non-bearer (e.g. holder-of-key) confirmation
is consumed as if it were a bearer token, with no key-possession proof. Low, Web-SSO IdPs
emit bearer, and the assertion is signed and audience-bound, but the Web Browser SSO profile
is a bearer profile and the method should be asserted.

- Fix: select `SubjectConfirmation[@Method='urn:oasis:names:tc:SAML:2.0:cm:bearer']` and reject when none is present.

### Documented residual (no change requested here)
The assertion-ID replay cache and the outstanding-AuthnRequest cache are in-process static
state (`SSO-Auth/Api/SamlReplayCache.cs`, `SamlRequestCache.cs`) and reset on a Jellyfin
restart, so a captured, still-within-window bearer assertion could be replayed once after a
restart inside its (short, skew-bounded) acceptance window. This is inherent to the in-memory
design, not a logic bug; noted for the threat model rather than fixed short-term. Persisting
the consumed-ID set would close it if ever elevated.

### Acceptance
- Each of (1)-(3) rejects fail-closed; add characterization tests (coordinate with #153) for
  the dual-signature, multi-restriction, and non-bearer-confirmation shapes.
- Runs the adversarial-review gate (touches the SAML validation core).

iderex:
# What & why

A standing adversarial SAML review found the core fail-closed with no exploitable bypass on the default posture, but flagged three points where the validator was more permissive than the SAML 2.0 profile requires. Each is a self-contained, fail-closed tightening. Closes #238.

1. **Every position-bound signature must validate.** `IsValid()` used to validate only the first `ds:Signature` in document order and ignore a second one, so which signature was authoritative was decided by position. It now requires ≥1 signature and validates **every** position-bound signature (reference covers the root or the single assertion, allowed algorithms, verifies against the pinned cert).
2. **Audience is AND across restrictions.** `IsValid(expectedAudience)` flattened every `<Audience>` across every `<AudienceRestriction>` and accepted on any match. SAML 2.0 (core §2.5.1.4) requires the SP in **every** `<AudienceRestriction>` (AND across restrictions, OR within one); `IsAddressedTo` now enforces that and fails closed when there is no restriction at all.
3. **A bearer `SubjectConfirmation` is required.** The Web Browser SSO profile (§4.1.4.2) is a bearer profile, but the readers took the first `SubjectConfirmationData` regardless of `Method`. `IsValid()` now requires a `SubjectConfirmation[@Method='...bearer']` with its data, and the Recipient / InResponseTo / time-bound reads are scoped to it, so a non-bearer (holder-of-key) confirmation is never consumed as the bearer token.

**Divergence from the issue's fix #1, made on review evidence:** the issue proposed rejecting `Count != 1`. The adversarial gate (availability + saml-domain lenses) flagged that as a **mass-lockout**: the profile explicitly permits signing **both** the Response and the Assertion (Keycloak "Sign Documents"+"Sign Assertions", Azure AD "Sign response and assertion", ADFS `MessageAndAssertion`), and rejecting `Count>1` would break every login from such a deployment with no security gain (the existing reference-coverage/enveloped-within anchoring already makes the read content signature-covered). The delivered "validate **all** signatures" approach removes the first-wins ambiguity **and** keeps doubly-signing IdPs working, strictly stronger than the old first-wins, with no availability regression. A focused 3-lens re-review of the revision returned PASS across the board.

**No user-facing docs change:** conformant IdPs are unaffected (one authoritative signature set, or both, still accepted, a single AudienceRestriction, and a bearer confirmation), so no `providers.md`/README change is needed.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green (851/851; Debug, the Release test host hit the known Smart App Control `0x800711C7` launch flake, which the config-flip confirmed is not a test failure).

## Notes

- **Adversarial review, full 4-lens + SAML-domain gate, then a focused 3-lens re-review of the fix-1 revision: all PASS.** The gate caught the fix-1 mass-lockout (see divergence above) and drove the revision; the re-review confirmed the doubly-signed-honest case is now accepted, the doubly-signed-with-a-decoy case is rejected, and the change is strictly stronger than the pre-existing first-wins with no fail-open. FIX 2 and FIX 3 were verified spec-correct (SAML 2.0 core §2.5.1.4 / Web Browser SSO §4.1.4.2) and cannot lock out a conformant IdP.
- Characterization tests (coordinated with #153, in `SamlAttackShapeTests`): honest doubly-signed → accepted; doubly-signed with one signature's bytes flipped → rejected (pins validate-all, and the corruption reaches the real `CheckSignature`, not just base64 well-formedness); audience present in every restriction → accepted; audience present in only one of two → rejected; non-bearer and no-Method confirmations → rejected (isolated from the time-bound check); bearer positive control → accepted. `SamlTestFactory` gained `subjectConfirmationMethod`, `audienceRestrictions`, and `CreateDoublySigned` to build these shapes against the real signature path (no crypto mock).
- The documented residual in #238 (in-process replay/request caches reset on restart) is unchanged and out of scope here.

## Summary by CodeRabbit

* **Security Enhancements**
  * Strengthened SAML signature validation to check all position-bound signatures and reject invalid or tampered signatures.
  * Improved audience validation to require the service provider in every audience restriction.
  * Enforced bearer-only subject confirmations and rejected missing or unsupported confirmation methods.
  * Added stricter validation for bearer confirmation time limits.

### DIFF
--- a/SSO-Auth.Tests/SamlAttackShapeTests.cs
+++ b/SSO-Auth.Tests/SamlAttackShapeTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Xml;
 using Jellyfin.Plugin.SSO_Auth;
 using Xunit;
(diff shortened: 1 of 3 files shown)


iderex (commented):
My sign-off — consolidated pre-merge review.

**Nature of change:** security hardening (#238) on the SAML validation core — three strict-conformance tightenings (validate every position-bound signature; audience AND across restrictions; require a bearer SubjectConfirmation). All fail-closed; no exploitable bypass existed on the default posture, this moves the reading further toward the SAML 2.0 profile.

**Internal adversarial review — full 4-lens + SAML-domain gate, then a focused 3-lens re-review of the fix-1 revision:**

- First pass: FIX 2 (audience AND, SAML 2.0 core §2.5.1.4) and FIX 3 (bearer required, Web Browser SSO §4.1.4.2) → PASS, spec-correct, no conformant-IdP lockout. FIX 1 as originally written ("reject signature Count != 1") → **NEEDS_IMPROVEMENT** from the availability and saml-domain lenses: it would mass-lock-out IdPs that legitimately sign both the Response and the Assertion (Keycloak/Azure AD/ADFS), with no security gain.
- Revised FIX 1 to "require ≥1 signature and validate that EVERY position-bound signature independently passes" (the lenses' recommended approach). Focused re-review (availability + bypass + correctness) → **PASS across the board**: mass-lockout resolved (honest doubly-signed accepted), strictly stronger than the pre-existing first-wins ({new accepts} ⊆ {old accepts}, no fail-open), reference-coverage still guarantees a validated signature covers the read content, tests non-vacuous. A correctness note (the corrupt-signature test should reach CheckSignature, not just base64 well-formedness) was applied — the test now flips a signature byte and reaches the real cryptographic verification.

The gate did its job: it caught a real availability regression in the issue's literal proposal and drove the better fix. The divergence is documented on #238 and in the PR.

**Gates:** local Debug + Release builds green (`--warnaserror`, 0 warnings), 851/851 tests (the Release test host hit the known Smart App Control `0x800711C7` launch flake; the config-flip to Debug confirmed it is a launch flake, not a test failure), Jellyfin compat check green, Prettier n/a. CI all green — CodeQL, abi-floor, package (JPRM), and the SonarCloud quality gate (passed outright). CodeRabbit: no actionable comments; final-commit status "Review completed".

**Docs:** none needed — conformant IdPs are unaffected (one authoritative signature set, or both, still accepted; a single AudienceRestriction; a bearer confirmation).

Overall verdict: **ready to merge** — three fail-closed conformance tightenings, the fix-1 mass-lockout caught and corrected, characterization tests pin every shape, all gates green.

PR #443, opened 2026-07-17
Harden the controller link-map conformance scan against split and rename

### LINKED ISSUE #388: Harden the controller-link-map conformance scan against the controller split and property rename
From the #387 review (TheRealStroopwafel, finding 3). `Controller_NeverTouchesProviderLinkMaps` scans only `SSO-Auth/Api/SSOController.cs`. Two future-proofing gaps:
- **The planned controller split** (#318) would move endpoints into a second file the scan never reads. Widen the scan to all controller files when the split lands.
- **A property rename** (`CanonicalLinks` → anything) would make the zero-occurrence scan pass vacuously. Add a one-line reflection assert that `ProviderConfigBase.CanonicalLinks` still exists, so a rename forces a conscious rule update.

Neither is a live gap today (single controller file, property named `CanonicalLinks`); do this as the split step approaches.

Refs #383, #372

iderex:
# What & why

Closes #388.

The two controller source-scan fitness functions, `Controller_NeverTouchesProviderLinkMaps`
and `Controller_NeverTouchesRawSocketsOrDns`, read a single hardcoded
`SSO-Auth/Api/SSOController.cs`. They are zero-occurrence invariants, so they could
pass for the wrong reason in two ways the #387 review (finding 3) flagged:

- **Controller split (#318).** Moving endpoints into a second file, a partial-class
  split or several controllers, leaves them in a file the scan never reads. This is
  not purely hypothetical: a second controller (`SSO-Auth/Views/SSOViewsController.cs`)
  already sits outside the scanned path.
- **Property rename.** Renaming `CanonicalLinks` would make the link-map scan match
  nothing and pass vacuously, silently dropping the guard on a security boundary
  (the account-link map the controller must never touch directly).

This change hardens both, and makes each sentinel-guarded so it fails loudly if its
target moves, a conformance rule that can only pass is the exact bug the issue is about:

- A shared `ControllerSourceFiles()` helper discovers **every** controller source file
  from reflection: it lists the controller *types* (deriving from `ControllerBase`) and
  finds the files whose body declares them. A partial-class split (one type, several
  files) and a multi-controller split (more types) are both covered, and matching the
  declaration in the file body rather than the file name also survives a controller file
  rename. Both controller scans now use it, so widening benefits both.
- The helper asserts the discovered set is **non-empty**, so a controller that reflection
  can no longer find (base type removed, moved out of `SSO-Auth`) fails loudly instead of
  turning the rules into silent no-ops.
- The link-map rule pins `ProviderConfigBase.CanonicalLinks` by reflection, so a rename
  fails at a clear assert and forces a conscious update of the scanned token.
- Findings now report the file name alongside the line number, since more than one file
  is scanned.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. — N/A to production code; the *rule itself* is now
      fail-closed against a vacuous pass (empty file set / renamed property both fail).
- [x] No secrets logged; secrets are redacted on config export., N/A; test-only, no logging.
      config persistence, or the release pipeline. — N/A; no production login/crypto/config/pipeline
      code is touched. This guards the controller-vs-link-map boundary at CI time.
- [x] Log inputs from the IdP are sanitized inline at the log call., N/A; no logging.

## Quality checklist

      testable unit. — The two scans' duplicated file-read is consolidated into one
      `ControllerSourceFiles()` helper.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green., Debug and Release, 0 warnings.
- [x] `dotnet test` is green., 851/851.
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change., N/A; `.cs` only.

## Notes

Test/tooling change, no production behavior. Proven **non-vacuous** by breaking each
guarded target locally and confirming the rule trips, then restoring:

- Split-widening + detection: a temporary `.CanonicalLinks` access added to
  `SSOViewsController` (a file the old `SSOController.cs`-only scan never read) failed
  `Controller_NeverTouchesProviderLinkMaps`, naming that file - 
  `Found: SSOViewsController.cs line 51: _ = new ...OidConfig().CanonicalLinks.Count;`.
- Property-rename sentinel: pointing the const at a non-existent property name failed the
  reflection assert, `ProviderConfigBase.CanonicalLinksRenamedProbe was renamed or
  removed; point this rule at the new provider link-map property ... (#388)`.

Out of scope / follow-up: `SSOViewsController` being outside the previously-scanned path
means the socket/DNS and link-map invariants were, until now, unenforced on it; this PR
closes that. No violation exists there today.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -26,7 +26,11 @@ namespace Jellyfin.Plugin.SSO_Auth.Tests;
 /// link-map access to CanonicalLinkService (the login/admin workflow) and ServerManagedFields.Preserve
 /// (the #157 server-managed re-injection the config tier owns), a boundary worth failing CI on, not just
 /// review; #383 retired the controller's last two inline re-injection sites into that shared Preserve, so
-/// the scan is now a plain zero-occurrence invariant on the controller.
+/// the scan is now a plain zero-occurrence invariant on the controller. Both source scans discover EVERY
+/// controller source file from reflection (<see cref="ControllerSourceFiles"/>) rather than one hardcoded
+/// path, so the planned #318 controller split — into partial-class files or several controllers — cannot
+/// hide an endpoint from them, and each is sentinel-guarded against a vacuous pass: the file set must be
+/// non-empty, and the link-map scan pins its target property by reflection so a rename fails loudly (#388).
 /// </summary>
 public class ArchitectureConformanceTests
 {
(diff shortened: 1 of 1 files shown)


iderex (commented):
My sign-off — tests/tooling only (no production behavior change), independently reviewed. Both controller source-scan fitness functions now discover EVERY controller source file via reflection (ControllerSourceFiles: types deriving from ControllerBase, matched by their declaration in the file body — survives the #318 partial/multi-controller split and a file rename), sentinel-guarded against a vacuous pass (non-empty file set + the link-map property pinned by reflection so a rename fails loudly). Verified the logic and the empirical non-vacuity demonstration (a temporary .CanonicalLinks in SSOViewsController tripped the rule naming that file; the property sentinel trips on a renamed target). Genuinely valuable finding: SSOViewsController.cs sat OUTSIDE the old hardcoded scan, so the link-map and socket/DNS invariants were unenforced on it until now — the widening closes that hole (no violation exists there today). Gates: Debug+Release --warnaserror green, 851 tests, compat green; all 16 CI checks green. Ready to merge.

Issue #400, opened 2026-07-16
Serialize per-user avatar store against concurrent same-user logins


iderex:
Surfaced by review on #399. Two concurrent SSO logins for the same user can both pass `SaveImage` and then interleave the profile-image check/clear/assign in `AvatarService.StoreAsync`, one invocation can clear or overwrite the newer invocation's record. The interleaving surface is pre-existing (the pre-#399 ordering had the same unsynchronized sequence, as does Jellyfin's own image-replace endpoint) and the outcome self-heals on the next login, but a keyed per-user lock around the write-and-update sequence would remove it.

**Acceptance**
- [ ] The full `SaveImage` + profile-image transition sequence is serialized per user (keyed async lock; no global serialization of unrelated users).
- [ ] A deterministic overlapping-call test proves the final record is consistent under two concurrent `StoreAsync` calls for the same user.
- [ ] No lock is held across the avatar HTTP fetch (only the store step).

Refs #399, #377.

Issue #377, opened 2026-07-15
Avatar save failure leaves a dangling ProfileImage and discards the previous avatar


rekamer:
Surfaced by review on #375 (the AvatarService extraction, where the code moved verbatim, this is **pre-existing** behavior, present on `main` in `SSOController.TrySetUserAvatarAsync`, not introduced by that refactor).

## Problem (data integrity)

In the avatar fetch/store, `user.ProfileImage` is mutated **before** `IProviderManager.SaveImage` is awaited:

1. `ClearProfileImageAsync(user)` clears the existing profile image (if any).
2. `user.ProfileImage = new ImageInfo(newPath)` points the user at the new path.
3. `await SaveImage(stream, mediaType, newPath)` writes the file.

If `SaveImage` throws (it is caught by the best-effort `catch`), the user object is left with its old image cleared **and** `ProfileImage` pointing at a path that was never written. `Authenticate` then calls `UpdateUserAsync(user)` unconditionally, persisting the broken reference, so a transient save failure downgrades the user from a working previous avatar to a dangling reference to a nonexistent file.

## Fix direction

Reorder so the user object is only mutated after a successful write: build the target `ImageInfo`, `SaveImage` to its path, and only then (on success) clear the old image and assign `user.ProfileImage`. On failure the previous avatar is left intact. A characterization test can assert that a throwing `SaveImage` leaves `ProfileImage` unchanged (this becomes unit-testable now that the logic is in `AvatarService`).

Refs #375, #318

shippingToken:
Plan:

1. Extract the store step out of `TrySetAsync` into an internal `StoreAsync(user, image, mediaType, extension)`, the fetch path stays as is (#385 tracks its testability separately); the store step becomes unit-testable via InternalsVisibleTo.
2. Reorder inside `StoreAsync`: `SaveImage` writes the bytes FIRST; only on success is the user object touched. Three success arms:
   - no previous image → assign `new ImageInfo(newPath)`;
   - previous path == new path (the common same-extension re-login) → the file was overwritten in place, so `ClearProfileImageAsync` must NOT run (it would delete the bytes just written); keep the record and refresh `LastModified` so clients re-fetch;
   - previous path != new path (content type changed) → clear the old record+file only now that the new bytes are safely on disk, then assign.
3. On any `SaveImage` throw the best-effort catch in `TrySetAsync` fires and the previous avatar, record and file, is left fully intact. That is the fix.
[...]

PR #459, opened 2026-07-17
Reject unsafe usernames when building the avatar profile path

### LINKED ISSUE #447: Avatar profile path is built from the IdP-controlled username without sanitization (path-traversal candidate)
Surfaced by the independent review of PR #445 (#400), bypass lens. Pre-existing, NOT introduced by #445, which only added a lock around the same path construction.

## Finding

`AvatarService` builds the profile-image path directly from `user.Username`:
```csharp
// SSO-Auth/Api/AvatarService.cs:156-159
var newPath = Path.Combine(
    _serverConfigurationManager.ApplicationPaths.UserConfigurationDirectoryPath,
    user.Username,
    "profile" + extension);
```
For an SSO login, `user.Username` originates from the IdP-controlled `preferred_username` (OIDC) / SAML `NameID` and is passed straight into `_userManager.CreateUserAsync(username)` (`SSO-Auth/Api/Linking/CanonicalLinkService.cs:210`) with **no plugin-side sanitization**. The only guard is Jellyfin's own `CreateUserAsync` username validation.

If Jellyfin's validator rejects `/` and `\` but **allows `.`** (needs empirical confirmation), a username equal to `..` yields `Path.Combine(root, "..", "profile.png")`, a single-level directory escape that writes an attacker-fetched raster image (size-capped, fixed filename `profile.<ext>`) into the **parent** of `UserConfigurationDirectoryPath`. An attacker who controls their own IdP account's `preferred_username`/`NameID` controls the path component.

Impact is constrained (single-level escape, fixed filename, image content), but it is a write outside the intended directory on the login path, so it warrants a fail-closed fix.

## Acceptance
- [ ] **Empirically determine reachability**: does Jellyfin's `CreateUserAsync` / username validation accept `..` and/or path separators? (Check the referenced Jellyfin assemblies or a probe; the plugin references `Jellyfin.Controller`/`Model` only.) Document the finding either way.
- [ ] **Fail-closed fix regardless of the host's policy** (don't depend on it): either derive the profile directory from the immutable `user.Id` (GUID, no metacharacters), or explicitly reject/normalize username path metacharacters (`.`, `..`, `/`, `\`, any `Path.GetInvalidFileNameChars`) before path construction. Prefer the approach that matches how Jellyfin itself stores the profile image so behavior is unchanged for legitimate usernames.
- [ ] A test proving a username of `..` (and one containing a separator) cannot escape the user-config directory.
- [ ] Runs the adversarial-review gate (login path).

## Notes
Broader hardening question worth a look while here: is the raw IdP-controlled username used unsanitized at any OTHER filesystem/sink? Keep this issue scoped to the avatar path; file separately if another sink turns up. Refs #400, #399.

iderex:
# What & why

The SSO avatar store built the profile-image path directly from
`user.Username`, which for an SSO login is IdP-controlled (OIDC
`preferred_username` / SAML `NameID`). A username of `..` made
`Path.Combine(root, "..", "profile.<ext>")` write the fetched (size-capped,
fixed-name) image into the **parent** of `UserConfigurationDirectoryPath` - 
a single-level directory escape on the login path.

We guard the username fail-closed before the path is built, in two independent
layers inside `AvatarService`:

- **Character check**, reject `.` / `..`, either separator (`/` or `\`), and
  any `Path.GetInvalidFileNameChars()`. `GetInvalidFileNameChars()` excludes
  `\` on Linux, so both separators are rejected explicitly on every platform.
- **Round-trip containment (belt-and-suspenders)** - 
  `Path.GetFullPath(root/username)` must equal `root/username` exactly. This
  keeps the write under the root and also rejects normalisation the character
  check can't see: Windows silently strips trailing dots/spaces, so `victim.`
  would otherwise fold onto another user's `victim/` directory (an in-root
  cross-user overwrite); absolute / drive-relative / device forms would resolve
  elsewhere.

A rejected username skips only the **best-effort** avatar and logs a warning
(username line-ending-sanitised inline at the call, per the CodeQL
`cs/log-forging` invariant); the login still succeeds with no avatar. The guard
never throws into the login path, and the stored path is byte-for-byte
unchanged for every legitimate username (the write still uses the original
`Path.Combine(root, user.Username, "profile"+extension)`; `GetFullPath` is used
only for the decision).

Closes #447

## Empirical reachability (acceptance item)

We confirmed the traversal is genuinely reachable, and that the fix does not
depend on the host:

- Jellyfin's `UserManager.CreateUserAsync` validates usernames with
  `^(?!\s)[\w \-'._@+]+(?<!\s)$`. That regex **admits `.` and `..`** (period is
  in the allowed class) and **rejects `/` and `\`** (not in the class). So `..`
  passes host validation, the escape is real.
- The `User` entity constructor only does `ArgumentException.ThrowIfNullOrEmpty`
 , no character validation. (Verified against the pinned Jellyfin 10.11
  sources.)
- The validator lives in `Jellyfin.Server.Implementations`, which is **off the
  plugin's reference surface** (the plugin references Controller / Model /
  Database.Implementations only), the plugin cannot even call it, which is why
  the guard is fail-closed regardless of host policy.
- The avatar profile path is the **only** username-keyed filesystem sink in the
  plugin (full-plugin scan for `Path.Combine` / `File.*` / `SaveImage` /
  stream-to-file keyed on the username; the sole other `StreamWriter` wraps an
  in-memory SAML buffer). No separate follow-up sink issue is warranted (issue
  #447 Notes).

## Adversarial review (4-lens gate, login-path change)

- **Availability, PASS.** The guard is avatar-only and fully enclosed in
  `TrySetAsync`'s best-effort catch, so it can never fail or throw into login;
  every realistic host-admitted username (email-style, dotted, `@`/`+`/hyphen/
  underscore/unicode/internal-space) is untouched; no lock contention or
  migration. Only degenerate traversal payloads lose their avatar.
- **Bypass, PASS (on re-review).** First pass flagged a residual: the initial
  `StartsWith` containment accepted the Windows trailing-dot fold (`victim.` →
  `victim/`, an in-root cross-user overwrite) and `...` → root. We switched to
  the exact round-trip equality, which empirically rejects `victim.`,
  `victim `, `...`, whitespace-cloaked escapes like `.. `, and the `NUL` device
  form, while being strictly tighter than the prefix check (no new
  false-accepts). No write outside the exact intended per-user directory and no
  cross-user clobber could be constructed.
- **Correctness, PASS (on re-review).** First pass flagged (1) a too-narrow
  `catch (ArgumentException)` that let `PathTooLongException` (a ~32K-char
  username) escape the "never throws" contract, and (2) the containment-reject
  branch being untested. We widened the catch to
  `when (e is ArgumentException or IOException or NotSupportedException)`
  (covers the bare `IOException`/`PathTooLongException` empirically thrown by
  `GetFullPath`; the catch wraps only pure string ops, so it can't mask a disk
  error) and added a deterministic layer-2-reject test. `#377` write-then-
  transition ordering, the `#400` per-user lock release, non-destructive
  fail-closed, and null/empty-safety all re-confirmed.
- **Integration, PASS.** The plugin owns the profile path (`SaveImage` takes an
  explicit caller path; the host does not re-derive it), so skipping is
  non-corrupting; the lock-key (Ordinal) vs containment (OrdinalIgnoreCase on
  Windows) difference is inert; correct on Linux and Windows.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: an unsafe (IdP-controlled) username is rejected and
      the avatar skipped, never written to an unintended path.
- [x] No secrets logged; the username is not a secret and is line-ending-
      sanitised inline at the log call.
- [x] A security review was performed (4-lens adversarial gate, above).
      (`user.Username?.ReplaceLineEndings(string.Empty)`).

## Quality checklist

- [x] No duplicated logic; the guard is one small, single-purpose static helper
      within `AvatarService`.
- [x] The code is self-documenting; the comments explain the *why* (IdP-
      controlled component → fail closed) and match the target architecture.
- [x] Architecture-conformance tests pass; this change establishes no new
      structural property to lock in.

## Verification

- [x] `dotnet test` is green (867 tests, 0 failures).
- [x] Prettier: N/A, only `.cs` files changed.

## Notes

- We deliberately did **not** take the alternative fix of deriving the profile
  directory from the immutable `user.Id` (GUID): that changes **where** avatars
  are stored for existing users (an avatar-migration concern) and is a storage-
  semantics decision. The validate-and-fail-closed approach keeps legitimate
  usernames' storage paths unchanged, no migration.
- The new layer-2 test branches on `OperatingSystem.IsWindows()`: on Windows
  `victim.` is rejected (no write); on POSIX it is a distinct literal directory
  and stores normally there, either way no escape and no cross-user write.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved avatar storage protection against unsafe usernames and path traversal.
  * Unsafe usernames now skip avatar writes without interrupting login or removing existing profile images.
  * Added safeguards for platform-specific path normalization issues on Windows.
  * Preserved avatar storage for legitimate usernames containing dots and `@` symbols.

### DIFF
--- a/SSO-Auth.Tests/AvatarServiceTests.cs
+++ b/SSO-Auth.Tests/AvatarServiceTests.cs
@@ -283,6 +283,89 @@ public class AvatarServiceTests
         await providers.Received(1).SaveImage(Arg.Any<Stream>(), "image/png", ProfilePath("alice", ".png"));
     }
 
+    [Theory]
+    [InlineData("..")] // parent-directory escape: Path.Combine(root, "..", ...) writes into the PARENT of the user-config dir
+    [InlineData(".")] // current-directory component
+    [InlineData("../evil")] // separator + traversal
+    [InlineData("a/b")] // forward slash
+    [InlineData("a\\b")] // backslash
+    public async Task StoreAsync_UnsafeUsername_SkipsWriteWithoutThrowingOrEscaping(string username)
+    {
+        // #447: user.Username is IdP-controlled (OIDC preferred_username / SAML NameID) and only the host's
+        // own CreateUserAsync validates it — a regex that ADMITS '.' and '..'. A username of ".." would make
+        // Path.Combine write the fetched image into the PARENT of the user-config directory. The store must
+        // fail closed: no out-of-directory write (SaveImage is never called with any path), no throw (login
+        // is best-effort and must still complete), and no profile-image record.
+        var (service, providers, users, log) = Build();
+        var user = UserNamed(username);
+
+        await service.StoreAsync(user, new MemoryStream(new byte[] { 1 }), "image/png", ".png");
+
+        await providers.DidNotReceive().SaveImage(Arg.Any<Stream>(), Arg.Any<string>(), Arg.Any<string>());
+        await users.DidNotReceive().ClearProfileImageAsync(Arg.Any<User>());
+        Assert.Null(user.ProfileImage);
+        Assert.Contains(log.Entries, e => e.Level == LogLevel.Warning && e.Message.Contains("safe path component"));
+    }
+
+    [Fact]
+    public async Task StoreAsync_UnsafeUsername_LeavesAnExistingProfileImageUntouched()
+    {
+        // Fail-closed must not be destructive either: a rejected ".." username skips the write AND leaves
+        // any previously set profile-image record exactly as it was (no clear, no re-point).
+        var (service, providers, users, _) = Build();
+        var user = UserNamed("..");
+        var previous = new ImageInfo(ProfilePath("alice", ".jpg"));
+        user.ProfileImage = previous;
+
+        await service.StoreAsync(user, new MemoryStream(new byte[] { 1 }), "image/png", ".png");
+
+        Assert.Same(previous, user.ProfileImage);
+        await providers.DidNotReceive().SaveImage(Arg.Any<Stream>(), Arg.Any<string>(), Arg.Any<string>());
+        await users.DidNotReceive().ClearProfileImageAsync(Arg.Any<User>());
+    }
+
+    [Fact]
+    public async Task StoreAsync_WindowsTrailingDotUsername_DoesNotWriteIntoAnotherUsersDirectory()
+    {
+        // #447 layer-2 (the round-trip containment check, the branch the character check can't reach):
+        // the host username regex admits a trailing dot, and Windows strips it, so "victim." would fold
+        // onto another user's "victim" directory — an in-root cross-user avatar overwrite. The guard must
+        // reject any username whose resolved path is not exactly root/username. On POSIX nothing is
+        // stripped, so "victim." is a distinct literal directory and the avatar stores there normally.
+        // Either way there is no cross-user write and no escape.
+        var (service, providers, users, log) = Build();
+        var user = UserNamed("victim.");
+
+        await service.StoreAsync(user, new MemoryStream(new byte[] { 1 }), "image/png", ".png");
+
+        if (OperatingSystem.IsWindows())
+        {
+            await providers.DidNotReceive().SaveImage(Arg.Any<Stream>(), Arg.Any<string>(), Arg.Any<string>());
+            await users.DidNotReceive().ClearProfileImageAsync(Arg.Any<User>());
+            Assert.Null(user.ProfileImage);
+            Assert.Contains(log.Entries, e => e.Level == LogLevel.Warning && e.Message.Contains("safe path component"));
+        }
+        else
+        {
+            await providers.Received(1).SaveImage(Arg.Any<Stream>(), "image/png", ProfilePath("victim.", ".png"));
+            Assert.Equal(ProfilePath("victim.", ".png"), user.ProfileImage?.Path);
+        }
+    }
+
+    [Fact]
+    public async Task StoreAsync_LegitimateUsernameWithDotAndAt_StillStoresUnchanged()
+    {
+        // Behavior preserved for legitimate usernames the host allows (dots and '@' are valid — e.g. an
+        // email-style preferred_username): the store path is unchanged and the write happens as before.
+        var (service, providers, _, _) = Build();
+        var user = UserNamed("first.last@example.com");
+
+        await service.StoreAsync(user, new MemoryStream(new byte[] { 1 }), "image/png", ".png");
+
+        await providers.Received(1).SaveImage(Arg.Any<Stream>(), "image/png", ProfilePath("first.last@example.com", ".png"));
+        Assert.Equal(ProfilePath("first.last@example.com", ".png"), user.ProfileImage?.Path);
+    }
+
     [Fact]
     public async Task ReadCappedAsync_BodyOverTheCap_Throws()
     {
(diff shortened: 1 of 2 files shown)


iderex:

## Adversarial review (4-lens gate, mandatory, login-path change)

- **Availability, PASS.** The guard is avatar-only and fully enclosed in the best-effort `TrySetAsync` catch, so it can never fail or throw into login; every realistic host-admitted username (email-style, dotted, `@`/`+`/hyphen/underscore/unicode/internal-space) is untouched; no lock contention or migration, no mass-lockout surface.
- **Bypass, PASS (on re-review).** First pass flagged a residual in the initial `StartsWith` containment: Windows trailing-dot/space stripping folded `victim.` onto another user's `victim/` directory (in-root cross-user overwrite), and `...` onto the root. We switched to an exact round-trip equality (`GetFullPath(root/username) == root/username`), which empirically rejects `victim.`, `victim `, `...`, whitespace-cloaked escapes like `.. `, and the `NUL` device form, while being strictly tighter than the prefix check (no new false-accepts). No write outside the exact intended per-user directory, and no cross-user clobber, could be constructed.
- **Correctness, PASS (on re-review).** First pass flagged (1) a too-narrow `catch (ArgumentException)` letting `PathTooLongException` (a ~32K-char username) escape the "never throws" contract, and (2) the containment-reject branch being untested. We widened the catch to `when (e is ArgumentException or IOException or NotSupportedException)`, it wraps only pure string ops, so it cannot mask a disk error, and added a deterministic layer-2-reject test. The `#377` write-then-transition ordering, `#400` per-user lock release, non-destructive fail-closed, and null/empty-safety were all re-confirmed.
- **Integration, PASS.** The plugin owns the profile path (`SaveImage` takes an explicit caller path; the host does not re-derive it), so skipping is non-corrupting; the lock-key (Ordinal) vs containment (OrdinalIgnoreCase on Windows) difference is inert; correct on Linux and Windows. Confirmed the avatar path is the only username-keyed filesystem sink in the plugin, no separate follow-up sink issue is warranted (#447 Notes).

## Empirical reachability

Jellyfin's `CreateUserAsync` validates usernames with `^(?!\s)[\w \-'._@+]+(?<!\s)$`, which **admits `.` and `..`** and rejects `/` / `\`; the `User` entity constructor only checks null/empty. So `..` passes host validation, the parent-directory escape is genuinely reachable. The validator lives in `Jellyfin.Server.Implementations`, off the plugin's reference surface, so the fix is fail-closed independent of host policy.

## CI / quality gates

- `dotnet build --no-restore --warnaserror`: green (0 warnings, 0 errors).
- `dotnet test`: green (867 tests, 0 failures).
- `build`, `CodeQL` / `Analyze (csharp)`, `package`, `abi-floor`, `dependency-review`, `zizmor`, greppable-invariants, Trojan-source, prettier: all pass.
- **SonarCloud quality gate: OK**, all conditions green (new_security_rating A, new_reliability_rating A, new_maintainability_rating A, new_coverage 85.0% ≥ 80, new_duplicated_lines 0.0%, hotspots-reviewed 100%). No coverage exception needed.
- Prettier: N/A (only `.cs` changed).

Overall verdict: all four adversarial lenses PASS and every required CI/quality gate is green. Merging.

PR #462, opened 2026-07-17
Add a positive-control sentinel to the socket/DNS conformance scan

### LINKED ISSUE #444: Add a positive-control sentinel to Controller_NeverTouchesRawSocketsOrDns
Surfaced while delivering #388 (PR #443).

#388 hardened both controller source-scan fitness functions in `SSO-Auth.Tests/ArchitectureConformanceTests.cs` to (a) discover every controller source file via reflection and (b) fail loudly on a vacuous pass. The link-map rule (`Controller_NeverTouchesProviderLinkMaps`) additionally pins its scanned token by reflection, `ProviderConfigBase.CanonicalLinks` must exist, so a property rename fails at a clear assert instead of turning the zero-occurrence scan into a silent no-op.

`Controller_NeverTouchesRawSocketsOrDns` has no equivalent positive control: it is a pure zero-occurrence scan of BCL marker strings (`System.Net.Sockets`, `SocketsHttpHandler`, `NetworkStream`, `Dns.`, `System.Net.Dns`). If those markers ever stopped matching how the socket/DNS surface is referenced (a refactor, an alias, a new BCL spelling), the rule would pass for the wrong reason.

Low value, the markers are stable, long-lived BCL identifiers, but for parity with the link-map rule's sentinel and to keep the invariant non-vacuous by construction, add a marker-liveness check (e.g. assert at least one marker matches a known socket/DNS reference in the avatar tier, the invariant's legitimate home, so a marker that stops matching anything fails loudly).

Scope: test/tooling only, no production change. Coordinate with the #388 helpers already in place (`ControllerSourceFiles`).

iderex:
# What & why

`Controller_NeverTouchesRawSocketsOrDns` (`SSO-Auth.Tests/ArchitectureConformanceTests.cs`) is a
pure zero-occurrence scan of BCL marker strings (`System.Net.Sockets`, `SocketsHttpHandler`,
`NetworkStream`, `Dns.`, `System.Net.Dns`). It had no positive control: if the markers ever stopped
matching how the socket/DNS surface is referenced (a refactor, an alias, a different BCL spelling),
the rule would keep "passing" for the wrong reason, a silent no-op instead of the invariant it
claims to enforce. The sibling link-map rule (`Controller_NeverTouchesProviderLinkMaps`, #388) has
exactly this kind of sentinel already: it pins its target property by reflection so a rename fails
loudly. This closes the same gap for the socket/DNS rule.

Unlike the link-map rule, these markers are BCL identifiers, not a token this codebase owns, there
is no single property to pin. So the sentinel instead pins the marker *set* against reality: it
asserts at least one marker still matches a real line in the raw socket/DNS surface's one legitimate
home (`AvatarService`, `AvatarUrlValidator`, `SsoRateLimiter`, per #375), pinned by `typeof(...)` so a
rename or move of that tier fails the check with a clear message. If a refactor ever changed how that
tier references sockets/DNS so that *none* of the markers matched it any more, this is the assertion
that would actually catch it.

Deliberately "at least one" marker, not "every" marker: `"System.Net.Dns"` is a defensive marker for
the fully-qualified/static-import spelling, which the codebase does not use anywhere today
(`Dns.GetHostAddressesAsync` resolves through the `using System.Net;` form, caught by the `"Dns."`
marker instead), that one marker having no live match is expected, not a liveness failure.

While touching `ControllerSourceFiles`, extracted the shared "find source files declaring these
types" scan into `SourceFilesDeclaring(IEnumerable<Type>)`, reused by both `ControllerSourceFiles`
and the new liveness check, instead of duplicating the file-walk for a second type set.

Test/tooling only, no production code changed.

Closes #444

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

Not applicable, test/tooling only, no login path, crypto, config persistence, or release pipeline
code touched.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose, testable unit (the shared
      `SourceFilesDeclaring` helper removes the duplication the new check would otherwise have added).
- [x] The code is self-documenting (readable without comments; comments explain why, not what) and
      matches the target architecture (#318).
- [x] Architecture-conformance tests pass, and the new structural property (the liveness sentinel) is
      itself the new rule locked into `ArchitectureConformanceTests`.

## Verification

- [x] `dotnet build -c Debug --warnaserror` is green (0 warnings, 0 errors).
- [x] `dotnet build -c Release --warnaserror` is green (0 warnings, 0 errors).
- [x] `dotnet test` is green: 887/887.
- [x] Prettier: not applicable, no `.js`/`.html`/`.md`/`.css` touched.
- [x] `sh scripts/check-jellyfin-compat.sh` is green.

Proved the new sentinel is non-vacuous locally: temporarily replaced the marker set with a string
that matches nothing in the tree and reran the rule in isolation - only the new liveness assertion
failed, with the expected message ("None of the socket/DNS markers match any line in their legitimate
home..."); the pre-existing zero-occurrence assertion above it still passed (as it should, an empty
marker match set trivially has zero controller occurrences too). Restored the real markers; full
suite back to 887/887.

## Notes

Scope is intentionally narrow per #444: test/tooling only, no production behavior change. The
existing `Controller_NeverTouchesProviderLinkMaps` and `Controller_NeverTouchesRawSocketsOrDns`
zero-occurrence assertions are unchanged; only the liveness sentinel and the small shared-helper
extraction are new.

## Summary by CodeRabbit

- **Tests**
  - Strengthened architecture validation to ensure controller code does not directly access raw sockets or DNS functionality.
  - Added safeguards to confirm these checks remain active and detect legitimate networking usage elsewhere in the application.
  - Improved source discovery used by architecture rules, making future validation more reliable and reducing the risk of checks passing without meaningful coverage.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -6,6 +6,7 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text.RegularExpressions;
 using Jellyfin.Plugin.SSO_Auth;
+using Jellyfin.Plugin.SSO_Auth.Api;
 using Jellyfin.Plugin.SSO_Auth.Config;
 using MediaBrowser.Model.Plugins;
 using Microsoft.AspNetCore.Mvc;
@@ -31,6 +32,10 @@ namespace Jellyfin.Plugin.SSO_Auth.Tests;
 /// path, so the planned #318 controller split — into partial-class files or several controllers — cannot
 /// hide an endpoint from them, and each is sentinel-guarded against a vacuous pass: the file set must be
 /// non-empty, and the link-map scan pins its target property by reflection so a rename fails loudly (#388).
+/// The socket/DNS scan's markers are BCL identifiers rather than a token this codebase owns, so its
+/// sentinel instead pins the marker SET against the surface's legitimate home — AvatarService,
+/// AvatarUrlValidator, SsoRateLimiter — asserting at least one marker still matches real usage there,
+/// so a marker set that stops matching anything real fails loudly too (#444).
 /// </summary>
 public class ArchitectureConformanceTests
 {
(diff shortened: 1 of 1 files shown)


iderex (commented):
My sign-off — test/tooling only, delivered by a delegated agent and independently reviewed (more-eyes). The independent review confirmed: the liveness sentinel is genuinely non-vacuous (empirically — Dns./System.Net.Sockets/SocketsHttpHandler/NetworkStream have live matches in the avatar/rate-limiter tier via 'using System.Net;', while System.Net.Dns has zero occurrences, so 'at least one marker' is the empirically-correct choice, not a weakening); the SourceFilesDeclaring extraction is behavior-preserving (both scans keep their exact discovery + failure semantics); the typeof home-type pin is a compile-time guard (a rename/move fails the build, stronger than a runtime assert). No production code touched. All CI green. Ready to merge.

PR #465, opened 2026-07-17
Harden IntervalGate against stale-sample re-admission

### LINKED ISSUE #334: Harden IntervalGate against stale-sample re-admission

iderex:
# What & why

`IntervalGate.TryEnter` treated every sample older than the cursor as a wall-clock correction and re-anchored (the #246 self-heal). A caller whose captured `now` was merely *stale*, descheduled between reading the clock and entering, landing just behind the cursor, was indistinguishable from a correction, so it CASed the cursor backward and was admitted: a second admission inside one interval. This is the fail-open residual the old characterization test pinned deliberately (the throttles guard anonymous-path DoS amplification, so an extra sweep/log line per race is bounded, but it is still a fail-open we can now close cleanly).

We replace the one-directional guard with a **symmetric** one: suppress when `|now - last| < interval`.

- **Stale sub-interval sample (the bug):** a backward span shorter than the interval is now suppressed *without* re-anchoring. No second admission; the cursor is left untouched so the next admission stays capped at one interval after the real entry.
- **Genuine large correction (DST fall-back, NTP step ≥ interval):** still fails the guard, so it enters and re-anchors, the #246 self-heal is preserved, and no capped store can be stranded.
- **Forward direction / first call / exactly-the-interval boundary:** bit-identical to before.

We picked the symmetric guard over the monotonic-clock alternative (issue candidate 2) because it is one line, keeps each caller's wall-clock `now` shared with the throttled action's expiry math (a `TickCount64` gate would decouple the two), keeps the injectable clock the characterization tests rely on, and moves only in the safe direction: it can never lengthen the wait beyond one interval, so it opens no stall.

Closes #334
Refs #318, #331

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. *(N/A to this diff; the change closes a fail-open in a DoS-guard throttle without touching any accept/reject decision.)*
- [x] No secrets logged; secrets are redacted on config export. *(Unchanged; the gate carries no request data.)*
      config persistence, or the release pipeline. *(Four-lens adversarial gate — see Notes.)*
- [x] Log inputs from the IdP are sanitized inline at the log call. *(Unchanged; no log inputs touched.)*

## Quality checklist

      testable unit. *(The change lands once in `IntervalGate`, shared by all 5 throttle sites.)*
- [x] The change adds no more code than the problem requires. *(One-line guard; comment + test updates.)*
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. *(No new structural property; the existing `no-inline-interval-throttle-cursor` opengrep rule stays accurate — confirmed by the integration lens.)*

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. *(Debug + Release, 0 warnings.)*
- [x] `dotnet test` is green. *(889 passed, 0 failed.)*
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. *(No such files changed.)*

### Behavior: before → after

| Case | Before | After |
| --- | --- | --- |
| Forward, ≥ interval elapsed | enters, re-anchors | unchanged |
| Forward, < interval elapsed | throttled | unchanged |
| Exactly interval (forward) | enters | unchanged |
| First call (cursor at MinValue) | enters | unchanged |
| **Stale sample, < interval behind cursor** | **enters + re-anchors backward (2nd admission)** | **suppressed, cursor untouched (no 2nd admission)** |
| Genuine correction, ≥ interval behind cursor | enters + re-anchors | unchanged (self-heals) |

Across all five sites (OidcStateStore prune + cap-warn, SamlRequestCache prune + cap-warn, SsoRateLimiter signal + prune) this only changes the stale-sample row; the DoS-amplification protection (single CAS winner per interval) is intact.

### Adversarial self-review (four refute-by-default lenses)

- **Availability / mass-lockout, VERDICT: PASS.** No new stall at any site. Suppressing a sub-interval backward blip leaves the cursor put, so admission is still guaranteed one interval after the last real entry regardless of how many stale blips arrive; a capped store's prune can never stop running into a cap-fill lockout. Genuine ≥-interval corrections still self-heal; no first-call regression.
- **Security-bypass / fail-open, VERDICT: PASS.** The fail-open race is actually closed: no stale sub-interval sample can produce a second admission, and boundary-straddling timing cannot amplify the throttled action beyond one action per interval. The single-CAS-winner property holds. No `Math.Abs(long.MinValue)` overflow, the tick difference for real `DateTime` values is bounded well below `long` range.
- **Correctness, VERDICT: PASS.** Forward/first-call/exactly-the-interval semantics are bit-identical; backward-exactly-the-interval enters + re-anchors, sub-interval-back is suppressed without re-anchor. The updated/added tests' arithmetic assertions hold.
- **Integration, VERDICT: NEEDS_IMPROVEMENT → resolved.** The lens found documentation drift: four downstream call-site comments still described the gate as an unconditional backward self-heal. **Resolved** in commit `64e1cd1`, each comment now qualifies the self-heal with "≥ interval (re-anchors)" and notes the sub-interval case is suppressed with the cursor untouched, while preserving the "never stalls like the hand-rolled predecessor" property. The opengrep `no-inline-interval-throttle-cursor` rule needs no change (confirmed accurate). No behavior change from the reconciliation.

### Tests

Re-pinned the stale-sample characterization test from the fail-open residual to the new suppressed behavior, and added: exactly-one-interval-backward self-heals (boundary), and no-stall-under-repeated-sub-interval-blips (admission still opens one interval after the real entry). The genuine large-correction self-heal test is unchanged (30 min back ≫ 1 min interval).

### DIFF
--- a/SSO-Auth.Tests/IntervalGateTests.cs
+++ b/SSO-Auth.Tests/IntervalGateTests.cs
@@ -10,8 +10,10 @@ namespace Jellyfin.Plugin.SSO_Auth.Tests;
 /// Characterization tests pinning IntervalGate's once-per-interval semantics — the shared throttle the
 /// anonymous-path sweeps and log signals rely on to not amplify a flood into CPU or log volume (#246,
 /// #195). The wall-clock cases (boundary, backward step) are the primary evidence for the #318 step 2b
-/// unification because CI cannot move real time, and each adopting site's Interlocked original had exactly
-/// these semantics: guard on a non-negative sub-interval span, then a single CAS winner per interval.
+/// unification because CI cannot move real time. The guard suppresses a sub-interval span in either
+/// direction (|now - last| &lt; interval), then a single CAS winner enters per interval: a genuine
+/// correction (backward span &ge; interval) self-heals, while a stale sub-interval sample is suppressed
+/// rather than re-admitted (#334), closing the original fail-open residual without opening a stall.
 /// </summary>
 public class IntervalGateTests
 {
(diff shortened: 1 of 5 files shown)


iderex (commented):
My sign-off — security hardening (anonymous-path throttle). Delivered by a delegated agent; independently gated by the four-lens adversarial review — availability, bypass, correctness PASS; integration returned NEEDS_IMPROVEMENT for four downstream call-site comments still describing an unconditional backward self-heal, now reconciled (commit 64e1cd1) to '>= interval re-anchors / sub-interval suppressed, cursor untouched' — comment-only, no behavior change. Confirmed: the symmetric |now-last| < interval guard closes the sub-interval backward re-admission fail-open in both directions, keeps forward/first-call/exactly-the-interval semantics bit-identical, cannot overflow Math.Abs for bounded DateTime.Ticks, preserves the single-CAS-winner property, and cannot stall any of the five throttle sites into a cap-fill lockout (admission stays pinned at last+interval); the >= interval clock correction (#246 self-heal) is preserved. Design chose the one-line symmetric guard over a monotonic clock to keep each caller's wall-clock now shared with its expiry math. 889 tests green (Debug+Release --warnaserror); CodeQL C# green. Ready to merge.

PR #466, opened 2026-07-17
Collapse repeated PluginPageInfo construction in SSOPlugin

### LINKED ISSUE #458: Collapse the repeated PluginPageInfo construction in SSOPlugin.GetPages/GetViews
`SSOPlugin.GetPages` (SSO-Auth/SSOPlugin.cs:~115) and `GetViews` (:~152) each build an array of `PluginPageInfo` where every entry repeats the same shape:
`new PluginPageInfo { Name = ..., EmbeddedResourcePath = $"{GetType().Namespace}.<Config|Views>.<asset>" }`
 - 5 entries in GetPages, 6 in GetViews (11 total), with `style.css`, `linking.html` and `linking.js` registered verbatim in both methods.

A tiny local factory collapses each entry to one line:
```csharp
private PluginPageInfo Page(string name, string resource) =>
    new() { Name = name, EmbeddedResourcePath = $"{GetType().Namespace}.{resource}" };
```
then e.g. `Page(Name, "Config.configPage.html")`, `Page(Name + ".js", "Config.config.js")`, `Page("style.css", "Config.style.css")`, … Roughly 40 lines removed.

Behaviour unchanged: the registered `Name` and `EmbeddedResourcePath` strings stay byte-identical, so Jellyfin resolves exactly the same embedded assets. No test pins the internal construction shape (the contract is the Name/resource pairs, which are preserved verbatim). Security: none, pure construction refactor; the embedded resources and the `<EmbeddedResource>` entries in SSO-Auth.csproj are untouched. Surfaced by the completeness critic and verified against origin/main; net-new (no open issue covers the page-manifest methods).

iderex:
# What & why

`SSOPlugin.GetPages`/`GetViews` built each `PluginPageInfo` by hand, repeating
the same `Name`/`EmbeddedResourcePath` shape 11 times (5 pages + 6 views). We
collapse each entry to a single call against a new private `Page(name,
resource)` factory. Closes #458.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

Not applicable, this touches `SSOPlugin.cs` (flagged as a sensitive file
generally) but only its page/view manifest construction, not the login path,
crypto, or config persistence. No user input, no auth decision, no secret
handling involved; we did not run the full adversarial-review gate for this
reason, consistent with the issue's own security note ("Security: none, pure
construction refactor").

- [ ] Fail-closed preserved: a missing signature, time-bound, or audience is
- [ ] No secrets logged; secrets are redacted on config export.
- [ ] Log inputs from the IdP are sanitized inline at the log call.

## Quality checklist

      testable unit.
      (No new structural property here — `Page` is a private helper, already
      covered by the existing `SSOPlugin_DeclaresNoConfigurationLogicBeyondTheStoreFacade`
      rule, which we confirmed still passes.)

## Verification

- [x] `dotnet test` is green (Debug and Release, 915/915).
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change (none
      touched — this PR is C# only).

## Notes

Self-review: the emitted `PluginPageInfo` set is byte-identical before and
after, same 5 page names/paths, same 6 view names/paths, same order, same
`{GetType().Namespace}.<resource>` interpolation. `SSOPluginPageManifestTests`
(new) asserts the exact name/resource pairs and their order for both
`GetPages()` and `GetViews()`, locking the behavior-preserving contract in.
Net: 84 lines changed, 42 fewer lines in `SSOPlugin.cs`. Scope: `SSOPlugin.cs`
plus the new test file only, nothing else touched.

### DIFF
--- a/SSO-Auth.Tests/SSOPluginPageManifestTests.cs
+++ b/SSO-Auth.Tests/SSOPluginPageManifestTests.cs
@@ -1,0 +1,69 @@
+using System.IO;
+using System.Linq;
+using Jellyfin.Plugin.SSO_Auth;
+using MediaBrowser.Common.Configuration;
+using MediaBrowser.Model.Plugins;
+using MediaBrowser.Model.Serialization;
+using Microsoft.Extensions.Logging;
+using NSubstitute;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Pins the <see cref="PluginPageInfo"/> set <see cref="SSOPlugin.GetPages"/> and
+/// <see cref="SSOPlugin.GetViews"/> emit — the (name, embedded resource) contract Jellyfin resolves
+/// static assets by — so the #458 factory-collapse refactor stays behavior-preserving.
+/// </summary>
+[Collection("SSOController")]
+public class SSOPluginPageManifestTests
+{
+    private static SSOPlugin CreatePlugin()
+    {
+        var appPaths = Substitute.For<IApplicationPaths>();
+        appPaths.PluginConfigurationsPath.Returns(Path.Combine(Path.GetTempPath(), "sso-test-" + System.Guid.NewGuid()));
+        var xml = Substitute.For<IXmlSerializer>();
+        return new SSOPlugin(appPaths, xml, Substitute.For<ILogger<SSOPlugin>>());
+    }
+
+    [Fact]
+    public void GetPages_EmitsTheExpectedNameAndResourcePairsInOrder()
+    {
+        var plugin = CreatePlugin();
+        const string ns = "Jellyfin.Plugin.SSO_Auth";
+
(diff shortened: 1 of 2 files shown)


iderex (commented):
My sign-off — pure construction refactor, behavior-preserving. Verified directly from the diff: the original literals already used the $"{GetType().Namespace}.…" interpolation, so the new Page(name, resource) factory ($"{GetType().Namespace}.{resource}") produces byte-identical EmbeddedResourcePath values for all 5 pages and 6 views, with every Name preserved exactly. The new SSOPluginPageManifestTests pins the emitted (Name, EmbeddedResourcePath) set + order as fresh coverage. No login/crypto/config-persistence surface, so the full 4-lens gate is correctly not required here. 915 tests green (Debug+Release --warnaserror); conformance rule still passes. Ready to merge.

Issue #232, opened 2026-07-13
Close the in-flight SSO login race on Unregister (revocation epoch / session termination)


iderex:
## Summary

`Unregister` (#213) removes a user's canonical links so future SSO logins fail closed, but it does not stop an **already in-flight** login: a concurrent SSO callback that has already resolved the user (via `CreateCanonicalLinkAndUserIfNotExist`, under the config lock) can proceed to `Authenticate` and mint a session *after* the links are removed. A TOCTOU between link-resolution and session-mint.

## Impact

- Window: milliseconds within a single callback request. Link-resolution runs under `ConfigMutationLock`; the session mint (`AuthenticateDirect`) happens after the lock is released. If `Unregister`'s removal lands in that gap, one session is minted for the just-revoked identity.
- Not attacker-controllable: the race depends on the admin's revocation timing coinciding with the target's in-flight login, not on attacker action; the result is at most one lingering session (which an admin revoking a compromised user would typically also terminate via Jellyfin session management).
- Real but narrow, surfaced by CodeRabbit on #213 (CWE-863). Tracked here rather than bolted onto the minimal #213 fix (a proper fix is a "heavy lift").

## Fix options (research the lightest)

1. **Revocation epoch/version** on the user or plugin config, stamped on `Unregister`, re-checked immediately before session mint under the config lock, reject a mint whose identity's revocation state changed since resolution. Cleanest but adds state + a mint-time gate.
2. **Terminate the user's active sessions** on `Unregister` (Jellyfin `ISessionManager` token revocation), kills existing sessions and narrows (not fully closes) the in-flight window; cheap complementary measure.
3. Combine: (2) for immediate effect + (1) for the race.

Prefer the lightest approach that makes revocation effective against an in-flight login without a heavy per-mint cost.

## Notes

Surfaced during #213 delivery (PR #231). The #213 fix is a strict improvement (revocation previously did nothing); this closes the remaining in-flight race. Per the KRITIS-grade security bar, worth a proper fix even though the current window is narrow.

PR #467, opened 2026-07-17
Revoke active tokens when unregistering a user

### LINKED ISSUE #440: Terminate the user's active sessions on Unregister (revoke existing tokens)
## Summary

`Unregister` (SSOController.cs, #213) revokes a user's SSO by removing their canonical links so future SSO logins fail closed, and #232 (the in-flight re-check before the mint) closes the race that let one *new* session slip through during revocation. Neither, however, terminates the user's **already-established** sessions: a token minted before the revocation stays valid until it expires or the admin kills it manually via Jellyfin's session management.

For a hard revoke (e.g. a compromised account), the operator almost always wants the existing sessions gone too. `ISessionManager` exposes this and the controller already injects it.

## Proposal

In `Unregister`, after removing the canonical links, terminate the user's active sessions:

```csharp
await _sessionManager.RevokeUserTokens(user.Id, null).ConfigureAwait(false);
```

(`RevokeUserTokens(Guid userId, string? currentAccessToken)` is present in both the ABI-floor `Jellyfin.Controller` 10.11.0 and the 10.11.11 build target; `null` revokes all of the user's tokens. Retain `_sessionManager` as a field, it is currently constructor-injected but only forwarded to `SessionMinter`.)

## Scope / notes

- This is a **complement to #232, not a substitute**: it does not close the in-flight mint race (a token minted after the sweep but before the request completes survives), #232's re-check is what closes that. This issue is about killing *existing* sessions on revoke.
- It is a behaviour change to revoke semantics (existing sessions now end immediately), so it warrants its own E2E checklist note and a harness test asserting `Unregister` calls `RevokeUserTokens(user.Id, null)`.
- Related, out of scope here (a possible separate belt-and-suspenders): the #232 re-check is the *last* gate, so a revoked-in-flight login still persists its permission/avatar write before the session is denied. A pre-write re-check would avoid that write at the cost of widening the mint window; noted for consideration, not proposed here.

Surfaced by the #232 design panel (approach 2). Priority low, a convenience/hardening improvement on top of the now-closed race.

iderex:
# What & why

Closes #440.

`Unregister` removes a user's canonical SSO links so future SSO logins fail
closed (#213), and #232 closes the in-flight mint race, but neither terminates
the sessions **already established** for that user: a token minted before the
revoke stays valid until it expires or an admin kills it by hand. For a hard
revoke (a compromised account) the operator wants those sessions gone too.

This adds one step to the admin-only `Unregister`: after the canonical links are
removed and the provider switch persisted, it calls
`RevokeUserTokens(user.Id, null)` to terminate the target user's active
sessions and tokens.

Design:

- **Which endpoint revokes.** Only the admin-elevated `POST /sso/Unregister/{username}`
  (the hard "lock this account down" path) revokes. The per-link
  `DeleteCanonicalLink` unlink and per-provider disable are left as-is, before
  this change nothing on any path revoked tokens, and this strictly adds
  revocation to the hardest path. Extending revocation to the single-link unlink
  / provider-disable surfaces is a reasonable follow-up (candidate issue below),
  out of #440's scope.
- **API.** `ISessionManager.RevokeUserTokens(Guid userId, string? currentAccessToken)`
  returning `Task`; the second arg is the token to *preserve*, so `null` revokes
  all of the user's tokens. Verified empirically (metadata-reader probe) that
  this exact signature is present in **both** the ABI floor `Jellyfin.Controller`
  10.11.0 and the 10.11.11 build target, so no `MissingMethodException` risk (the
  `abi-floor` CI job also rebuilds against the floor with `--warnaserror`).
- **Scope.** Strictly the single resolved `user.Id` (an early `NotFound()` guards
  the unknown-user case), so it cannot mass-revoke other users.
- **Fail-closed + availability.** The revoke runs **last**, after the durable
  link removal (persisted synchronously under the config-store lock) and the
  awaited provider-switch persist. So if it throws, the unregister's
  security-critical part is already committed, the account fails future SSO
  logins closed, rather than being left half-done. The exception surfaces as a
  500 (the admin learns the token kill failed and can retry) instead of a false
  `Ok`. No login-path lock is held across the revoke I/O. On a self-unregister
  the admin still gets the `200` and is logged out on the next request; the
  account switches to local (`Jellyfin`) auth so password login and the
  last-admin path survive, no permanent lockout.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the durable revoke commits before the token revoke,
      so a throw over-locks (future SSO logins already fail closed) rather than
      under-locks; the revoke runs unconditionally on the resolved user.
- [x] No secrets logged; only `user.Id` (a Guid) and an int count are logged - no
      IdP-controlled string, so no sanitization needed.
- [x] A security review was performed - four refute-by-default lenses (verdicts
      under Notes).
- [x] Log inputs from the IdP are sanitized inline at the log call - N/A, no
      user-controlled string is logged.

## Quality checklist

- [x] No duplicated logic; the change reuses the already-injected
      `ISessionManager` and adds one awaited call plus a field.
- [x] The code is self-documenting; the one comment explains *why* the call is
      ordered last and why `null` is safe.
- [x] Architecture-conformance tests pass; this change establishes no new
      structural property to lock in.

## Verification

- [x] `dotnet build -c Debug --warnaserror` and `-c Release --warnaserror` are
      green (0 warnings).
- [x] `dotnet test` is green (921 tests, incl. 4 new).
- [x] Prettier, N/A, only `.cs` files changed.
- [x] `scripts/check-jellyfin-compat.sh` is green.

## Notes

**Tests** (`SSO-Auth.Tests/SSOControllerUnregisterTests.cs`):
`RevokesTheTargetUsersActiveTokens` (revoked once with `(UserId, null)`),
`DoesNotRevokeTokensForOtherUsers` (scope), `TokenRevokeNoOp_StillCompletesOk`
(no-op revoke still returns `Ok`), `WhenTokenRevokeFails_LinksAlreadyRemovedAndProviderPersisted`
(revoke-last ordering: a throw leaves the links removed and provider persisted).

**Adversarial review, four refute-by-default lenses:**

- Availability / mass-lockout, **PASS**: single-user, elevation-gated,
  last-in-sequence; password login survives, no login-path lock or cache added.
- Security bypass / fail-open, **PASS**: revoke is unconditional, ordered
  fail-closed (over-locks on failure), `null`=revoke-all is the stronger choice,
  and the #232 boundary is represented honestly.
- Correctness, **PASS**: `null` valid, awaited correctly, wiring correct, the
  ordering claim holds statement-by-statement, tests assert what they claim.
- Integration, **NEEDS_IMPROVEMENT → resolved at the process level**: the only
  finding was the missing E2E-CHECKLIST note the issue itself calls for.
  `docs/E2E-CHECKLIST.md` is a git-ignored local doc (never part of a PR) and the
  isolated worktree cannot edit the canonical copy, so the exact append text is
  handed over below for the local checklist. The code integration was verified
  clean (ABI floor + target, DI, self-unregister, WebSocket teardown expected).

**E2E-CHECKLIST note to add locally** (under "Account linking & safety"):

> - [ ] Unregister terminates active sessions (#440): with the target user signed
>   in on one or more clients, `POST /sso/Unregister/{username}` force-logs-out
>   those clients, their WebSockets drop and a previously-issued token now
>   returns 401 (mocks cannot exercise the real teardown). Scoped to the one
>   user: a second signed-in user's sessions are untouched. An admin
>   unregistering their OWN account still gets the 200 but is logged out on the
>   next request. Complement to #232 (mint race); this kills already-issued tokens.

**Out of scope / follow-up candidate:** extend token revocation to
`DeleteCanonicalLink` (single-link unlink) and per-provider disable, a
compromised-IdP-identity unlink arguably should also kill live tokens.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerUnregisterTests.cs
+++ b/SSO-Auth.Tests/SSOControllerUnregisterTests.cs
@@ -50,6 +50,69 @@ public class SSOControllerUnregisterTests
         Assert.False(links.ContainsKey("sub-alice"));
     }
 
+    [Fact]
+    public async Task Unregister_KnownUser_RevokesTheTargetUsersActiveTokens()
+    {
+        var harness = new SsoControllerHarness();
+        SeedUser(harness);
+
+        await harness.Controller.Unregister("alice", "Jellyfin");
+
+        // A hard revoke must also terminate the user's already-issued tokens, scoped to this one user; null
+        // revokes all of their tokens (#440).
+        await harness.SessionManager.Received(1).RevokeUserTokens(UserId, null);
+    }
+
+    [Fact]
+    public async Task Unregister_DoesNotRevokeTokensForOtherUsers()
+    {
+        var harness = new SsoControllerHarness();
+        SeedUser(harness);
+        var otherUser = Guid.Parse("55555555-5555-5555-5555-555555555555");
+
+        await harness.Controller.Unregister("alice", "Jellyfin");
+
+        // The revoke is scoped strictly to the resolved target — no other user's tokens may be swept.
+        await harness.SessionManager.DidNotReceive().RevokeUserTokens(otherUser, Arg.Any<string?>());
+    }
+
+    [Fact]
+    public async Task Unregister_TokenRevokeNoOp_StillCompletesOk()
+    {
+        // With the mock's default (a completed no-op task) the revoke changes nothing; the unregister must
+        // still persist the provider switch and return Ok.
+        var harness = new SsoControllerHarness();
+        var user = SeedUser(harness);
+
+        var result = await harness.Controller.Unregister("alice", "Jellyfin");
+
+        Assert.IsType<OkResult>(result);
+        await harness.UserManager.Received(1).UpdateUserAsync(user);
+        await harness.SessionManager.Received(1).RevokeUserTokens(UserId, null);
+    }
+
+    [Fact]
+    public async Task Unregister_WhenTokenRevokeFails_LinksAlreadyRemovedAndProviderPersisted()
+    {
+        // The token revoke runs LAST, after the durable revoke is committed, so a failure there cannot leave
+        // the unregister half-done: the links are already dropped and the provider switch persisted (#440).
+        var harness = new SsoControllerHarness(c =>
+            c.OidConfigs["keycloak"] = new OidConfig
+            {
+                CanonicalLinks = new SerializableDictionary<string, Guid> { ["sub-alice"] = UserId },
+            });
+        var user = SeedUser(harness);
+        harness.SessionManager.RevokeUserTokens(Arg.Any<Guid>(), Arg.Any<string?>())
+            .Returns(Task.FromException(new InvalidOperationException("session store unavailable")));
+
+        await Assert.ThrowsAsync<InvalidOperationException>(() => harness.Controller.Unregister("alice", "Jellyfin"));
+
+        var links = SSOPlugin.Instance.ReadConfiguration(c => c.OidConfigs["keycloak"].CanonicalLinks);
+        Assert.False(links.ContainsKey("sub-alice"));
+        Assert.Equal("Jellyfin", user.AuthenticationProviderId);
+        await harness.UserManager.Received(1).UpdateUserAsync(user);
+    }
+
     // Registers a user with the harness's mocked IUserManager so GetUserByName resolves it.
     private static User SeedUser(SsoControllerHarness harness)
     {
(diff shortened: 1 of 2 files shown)


iderex (commented):
My sign-off — security, session/logout path. Delivered by a delegated agent; independently gated by the four-lens adversarial review — availability, bypass, correctness PASS; integration NEEDS_IMPROVEMENT was solely the E2E-CHECKLIST note the issue calls for (that doc is git-ignored/local, unreachable from the isolated worktree — appended to the canonical copy separately), no code defect. Confirmed: revocation uses ISessionManager.RevokeUserTokens(userId, null) (second arg = token to preserve; null revokes all), signature verified empirically against BOTH the 10.11.0 ABI floor and the 10.11.11 target; it runs LAST in Unregister after the durable link removal + provider-switch persist, so a throw leaves the unregister complete and fail-closed (never a false Ok, never half-done); strictly scoped to the single resolved user.Id; self-unregister and last-admin stay recoverable via surviving local password auth. 4 new tests, 921/923 green (Debug+Release --warnaserror); CodeQL C# green. Ready to merge.

Issue #32, opened 2026-07-11
Replay of a captured SAML assertion


iderex:
A validly-signed SAML assertion can currently be submitted more than once: nothing records that an assertion ID has already been used to mint a session, so a captured assertion can be replayed within its validity window to create additional sessions.

Enforce one-time use of the assertion ID.

Acceptance criteria:

- [ ] An assertion ID that already minted a session is rejected on reuse (HTTP problem response).
- [ ] Fail-closed: an assertion with no usable ID is rejected.
- [ ] A consumed ID is retained for at least the whole window the assertion would still be accepted (its NotOnOrAfter plus the validation clock skew), with a sensible floor.
- [ ] Enforced at the session-minting endpoint only, so the two-step post-then-auth flow consumes the ID exactly once.
- [ ] Pure, unit-tested cache; the assertion ID is read from the signed assertion.

PR #469, opened 2026-07-17
Bound SamlReplayCache with a hard cap and throttled prune

### LINKED ISSUE #452: Give SamlReplayCache the hard cap and shared IntervalGate prune its sibling caches use
## What & why

`SamlReplayCache` (`SSO-Auth/Api/SamlReplayCache.cs`) is the only cache in the
login-path family that diverges from the pattern the other three converged on:

| Cache | Expired-entry prune | Global cap | Per-client sub-cap |
|-------|---------------------|-----------|--------------------|
| `OidcStateStore` | `IntervalGate`-throttled | 100k | yes |
| `SamlRequestCache` | `IntervalGate`-throttled | 100k | yes |
| `SsoRateLimiter` | `IntervalGate`-throttled | 100k | n/a |
| `SamlReplayCache` | **unthrottled full-dictionary sweep every call** (`TryConsume`, lines 50-56) | **none** | none |

`TryConsume` walks the whole `_consumed` dictionary on every consume, and
`_consumed` carries no size bound. The single instance is process-wide
(`SSOController.SamlReplays`).

This is **not** a performance problem and we should not frame it as one.
`TryConsume` runs only after signature validation on the SAML mint/link path
(`SSOController` ~1016 / ~1252), so it is not anonymously floodable; the retained
set is bounded by legitimate SAML login volume over the ~1h retention floor
(`ComputeRetention`); and the O(n) `DateTime` compare is microseconds against the
RSA/XML-DSig verification the same request already paid. Throttling the sweep is
close to a wash, it trades a little CPU for a little memory.

The value is **consistency and defense-in-depth** (directives 1 and 4):

- one uniform cache pattern instead of three-plus-one, reuse the `IntervalGate`
  primitive and delete the ad-hoc prune loop, so there is less code, not more;
- a hard memory bound (CWE-400 parity) so the consumed-assertion set cannot grow
  without limit, matching every sibling.

## The design subtlety (why this is tracked, not a drive-by fix)

A cap here is **not** the same as in the siblings and has to go through the
security-review gate:

- The sibling caches refuse-at-cap because refusing fails **one login closed**.
  The replay cache is the opposite: refusing to record a consumed assertion fails
  the replay check **open**, the assertion is never marked used, so it stays
  replayable. A naive "reject the insert at the cap" would be a fail-open
  regression of the exact guarantee this cache exists for (#32).
- Rejecting the *login* at the cap instead is a mass-lockout / availability
  hazard (the availability lens would rightly flag it).
- Evicting the soonest-to-expire entry to free a slot reopens the replay window
  for that one assertion.

So the correct cap behaviour needs a deliberate decision, evict expired-first
and only then bound, or fail the login closed behind a bounded, observable
signal, and it must preserve the invariant that every within-window consumed
assertion stays recorded for its full validity window.

## Minimal-code sketch

- Add `private readonly IntervalGate _pruneGate = new(TimeSpan.FromMinutes(1));`
  and wrap the sweep in `if (_pruneGate.TryEnter(nowUtc)) { ... }`, mirroring
  `SamlRequestCache.PruneExpired`. When the sweep is throttled, `TryConsume` must
  treat an existing **expired** entry as absent (replace it) so a lingering
  unswept entry cannot false-reject a fresh login, moot in practice (assertion
  IDs are unique `xsd:ID`s) but correct.
- Decide and implement the cap semantics per the above (fail-closed, no
  replay-window reopen). Run `/security-review` (availability + bypass lenses)
  since it touches the replay invariant.

No behaviour change on the happy path; the replay guarantee and the upstream
time-bounds rejection are unchanged.

## Notes

Surfaced by a performance review of the login hot path. That review found no
measurable per-login speed wins, the path is already well optimised
(`IHttpClientFactory` for the OIDC client, discovery/JWKS reuse from #247,
`IntervalGate`-throttled caches, bounded rate limiter and per-client budgets).
This is the one residual code-consistency/hardening gap worth tracking, and it is
explicitly a quality/hardening item, not a perf fix.

iderex:
# What & why

Bring `SamlReplayCache` (`SSO-Auth/Api/SamlReplayCache.cs`), the SAML
assertion-ID replay cache, in line with the sibling login-path caches
(`SamlRequestCache`, `OidcStateStore`): an `IntervalGate`-throttled
expired-entry sweep plus a hard 100k global cap, replacing the unthrottled
full-dictionary sweep on every consume and the unbounded set it carried
before. This is the last cache in the family that diverged from the pattern
the other three converged on, a consistency / defense-in-depth item
(directives 1 and 4), not a performance fix.

## Design (current vs sibling structure, cap + eviction semantics)

**Before:** `TryConsume` walked the whole `_consumed` dictionary on every
call (unthrottled O(n) sweep) and the set had no size bound.

**After:** the sweep is throttled through a `_pruneGate` `IntervalGate`
(1-minute interval), and a `DefaultMaxEntries = 100_000` cap bounds the set - 
the same cap value and interval the siblings use.

The cap had to go through the security-review gate because a replay cache is
the opposite of the siblings at capacity:

- The sibling caches refuse-at-cap because refusing fails **one login
  closed**. For a replay cache, *failing to record* a consumed assertion
  would fail the replay check **open** (the assertion is never marked used, so
  it stays replayable). A naive LRU/oldest-first eviction of a *live* entry
  reopens the replay window for exactly that assertion (#32).
- Rejecting the *login* at the cap is fail-closed, but a blind reject is an
  availability hazard.

**Resolution, eviction is tied strictly to expiry; the cap fails the login
closed, never a live entry:**

- `SweepExpired` removes only entries whose recorded retention has elapsed
  (`value <= nowUtc`), using the value-matched `TryRemove(KeyValuePair)`
  overload so a concurrently-refreshed entry is never dropped.
- At the cap, only a **new** id would grow the set. The cap block first runs
  an unthrottled reclaim of expired entries (so a cache full of
  expired-but-unswept entries never false-rejects a fresh login, strictly
  *more* available than the siblings), and only if the cache is **still full
  of unexpired entries** does `TryConsume` return `false`, which the caller
  treats as a rejected login (fail closed). No live entry is ever evicted.
- The one-time claim is an atomic `TryAdd`, so two concurrent presentations of
  the same assertion cannot both succeed. A lingering **expired** entry with a
  reused xsd:ID is replaced in place via a value-matched `TryUpdate` (moot in
  practice, assertion IDs are unique, but correct once the sweep is
  throttled).

**Why replay is not re-opened:** the only removals are (a) expired entries and
(b) an in-place replace of an already-expired entry. A within-window
(unexpired) consumed id is never dropped, overwritten, or refreshed. So a
replay of any assertion still inside its validity window always finds its
entry and is rejected. `ComputeRetention` (unchanged) retains until
`max(NotOnOrAfter + skew, now + 1h)`, i.e. at least the whole window the
assertion is still acceptable, so an entry never expires while the assertion
is still replayable.

The caller (`SSOController.TryConsumeSamlReplay`) is unchanged, the
`internal bool TryConsume(string, DateTime, DateTime)` signature is
byte-identical, and `false` still rejects the login.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (Here: a missing assertion ID and a
      cap refusal both return `false` → login rejected; a live entry is never
      evicted, so replay is never re-admitted.)
- [x] No secrets logged; secrets are redacted on config export. (No logging in
      this type.)
      log calls.)

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. (Debug + Release, 0 warnings.)
- [x] `dotnet test` is green. (923 tests, Debug + Release.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (None touched.)

## Notes

**Adversarial self-review (4 refute-by-default lenses), all PASS.**

- **Security-bypass**, PASS. No cap path ever evicts or refreshes a *live*
  entry; `SweepExpired` and the `TryUpdate` replace touch only entries with
  `value <= nowUtc`; `ComputeRetention` retains until `max(bounds)+skew >=` the
  acceptance window, so replay is never re-opened. Concurrent identical-id
  presentations cannot both win (atomic `TryAdd`). A cap refusal is fail-closed
  (caller rejects the login), not fail-open.
- **Availability / mass-lockout**, PASS. The 100k live-entry cap is
  unreachable by legitimate post-signature, rate-limited traffic; the
  unthrottled cap-path reclaim makes this cache *more* available than the
  siblings; no livelock in the claim loop; less hot-path work than the code it
  replaces.
- **Correctness**, PASS. The `while` loop terminates and returns the right
  result under concurrency; `TryRemove(KeyValuePair)` value-match is safe;
  boundaries (`value <= nowUtc`, `Count >= _maxEntries`) are consistent; the
  new tests exercise what they claim (accounting for `IntervalGate` entering on
  its first call).
- **Integration**, PASS. Caller signature byte-identical; the single static
  instance and the parameterless ctor surface are intact; blast radius is 3
  files; the conformance rule is non-vacuous.

**Conformance-rule tightening (from the integration lens):** the new rule
`LoginPathCaches_ThrottleTheirExpiredEntrySweepThroughIntervalGate` originally
asserted only that each cache declares *some* `IntervalGate` field, but the
siblings carry two (`_pruneGate` + `_capWarnGate`), so that would not catch a
sibling that dropped its prune gate while keeping cap-warn. Tightened to key on
the `_pruneGate` field specifically (the field name is uniform across all three
caches), so it tracks the prune throttle the rule's name promises.

**Out of scope / follow-up:** the siblings emit a throttled capacity warning at
their cap for observability. This change keeps the `TryConsume` signature
byte-identical and does not wire an equivalent signal, because doing so would
require touching `SSOController` (a parallel PR owns that file). The security
property does not depend on the warning. A follow-up can plumb a throttled
cap-refusal signal from `SamlReplayCache` through the controller once that PR
lands.

Closes #452

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -162,6 +162,29 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void LoginPathCaches_ThrottleTheirExpiredEntrySweepThroughIntervalGate()
+    {
+        // Locked in by #452: the login-path caches converged on one bounding pattern — an
+        // IntervalGate-throttled expired-entry sweep plus a hard global cap — so none can regress to the
+        // unthrottled full-dictionary sweep (or the unbounded set) SamlReplayCache carried before #452.
+        // Each named cache must declare the PRUNE gate specifically (an IntervalGate field named
+        // "_pruneGate"), not merely some IntervalGate: the siblings also carry a "_capWarnGate", so keying
+        // on the field type alone would miss a sibling that dropped its prune gate but kept cap-warn.
+        // SamlRequestCache and OidcStateStore already had it (#246, #327); SamlReplayCache adopted it (#452).
+        const string pruneGateField = "_pruneGate";
+        var caches = new[] { typeof(SamlReplayCache), typeof(SamlRequestCache), typeof(OidcStateStore) };
+        var missing = caches
+            .Where(t => !t.GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.DeclaredOnly)
+                .Any(f => f.FieldType == typeof(IntervalGate) && f.Name == pruneGateField))
+            .Select(SimpleName)
+            .ToList();
+
+        Assert.True(
+            missing.Count == 0,
+            "Every login-path cache must throttle its expired-entry sweep through an IntervalGate field (#452): " + string.Join(", ", missing));
+    }
+
+    [Fact]
     public void Controller_NeverTouchesProviderLinkMaps()
     {
         // Locked in by the link/unlink admin-surface extraction (#372) and completed by #383: the two
--- a/SSO-Auth.Tests/SamlReplayCacheTests.cs
+++ b/SSO-Auth.Tests/SamlReplayCacheTests.cs
@@ -54,6 +54,89 @@ public class SamlReplayCacheTests
         Assert.True(cache.TryConsume("_a", later.AddMinutes(10), later));
     }
 
+    // --- Hard cap + throttled prune (#452) ---
+
+    // A cache whose cap is small and reachable (production 100k is not).
+    private static SamlReplayCache SmallCache(int cap = 3) => new SamlReplayCache(cap, TimeSpan.FromMinutes(1));
+
+    [Fact]
+    public void TryConsume_ReplayStillRejected_UnderCapPressure()
+    {
+        // The security regression: a live consumed assertion must NEVER be evicted to make room, or its
+        // replay would be re-admitted. Record X (live), fill the rest of the cap with other live logins,
+        // then push a new distinct login (refused fail-closed), then replay X — X must still be rejected,
+        // proving its live entry survived the cap pressure.
+        var cache = SmallCache(cap: 3);
+        var expiry = Now.AddMinutes(10);
+
+        Assert.True(cache.TryConsume("_x", expiry, Now));
+        Assert.True(cache.TryConsume("_f1", expiry, Now));
+        Assert.True(cache.TryConsume("_f2", expiry, Now)); // cache now full of live entries
+
+        Assert.False(cache.TryConsume("_new", expiry, Now)); // fail closed: no live entry evicted
+        Assert.False(cache.TryConsume("_x", expiry, Now)); // X still recorded -> replay still caught
+    }
+
+    [Fact]
+    public void TryConsume_AtCap_NewDistinctLogin_IsRejectedFailClosed()
+    {
+        var cache = SmallCache(cap: 2);
+        var expiry = Now.AddMinutes(10);
+        Assert.True(cache.TryConsume("_a", expiry, Now));
+        Assert.True(cache.TryConsume("_b", expiry, Now)); // at cap with live entries
+
+        Assert.False(cache.TryConsume("_c", expiry, Now));
+        Assert.Equal(2, cache.Count); // the refusal did not grow or evict
+    }
+
+    [Fact]
+    public void TryConsume_AtCap_WithExpiredEntries_ReclaimsThenAdmits()
+    {
+        // At the cap but full of EXPIRED (not yet swept) entries: the unthrottled reclaim at the cap frees
+        // the slots and the fresh login is admitted, even though the routine sweep is still throttled.
+        var cache = SmallCache(cap: 3);
+        var shortExpiry = Now.AddSeconds(30);
+        Assert.True(cache.TryConsume("_a", shortExpiry, Now)); // first call enters the prune gate
+        Assert.True(cache.TryConsume("_b", shortExpiry, Now));
+        Assert.True(cache.TryConsume("_c", shortExpiry, Now)); // at cap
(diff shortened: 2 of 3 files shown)


iderex (commented):
My sign-off — security hardening (SAML replay defense + DoS bound). Delivered by a delegated agent; independently gated by the four-lens adversarial review — all PASS. Confirmed the load-bearing property: every removal path touches only expired entries (value <= nowUtc) — SweepExpired uses the value-matched TryRemove(KeyValuePair) overload and the only in-place TryUpdate fires solely on an already-expired entry — so a within-window consumed id is never dropped, overwritten, or refreshed and a replay inside the validity window always finds its entry and is rejected. At the cap only a NEW id would grow the set; the cap path reclaims expired entries (unthrottled) first and, if still full of unexpired entries, returns false = a rejected login (fail closed), never evicting a live entry. ComputeRetention (unchanged) retains until max(NotOnOrAfter+skew, now+1h) >= the acceptance window. Sibling parity (100k cap, 1-min IntervalGate-throttled prune); caller signature byte-identical; the new conformance rule was tightened to key on the _pruneGate field specifically (so a dropped prune gate is caught). 923 tests green (Debug+Release --warnaserror); CodeQL C# green. Ready to merge.

Issue #218, opened 2026-07-13
Gate account adoption on a verified stable claim, not preferred_username


iderex:
## Summary

Adoption of a pre-existing unlinked account resolves by the mutable `preferred_username` (`GetUserByName(username)`); `email_verified` and a stable/verified claim are never consulted. When an admin opts into `AllowExistingAccountLink` for a provider whose `preferred_username` is neither unique nor verified, a new principal that asserts an existing Jellyfin user's name is routed to adoption and takes it over.

## Impact, opt-in residual risk

The link is correctly keyed on the stable `sub` (#155), but `existingAccountUserId = _userManager.GetUserByName(username)?.Id` (`SSO-Auth/Api/SSOController.cs:962`) matches on display name only. This is the documented purpose of the `AllowExistingAccountLink` opt-in, so it is a residual-risk/hardening note rather than a default-config vulnerability, but the opt-in currently assumes, without enforcing, that the IdP guarantees unique, non-reassignable usernames.

## Affected code

- `SSO-Auth/Api/SSOController.cs:962`

## Fix sketch

Port V2's `EmailVerifiedResolver` + `PrivilegedAdoptionResolver`: require `email_verified = true` (and/or match on a verified email) before adoption, and refuse adoption of privileged accounts. Document that `preferred_username`-based adoption assumes IdP-guaranteed unique, non-reassignable usernames.

## V2 status

Partial in V2: adoption is still keyed on display name in `AccountLinkResolver.Resolve`, but V2 adds `EmailVerifiedResolver.cs` and `PrivilegedAdoptionResolver.cs` as gates. Port those helpers to close the gap here.

## Notes

Surfaced by the 2026-07-13 standing security audit. Pre-production, tracked publicly per the current triage policy.

Issue #133, opened 2026-07-12
Make SSO account adoption atomic (check-then-act across MutateConfiguration)


iderex:
Surfaced by review on #132 (audit log). Pre-existing, not introduced there: the account-adoption path checks then acts non-atomically. `CreateCanonicalLinkAndUserIfNotExist` reads the existing link / `GetUserByName` OUTSIDE the config lock, then `AccountLinkResolver.Resolve` decides, then `CreateCanonicalLink` writes inside `MutateConfiguration`. The individual writes are serialized (P2#8), but the check→decide→write sequence is not one atomic operation, so two concurrent first-logins for the same identity can both resolve to "adopt", double-emit the new `[SSO Audit]` AccountAdopted line, and one link write can overwrite the other. The adoption case also discards the `ActionResult` from `CreateCanonicalLink`, so a failed link could fall through as a successful adoption.

Fix: move the existence check and the insert into a single `MutateConfiguration` read-modify-write, emit the audit event only when that operation actually performed the adoption, and preserve the error contract (propagate a failed link instead of ignoring it). Add a concurrency characterization test. Path-instruction requirement: login-path check-then-act on shared collections must be atomic.

PR #477, opened 2026-07-17
Fold legacy-link re-key into one authoritative migration transaction

### LINKED ISSUE #363: Fold legacy-link resolution and migration into one config mutation
Raised by review on #358 (CodeRabbit, adversarially verified there). `ResolveOrCreateAsync` resolves the subject/legacy link candidates in one locked `Read`, then performs the #155 legacy re-key in a **separate** `Mutate`, two lock acquisitions, so a concurrent writer can interpose between snapshot and migration.

Verified blast radius (why this is hardening, not a vulnerability):
- The harmful outcome is not reachable: the request only ever binds to the value from its **own** verified snapshot (both candidates were live links keyed to this login's own identity at snapshot time); no interleaving hands it a concurrent writer's account.
- `MigrateCanonicalLinkKey` never overwrites a live subject key (its guard no-ops when `newKey` exists and is live), so persistent state converges on the authoritative subject key at the next login.
- The deleted-user race fails closed downstream (`Authenticate` re-fetches by id and throws when the user is gone) and is common to every login path, not migration-specific.
- Pre-existing #155/#353 structure; #358 only gated migration behind `AllowExistingAccountLink`, so it now runs strictly less often.

Still worth closing: the repo's own invariant is that login-path check-then-act on shared config state is a single atomic read-modify-write (#133 did exactly this for link creation). Fold the candidate resolution + `ResolveCanonicalLink` decision + re-key into ONE `Mutate` (or have the migration mutation return the authoritative mapping and use that), removing the window entirely instead of reasoning about its safety.

Needs the full adversarial gate (login path) and a concurrency characterization test if practical.

Refs #155, #354, #358, #133

iderex:
# What & why

Closes #363.

The #155 legacy-link (pre-#155 username-keyed) migration in
`CanonicalLinkService.ResolveOrCreateAsync` resolved the subject/legacy link
candidates in one locked `Read`, then performed the re-key in a **separate**
`Mutate`, two lock acquisitions, so a concurrent login for the same identity
could interpose between the snapshot and the re-key. The harmful outcome was
not reachable (the re-key guard is idempotent, a live subject key is never
overwritten, and the deleted-target race fails closed downstream at the #232
mint re-check), but the repo invariant is that a login-path check-then-act on
shared config state is a single atomic read-modify-write (#133 did exactly this
for link creation).

Rather than keep reasoning about the window's safety, we take the issue's
second suggested route: **have the migration mutation return the authoritative
mapping and bind the login to that.**

Current path (two acquisitions, stale-snapshot bind):
`Read` candidates -> decide migrate -> `MigrateCanonicalLinkKey` (void re-key,
second `Mutate`) -> the login keeps binding to the **pre-migration** snapshot's
`linkedUserId`.

New path:
- `void MigrateCanonicalLinkKey` becomes `Guid? MigrateAndResolveCanonicalLink`
 , the *same* idempotent re-key, inside one config transaction, that now
  **returns** the authoritative user id the identity resolves to. The caller
  binds `linkedUserId` to that return, so a concurrent winner's live subject
  link, or a since-deleted target, drives the final decision, never a stale
  read.
- The preliminary candidate `Read` stays as the fast common path: an ordinary
  login that resolves a live subject link (or needs no migration) remains a
  pure read and **persists nothing**, so the fold does not turn every login
  into a config write. The migrate transaction is entered only when a re-key is
  actually due.

Atomicity argument: the value the terminal switch acts on (`linkedUserId`) now
originates *inside* the same one-lock transaction as the re-key write. The
preliminary Read no longer supplies the bound id, it only selects the
migrate/no-migrate branch and the admin gate, both of which are re-validated
authoritatively inside the transaction (the `requireEnabled` guard and the
re-key guard itself). No stale snapshot can drive a wrong write. Concurrent
same-identity logins serialize on the one lock and resolve to exactly one
winner with no double-write.

Behaviour-preserving: the re-key guard is the old one with renamed locals
(no key-direction swap); SAML passes key == name so it never enters the migrate
path (unchanged); the #218 admin-adoption gate, the #354/#361 name-trust
guards, and the #373/#380 provider-vanished fail-closed throw all fire exactly
as before, ahead of any write. The one divergence is strictly *more*
fail-closed: in the rare concurrent-full-revoke-during-migration edge the new
code returns `null` and routes to the create/adopt gate (re-adopting through
the full eligibility gate) instead of stale-binding a just-revoked id and then
failing at the mint re-check.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

Adversarial review gate (login/linking path), four refute-by-default lenses:

- Availability / mass-lockout: **PASS**, migration only entered when due; hot
  path stays a pure read; no new refusal path (the migrate branch requires
  `allowExistingAccountLink == true`, so a `null` return falls through to
  adopt/create, never `RejectNameTaken`); every divergent race is equal-or-more
  available than the committed code.
- Security-bypass / fail-open: **PASS**, the #218 admin gate still throws
  before any re-key; the returned id is either the just-cleared non-admin legacy
  target or the authenticated subject's own link; the else-branch never
  overwrites a live subject link and never returns a foreign subject's id;
  deleted-target handling defers to the intact #232 re-check.
- Correctness / atomicity / concurrency: **PASS**, the re-key and authoritative
  re-resolution are one lock acquisition whose returned id (not the preliminary
  snapshot) drives the bind; guard logic preserved; concurrent same-identity
  logins resolve to exactly one winner with no double-write.
- Integration: **PASS**, public signature unchanged; both callers
  (`SSOController` OID/SAML) untouched; SAML is a genuine no-op; persist and
  locking semantics preserved; no dangling references to the removed method.

## Quality checklist

      testable unit.
      change establishes is locked in as a rule in `ArchitectureConformanceTests`
      (`LegacyLinkMigration_ReturnsTheAuthoritativeMapping_NotAVoidReKey`).

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green (Debug + Release, 0
      warnings).
- [x] `dotnet test` is green (931/931, both configs).
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change (none
      touched).

## Notes

Scope is confined to `CanonicalLinkService` (+ tests, + conformance) per the
#318 step-2 plan; `SSOController`, `IntervalGate`, and the caches are untouched.
`ProviderConfigStore` is deliberately not modified: with only `Read` (no
persist) and `Mutate` (always persists) primitives, the fold keeps a
preliminary `Read` fast path so the hot login path stays write-free, and
escalates to a single authoritative `Mutate<Guid?>` only when a re-key is due.

The new conformance rule pins the *helper's* `Guid?` return as a proxy for
"the login binds to the authoritative value"; the caller's consumption of that
return is covered by the behavioural tests (hot-path-no-write,
migrate-persists-once, concurrent-winner-no-overwrite).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -298,6 +298,38 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void LegacyLinkMigration_ReturnsTheAuthoritativeMapping_NotAVoidReKey()
+    {
+        // Locked in by #363: the #155 legacy-link re-key runs as a SECOND lock acquisition after the
+        // candidate-resolving read, so a concurrent login could migrate the same identity between the two.
+        // The fix folds the re-key and the re-resolution into one config transaction that RETURNS the
+        // authoritative user id, and the caller binds the login to that returned id rather than the
+        // pre-migration snapshot. Structurally that means the migration helper must be a value-returning
+        // mutation (Guid?), never a fire-and-forget void re-key whose result the caller ignores — a
+        // revert to void would silently reopen the window. Reflection over the service's own private
+        // methods pins it: any migration helper (name contains "Migrate") must return Guid?.
+        var migrationHelpers = typeof(CanonicalLinkService)
+            .GetMethods(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.DeclaredOnly)
+            .Where(m => m.Name.Contains("Migrate", StringComparison.Ordinal))
+            .ToList();
+
+        // Sentinel against a vacuous pass: a rename that drops "Migrate" from the helper's name would make
+        // the scan match nothing and pass for the wrong reason, so require at least one to exist and force
+        // a conscious update of this rule if the naming changes.
+        Assert.True(
+            migrationHelpers.Count > 0,
+            "No legacy-link migration helper (a private method whose name contains \"Migrate\") was found on CanonicalLinkService; it was renamed, so point this rule at the new name so the return-type invariant keeps guarding #363.");
+
+        var voidReKeys = migrationHelpers
+            .Where(m => m.ReturnType != typeof(Guid?))
+            .Select(m => $"{m.Name} -> {m.ReturnType.Name}")
+            .ToList();
+        Assert.True(
+            voidReKeys.Count == 0,
+            "The legacy-link migration must return the authoritative mapping (Guid?) so the login binds to the post-migration state, not a pre-migration snapshot (#363); these do not: " + string.Join(", ", voidReKeys));
+    }
+
+    [Fact]
     public void SSOPlugin_DeclaresNoConfigurationLogicBeyondTheStoreFacade()
     {
         // Locked in by the ProviderConfigStore extraction (#318): SSOPlugin is bootstrap + page
--- a/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
@@ -348,6 +348,93 @@ public class CanonicalLinkServiceTests
         Assert.Contains("refusing to migrate the account link", ex.Message, StringComparison.Ordinal);
     }
 
+    [Fact]
+    public async Task ResolveOrCreateAsync_LiveSubjectLink_PersistsNothing()
+    {
+        // #363 guard: an ordinary login that resolves an existing subject-keyed link must stay a pure
+        // read. Folding the legacy migration into the resolution must NOT turn every login into a config
+        // persist — the common (no-migration) path writes nothing, so the fold cannot regress the hot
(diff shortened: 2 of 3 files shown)


iderex (commented):
My sign-off — security/login-linking path (#318 plan, step 2). Delivered by a delegated agent (which then hit the session limit right after opening the PR — the delivery itself, gates, self-review and metadata were all complete). Independently gated by the four-lens adversarial review, all PASS: availability (migration only entered when due; no mass-lockout), bypass (the #218 admin-adoption gate still throws; no fail-open re-opened), correctness/atomicity (the re-key + authoritative switch are one atomic read-modify-write on linkedUserId, #133-style; SAML passes key==name so never enters the migrate path; the #373/#380 provider-vanished fail-closed throws all fire; the concurrent-full-revoke-during-migration edge fails closed at the #232 mint re-check), integration (public signature unchanged, both callers updated). Fail-closed preserved, no refusal lost. CodeQL C# + build + prettier + dependency-review all green. Ready to merge.

Issue #448, opened 2026-07-17
Bound the per-user avatar store-lock acquire with a cancellation token


iderex:
Surfaced by the review of PR #445 (#400), correctness + availability + bypass lenses (all LOW, non-blocking).

`AvatarService.StoreAsync` acquires the per-user `KeyedLockStore` gate with `CancellationToken.None`, so the same-user acquire wait is unbounded. If one user's store step stalls while holding the gate (e.g. `IProviderManager.SaveImage` blocks on disk I/O, or `ClearProfileImageAsync` on the host DB), concurrent logins for that SAME user park indefinitely.

Blast radius is strictly per-user (the keyed lock cannot stall other users - confirmed no mass-lockout), the operations under the gate are the same local disk/DB calls the login already awaits, and the attacker-controlled slow path (the avatar HTTP fetch) is deliberately OUTSIDE the lock, so this does not materially worsen availability versus the pre-#400 code, which awaited the same operations without a token. Hence LOW.

## Improvement
Thread an ambient request-abort / bounded `CancellationToken` through `TrySetAsync`/`StoreAsync` into `KeyedLockStore.AcquireAsync` so a stalled critical section cannot park same-user logins forever (`AcquireAsync` already honors the token and cleans up on cancel, the plumbing is the only missing piece). Add a test that a cancelled acquire wait releases the waiter and leaks no refcount/permit (KeyedLockStore already has the primitive-level test; this adds the AvatarService wiring).

Refs #400, #445.

PR #478, opened 2026-07-17
Trim redundant per-login work in the avatar fetch

### LINKED ISSUE #248: Trim redundant per-login work in Authenticate: HTTP stack churn, unconditional avatar re-download, double user write
## Problem

`Authenticate` repeats avoidable work on every successful login.

**Avatar fetch rebuilds the whole HTTP stack and re-downloads every time.** `TrySetUserAvatarAsync` (`SSOController.cs:1503-1588`) constructs a new `SocketsHttpHandler` + `HttpClient` per login (`:1522-1532`), a fresh connection pool, so a full TCP+TLS handshake each time and zero reuse, then unconditionally downloads the avatar (up to 10 MB), clears the old profile image, and rewrites it to disk even when nothing changed. It is awaited before the session is minted (`:1470`), so a slow avatar host adds up to the full 10 s deadline to every login.

**Double user write.** The user record is persisted (`:1472`), then when `DefaultProvider` is configured, `AuthenticationProviderId` is assigned and the record persisted again (`:1490-1494`), two full user-store writes per login.

## Fix

- Hoist one `static readonly HttpClient` built with the same handler settings. `AvatarConnectCallback` is already a static, stateless method (`:1714`) and the handler/client are thread-safe, so this is a pure move; set `PooledConnectionLifetime` (~2 min) so DNS changes are honored. The SSRF `ConnectCallback`, redirect bound, size cap, and read deadline stay exactly as they are. Net negative code.
- Skip the fetch when `user.ProfileImage` exists and the resolved avatar URL is unchanged (one comparison). Behavior note to decide at fix time: changed avatar bytes behind the same URL would then not refresh on every login, an acceptable cadence trade, or gate it with `If-Modified-Since` instead.
- Assign `AuthenticationProviderId` before the first `UpdateUserAsync` and delete the second call, one write, identical end state, three fewer lines.

While in the area: `SamlAuth` evaluates `samlResponse.GetCustomAttributes("Role")` twice per login (`SSOController.cs:867`, `:887`), hoist to a local.

iderex:
# What & why

Closes #248.

`Authenticate` repeated avoidable work on every successful login. This trims two of the three items in the issue inside `AvatarService`; the third was already resolved upstream.

**1, HTTP stack churn.** The controller builds a fresh `AvatarService` per request (`SSOController.cs:121`), and `TrySetAsync` built a new `SocketsHttpHandler` + `HttpClient` per fetch, a fresh connection pool, so a full TCP+TLS handshake and zero reuse every login. We hoist one process-wide `static readonly HttpClient` (`SharedClient`) built from the same `CreateHardenedHandler()`, reused across all per-request instances. The hardening is unchanged, SSRF `ConnectCallback` (per-connection + per-redirect-hop IP validation), `MaxAutomaticRedirections = 5`, `UseProxy = false`, the 10s deadline, the `ContentLength` pre-check, and the streamed `ReadCappedAsync` byte cap all stay exactly as they were. We add `PooledConnectionLifetime = 2 min` so DNS changes are still honored despite pooling. User-Agent and the conditional header move onto the per-request `HttpRequestMessage`, so the shared client holds no mutable per-login header state (thread-safe under concurrent logins). The test seam changes from a `Func<HttpMessageHandler>` factory to an injected `HttpClient`.

**2, Unconditional re-download → conditional (the decided cadence).** Per my decision on #248, the fetch is now conditional via **`If-Modified-Since`**, not skip-on-unchanged-URL and not always-download. The header value is the user's last-store timestamp (`user.ProfileImage.LastModified`), exactly "when we last fetched this representation". A **304** keeps the existing avatar and stores nothing; a **200** refreshes it as before. First login (no stored image) sends no conditional header and fetches unconditionally. An origin that ignores the header answers 200, so the flow degrades safely to the old always-download.

Design detail on the `If-Modified-Since` value: we deliberately do **not** advance `LastModified` on a 304, so it stays the stable anchor for the next login's conditional. `new DateTimeOffset(DateTime.SpecifyKind(lastStored, DateTimeKind.Utc))` is total for any stored `Kind` (values are written as `DateTime.UtcNow`, so no time is shifted). The 304 branch is checked **before** `EnsureSuccessStatusCode`, which treats 304 as a non-success throw. Clock-skew degrades safely in both directions (a needless re-fetch, or a bounded stale window that self-corrects), never a broken login.

**3, Double user write.** Already collapsed upstream by #391 (commit `3f03475`): the mint path (`SessionMinter.MintAsync`) now performs a single `UpdateUserAsync` with `AuthenticationProviderId` set beforehand. The remaining second `UpdateUserAsync` in the codebase (`SSOController.cs:1128`) is the `Unregister` admin endpoint, not the login path. Verified empirically, no change needed here, and `SessionMinter.cs` is untouched on this branch.

Structural lock-in: `ArchitectureConformanceTests.AvatarService_HoldsAStaticSharedHttpClient` pins that the outbound stack is a static shared client (rebuild-per-login cannot regress silently); a behavioural test proves two production-built instances reference the same client.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (Avatar fetch is best-effort throughout; all SSRF/content-type/size guards still fail closed — no fetch, no store.)
- [x] No secrets logged; secrets are redacted on config export. (No secret handling touched; IdP-controlled URL and content-type are still sanitized inline at the log call.)
      config persistence, or the release pipeline. (Four-lens adversarial review — see Notes.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (Unchanged: `avatarUrl` and `mediaType` use `ReplaceLineEndings(string.Empty)` at the call site.)

## Quality checklist

      testable unit.
- [x] The change adds no more code than the problem requires. (Net: the shared client removes the per-fetch construct; the conditional adds one header + one early return.)
      why, not what) and matches the target architecture (#318). (Step 15 of #318; changes confined to AvatarService.)

## Verification

- [x] `dotnet test` is green. (933 passed, 0 failed.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (None touched, C# only.)

## Notes

**Adversarial four-lens review (refute-by-default), all PASS:**

- **Availability / mass-lockout, PASS.** No new stall/deadlock/serialization vector. `SocketsHttpHandler` keeps its default `MaxConnectionsPerServer` (int.MaxValue), so concurrent logins never queue behind one another; the per-request `CancellationTokenSource(10s)` still bounds the whole fetch end to end (a trickle-body attack cannot hang a login past ~10s); the shared client is never disposed so it cannot self-brick, and a throw is swallowed by the best-effort catch. The 304 early-return is a net availability win. The per-user store lock (#400) is unchanged and never held across the network fetch.

- **Security-bypass, PASS.** Connection pooling cannot skip the per-connection SSRF `ConnectCallback`: the pool is keyed by request authority (same host only) and a pooled connection is pinned to the already-validated public IP, so a DNS rebind cannot hijack it; reaching an internal IP needs a new connection, which re-runs the callback → `IsBlockedAddress` → reject. `PooledConnectionLifetime = 2 min` tightens rebind staleness relative to the default (infinite). Moving User-Agent to the per-request message is the correct thread-safety fix, same value. A malicious 304 (even with a body) only keeps the user's existing, previously-validated avatar, no new/unvalidated content is stored (body never read). All guards intact and reachable: content-type raster allow-list (#217), ContentLength + streamed byte cap (#220), #447 username path-traversal guard, #384 dotted extension, #400 per-user lock, #448 bounded acquire. No user-write field dropped, `SessionMinter.cs` is byte-identical to `origin/main`.

- **Correctness, PASS** (was NEEDS_IMPROVEMENT, resolved). `If-Modified-Since` ordering, `DateTimeOffset` totality, the first-login guard, and shared-client thread-safety are all correct and fail-safe. The initial pass flagged that the 304 test did not distinguish the correct early-return from a regression (a deleted early-return throws on 304, is swallowed, and leaves the same observable state), fixed by asserting no `Error` was logged, which now fails on that regression (empirically, 304 → `HttpRequestException` into the Error-logging catch). Re-reviewed: PASS.

- **Integration, PASS.** Constructor seam change compiles at every call site (production 5-arg constructor unchanged at `SSOController.cs:121`; `SessionMinter` untouched). `ImageInfo.LastModified` confirmed (against the pinned Jellyfin 10.11.11 package) to be a real settable property initialized to `DateTime.UtcNow`. A never-disposed static `HttpClient` is the correct process-lifetime pattern (Jellyfin does not hot-unload plugins). The `Controller_NeverTouchesRawSocketsOrDns` liveness marker still matches AvatarService, and the new conformance rule is non-vacuous. Two low-severity best-effort notes (below).

**Out-of-scope / follow-up candidates (not addressed here):**

- The issue's "while in the area" note, `SamlAuth` evaluates `samlResponse.GetCustomAttributes("Role")` twice per login (`SSOController.cs`), is in the controller hot file (#318 serial), so it is left out of this PR. Candidate for a separate small refactor.
- Best-effort self-heal edge (integration lens): if the on-disk `profile.*` file is deleted out-of-band while the DB `ImageInfo` record survives, a 304 now skips the re-download so the avatar no longer self-heals on the next login. Low severity, best-effort path, candidate follow-up if self-heal is desired.
- The 304-vs-200 behaviour against real conditional / non-conditional avatar hosts is a browser/real-server behaviour CI cannot exercise, belongs on the manual E2E checklist, not a merge blocker.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -268,6 +268,23 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void AvatarService_HoldsAStaticSharedHttpClient()
+    {
+        // Locked in by the per-login churn trim (#248): the controller builds a fresh AvatarService per
+        // request, so the outbound HTTP stack must be a STATIC shared client — one connection pool for the
+        // whole process — not a per-instance client that would open a new pool (a full TCP+TLS handshake)
+        // on every login. The reference field the constructor reads (_httpClient) points at this shared
+        // client in production; the reference-equality across two production instances is proven behaviorally
+        // in AvatarServiceTests, and this rule locks in that the shared field it points at exists at all.
+        var staticClient = typeof(AvatarService)
+            .GetFields(BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly)
+            .Where(f => !f.Name.Contains('<', StringComparison.Ordinal))
+            .Any(f => typeof(System.Net.Http.HttpClient).IsAssignableFrom(f.FieldType));
+
+        Assert.True(staticClient, "AvatarService must hold a static HttpClient so the outbound stack is reused across the controller's per-request instances rather than rebuilt per login (#248).");
+    }
+
+    [Fact]
     public void SessionMinter_RechecksRevocationImmediatelyBeforeTheMint()
     {
         // Locked in by the in-flight revocation gate (#232): MintAsync must evaluate the caller-supplied
--- a/SSO-Auth.Tests/AvatarServiceTests.cs
+++ b/SSO-Auth.Tests/AvatarServiceTests.cs
@@ -39,7 +39,8 @@ public class AvatarServiceTests
     }
 
     // Builds a service whose HTTP fetch is served by a stub handler returning the given response, so the
-    // fetch path (content-type gate, size cap, happy path) runs without live HTTP (#385 seam).
+    // fetch path (content-type gate, size cap, happy path, conditional 304) runs without live HTTP (#385
+    // seam). The client wraps the stub and stands in for the process-wide shared client (#248).
     private static (AvatarService Service, IProviderManager Providers, IUserManager Users, CapturingLogger Log) Build(HttpResponseMessage response)
     {
         var users = Substitute.For<IUserManager>();
@@ -47,7 +48,7 @@ public class AvatarServiceTests
         var serverConfig = Substitute.For<IServerConfigurationManager>();
         serverConfig.ApplicationPaths.UserConfigurationDirectoryPath.Returns(UserDataRoot);
         var log = new CapturingLogger();
-        var service = new AvatarService(users, providers, serverConfig, log, "test-agent/1.0", () => new StubHandler(response));
+        var service = new AvatarService(users, providers, serverConfig, log, "test-agent/1.0", new HttpClient(new StubHandler(response)));
         return (service, providers, users, log);
     }
 
@@ -61,7 +62,7 @@ public class AvatarServiceTests
         var serverConfig = Substitute.For<IServerConfigurationManager>();
         serverConfig.ApplicationPaths.UserConfigurationDirectoryPath.Returns(UserDataRoot);
         var log = new CapturingLogger();
-        var service = new AvatarService(users, providers, serverConfig, log, "test-agent/1.0", () => new StubHandler(new HttpResponseMessage()), userStoreLocks);
+        var service = new AvatarService(users, providers, serverConfig, log, "test-agent/1.0", new HttpClient(new StubHandler(new HttpResponseMessage())), userStoreLocks);
         return (service, providers, users, log);
     }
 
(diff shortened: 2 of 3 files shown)


iderex (commented):
My sign-off — login-path + network fetch (#318 plan, step 15; the decided If-Modified-Since cadence). Delivered by a delegated agent (resumed after a session-limit crash). Independently gated by the four-lens review, all PASS (correctness was NEEDS_IMPROVEMENT — the 304 test didn't distinguish the correct early-return from a swallowed-exception regression — fixed with a 'no Error logged' assertion, re-reviewed PASS). Confirmed: one process-wide static readonly HttpClient built from the SAME CreateHardenedHandler (SSRF ConnectCallback re-validates per new connection; pooling pins reuse to the already-validated IP, so it cannot skip the SSRF check; 2min PooledConnectionLifetime tightens the rebind window); If-Modified-Since sourced from ProfileImage.LastModified, 304 keeps the existing image and stores nothing (checked before EnsureSuccessStatusCode), 200 refreshes, first login unconditional; the 'double user write' was already resolved by #391 (the second write is the Unregister endpoint, not login — verified empirically, no change). All prior avatar hardening + persisted fields intact; SSOController.cs and SessionMinter.cs untouched. 933 tests; new AvatarService_HoldsAStaticSharedHttpClient conformance rule; CodeQL + required checks green. Ready to merge.

PR #484, opened 2026-07-17
Surface AllowExistingAccountLink in the admin provider form

### LINKED ISSUE #432: Surface AllowExistingAccountLink in the admin provider form (currently config-XML-only)
`AllowExistingAccountLink` (`ProviderConfigBase`) governs whether an SSO login may adopt a pre-existing, unlinked Jellyfin account whose name matches the SSO name, the fail-closed gate behind #354/#358. It is a real, security-relevant toggle, but it has **no field in the admin provider form**: `configPage.html` / `config.js` never reference it, so `listArgumentsByType` never persists it. This was confirmed while delivering #365, the save-contract fitness function's persisting set is exactly the other 31 `OidConfig` properties, and `AllowExistingAccountLink` is not among them. The only way to turn it on today is hand-editing the config XML.

This matters because #354's upgrade runbook explicitly tells admins to enable the flag during the legacy-link migration window, an operation with no UI affordance. An admin following the runbook has to locate and edit `SSO-Auth.xml` by hand, per provider.

Proposal:

- Add an `sso-toggle` checkbox for `AllowExistingAccountLink` to `#sso-new-oidc-provider`, with its `id` spelled exactly `AllowExistingAccountLink` so the #365 save contract persists it. Give it a `fieldDescription` that states the security implication (it lets a first SSO login adopt a same-named existing account) and points to the migration runbook.
- Decide whether the SAML side needs the same affordance (the flag lives on the shared base).

Out of scope: the stale-name superset residual of the flag-on arm is tracked separately in #361.

Refs #354, #358, #365

iderex:
# What & why

Closes #432

`AllowExistingAccountLink` (`ProviderConfigBase`, inherited by `OidConfig`)
gates whether a first SSO login may adopt a same-named, pre-existing, unlinked
Jellyfin account, the fail-closed gate behind #354/#358. It is a real,
security-relevant toggle, but it had no field in the admin provider form:
`configPage.html` / `config.js` never referenced it, so `listArgumentsByType`
never persisted it. The only way to turn it on was hand-editing `SSO-Auth.xml`
per provider, despite the #354 legacy-link migration runbook explicitly telling
admins to enable it during the migration window.

We add an `sso-toggle` checkbox for it to `#sso-new-oidc-provider`, wired through
the existing generic save path.

**How the flag is rendered + saved (mirrors `EnableAllFolders`/`RequirePkce`):**
the checkbox carries `class="sso-toggle"` and `id="AllowExistingAccountLink"`,
spelled exactly like the `OidConfig` property. That is the entire #365 save
contract, no `config.js` change is needed because the save/load path is generic
over the marker classes:

  `#AllowExistingAccountLink.checked = Boolean(provider["AllowExistingAccountLink"])`
  `current_config["AllowExistingAccountLink"] = checkbox.checked` onto the
  existing provider object and PUTs the whole config, it WRITES the value on
  save. The server keeps the JSON member because it is a real `OidConfig`
  property.

default (`false`, fail closed) and its semantics are unchanged. An unchecked box
persists `false`, which equals the property default, so a provider with no
explicit value stays `false` after a save. The `fieldDescription` states the
security implication (a first SSO login can adopt a same-named existing account),
points to the migration runbook, and records that a name-matched administrator
account is never adopted and that each adoption is an audit event.

persists it (`ConfigPreservationTests` already asserts `AllowExistingAccountLink`
survives an XML round-trip) -> `loadProvider` re-reads and re-checks the box.
Other server-managed/XML-only fields are untouched: `ServerManagedFields.Preserve`
only re-injects `CanonicalLinks` and `OidSecret`, and this flag is neither, so the
overlay and server-managed preservation are intact.

The **SAML side needs no change**: there is no SAML provider form in the admin UI
at all (SAML config is entirely XML-only), so there is nothing to add the
affordance to. Building a SAML provider form is separate, out-of-scope work.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. (No change to auth logic; the flag's
      `false` default and semantics are untouched. An unchecked box persists
      `false`, the fail-closed value.)
      secret handling touched.)
      config persistence, or the release pipeline. (Four-lens adversarial review
      below; config-persistence surface.)
      logging.)

## Quality checklist

      (Added `AllowExistingAccountLink` to the security-critical reverse-pin
      roster, as that test's comment instructs.)

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. (Debug, 0 warnings.)
- [x] `dotnet test` is green. (937 passed, incl. the extended
      (`configPage.html` formatted; the local CRLF working tree flags it as a
      false positive, so verified by `prettier --write` leaving the content
      unchanged beyond the addition. CI runs on LF.)

## Notes

Adversarial self-review, four refute-by-default lenses (config-persistence + a
security-relevant flag):

  the admin page renders one more toggle; `listArgumentsByType` simply gains one
  `check_fields` entry, no iteration/limit concern. A provider with no explicit
  value loads unchecked and, if saved, persists `false` - the existing default - 
  so no adoption is silently enabled and nobody is locked out.
  stored flag without an explicit admin click is the pre-existing "save without
  Load Provider first" behavior shared by every toggle, and here that direction
  is `true -> false`, the fail-closed (safer) direction, and only reachable by an
  admin re-saving the same provider name without loading it. Before this change a
  stale XML `true` could never be turned off in the UI; now it can, a net
  improvement. No server-managed field is touched: `ServerManagedFields` preserves
  only `CanonicalLinks` and `OidSecret`.
- **Correctness (read-on-load / write-on-save round-trip), PASS.** `id` matches
  the `OidConfig` property exactly and `class="sso-toggle"` puts it in
  `check_fields`; `loadProvider` sets `checked = Boolean(provider[id])` and
  `saveProvider` writes `current_config[id] = checked`. Default unchecked -> `false`
  = property default. The conformance test now asserts it persists.
- **Integration (the `saveProvider` overlay + `ServerManagedFields` intact) - 
  PASS.** The checkbox is inside `#sso-new-oidc-provider` (between the RequirePkce
  and Scheme Override blocks), so the form-scoped scan picks it up; the overlay
  (seed from existing provider, overlay form fields, PUT whole object) is
  unchanged, and the added field is orthogonal to server-managed re-injection.

Out of scope: the stale-name superset residual of the flag-on arm (#361); the
`RequireVerifiedEmailForAdoption` companion flag (#218), still XML-only; a SAML
provider admin form (does not exist yet).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -448,7 +448,7 @@ public class ArchitectureConformanceTests
         {
             "EnableAuthorization", "OidSecret", "DisableHttps", "DisablePushedAuthorization",
             "DoNotValidateEndpoints", "DoNotValidateIssuerName", "DoNotValidateResponseIssuer",
-            "DoNotLoadProfile", "RequirePkce",
+            "DoNotLoadProfile", "RequirePkce", "AllowExistingAccountLink",
         };
         var unsaved = securityCritical.Where(p => !matchedIds.Contains(p)).ToList();
         Assert.True(
--- a/SSO-Auth/Config/configPage.html
+++ b/SSO-Auth/Config/configPage.html
@@ -740,6 +740,37 @@
                   </div>
                 </div>
 
+                <div
+                  class="checkboxContainer checkboxContainer-withDescription"
+                >
+                  <label>
+                    <input
+                      is="emby-checkbox"
+                      id="AllowExistingAccountLink"
+                      name="AllowExistingAccountLink"
+                      type="checkbox"
+                      class="sso-toggle"
+                    />
+                    <span>Allow Adopting Existing Accounts (Sensitive)</span>
+                  </label>
+                  <div class="fieldDescription checkboxFieldDescription">
+                    ⚠ Sensitive: lets a first SSO login adopt (take over) a
+                    pre-existing, unlinked Jellyfin account whose username
+                    matches the SSO name, instead of rejecting it. Off by
+                    default (fail closed). Enable only for the one-time
+                    legacy-link migration window described in the
+                    account-linking runbook on the
+                    <a
+                      is="emby-linkbutton"
+                      href="https://github.com/iderex/jellyfin-plugin-sso/wiki"
+                      class="button-link"
+                      >wiki</a
+                    >, then turn it back off. A name-matched administrator
+                    account is never adopted regardless of this setting, and
+                    each adoption is recorded as an audit event.
+                  </div>
+                </div>
+
                 <div class="inputContainer">
                   <label
                     class="inputLabel inputLabelUnfocused"


iderex (commented):
My sign-off — config-persistence + surfaces a security-relevant flag in the admin UI (no semantics change). Delivered by a delegated agent; four-lens review all PASS. Confirmed: one .sso-toggle checkbox with id exactly AllowExistingAccountLink rides the generic #365 save contract (loadProvider reads Boolean(provider[id]); saveProvider writes current_config[id]=checked) so no config.js change was needed; unchecked persists false = the property default, so a provider with no explicit value stays false after a save (fail-closed posture + adoption semantics unchanged); the form-contract is now pinned in the ArchitectureConformanceTests security-critical reverse-pin roster and the XML round-trip by ConfigPreservationTests; ServerManagedFields (CanonicalLinks/OidSecret) untouched. Net safety improvement: a stale XML true can now be disabled via the UI, which was impossible before. SAML has no admin form (XML-only) — correctly out of scope. 937 tests; CodeQL + required checks green. Ready to merge.

Issue #473, opened 2026-07-17
Introduce VerifiedIdentity: only protocol validators can construct a verified login identity


iderex:
Part of the #318 unified-architecture plan ([design note](https://github.com/iderex/jellyfin-plugin-sso/issues/318#issuecomment-5005931010), step 10 / new sub-issue 2). The fail-closed keystone of the target.

## Proposal

Generalize the existing OIDC `RedeemedState` into a protocol-agnostic `VerifiedIdentity` record that **only the two protocol validators can construct** (private ctor; factory on `OidcStateStore.TryRedeem` and on the SAML validator, each yielding it only after ALL validation has passed). The shared login-completion path (the future `LoginCompletionService`, and today `ResolveOrCreateAsync` → `SessionParameters` → `SessionMinter`) accepts **only** a `VerifiedIdentity`.

## Why (security first)

Reaching account-resolution or session-minting with an unvalidated protocol response becomes a **compile error, not a review catch**, structurally stronger than any test against the throwing-accessor→guarded-Try fail-open hazard. It also collapses the two protocols' divergent `SessionParameters` construction into one path (minimal-code win).

## Acceptance
- [ ] `VerifiedIdentity` has a private ctor; `new VerifiedIdentity(...)` appears only in `OidcStateStore` and the SAML validator.
- [ ] The completion path signature takes `VerifiedIdentity` (no raw claims/response reach it).
- [ ] Conformance rule `VerifiedIdentity_IsConstructedOnlyByProtocolValidators` locks it in.
- [ ] Behaviour-preserving: existing login/mint tests stay green; a fail-closed test proves an unvalidated path cannot mint.

Security-sensitive (login path) → adversarial multi-lens gate applies. Sequence after #341.

Issue #326, opened 2026-07-14
Bind the OpenID authorize state to the initiating user-agent (login CSRF / session fixation)


iderex:
The OpenID `state` is stored process-globally keyed by the state token and is never bound to the browser that started the flow. `OidChallenge` sets no cookie (`SSO-Auth/Api/SSOController.cs:329` redirects with no `Set-Cookie`), and both `OidPost` (`SSOController.cs:180-187`) and the anonymous `OidAuth` (`SSOController.cs:622-650`) locate the pending login solely by the state value. RFC 6749 §10.12 and the OAuth 2.0 Security BCP assign `state` the job of tying the callback to the user-agent that initiated authorization; here it only serves as an unguessable server-side lookup key. This is the cross-user session-fixation gap #176 asked to characterize, confirmed present.

## Exploit (forced login as the attacker)

1. Attacker starts a login (`GET /sso/OID/p/{provider}`), creating an authorize state in the global store, and obtains a fresh unused authorization `code` from the IdP for their own identity (intercepting the IdP redirect before it reaches the callback).
2. Attacker lures the victim, a plain top-level GET, no anti-CSRF on the endpoint, to `/sso/OID/redirect/{provider}?code=<attacker_code>&state=<attacker_state>&iss=<provider>`.
3. `OidPost` finds the attacker's state, redeems the attacker's code (PKCE passes: the verifier is the attacker's, held server-side in the state; PKCE binds the code to the authorize request, not to a browser), validates the attacker's id_token, marks the state `Valid` with the attacker's subject, and returns the auto-submitting auth page to the **victim's** browser.
4. That page runs on the Jellyfin origin, POSTs the state to the anonymous `OID/Auth` endpoint, receives the **attacker's** `AuthenticationResult`, and writes its `AccessToken`/`UserId` into `localStorage['jellyfin_credentials']` (`SSO-Auth/WebResponse.cs:529-537`).

Result: the victim's browser is silently signed in as the attacker. Anything the victim then does, uploads, profile data, watch history, lands in the attacker-controlled account. PKCE, the RFC 9207 `iss` check (#125), and provider-scoping do not stop this; none binds the completing browser to the initiating one. Bounded by the authorization code's lifetime and the need to land the lure. Not a takeover of the victim's existing account (the `Link` path is `[Authorize]`-gated and Jellyfin uses header bearer tokens, not ambient cookies), so impact is forced-login, not existing-account compromise.

## Fix (fail closed)

[...]

PR #485, opened 2026-07-17
Swap the OpenID authorize state atomically instead of promoting it in place

### LINKED ISSUE #341: Replace the in-place authorize-state promotion with an atomic Pending/Ready swap
Step 4b of the #318 state-store consolidation. Today the OpenID callback promotes a stored pending state in place: `PendingState.Complete` copies the eight derived fields onto the shared `TimedAuthorizeState` instance the store already holds. That relocated the pre-consolidation behavior verbatim, but it keeps a theoretical window: a redeem racing the promotion can observe a half-applied field set (it must still win the atomic claim and pass the `Valid` check, so exploitation requires the promotion to be mid-flight at exactly that moment, narrow, but structurally possible).

The target design replaces `TimedAuthorizeState` with an `AuthorizeSession` sum: `Pending(oidcClientState, provider, createdAt)` stored at challenge, swapped atomically for `Ready(derived, provider, createdAt)` at the callback. The swap closes the half-applied-promotion window, deletes `TimedAuthorizeState` and the eight-field copy, and makes "role gate passed" which variant you hold instead of a mutable `Valid` flag.

This is a concurrency-semantics change on the login path, so it needs its own adversarial review and characterization tests (racing promotion vs redeem), separate from the behavior-preserving consolidation that landed before it.

Refs #318

iderex:
# What & why

Closes #341. Refs #318 (step 4b of the state-store consolidation).

Today the OpenID callback promotes a stored pending state **in place**:
`PendingState.Complete` copied the eight derived fields (username, subject,
validity, admin, Live TV, folders, avatar) onto the shared `TimedAuthorizeState`
instance the store already held. A redeem racing that copy could observe a
half-applied field set, it still had to win the atomic claim and pass the
validity check, so the window was narrow but structurally real (e.g. `Valid`
written before `Folders`, exposing a redeemable state with a null folder list).

We replace `TimedAuthorizeState` with a **closed, immutable `AuthorizeSession`
sum**, swapped atomically instead of mutated:

- **`Pending`**, registered at the challenge. Built complete: the reused
  discovery metadata (#247) and the RFC 9207 response-`iss` requirement (#210)
  are folded into construction, so registration is a single atomic
  `ConcurrentDictionary.TryAdd` with no post-hoc mutation of a stored entry.
- **`Ready`**, swapped in at the callback via a single `TryUpdate`
  (compare-and-set on the exact peeked `Pending`) once the role gate passes.

"Role gate passed" is now **which variant the store holds**, not a mutable
`Valid` flag, and the swap replaces one immutable value with another, so a
redeemer racing the promotion observes either the whole `Pending` (structurally
not redeemable) or the whole `Ready`, never a torn read. `TimedAuthorizeState`,
`PendingState`, and `RedeemedState` are deleted (the two variants subsume them),
along with the `SetProviderInformation` / `MarkResponseIssuerRequired`
enrichment methods (folded into `Pending` construction).

The invariants are preserved: the exactly-once claim (atomic
`TryRemove(KeyValuePair)`), the disabled-provider guard before consumption, the
browser-binding gate before the atomic remove (#326), the cross-provider replay
guard, and the fail-closed miss/expiry paths. One property is strengthened: an
invalid login now leaves an **unpromoted `Pending` that is never redeemable**,
rather than a completed entry carrying `Valid=false`.

## Type of change

- [ ] Bug fix
- [ ] Feature

(Refactor on the login/state path that closes a torn-read window, hardening in
effect; labelled per the issue as `hardening` under P3.)

## Security checklist

- [x] Fail-closed preserved: a miss, expired, provider-mismatched, binding-
      mismatched, or not-yet-promoted state returns a clean reject and consumes
      nothing; no fall-through to a mint.
- [x] No secrets logged; `Summaries` still projects only Provider/Created/
      Valid/IsLinking (the token and PKCE verifier/nonce never serialize out).
- [x] A security review was performed (four adversarial lenses below).
- [x] Log inputs from the IdP are sanitized inline at the log call (unchanged;
      the callback's `ReplaceLineEndings` sites are untouched).

## Quality checklist

- [x] No duplicated logic; the pending->ready transition is a single atomic swap
      in one place (`OidcStateStore.Promote`), replacing a nine-line field copy.
- [x] The change adds no more code than the problem requires - net simplification
      (three types + two enrichment methods removed, one sum type added).
- [x] The code is self-documenting (comments explain the atomicity/why); converges
      on the #318 target (closed sum, thin controller, store owns the idiom).
- [x] Architecture-conformance tests pass; the new structural property is locked
      in as `ArchitectureConformanceTests.AuthorizeStates_AreImmutableVariants`
      (closed sum, sealed variants, no settable property / writable field).

## Verification

- [x] `dotnet build --warnaserror` green, Debug and Release, 0 warnings.
- [x] `dotnet test` green, 940 passed, Debug and Release.
- [x] `sh scripts/check-jellyfin-compat.sh` green.
- [x] Prettier: N/A (no `.js`/`.html`/`.md`/`.css` changes).

## Notes

**Adversarial self-review (login/state path, four refute-by-default lenses):**

- **Availability / mass-lockout, VERDICT: PASS.** Behaviour-preserving on the
  happy path and the denial paths. A `Promote` that loses the CAS (concurrent
  callback, or a state pruned/redeemed since the peek) still returns the auth
  page; the redeem is the real gate, so no legitimate single-user login breaks.
  A retried callback GET after promotion now gets a clean "Invalid or expired
  state" 400 instead of a re-run token exchange that the IdP rejects for code
  reuse, same outcome, no new lockout. Cap/per-client/prune logic unchanged.

- **Security-bypass (exactly-once? double-mint? fail-open?), VERDICT: PASS.**
  Exactly-once redemption is intact (atomic `TryRemove`; `ConcurrentRedemption_
  OfSameState_ClaimsExactlyOnce`). Promotion is single-winner (`TryUpdate` CAS;
  `Promote_SecondPromote_IsRejected_SingleWinner`). The torn-read window is the
  thing being closed, `PromoteRacingRedeem_...NeverAPartialOne` asserts any
  redeemed `Ready` is fully populated. Disabled-provider, binding, and cross-
  provider guards all still precede consumption. An invalid login creates no
  `Ready` at all (stronger than before). No throwing path was turned into a
  fail-open Try.

- **Correctness (atomicity / torn-read / ABA), VERDICT: PASS.** `TryUpdate`
  compares the stored value by reference against the exact peeked `Pending`
  (`PeekCurrent` returns the stored reference, not a copy). No ABA: tokens are
  single-use CSPRNG values (duplicate `TryAdd` fails) and the state is monotonic
  (Pending -> Ready -> removed), so a comparand `Pending` can exist at most once.
  The `(Ready)` cast in `TryRedeem` runs only after `IsRedeemableBy` proved the
  variant and the atomic remove won.

- **Integration (Jellyfin contracts / config / call sites), VERDICT: PASS.**
  All three call sites (OidPost callback, OidAuth mint, OidLink) updated and
  green; the end-to-end `OidChallengeToCallback_*` tests confirm discovery reuse
  (#247) and the response-`iss` requirement (#210) still flow after folding them
  into `Pending`. The admin-debug `Summary` shape is unchanged. No config-
  persistence, reverse-proxy, or public API surface change. `SeedOidStateForTests`
  (test-only, `InternalsVisibleTo`) took the new type; no other callers of the
  removed methods exist.

**Out of scope:** the controller touch is confined to the challenge (build the
`Pending`, one `TryAdd`) and the callback (replace the in-place `Complete` with
`Promote`, reordered so promotion runs only for a valid, subject-bearing login);
no change to `SessionMinter`/`AvatarService`, `providers.md`, or unrelated
controller logic. This unblocks #473 (VerifiedIdentity).


iderex (commented):
My sign-off — security/login-path, OIDC authorize-state atomicity (#318 plan, step 1; unblocks #473 VerifiedIdentity). Delivered by a delegated agent; four-lens adversarial review all PASS. Confirmed: the in-place TimedAuthorizeState promotion (which wrote Valid as the 4th of 9 field copies — a real, if narrow, torn-read window) is replaced by a closed immutable AuthorizeSession sum (Pending → Ready) swapped atomically via a single ConcurrentDictionary.TryUpdate CAS on the exact peeked Pending. Exactly-once holds (single-winner CAS + single-winner TryRemove; a redeemer's TryGetValue returns either the whole Pending — not redeemable, no consume — or the whole Ready, fully built before the swap; no ABA since tokens are single-use CSPRNG and state is monotonic). All fail-closed guards still precede consumption (disabled-provider, binding #326, cross-provider replay, clean reject on miss/expiry) and the posture is STRENGTHENED — an invalid login now leaves an unpromoted Pending instead of a completed Valid=false entry. #247 discovery-reuse and #210 response-iss enrichments folded into Pending construction, deleting the sub-ms enrichment gap. Net simplification: TimedAuthorizeState/PendingState/RedeemedState + two store methods deleted, one sum type added. 940 tests (Debug+Release --warnaserror); new AuthorizeStates_AreImmutableVariants conformance rule; CodeQL + required checks green; mergeable clean. Ready to merge.

Issue #491, opened 2026-07-17
SAML multi-certificate rotation (signing rollover + multiple inbound verification certs)


iderex:
Split off from #167 (which delivered single-certificate outgoing AuthnRequest signing) to keep that PR focused. Two independent rotation capabilities remain, both for zero-downtime key rotation:

**1. Outbound signing-key rollover (SP side).** Today `SamlConfig.SamlSigningKeyPfx` holds one service-provider signing key. Support a primary + rollover key so the SP can rotate its own signing certificate without a hard cutover: sign with the primary, and publish/allow both public certificates at the identity provider during an overlap window.

**2. Inbound verification-cert rotation (IdP side).** This is the remaining acceptance criterion of #167: accept a response whose signature verifies against EITHER the primary `SamlCertificate` or an optional secondary certificate, so the identity provider can roll its signing key over an overlap window instead of breaking every login at the cutover; reject expired certs. Mirrors the V2 `SamlSecondaryCertificate` field. This is the higher-value half (IdP key rotation is the common operational event).

Both should reuse the existing algorithm allowlist (no SHA-1) and the fail-closed posture. Update `providers.md` with the documented rotation procedure (the note in #167's docs points here).

Refs #167

iderex:
Both parts delivered: part 1 (outbound SP signing-key rollover) in #573, part 2 (inbound IdP verification-cert rotation) in #582, both merged into the 4.2 feature-dev branch. Closing manually (auto-close does not fire off a non-default branch).

PR #492, opened 2026-07-17
Sign outgoing SAML AuthnRequests (HTTP-Redirect binding)

### LINKED ISSUE #167: Outgoing SAML request signing and multi-certificate rotation
Sign outgoing AuthnRequests (for IdPs that require signed requests) and support multiple SP/IdP certificates to allow zero-downtime certificate rotation. Feature parity with V2.

Acceptance criteria:
- [ ] Optional signing of outgoing AuthnRequests with the SP key, using the existing algorithm allowlist (no SHA-1/MD5).
- [ ] Multiple valid IdP verification certificates accepted during a rotation window; expired ones rejected.
- [ ] Documented rotation procedure in providers.md.

iderex:
# What & why

Some identity providers require the service provider to sign its outgoing SAML `AuthnRequest` ("signed authentication requests" / "client signature required"). We add opt-in signing of the outgoing request so those providers can be used, without affecting any existing deployment.

**Design, where the signature goes.** The plugin sends its `AuthnRequest` over the **HTTP-Redirect binding** (`SamlChallenge` DEFLATE-compresses the request into the `SAMLRequest` query parameter and issues a redirect). For that binding, SAML 2.0 mandates a **detached query-string signature** (Bindings §3.4.4.1), not an enveloped `<ds:Signature>`, identity providers ignore an embedded XML signature on a redirect-binding message. So the new `SamlRedirectSigner` computes the signature over the URL-encoded `SAMLRequest`[`&RelayState`]`&SigAlg` string, in that mandated order, and appends `&Signature=`. The same `Uri.EscapeDataString` encoding is used both to compute the signed octets and to emit the URL, so the IdP verifies over exactly what is transmitted.

**Algorithm, allowlist reuse, no SHA-1.** Signing is **RSA-SHA256**. The signer pins its `SigAlg` and guards it through `SamlSignatureAlgorithms.IsSignatureMethodAllowed`, the same allowlist the inbound response validator enforces, so the outgoing path can never sign weaker than the SP demands of the IdP. A test asserts the emitted `SigAlg` is on the allowlist and is not SHA-1 (the structural "reuse the allowlist" property is locked there rather than in `ArchitectureConformanceTests`, whose rules are reflection/source-scan structural checks that a value assertion would not fit).

**Config, default off, secret handling.** Two per-provider fields on `SamlConfig`:
- `SignAuthnRequests` (bool, **default false**), off means the request is sent exactly as before (unsigned), byte-for-byte; existing deployments are unaffected.
- `SamlSigningKeyPfx`, the SP signing key as a Base64, unencrypted PKCS#12 (PFX) blob (cert + RSA private key); the operator supplies the keypair whose public certificate the IdP is configured to trust. It is a **write-only secret**: withheld from every JSON response via `WriteOnlySecretConverter` (never reaches the admin browser / a config export), preserved on a blank save via `ServerManagedFields`, validated on `SAML/Add` and on the config-page save, and persisted only to the on-disk XML. Same pattern as the OpenID `OidSecret`. It is never logged.

**Fail-closed.** With signing enabled, a missing / unloadable / private-key-less key, or an empty endpoint, or a platform key-store refusal, refuses the challenge with a clean handled 500. It never silently falls back to sending an unsigned request, so an operator who turned signing on cannot get a silent downgrade. A garbage key is also rejected at save time.

**Scope split.** This PR is single-certificate signing only. Multi-certificate rotation, SP signing-key rollover and accepting multiple inbound IdP verification certificates during an overlap window (the second half of #167's acceptance criteria), is split to #491 to keep this focused. `providers.md` documents the new options and points the rotation procedure at the follow-up.

ECDSA SP keys are not supported (the signer is RSA-SHA256); a non-RSA PFX is rejected as invalid. This is a documented limitation, not a silent failure.

Closes #167

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] Fail-closed preserved: signing enabled + missing/invalid key (or empty endpoint / key-store refusal) refuses the challenge with a 500; it never emits an unsigned request. Default off is unchanged behavior.
- [x] No secrets logged; the signing key is redacted on config export (write-only across JSON) and never written to a log, the 500 log message carries only the provider name (inline line-ending-stripped) and a key-free reason.
- [x] A security review was performed: the four-lens adversarial gate (availability / bypass / correctness / integration) ran on the diff, all PASS; see Notes.
- [x] Log inputs from the IdP are sanitized inline at the log call (`provider?.ReplaceLineEndings(string.Empty)`).

## Quality checklist

- [x] No duplicated logic; the signer and key loader are small single-purpose helpers; the allowlist check is shared, not re-implemented.
- [x] The change adds no more code than the problem requires (redirect-binding signing reuses the existing transport; no new POST binding or metadata surface).
- [x] The code is self-documenting; comments explain the why (binding choice, fail-closed, secret handling).
- [x] Architecture-conformance tests pass; no new reflection/source-scan structural property is introduced (the allowlist-reuse property is locked by `SamlRedirectSignerTests`).

## Verification

- [x] `dotnet build --warnaserror` green, Debug and Release (incl. the solution).
- [x] `dotnet test` green, 988 tests, 0 failed.
- [x] Prettier clean for the `providers.md` change.
- [x] `scripts/check-jellyfin-compat.sh` green.

## Notes

**Adversarial review gate (four refute-by-default lenses, on the committed diff):**

- **Availability, PASS.** Default-off path is byte-identical (`GetRedirectUrl` untouched); only opt-in providers are affected; the write-only key cannot be wiped by a blank save (preserve-on-blank); fail-closed 500 is per-provider and bounded (request-cache is capped/#327 and self-expiring). Flagged the narrowed catch → hardened (below).
- **Security-bypass, PASS.** No SHA-1 downgrade (pinned + allowlist-guarded); no silent-unsigned path when signing is on; the private key is write-only, never logged, never transmitted (used only to sign locally, so preserve-on-blank cannot exfiltrate it, the OIDC identity-guard is not needed); the signature covers the correct octets in the mandated order.
- **Correctness, PASS.** Signed octets are the *same string* that is emitted (structurally cannot diverge); RelayState included iff non-empty and correctly ordered; separator handling correct; the round-trip tests verify via `RSA.VerifyData` with a foreign-key negative (non-vacuous); cert/key disposed strictly after signing.
- **Integration, PASS.** BCL crypto APIs exist on net9.0 and the 10.11 ABI floor (ran the abi-floor build); write-only secret round-trips XML + System.Text.Json + camelCase; `EphemeralKeySet` yields a SignData-usable key on Windows (proven by the controller test); no existing serialization/arch behavior broken; the single-file `#nullable enable` compiles clean under warnaserror + analyzers.

**Review-driven hardening (2nd commit).** Two lenses noted the challenge's `catch` was narrowed to `InvalidOperationException`, so an empty-endpoint (`ArgumentException`) or a `CryptographicException` from `SignData` would escape as a raw host 500 (still fail-closed, just not clean). Broadened the catch to cover all three and added an empty-endpoint fail-closed test.

**E2E items (outside CI's reach, for the checklist):** the macOS `DefaultKeySet` fallback branch is Mac-only; real-IdP acceptance of the detached redirect-binding signature (ADFS / Keycloak / Entra query canonicalization) is live-interop.

**Out of scope / follow-up:** multi-certificate rotation → #491.

### DIFF
--- a/SSO-Auth.Tests/ConfigPreservationTests.cs
+++ b/SSO-Auth.Tests/ConfigPreservationTests.cs
@@ -235,6 +235,93 @@ public class ConfigPreservationTests
         Assert.True(string.IsNullOrEmpty(ServerManagedFields.ResolveUpdatedSecret(incoming, live)));
     }
 
+    // --- SAML signing key: write-only secret + preserve-on-blank (#167) ---
+
+    [Fact]
+    public void SamlSigningKey_SerializedValueIsHidden_ButStaysDeserializableAndInXml()
+    {
+        var config = new SamlConfig { SamlClientId = "sp", SamlSigningKeyPfx = "pfx-secret-blob" };
+
+        // JSON responses must not leak the private-key blob; the property is emitted as null (write-only).
+        var json = System.Text.Json.JsonSerializer.Serialize(config);
+        Assert.DoesNotContain("pfx-secret-blob", json, StringComparison.Ordinal);
+        Assert.Contains("\"SamlSigningKeyPfx\":null", json, StringComparison.Ordinal);
+
+        // Core serializes with a camelCase policy; the value must stay hidden there too.
+        var camel = System.Text.Json.JsonSerializer.Serialize(
+            config,
+            new System.Text.Json.JsonSerializerOptions { PropertyNamingPolicy = System.Text.Json.JsonNamingPolicy.CamelCase });
+        Assert.DoesNotContain("pfx-secret-blob", camel, StringComparison.Ordinal);
+
+        // XML (on-disk config) must still persist it, so signing survives a restart.
+        var serializer = new XmlSerializer(typeof(SamlConfig));
+        using var writer = new System.IO.StringWriter();
+        serializer.Serialize(writer, config);
+        var xml = writer.ToString();
+        Assert.Contains("SamlSigningKeyPfx", xml, StringComparison.Ordinal);
+        Assert.Contains("pfx-secret-blob", xml, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public void SamlSigningKey_IsDeserializedFromJson_SoItCanBeSetAndRotated()
+    {
+        const string body = "{\"SamlClientId\":\"sp\",\"SamlSigningKeyPfx\":\"typed-pfx\"}";
+        var parsed = System.Text.Json.JsonSerializer.Deserialize<SamlConfig>(body);
+        Assert.Equal("typed-pfx", parsed!.SamlSigningKeyPfx);
+    }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    [InlineData("   ")]
+    public void Preserve_BlankIncomingSigningKey_KeepsLiveKey(string? incomingKey)
+    {
+        // A config-page save arrives with the key blank (withheld from JSON) and must keep the stored one.
+        var live = new PluginConfiguration();
+        live.SamlConfigs["adfs"] = new SamlConfig { SamlSigningKeyPfx = "live-pfx" };
+        var incoming = new PluginConfiguration();
+        incoming.SamlConfigs["adfs"] = new SamlConfig { SamlSigningKeyPfx = incomingKey };
+
+        ServerManagedFields.Preserve(incoming, live);
+
+        Assert.Equal("live-pfx", incoming.SamlConfigs["adfs"].SamlSigningKeyPfx);
+    }
+
+    [Fact]
+    public void Preserve_NonBlankIncomingSigningKey_IsKeptAsRotation()
+    {
+        var live = new PluginConfiguration();
+        live.SamlConfigs["adfs"] = new SamlConfig { SamlSigningKeyPfx = "old-pfx" };
+        var incoming = new PluginConfiguration();
+        incoming.SamlConfigs["adfs"] = new SamlConfig { SamlSigningKeyPfx = "rotated-pfx" };
+
+        ServerManagedFields.Preserve(incoming, live);
+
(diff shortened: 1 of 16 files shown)


iderex (commented):
My sign-off — SECURITY feature, SAML crypto (outgoing AuthnRequest signing). Delivered by a delegated agent; four-lens adversarial review all PASS, with a review-driven hardening commit (broadened the challenge catch to ArgumentException/CryptographicException so an empty endpoint or crypto error fails closed as a clean 500, never a raw 500 or silent-unsigned). Load-bearing design call, empirically correct: the plugin sends its AuthnRequest over the HTTP-Redirect binding, which per SAML 2.0 Bindings §3.4.4.1 mandates a DETACHED query-string signature, not an enveloped <ds:Signature> (that is POST-binding; IdPs ignore an embedded signature on a redirect) — so the agent correctly implemented the detached signature (SamlRedirectSigner, RSA-SHA256, matching the V2 reference) instead of the enveloped form my brief wrongly described. Confirmed: SigAlg pinned rsa-sha256 and guarded by the same inbound allowlist (no SHA-1); the signed octets are the exact same string emitted (Uri.EscapeDataString for both — cannot diverge); opt-in SamlConfig.SignAuthnRequests defaults off = byte-identical current behavior; SamlSigningKeyPfx is a write-only secret (WriteOnlySecretConverter, ServerManagedFields preserve-on-blank, validated on Add + save, never logged — the OidSecret pattern); signing-on + missing/unloadable/private-key-less key refuses the challenge fail-closed. Multi-cert rotation split to #491. 988 tests (Debug+Release --warnaserror); CodeQL C# + required checks green; mergeable clean; providers.md documents the feature. Ready to merge.

PR #495, opened 2026-07-17
Introduce VerifiedIdentity as the session-mint keystone

### LINKED ISSUE #473: Introduce VerifiedIdentity: only protocol validators can construct a verified login identity

iderex:
# What & why

Both the OpenID and SAML login paths resolve the same identity and privilege values, then each built its own `SessionParameters` inline before minting the session. Nothing at the type level stopped a future caller from reaching the mint with values that had not been through full protocol validation.

This introduces **`VerifiedIdentity`**, a sealed record with a **private constructor** and two factories, each of which represents a completed protocol validation:

- **`FromOidcRedemption`**, built inside `AuthorizeSession.Ready`, which the store produces only for a promoted (role-gate-passed) state and hands out only through the one-time atomic redeem; and
- **`FromValidatedSaml`**, called only at the SAML session-minting endpoint after the response has passed signature, time-bound, audience, recipient, `InResponseTo`/binding, the login allow-list, replay one-time-use, and the non-empty-NameID guard.

The shared completion path, `CompleteLoginAsync` (`ResolveOrCreateAsync → SessionParameters → SessionMinter`), now accepts a `VerifiedIdentity` **and nothing else**, so minting from a raw, unvalidated response is a **compile error**: there is no overload that takes anything else, and the private constructor makes the type unforgeable from outside. The two protocols' divergent `SessionParameters` construction collapses into this one path; the only per-protocol input left is the adoption gate (OpenID may require a verified email #218; SAML passes `AdoptionGate.None`), passed as a parameter.

Behaviour is preserved field-for-field: HTTP response shapes, audit labels (`OpenID`/`SAML`), config reads, the ten `SessionParameters` members, and the fail-closed subject/username guards are all unchanged.

Closes #473

## Type of change

- [ ] Bug fix
- [ ] Feature

(Hardening-flavoured refactor: it tightens a login-path type invariant, but ships no behaviour change.)

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green.

### Design

- **Unforgeable keystone.** `VerifiedIdentity` has a private constructor; the only two construction sites are the two factories, each named for the validation it presupposes. Nothing outside `VerifiedIdentity.cs` can `new` one, so holding a `VerifiedIdentity` is proof the login was verified.
- **Type-level mint lock.** `CompleteLoginAsync` and `SessionMinter.MintAsync` are reachable only with a `VerifiedIdentity`; there is no path that mints from a raw response.
- **Protocol asymmetry, deliberately documented.** For OpenID the guarantee is structural (the store only produces `Ready` for a promoted state and only hands it out through the atomic redeem). For SAML the factory call sits after the full validation chain in the endpoint, pinned by the existing controller characterization tests plus the new conformance call-site scan. A dedicated SAML validator type is a later #318 step; the factory's allow-list moves with it.
- **No premature `SsoProtocol` abstraction.** The identity carries just the two protocol-facing labels the completion path branches on (`LinkMode` "oid"/"saml", `AuditProtocol` "OpenID"/"SAML"); the richer protocol type is deferred to its own #318 step.
- **Minimal churn.** `AuthorizeSession.Ready`'s constructor signature is unchanged (it now builds the identity internally), so existing `Ready(...)` construction in tests is untouched.

### Conformance rule (locked in)

`VerifiedIdentity_IsConstructedOnlyByProtocolValidators`: (1) reflection, no externally-reachable constructor (public/internal/protected-internal); (2) the two named factories exist; (3) source scan, each factory is invoked only from its protocol's validator (`FromOidcRedemption` → `AuthorizeSession.cs`; `FromValidatedSaml` → the controller's SAML endpoint). The private-ctor reflection assertion is the airtight lock; the call-site scan is the sharper "constructed only by the RIGHT validator" signal on top.

### Adversarial-review gate, four lenses, all PASS

- **Availability / mass-lockout, PASS.** Field-for-field equivalence against `origin/main`: link keys, adoption gates (SAML still `AdoptionGate.None`; OpenID still keyed on `EmailVerified`), session parameters, audit labels and the fail-closed subject/username guards all preserved. The null-forgiving `derived.Subject!`/`derived.Username!` is safe, `OidCallback` guarantees non-null subject and (via `Valid`) non-null username before promotion, so no new NRE/crash path.
- **Security-bypass / fail-open, PASS.** `CompleteLoginAsync` has exactly two callers (both post-validation); the SAML factory sits after every reject gate with no reordering; the link paths gain no mint capability; empirically refuted the `with`-expression/copy-ctor escape (`with { Admin = true }` fails to compile, all properties get-only; the sealed record's copy ctor is private).
- **Correctness, PASS.** Factory field mapping verified against the removed inline code (no swapped Subject/Username, correct label strings, `Folders` aliased not copied). Conformance helper verified: the qualified token does not match the method definition, path normalization is consistent, the reflection ctor check ignores the private copy ctor and is non-vacuous. One non-blocking residual (the substring scan would miss a `using static`/line-split spelling), dispositioned: the airtight lock is the private ctor; the scan is documented as belt-and-braces.
- **Integration, PASS.** HTTP response shapes and status codes unchanged for success and every rejection; the four `ProviderConfigBase` reads are genuinely base-class (no per-subtype shadow); the admin debug `Summaries()` reads only base members; no existing conformance rule trips on the new type; `InternalsVisibleTo` reaches the internal type/factories; all ten `required` `SessionParameters` members still set. Empirically green: build `--warnaserror` clean, 992/992 tests pass.

Two lens observations were addressed by tightening comments (the copy-ctor accessibility note; the substring-scan defense-in-depth note). CI still green after.

### Gates

- `dotnet build -c Debug --warnaserror` and `-c Release --warnaserror`: 0 warnings, 0 errors.
- `dotnet test`: 992 passed / 0 failed (989 prior + 3 new).
- No `.js`/`.html`/`.md`/`.css` touched, so Prettier is a no-op.

### Out of scope

- Extracting a dedicated SAML validator type (a later #318 step) that would become the natural home for the `FromValidatedSaml` call and the conformance allow-list.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -224,6 +224,87 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void VerifiedIdentity_IsConstructedOnlyByProtocolValidators()
+    {
+        // Locked in by #473: VerifiedIdentity is the keystone the session-minting path is keyed on, and it
+        // is unforgeable — its constructor is PRIVATE, so the only way to obtain one is a named factory that
+        // stands for "this protocol's validation has completed". Two properties are pinned:
+        //
+        // 1. Reflection: NO declared instance constructor is reachable from outside the type — none is
+        //    public, internal, or protected-internal. A sealed record's own compiler-generated copy
+        //    constructor is emitted PRIVATE (protected only for unsealed records), so it too is excluded by
+        //    this filter; the accessibility test is written to also exclude a plain `protected` ctor, which
+        //    is unreachable on a sealed type anyway (no derived type could invoke it). The C# compiler
+        //    guarantees such a constructor cannot be invoked outside the declaring type, so this alone
+        //    proves `new VerifiedIdentity(...)` can appear only inside VerifiedIdentity.cs (the two
+        //    factories) — no third construction path can compile. (An empty `with { }` on an existing
+        //    instance clones a valid identity verbatim; every property is get-only, so it can neither
+        //    mutate nor forge one.) A future `public`/`internal` ctor added to the type would reopen that
+        //    hole and fail HERE.
+        // 2. Source scan: each factory is INVOKED only from its protocol's validator. FromOidcRedemption
+        //    belongs to the OpenID redeem path — built inside AuthorizeSession.Ready, which the store hands
+        //    out only through the one-time atomic redeem — and FromValidatedSaml only at the SAML
+        //    session-minting endpoint after full response validation. A call from anywhere else (a link
+        //    endpoint, a new controller action) would mean an identity minted from something other than a
+        //    completed validation, so it fails the scan.
+        var ctors = typeof(VerifiedIdentity)
+            .GetConstructors(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly);
+        Assert.True(ctors.Length > 0, "VerifiedIdentity must declare an instance constructor to pin (it was removed or the type was renamed).");
+        var reachable = ctors
+            .Where(c => c.IsPublic || c.IsAssembly || c.IsFamilyOrAssembly)
+            .Select(c => c.ToString())
+            .ToList();
+        Assert.True(
+            reachable.Count == 0,
+            "VerifiedIdentity's constructor(s) must not be reachable from outside the type (no public/internal/protected-internal ctor) so it is constructible only through its validation factories (#473): " + string.Join(", ", reachable));
+
+        // Sentinel + call-site pins. The factory NAMES are the contract; require both to exist (a rename
+        // must consciously update this rule), then confine each factory's invocation to the file(s) that own
+        // its protocol's validation. AuthorizeSession is where the OpenID identity is built (from the
+        // role-gate result); the SAML factory is invoked from the controller's SAML endpoint until a later
+        // #318 step extracts a dedicated SAML validator — at which point this allow-list moves with it.
+        const string oidcFactory = "FromOidcRedemption";
+        const string samlFactory = "FromValidatedSaml";
+        var factoryMethods = typeof(VerifiedIdentity)
+            .GetMethods(BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly)
+            .Where(m => m.ReturnType == typeof(VerifiedIdentity))
+            .Select(m => m.Name)
+            .ToHashSet(StringComparer.Ordinal);
+        Assert.True(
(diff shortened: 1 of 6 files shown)


iderex (commented):
My sign-off — SECURITY keystone (#318 step 10), both protocols' hot path. Delivered by a delegated agent (CPU-capped, resumed after two overheat crashes); four-lens adversarial review all PASS. Confirmed the load-bearing property: VerifiedIdentity is a sealed record with a PRIVATE constructor and two factories only — FromOidcRedemption (built inside AuthorizeSession.Ready, which the store yields only for a promoted, role-gate-passed state via the one-time atomic redeem) and FromValidatedSaml (called only at the SAML mint endpoint after signature/time/audience/recipient/InResponseTo/allowlist/replay/non-empty-NameID). CompleteLoginAsync accepts ONLY a VerifiedIdentity, so minting from a raw/unvalidated response is a COMPILE error, not a review catch; the record is unforgeable (private copy-ctor;  fails to compile — get-only members). The two protocols' SessionParameters construction collapsed into the one path (minimal-code win), behaviour field-for-field preserved. New conformance rule VerifiedIdentity_IsConstructedOnlyByProtocolValidators (no externally-reachable ctor via reflection; each factory invoked only from its protocol's validator). 992 tests (Debug+Release --warnaserror); CodeQL C# + required checks green; mergeable clean. Ready to merge.

PR #498, opened 2026-07-17
Extract the OpenID discovery-facts cache into OidcDiscoveryCache

### LINKED ISSUE #449: Extract the OidcDiscovery facts cache from SSOController into its own probe/cache type
Surfaced by the #210 delivery (PR #446).

`DiscoveryFactsCache` is a raw dictionary living on `SSOController` under a documented `MutableKeyedState_LivesOnlyInsideStoreLikeTypes` exemption. After #141 (PKCE S256 discovery probe) and #210 (RFC 9207 `authorization_response_iss_parameter_supported` probe) it now caches TWO discovery-derived facts per provider, so the case to extract it into a small, testable `OidcDiscovery*` probe/cache type (a `*Store`/`*Cache` per the target layering), removing the controller exemption and consolidating the discovery-fetch/parse/cache logic in one place, is stronger. Part of the #318 thin-controller direction; coordinate with #160.

Scope: pure refactor, behavior-preserving; move the cache + the discovery-facts read (`ReadDiscoveryFactsAsync`) off the controller, keep the 15-min staleness window, retire the conformance exemption. Refs #318, #160, #141, #210.

iderex:
# What & why

Extract the OpenID discovery-facts cache out of `SSOController` into its own
internal sealed `OidcDiscoveryCache`, and retire the conformance exemption that
covered it. Closes #449 (part of the #318 thin-controller direction; coordinates
with #160/#141/#210).

The controller held a raw `ConcurrentDictionary` (`DiscoveryFactsCache`) plus a
`ReadDiscoveryFactsAsync` method that fetched a provider's discovery document and
cached the two facts the challenge reads in one fetch, PKCE-S256 support (#141,
RFC 9700 §2.1.1) and whether the AS advertises the RFC 9207 response-`iss`
parameter (#210). It lived behind a documented
`MutableKeyedState_LivesOnlyInsideStoreLikeTypes` exemption noting it would move
into its own type in a later #318 step. This is that step: now that the cache
holds two facts and owns the fetch/parse, the case to consolidate it is stronger.

What changed:

- New `OidcDiscoveryCache` (a `*Cache` type per the target layering) owns the
  keyed cache, the 15-minute TTL, the discovery-URL derivation, and the
  timeout-bounded fetch/parse. The controller keeps one process-wide
  `static readonly` instance (like the SAML caches) and delegates via
  `ReadFactsAsync(config, provider, httpClientFactory, logger)`.
- `DiscoveryFacts` moves to its own file.
- The `SSOController.DiscoveryFactsCache` exemption is removed from
  `MutableKeyedState_LivesOnlyInsideStoreLikeTypes`, so the rule now covers the
  extracted cache structurally (it ends in `Cache`) and still catches any new
  raw dictionary added to the controller.
- New focused unit tests for the cache; `ARCHITECTURE.md` updated.

Behaviour is preserved: the fetch/parse/URL-derivation/TTL logic moved verbatim,
a within-TTL read is served without a re-fetch, a stale entry re-fetches, only a
definitive result is cached (a transient failure returns the tolerant default
without caching, so the next login retries), and the #341/#247
Pending-construction reuse path is untouched.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the PKCE gate still reads the same
      `DiscoveryFacts` values, so `RequirePkce` + not-advertised (false) and
      `RequirePkce` + unreadable (null) both still reject; response-`iss` stays
      optional exactly as before.
- [x] No secrets logged; the cache holds only two booleans and a timestamp keyed
      by a discovery URL — no secrets cached or logged.
- [x] A security review was performed (adversarial 4-lens gate; see Notes).
- [x] Log inputs from the IdP are sanitized inline at the log call: the
      fetch-failure warning keeps `provider?.ReplaceLineEndings(string.Empty)`
      inline, so the sanitizer never crosses a helper boundary.

## Quality checklist

- [x] No duplicated logic; the discovery-fetch/parse/cache logic now lives in one
      small, single-purpose, testable unit.
- [x] The change adds no more code than the problem requires (a straight
      extraction plus its tests).
- [x] The code is self-documenting and matches the target architecture (#318);
      comments explain the bounding rationale and the fail-closed "why".
- [x] Architecture-conformance tests pass; the tightened
      `MutableKeyedState_LivesOnlyInsideStoreLikeTypes` (exemption removed) is
      the locked-in structural property this change establishes.

## Verification

- [x] `dotnet build --warnaserror` is green (Debug and Release, 0 warnings).
- [x] `dotnet test` is green (1004 passed, +8 new).
- [x] Prettier is clean for the `.md` change; no `.js`/`.html`/`.css` touched.
- [x] Jellyfin compatibility metadata check is consistent (no metadata changed).

## Notes

Bounding decision: the cache key is the admin-configured provider's resolved
discovery URL (reached only after `config is { Enabled: true }`), so the map
holds at most one entry per configured provider, never attacker-supplied. Unlike
the anonymous-flood-facing login caches (`OidcStateStore`, `SamlRequestCache`,
which key on CSPRNG tokens and carry a hard cap + `IntervalGate`-throttled sweep),
`OidcDiscoveryCache` needs neither: a stale entry is re-fetched and overwritten in
place. It is therefore deliberately NOT added to
`LoginPathCaches_ThrottleTheirExpiredEntrySweepThroughIntervalGate`; documented on
the type.

Adversarial self-review, four refute-by-default lenses, all PASS:

- Availability / mass-lockout, VERDICT: growth bound is unchanged (one entry per
  admin-configured provider, not attacker-supplied), the 15-min TTL still serves
  within-TTL hits without a fetch (no re-fetch storm), failures are not cached
  (retry, matching the prior tolerant behaviour), and the test-only zero-TTL
  constructor cannot reach production. PASS.
- Security-bypass, VERDICT: the PKCE gate and response-`iss` handling see the
  identical `DiscoveryFacts`; TLS trust is unchanged (same `SsoHttp.CreateClient`,
  same 10s timeout); no secrets cached/logged; the log-forging sanitizer stays
  inline. No weakened PKCE/discovery trust. PASS.
- Correctness, VERDICT: URL derivation, TTL comparison, only-definitive-caching,
  and the tolerant `(null, false)` default all moved verbatim; unit tests pin the
  cache-hit, stale-refetch, definite-false, unusable-endpoint, no-double-append,
  failure-not-cached, and clear paths. Cache key + reuse semantics identical. PASS.
- Integration, VERDICT: the #341/#247 reuse (Pending built with
  `ProviderInformation` + `ResponseIssuerAdvertised` at construction) is untouched;
  the conformance exemption removal leaves the full suite green; the harness reset
  still clears discovery state between tests. PASS.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -84,13 +84,13 @@ A handful of state stores live as `private static readonly` fields on
 `IPluginServiceRegistrator` in source, so a `static readonly` field is today's
 only way to get one process-wide instance):
 
-| Store                 | Guards                                                                                     |
-| --------------------- | ------------------------------------------------------------------------------------------ |
-| `OidcStateStore`      | in-flight OIDC authorize state; capacity cap, lifetime, throttled-sweep via `IntervalGate` |
-| `SamlReplayCache`     | one-time-use SAML assertion IDs (replay protection)                                        |
-| `SamlRequestCache`    | outstanding SAML `AuthnRequest` IDs for `InResponseTo` correlation                         |
-| `DiscoveryFactsCache` | per-discovery-URL PKCE-S256 / RFC 9207 `iss` facts, 15-minute TTL                          |
-| `SsoRateLimiter`      | opt-in per-client rate limiting on the anonymous login endpoints                           |
+| Store                | Guards                                                                                      |
+| -------------------- | ------------------------------------------------------------------------------------------- |
+| `OidcStateStore`     | in-flight OIDC authorize state; capacity cap, lifetime, throttled-sweep via `IntervalGate`  |
+| `SamlReplayCache`    | one-time-use SAML assertion IDs (replay protection)                                         |
+| `SamlRequestCache`   | outstanding SAML `AuthnRequest` IDs for `InResponseTo` correlation                          |
+| `OidcDiscoveryCache` | per-discovery-URL PKCE-S256 / RFC 9207 `iss` facts, 15-minute TTL; owns the fetch/parse too |
+| `SsoRateLimiter`     | opt-in per-client rate limiting on the anonymous login endpoints                            |
 
 ### The uniform outcome
 
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -139,16 +139,14 @@ public class ArchitectureConformanceTests
     {
         // Locked in by the OidcStateStore consolidation (#318): a raw dictionary holding runtime state
         // outside a *Store/*Cache/*Limiter type is how the pre-consolidation controller accumulated its
-        // scattered cap/lifetime/sweep conventions. Two documented exemptions:
-        // - SSOController.DiscoveryFactsCache: the per-discovery-URL facts cache (PKCE-S256 support #141 +
-        //   the RFC 9207 response-iss advertisement #210) still lives on the controller and moves into its
-        //   own probe type in a later #318 step; naming the exact field keeps anything new from hiding
-        //   behind the exemption.
+        // scattered cap/lifetime/sweep conventions. The former SSOController.DiscoveryFactsCache exemption
+        // was retired in #449 once that cache moved into OidcDiscoveryCache (a *Cache type), so the rule now
+        // covers it structurally. One documented exemption remains:
         // - ProviderConfigBase._canonicalLinks: the persisted account-link map — serialized plugin
         //   configuration mutated only under the config lock, so a runtime store type would be the
         //   wrong home; it is config state, not in-flight state.
         var storeLike = new[] { "Store", "Cache", "Limiter" };
-        var exemptions = new[] { "SSOController.DiscoveryFactsCache", "ProviderConfigBase._canonicalLinks" };
+        var exemptions = new[] { "ProviderConfigBase._canonicalLinks" };
 
         var offenders = PluginClasses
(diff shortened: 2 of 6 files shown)


iderex (commented):
My sign-off — login-path/config refactor (#318 step 5, Closes #449). Delivered by a delegated agent (CPU-capped); four-lens review all PASS. The OIDC discovery-facts cache (PKCE-S256 support #141 + RFC 9207 iss advertisement #210, keyed by discovery URL, 15-min TTL) is extracted from SSOController into a new internal sealed OidcDiscoveryCache; the controller keeps one static readonly instance and delegates; DiscoveryFacts moved to its own file. Well-reasoned divergence from the sibling caches: it deliberately omits the hard cap + IntervalGate sweep because it keys on ADMIN-configured provider endpoints (at most one entry per configured provider, overwritten in place), not attacker-supplied CSPRNG tokens like OidcStateStore/SamlRequestCache — so it is correctly excluded from LoginPathCaches_ThrottleTheirExpiredEntrySweepThroughIntervalGate, with the bounding rationale documented on the type. The DiscoveryFactsCache exemption is removed from MutableKeyedState_LivesOnlyInsideStoreLikeTypes, so the rule now covers the extracted cache structurally and still catches any new raw dict on the controller. Behaviour-preserving: fetch/parse/URL-derivation/TTL/definitive-only-caching/tolerant-default moved verbatim, the #341/#247 reuse path untouched, the fetch-failure log keeps its inline sanitizer; failures are not cached (retry). 8 new tests, 1004 total; ARCHITECTURE.md store table updated; CodeQL + required checks green; mergeable clean. Ready to merge.

PR #500, opened 2026-07-18
Extract the OpenID login flow into OidcLoginService

iderex:
# What & why

Continue thinning the `SSOController` monolith (#318 step 12): the whole OpenID
flow, the challenge, the redirect callback, the session-minting authenticate
leg, and the manual link redeem, moves out of the controller into a new
`internal sealed OidcLoginService` in the flow tier (`Api/Flows/`), matching the
shape `LoginCompletionService` already established.

- The two OpenID-specific process-wide caches (the in-flight authorize-state
  store and the discovery-facts cache) move onto the service as `static readonly`
  fields, keeping their single-process-wide instance semantics, the controller
  is per-request, so an instance field would drop in-flight state between a
  challenge and its callback. The `Reset*/Seed*ForTests` hooks move with them and
  stay `internal` via `InternalsVisibleTo`.
- The shared per-client rate limiter deliberately stays a controller static: it
  fronts **both** the OpenID and SAML flows, so both keep reaching it through the
  controller's unchanged rate-limit gate. **SAML rate-limiting is untouched.**
- The controller's OpenID endpoints become thin adapters: apply the shared
  rate-limit gate, then delegate. Challenge/callback are irreducibly HTTP, so they
  take the request/response plus a delegate for the two response shapes the
  controller still owns (the security-headered auth page and the shared
  manual-link write mapping); the authenticate leg stays HttpContext-free like the
  completion tail, taking only the model, the binding cookie, and the
  remote-endpoint resolver.

Behaviour-preserving and wire-for-wire: same redirect/authorize URLs, state
handling, callback outcomes, and 400/401/redirect bodies. The controller drops
from 1476 to 1058 lines. Refs #160.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the disabled-provider guard still precedes state
      consumption, a binding mismatch does not consume the state, a miss/expiry
      cleanly rejects, and the `VerifiedIdentity` factory still runs only after
      full OIDC validation. No throwing path became a guarded-Try that falls
      through to mint.
      changed; the moved log calls keep their inline `ReplaceLineEndings`
      sanitizers.)
- [x] A security review was performed, the four-lens adversarial gate below.
- [x] Log inputs from the IdP are sanitized inline at the log call (moved
      verbatim with the code).

## Quality checklist

      unit (`OidcLoginService`). The one deliberate copy — a `PlainTextError`
      helper mirroring the controller's `ReturnError`, which the SAML challenge
      still uses — is noted below; a shared home is deferred to the SAML
      extraction step.
- [x] The change adds no more code than the problem requires (it is a move; the
      controller shrinks by 418 lines).
      locked in as `Controller_DelegatesOidcFlowToTheFlowService` in
      `ArchitectureConformanceTests` (nameof-derived tokens, liveness-guarded).

## Verification

- [x] `dotnet test` is green (1014 passed, 0 failed).
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change (none in this PR).

## Notes

**Adversarial self-review, four refute-by-default lenses (login-path,
both-protocol-touching):**

- **Availability / mass-lockout, VERDICT: the shared rate limiter is untouched
  and still fronts both protocols (SAML `RateLimitCheck` unchanged); the OIDC
  caches keep single-process-wide lifetime (`static readonly`, not per-instance,
  so per-request service construction never re-initialises them, confirmed by
  the end-to-end challenge→callback test that persists state across two calls); no
  new lockout path. PASS.**
- **Security-bypass / fail-open, VERDICT: every fail-closed guard precedes state
  consumption exactly as before (enabled-check before TryAdd/PeekCurrent/TryRedeem
  in all four legs); the binding mismatch still does not consume the state; the
  `VerifiedIdentity` keystone is unweakened (still built inside
  `AuthorizeSession.Ready`, handed out only by the one-time atomic redeem, the
  `VerifiedIdentity_IsConstructedOnlyByProtocolValidators` conformance rule still
  passes); no throwing guard became a fall-through; no default-accept. PASS.**
- **Correctness / wire equivalence + static semantics, VERDICT: rejection bodies
  are byte-identical, preserving the `BadRequestObjectResult` vs `ContentResult`
  distinctions (callback "Missing state"/"Invalid or expired state" stay
  `BadRequestObjectResult`; challenge/auth rejections stay `ContentResult` via
  `LoginStatusMapper`; `PlainTextError` reproduces `ReturnError` exactly);
  `RedirectResult` for the challenge and text/html for the callback are unchanged;
  the base-URL argument mapping into `CanonicalBaseUrl.Resolve` is identical; the
  1014-test suite (RFC 9207, COAT, browser-binding, token-exchange-fail,
  sub-missing, end-to-end challenge→callback) passes empirically. PASS.**
- **Integration, VERDICT: the service is constructed after its collaborators in
  the controller ctor and receives the factories as ctor params; the per-request
  service / process-wide static lifetime split is sound; the test-isolation hooks
  moved to `OidcLoginService` and the harness/tests call the new location (build +
  tests green); both callbacks (redirect callback, session-mint) plus the link and
  states endpoints delegate correctly, with the CSP/security-header logic still
  owned by the controller's `HtmlAuthPage` (passed as a delegate, not duplicated).
  PASS.**

All four PASS.

**Out of scope:** `ReturnError` and the `MapWrite`/`HtmlAuthPage` helpers remain
shared with the SAML endpoints, so this PR passes the two render/mapping helpers
into the service as delegates and keeps a small `PlainTextError` twin rather than
extracting a shared home, that unification rides with the later SAML flow
extraction step (Refs #160).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -472,6 +472,46 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void Controller_DelegatesOidcFlowToTheFlowService()
+    {
+        // Locked in by the OpenID flow extraction (#160, #318 step 12): the OpenID challenge and redirect
+        // callback bodies, together with the OpenID-specific process-wide state (the in-flight authorize
+        // store and the discovery-facts cache), moved into OidcLoginService. The controller's OpenID
+        // endpoints now apply the shared rate-limit gate and hand the request to that service, so a
+        // CONTROLLER neither holds those OIDC caches nor drives the OidcClient challenge/callback protocol
+        // itself. Call-level property, so it is a source scan like the other controller rules above.
+        //
+        // The two cache tokens are nameof-derived, so a rename of either type fails to COMPILE this rule
+        // rather than passing vacuously; the two protocol tokens are the OidcClient methods the challenge
+        // (PrepareLoginAsync) and callback (ProcessResponseAsync) drive. The shared per-client rate limiter
+        // is deliberately NOT a marker — it fronts BOTH protocols, so it legitimately stays a controller
+        // static and its extraction is out of scope for this step.
+        var storeToken = nameof(OidcStateStore);
+        var cacheToken = nameof(OidcDiscoveryCache);
+        var markers = new[] { storeToken, cacheToken, "PrepareLoginAsync", "ProcessResponseAsync" };
+
+        var controllerHits = ControllerSourceFiles()
+            .SelectMany(path => File.ReadAllLines(path)
+                .Select((line, index) => (File: Path.GetFileName(path), Text: line.Trim(), Number: index + 1)))
+            .Where(l => markers.Any(m => l.Text.Contains(m, StringComparison.Ordinal)))
+            .Select(l => $"{l.File} line {l.Number}: {l.Text}")
+            .ToList();
+
+        Assert.True(
+            controllerHits.Count == 0,
+            "A controller must not hold the OpenID authorize/discovery caches or drive the OidcClient challenge/callback protocol; the OpenID flow lives in OidcLoginService (#160). Found: " + string.Join(" | ", controllerHits));
+
+        // Liveness against a vacuous pass: the OpenID flow must actually live in OidcLoginService — a move,
+        // not a silent removal — so the flow service's own source must contain every moved token.
+        var oidcSource = string.Join(
+            "\n",
+            SourceFilesDeclaring(new[] { typeof(OidcLoginService) }).Select(File.ReadAllText));
+        Assert.True(
+            markers.All(m => oidcSource.Contains(m, StringComparison.Ordinal)),
+            "OidcLoginService must own the OpenID challenge/callback flow and its authorize/discovery caches; otherwise the controller scan passes vacuously (#160).");
+    }
+
+    [Fact]
     public void RateLimiting_FlowsThroughLoginOutcome()
     {
         // Locked in by #474: the rate-limit rejection was the last login-path error that bypassed the single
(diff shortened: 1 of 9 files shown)


iderex (commented):
My sign-off — login-path refactor, the core OIDC controller-thinning (#318 step 12, Refs #160). Delivered by a delegated agent (CPU-capped); four-lens review all PASS. The entire OpenID flow moved from SSOController into a new internal sealed OidcLoginService under Api/Flows (ChallengeAsync/CallbackAsync/AuthenticateAsync/Link/StateSummaries + the OIDC-only helpers); the five OpenID endpoints are now thin (shared rate-limit gate, then a one-line delegate to _oidc, new'd in the ctor). OidcStateStore + OidcDiscoveryCache relocated onto the service as static readonly (process-wide instance semantics preserved — confirmed by the end-to-end challenge->callback test), with their Reset*/Seed*ForTests hooks moved along (internal via InternalsVisibleTo). Correct shared-state handling: the SsoRateLimiter deliberately STAYS a controller static reached by both protocols via the unchanged RateLimitCheck gate — SAML rate-limiting untouched. Fail-closed + VerifiedIdentity preserved wire-for-wire: guard order verbatim (disabled-provider precedes state consumption in all legs; binding mismatch does not consume; miss/expiry cleanly rejects), rejection body types exact (BadRequestObjectResult vs ContentResult, PlainTextError reproduces ReturnError), CSP HtmlAuthPage + MapWrite passed in as delegates (no security-header duplication). Controller 1476 -> 1058 lines (-418). 1014 tests; new conformance rule Controller_DelegatesOidcFlowToTheFlowService; CodeQL + required checks green; mergeable clean. The shared ReturnError/MapWrite/HtmlAuthPage home rides with the SAML flow extraction (next step). Ready to merge.

PR #501, opened 2026-07-18
Extract the SAML login flow into SamlLoginService

iderex:
# What & why

Extract the SAML login flow off the `SSOController` monolith into a new
`internal sealed SamlLoginService`, symmetric to the just-merged
`OidcLoginService` (#318 step 13). The controller's SAML endpoints become thin
rate-limit + delegate adapters, so the SAML protocol logic now lives in one
flow-tier collaborator instead of inline on the controller.

We also close the duplication PR #500 flagged: the HTTP result shapes both flow
services share, the CSP-hardened intermediate auth page, the plain-text flow
error, and the manual-link write mapping, are consolidated into one shared
`SSO-Auth/Api/Shared/FlowResponses.cs`, and `OidcLoginService` is switched to
call it instead of the passed-in controller delegates and its `PlainTextError`
twin.

Refs #160

## Type of change

- [ ] Security hardening / vulnerability fix
- [ ] Bug fix
- [ ] Feature
- [x] Refactor / code quality / docs

## Security checklist

Fill in for any change touching the login path, crypto (SAML/OIDC), config
persistence, or the release pipeline.

- [x] Fail-closed preserved: a missing signature, time-bound, or audience is
      rejected, never default-accepted. The full SAML validation chain
      (signature/time/audience/recipient -> InResponseTo correlation + browser
      binding -> allow-list -> one-time replay consume -> non-empty NameID) still
      runs, in the same order, before `VerifiedIdentity.FromValidatedSaml`.
- [x] No secrets logged; secrets are redacted on config export. The signing-key
      failure path logs only `ex.Message`, never key material -- moved verbatim.
- [x] A security review was performed for changes to SAML/OIDC validation,
      config persistence, or the release pipeline. Four-lens adversarial gate
      below -- all PASS.
- [x] Log inputs from the IdP are sanitized inline at the log call
      (`?.ReplaceLineEndings(string.Empty)` at every SAML log site -- moved
      verbatim, not routed through a helper boundary).

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose,
      testable unit. This removes duplication rather than adding it: the shared
      auth-page/error/write shapes now have one home in `FlowResponses`.
- [x] The change adds no more code than the problem requires. `SSOController`
      drops from 1169 to 728 lines (-441); the moved code is relocated, not
      rewritten.
- [x] The code is self-documenting and matches the target architecture (#318):
      thin controller + pure/flow-tier collaborators, mirroring the OpenID flow.
[...]

Issue #474, opened 2026-07-17
Add LoginOutcome.Throttled and route rate-limiting through the outcome/mapper


iderex:
Part of the #318 unified-architecture plan ([design note](https://github.com/iderex/jellyfin-plugin-sso/issues/318#issuecomment-5005931010), step 9 / new sub-issue 1).

## Proposal

Add a `LoginOutcome.Throttled(RetryAfter)` case and route rate-limiting through the outcome instead of the bare `ContentResult` that `RateLimitCheck` returns today. Rate-limiting becomes flow stage 0; `LoginStatusMapper` maps `Throttled` to a 429 and is the ONLY place `Retry-After` is set. `LoginOutcome`'s own doc comment already reserves this case.

## Why

`LoginOutcome` should be the single currency across a flow boundary and `LoginStatusMapper` the single error-emission point (the #318 standard). The bare `ContentResult` is the last login-path error path that bypasses the mapper.

## Acceptance
- [ ] `LoginOutcome.Throttled(RetryAfter)` exists; `RateLimitCheck`'s bare `ContentResult` is gone.
- [ ] The 429 + `Retry-After` are emitted only in `LoginStatusMapper`.
- [ ] Conformance: `RateLimiting_FlowsThroughLoginOutcome` (no bare rate-limit ContentResult on the controller).
- [ ] Behaviour-preserving: the rate-limit response (status, Retry-After, body) is byte-identical, pinned by a characterization test.

Sequence after step 4 (#456). area: oidc, saml.

Issue #382, opened 2026-07-15
Rate-limit the authenticated link/unlink admin endpoints


TheRealStroopwafel:
Raised by review on #373. `RateLimitCheck` covers only the anonymous challenge/callback/auth endpoints. The link/unlink/list admin endpoints have no rate limiting, so an authenticated user can force serialized config-XML disk writes at will (each DELETE against a nonexistent-or-name-miss provider runs a full persist under the global config lock). This predates #373 (the valid-provider + bogus-name DELETE already forced a no-op write) and is admin/self-scope gated, so it is low priority, but the endpoints are the one write surface with no throttle.

Direction: extend the limiter to the authenticated link/unlink endpoints (a distinct endpoint class / budget), or short-circuit the no-result admin arms before the persist so a miss does not write at all.

Refs #373

PR #503, opened 2026-07-18
Move the shared rate-limit gate off SSOController into the Shared tier

### LINKED ISSUE #502: Move the shared rate-limit gate off SSOController into the Shared tier
After the OpenID (#500) and SAML (#501) flow extractions, the only remaining mutable process-wide static on `SSOController` was the shared `SsoRateLimiter` instance, fronted by the controller's `RateLimitCheck` gate for both the OpenID and SAML anonymous endpoints. That was the last thing keeping the controller from being a stateless request dispatcher (the #160 static-state goal, part of the #318 target architecture).

## What we want
Relocate the shared rate-limit gate into the `Api/Shared/` tier so the controller holds no mutable static state:

- A small `SsoRateLimitGate` (Api/Shared) owns the one `static readonly SsoRateLimiter` and exposes the opt-in `Check(...)` returning the same `LoginOutcome.Throttled`/null.
- The controller's endpoints keep fronting the gate (wire-for-wire: same fire points, same 429/Retry-After via the single mapper #474, same IP classifier, same process-wide single instance).
- Lock it in with a `Controller_HoldsNoMutableStaticState` architecture-conformance rule that fails if a mutable static (a writable static, or a static readonly `*Store`/`*Cache`/`*Limiter`/dictionary) is dropped back onto any controller.

Behaviour-preserving only, no rate-limit behaviour, config, or security-posture change. Refs #160, #318.

iderex:
# What & why

After the OpenID (#500) and SAML (#501) flow extractions, the only remaining mutable process-wide static on `SSOController` was the shared `SsoRateLimiter`, fronted by the controller's `RateLimitCheck` gate for both protocols' anonymous endpoints. This relocates that gate into the `Api/Shared/` tier so the controller becomes a stateless request dispatcher, the last step of the #160 static-state goal in the #318 target architecture.

- New `SsoRateLimitGate` (Api/Shared) owns the one `static readonly SsoRateLimiter` and exposes the opt-in `Check(...)`. Its body is a byte-for-byte move of the controller's former `RateLimitCheck` (config read under the lock, IP classifier, endpoint-class keying, the #195 bounded observability signal, and the `LoginOutcome.Throttled` the single mapper renders, #474).
- The controller keeps a thin `RateLimitCheck` wrapper that supplies the three request-scoped inputs (endpoint class, `HttpContext.Connection.RemoteIpAddress`, `Response`) and delegates to the gate, so every fire point stays wire-for-wire.
- Locked in by a new `Controller_HoldsNoMutableStaticState` conformance rule.
- Stale rate-limiter remarks refreshed on both flow services, the two #318 delegation-rule comments, and the ARCHITECTURE.md process-wide-stores section.

Behaviour-preserving only: no rate-limit behaviour, config, or security-posture change. The rate limit fires at the same points, same 429 + `Retry-After`, same fail-open for unattributable/non-public clients, same single process-wide instance shared across OpenID and SAML.

Closes #502, Refs #160, #318.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (Unchanged — this only relocates the rate-limit gate; the limiter fails OPEN by design for unattributable/non-public clients, preserved verbatim.)
- [x] No secrets logged; secrets are redacted on config export. (The only log line is the #195 aggregate throttled-count, carrying no client key.)
      config persistence, or the release pipeline. (Four-lens adversarial review below.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (No IdP input reaches the moved log line; unchanged.)

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. (Debug + Release, 0 warnings.)
- [x] `dotnet test` is green. (1024 passed, 0 failed.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (ARCHITECTURE.md checked.)

## Notes

**Design.** The shared limiter instance + the whole check live in `SsoRateLimitGate` (Api/Shared). Both login flows reach it through the controller's endpoints, which keep the thin `RateLimitCheck` wrapper, we deliberately did NOT push the gate call down into `OidcLoginService`/`SamlLoginService`, because (a) that would change fire points (the flow mint legs are intentionally HttpContext-free, #177, taking a remote-IP `Func` rather than the `IPAddress` the classifier needs), and (b) the `Link` legs are not rate-limited today, so per-method placement would risk drift. Keeping the controller as the front is the wire-for-wire-safe, minimal change, and the gate is a genuinely shared home any caller (including the future admin-endpoint throttling in #382) can call. Controller: 728 -> 693 lines, now holding no mutable static state.

**Adversarial self-review (four refute-by-default lenses).**

- **Availability / mass-lockout**, VERDICT: The gate body is a verbatim move; the disabled-limiter early-return, the `windowSeconds < 1` / `maxAttempts < 1` disables, and the null-key fail-open (unattributable/non-public/proxy) are all unchanged. One `static readonly` process-wide instance is preserved, so a client's window is still shared across all endpoints and both protocols, no new lockout, no per-flow split. The two 429 pinning tests pass. **PASS**
- **Security-bypass**, VERDICT: No path reaches a flow service without first going through the unchanged `RateLimitCheck(...) is { } throttled` guard; the gate is `internal static`, adds no public surface, and `SsoRateLimiter` itself is untouched. Throttled still maps through `LoginStatusMapper.ToActionResult(Throttled, response)` to a 429 + `Retry-After` + fixed body, characterized byte-for-byte by `SamlChallenge_OverRateLimit_Returns429`. **PASS**
- **Correctness**, VERDICT: Same instance semantics (default `maxEntries = 100_000`), same six fire points with the same endpoint-class strings, same `DateTime.UtcNow` threaded into `IsAllowed` and `RecordThrottledHit`, byte-identical log line. Nothing reordered. **PASS**
- **Integration**, VERDICT: The controller fronts the gate for its anonymous endpoints; admin endpoints (unchanged) still don't throttle. The new `Controller_HoldsNoMutableStaticState` is non-vacuous: it flags a writable static or a static readonly state-container (`*Store`/`*Cache`/`*Limiter`/dictionary) on any controller, so the former `SsoRateLimiter` static would fail it, while correctly allowing `SSOViewsController`'s immutable `static readonly` asset ETag; a controller-exists sentinel plus a liveness pin (SsoRateLimitGate owns the SsoRateLimiter, which `IsStateContainer`) guard the vacuous cases. Full suite green. **PASS**

**Out of scope (follow-up).** The ARCHITECTURE.md "Process-wide stores" section had pre-existing drift from #500/#501 (it still described the OIDC/SAML caches as living on `SSOController`); we corrected the whole table's owner column while here, but any deeper #318-narrative refresh of that doc is separate.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -79,18 +79,24 @@ that tail into one shared collaborator is target-direction step 11
 
 ### Process-wide stores
 
-A handful of state stores live as `private static readonly` fields on
-`SSOController` (not dependency-injected — there is no
+A handful of state stores live as `private static readonly` fields — the OIDC
+and SAML caches on their owning flow service (`OidcLoginService` /
+`SamlLoginService`, #500/#501) and the shared rate limiter on the
+`SsoRateLimitGate` (Api/Shared, #160) — not dependency-injected (there is no
 `IPluginServiceRegistrator` in source, so a `static readonly` field is today's
 only way to get one process-wide instance):
 
-| Store                | Guards                                                                                      |
-| -------------------- | ------------------------------------------------------------------------------------------- |
-| `OidcStateStore`     | in-flight OIDC authorize state; capacity cap, lifetime, throttled-sweep via `IntervalGate`  |
-| `SamlReplayCache`    | one-time-use SAML assertion IDs (replay protection)                                         |
-| `SamlRequestCache`   | outstanding SAML `AuthnRequest` IDs for `InResponseTo` correlation                          |
-| `OidcDiscoveryCache` | per-discovery-URL PKCE-S256 / RFC 9207 `iss` facts, 15-minute TTL; owns the fetch/parse too |
-| `SsoRateLimiter`     | opt-in per-client rate limiting on the anonymous login endpoints                            |
+| Store                | Owner              | Guards                                                                                      |
+| -------------------- | ------------------ | ------------------------------------------------------------------------------------------- |
+| `OidcStateStore`     | `OidcLoginService` | in-flight OIDC authorize state; capacity cap, lifetime, throttled-sweep via `IntervalGate`  |
+| `SamlReplayCache`    | `SamlLoginService` | one-time-use SAML assertion IDs (replay protection)                                         |
+| `SamlRequestCache`   | `SamlLoginService` | outstanding SAML `AuthnRequest` IDs for `InResponseTo` correlation                          |
+| `OidcDiscoveryCache` | `OidcLoginService` | per-discovery-URL PKCE-S256 / RFC 9207 `iss` facts, 15-minute TTL; owns the fetch/parse too |
+| `SsoRateLimiter`     | `SsoRateLimitGate` | opt-in per-client rate limiting on the anonymous login endpoints                            |
+
+The controller itself now holds **no mutable static state** — every store above
+lives in a flow service or the Shared tier, pinned by
+`Controller_HoldsNoMutableStaticState`.
 
 ### The uniform outcome
 
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -8,6 +8,7 @@ using System.Text.RegularExpressions;
 using Jellyfin.Plugin.SSO_Auth;
 using Jellyfin.Plugin.SSO_Auth.Api;
 using Jellyfin.Plugin.SSO_Auth.Api.Flows;
+using Jellyfin.Plugin.SSO_Auth.Api.Shared;
 using Jellyfin.Plugin.SSO_Auth.Config;
 using MediaBrowser.Model.Plugins;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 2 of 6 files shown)


iderex (commented):
My sign-off — login-path rate-limit refactor, finishes the #160 static-state goal (Closes #502, Refs #160/#318). Delivered by a delegated agent (CPU-capped); four-lens review all PASS. The shared SsoRateLimiter + the RateLimitCheck body moved off SSOController into a new internal static SsoRateLimitGate under Api/Shared (owns the one private static readonly SsoRateLimiter; the body is a verbatim move — locked config read, NormalizeClientKey IP classifier, endpoint-class keying, the #195 throttled-count signal, the #474 LoginOutcome.Throttled). Sound placement call: the controller keeps a thin RateLimitCheck(endpointClass) wrapper as the front rather than pushing the gate into the flow services, because the flow mint legs are intentionally HttpContext-free (#177) and the Link legs do not rate-limit — so per-method placement would have changed fire points; the wrapper preserves the exact six call sites, keys, single process-wide instance, 429/Retry-After, and fail-open. New conformance rule Controller_HoldsNoMutableStaticState (flags a writable static or a static-readonly state-container on any controller; correctly excludes the immutable SSOViewsController.AssetETag), non-vacuous. Controller 728 -> 693, zero SsoRateLimiter references remain — it now holds NO mutable static state and is a stateless request dispatcher. 1024 tests; ARCHITECTURE.md stores table + stale comments refreshed; CodeQL + required checks green; mergeable clean. Ready to merge.

PR #504, opened 2026-07-18
Parse the oid/saml mode token once at the controller boundary

### LINKED ISSUE #369: Parse the oid/saml mode token once at the controller boundary

iderex:
# What & why

Closes #369. The route's `{mode}` token (`oid`/`saml`) travelled inward as a raw
string and was parsed/compared in several places that all had to agree:

- two independent switches that had to stay in lock-step, the controller's
  culture-sensitive `mode.ToLower()` dispatch in `AddCanonicalLink`, and
  `CanonicalLinkService.TryGetLinks`' invariant-lowercase `mode.ToLowerInvariant()`
  switch;
- per-site `string.Equals(mode, "saml"/"oid", …)` comparisons for the link
  projection (`LinksByUser`) and the adoption audit line;
- the DELETE link route forwarded the raw, unvalidated string all the way into
  the service, where the invalid-mode throw was reachable deep inside a config
  transaction.

A typo or a culture-sensitive lowercasing edge (e.g. an uppercase `OID` under a
Turkish locale) could make the two dispatches disagree.

This sequences after the flow extractions (#500/#501), so the boundary is now
its final shape. The change introduces a typed `ProviderMode` enum (`Saml`/`Oid`)
parsed exactly once, at the HTTP boundary, through a single
`ProviderModeParser.TryParse`, and threads the typed value inward:

- **Single parse at the boundary.** `SSOController.ParseMode` turns the `{mode}`
  route string into `ProviderMode` once, for both the add and delete link routes.
  An unknown token is rejected there, an `ArgumentException` with the same
  message the old dispatch produced, and never defaults to a protocol. No inner
  layer re-parses or re-compares the raw string.
- **Typed value threaded inward.** `CanonicalLinkService` takes `ProviderMode` on
  every mode-bearing method (`ResolveOrCreateAsync`, `TryCreateLink`,
  `TryRemoveLink`, `LinksByUser`, `IsIdentityStillLinked`, and the private
  helpers); the dispatch is an enum `switch`, and the link projection and adoption
  audit compare the enum. `VerifiedIdentity.LinkMode` carries the protocol as the
  typed `ProviderMode`, and the flow services pass the enum literal to
  `TryCreateLink`.
- **Log output unchanged.** The lowercase link-namespace token still renders in
  operator log lines via `ProviderMode.ToToken()`, the single string↔mode
  mapping home, so `{Mode}` log output is byte-identical.

Removed: the second (`TryGetLinks`) dispatch's string lowercasing, the two
`string.Equals(mode, …)` comparisons, and the controller's culture-sensitive
`ToLower()` switch.

Behaviour-preserving: same routes, same accepted modes (now case-insensitive and
culture-independent, so the two former dispatches can no longer disagree), same
fail-closed rejection of an unknown mode (same `ArgumentException`), same
downstream behaviour and log output.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

Touches the login path, so filled in even though it is security-neutral:

- [x] Fail-closed preserved: an unknown/unparseable mode is still rejected once,
      at the boundary, and never defaults to a protocol (the internal
      `TryGetLinks` default arm keeps a belt throw against an out-of-range enum).
      change beyond rendering the same lowercase mode token.)
- [x] A security review was performed - the four-lens adversarial-review gate;
      results under Notes.
- [x] Log inputs from the IdP are sanitized inline at the log call (unchanged;
      `ReplaceLineEndings` on provider/username stays inline).

## Quality checklist

- [x] No duplicated logic; the single string↔mode mapping lives once in
      `ProviderModeParser`, and the dispatch lives once behind the boundary parse.
- [x] The change adds no more code than the problem requires (net removes the
      second dispatch and the per-site comparisons).
- [x] Architecture-conformance tests pass; the new structural property (the
      linking surface threads the typed `ProviderMode`, never a raw string mode)
      is locked in as `ProviderMode_IsThreadedTyped_NotAsARawStringToken`.

## Verification

- [x] `dotnet test` is green (1040 tests).
- [x] Prettier: no `.js`/`.html`/`.md`/`.css` changed.

## Notes

New tests: `ProviderModeTests` (parse case-insensitivity, fail-closed rejection,
token round-trip), a delete-route invalid-mode pin and a mixed-case routing pin
in `SSOControllerLinkTests`, and the conformance rule above.

Minor deliberate behaviour delta, an improvement not a regression: the parse is
now culture-independent (`OrdinalIgnoreCase`), so a mixed-case token like `OID`
is accepted consistently on both routes instead of depending on the server
locale. Real traffic sends lowercase `oid`/`saml` (see `Config/linking.js`), so
it is unaffected.

Pre-merge review, the four-lens adversarial gate ran on the diff; verdicts are
posted as a follow-up comment. Ready for approval once green.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -712,6 +712,48 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void ProviderMode_IsThreadedTyped_NotAsARawStringToken()
+    {
+        // Locked in by #369: the route's {mode} token is parsed ONCE at the controller boundary into the
+        // ProviderMode enum, and the typed value is threaded inward — so no linking-tier method re-accepts
+        // the raw string to re-parse or re-compare it (the two former divergent dispatches, a
+        // culture-sensitive ToLower() switch and an invariant-lowercase one, that had to agree). Pin it
+        // structurally on the two types the token flows through:
+        //
+        // 1. CanonicalLinkService — the linking workflow: NO method (public or private) may take a parameter
+        //    named "mode" typed as string; it must be the ProviderMode enum. A revert to a string mode
+        //    parameter (reopening the re-parse-inward hole) fails HERE.
+        // 2. VerifiedIdentity.LinkMode — the identity the login path carries: must expose the protocol as the
+        //    typed ProviderMode, not a "oid"/"saml" string the mint path would have to re-compare.
+        const BindingFlags anyMethod = BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly;
+        var methods = typeof(CanonicalLinkService).GetMethods(anyMethod);
+
+        var stringModeParams = methods
+            .SelectMany(m => m.GetParameters().Select(p => (Method: m.Name, Param: p)))
+            .Where(x => x.Param.Name == "mode" && x.Param.ParameterType == typeof(string))
+            .Select(x => $"{x.Method}(string mode)")
+            .ToList();
+        Assert.True(
+            stringModeParams.Count == 0,
+            "CanonicalLinkService must take the parsed ProviderMode, never a raw string mode token, so the {mode} route string is parsed once at the boundary and threaded typed inward (#369): " + string.Join(", ", stringModeParams));
+
+        // Sentinel against a vacuous pass: a ProviderMode-typed mode parameter must actually exist on the
+        // surface, so a rename that dropped the parameter entirely does not pass for the wrong reason.
+        var typedModeExists = methods
+            .SelectMany(m => m.GetParameters())
+            .Any(p => p.Name == "mode" && p.ParameterType == typeof(ProviderMode));
+        Assert.True(
+            typedModeExists,
+            "No ProviderMode-typed \"mode\" parameter was found on CanonicalLinkService; the linking surface was renamed, so point this rule at the new shape so the typed-mode invariant keeps guarding #369.");
+
+        var linkMode = typeof(VerifiedIdentity).GetProperty("LinkMode", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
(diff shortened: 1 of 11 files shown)


iderex (commented):
My sign-off — login-path type-safety refactor (#318 step 17, Closes #369). Delivered by a delegated agent (CPU-capped); the four-lens adversarial review all PASS. The oid/saml {mode} route token is now parsed ONCE at the controller boundary (ParseMode → typed ProviderMode via ProviderModeParser.TryParse) and the enum threads inward; the scattered raw-string comparisons are gone, VerifiedIdentity.LinkMode is the enum set only by the two factories, and the flow services + link endpoints all pass the typed value. Behaviour-preserving statement-for-statement: routes byte-identical, config-dictionary routing + adoption audit strings + all six lowercase {Mode} log renders (via ToToken) unchanged, and unknown-mode rejection is the same ArgumentException (same message) at the boundary with the auth guard still evaluated first (no enumeration oracle). The one intended change — unifying the old culture-sensitive ToLower() / invariant ToLowerInvariant() to OrdinalIgnoreCase — is a strict correctness improvement (it fixes a latent Turkish-locale bug where OID under tr-TR would wrongly throw) and fails closed on every unknown/null/malformed token (TryParse=false makes ParseMode throw before the default(ProviderMode)==Saml out-value is ever read). New ProviderModeTests + invalid-mode DELETE test + the non-vacuous ProviderMode_IsThreadedTyped conformance rule; 1040 tests; CodeQL + required checks green; mergeable clean. Ready to merge.

Issue #237, opened 2026-07-13
Dead-code sweep: unused constructors, redundant defaults, leftover debug logging


iderex:
Small provably-dead items from a code-reduction review, ~60 lines total, none behavior-affecting:

- `SerializableDictionary.cs:24-69`: five constructor overloads are never called, every construction site repo-wide uses the parameterless one (`PluginConfiguration.cs:17,18,234,393`; `SSOController.cs:1298,1319`; tests use collection initializers). Either delete all five (~46 lines), or keep the `IDictionary` copy constructor and let `SnapshotConfigs` (`SSOController.cs:1317-1326`) use it instead of its manual foreach copy, `Dictionary(IDictionary)` performs the identical shallow copy with the default comparer, so the #157 detached-snapshot property (still taken under the config lock) is unchanged. XML serialization only requires the parameterless constructor, which stays.
- `SSOController.cs:1909-1914`: the `TimedAuthorizeState` constructor assigns six CLR defaults (`false`/`null`). Keep only `State` and `Created`.
- `SSOController.cs:1626-1629`: `Invalidate()` is a one-line wrapper with a single caller (`:257`); inline the `AuthStateStore.InvalidateExpired` call.
- `SSOController.cs:12`: `using System.Text.RegularExpressions;` is unused (the only Regex lives in `OidcAuthorizeStateBuilder`).
- `Config/config.js:394`: `initSsoConfigurationPage` calls `listArgumentsByType(view)` and discards the result; the function is a pure DOM query with no side effects (leftover from the original UI PR).
- Leftover debug `console.log` statements in the admin/linking UI: `config.js:430`, `linking.js:83`, `linking.js:186`, `Views/apiClient.js:3`, `Views/apiClient.js:123`. The linking-page ones print the user's provider link map to the console; plain removal. (The error-path `console.log(e)` in the served auth page stays, it is error reporting.)

Verified by grep across SSO-Auth and SSO-Auth.Tests; no security check, fail-closed branch, or serialization contract is touched.

PR #506, opened 2026-07-18
Sweep contributor-clarity traps: callback naming, dead accessor, IP-classifier home, asset casing

### LINKED ISSUE #370: Contributor-clarity sweep: callback naming, dead SAML accessor, IP-classifier home, asset casing

iderex:
# What & why

Four small, disjoint contributor-clarity traps bundled as one sweep (in the
style of #237). #370 was filed before the OIDC/SAML flow extraction
(#500/#501), the rate-limit gate move (#503) and the mode-token typing
(#504), so each item was re-verified against the code as it stands today
before touching anything. All four were still applicable; nothing has been
resolved out from under the issue by those refactors.

Closes #370

## Per-item disposition

- **Callback naming**, still applicable. `SSOController.OidPost` handled a
  GET, excused only by an inline comment explaining the misleading name.
  Renamed `OidPost` -> `OidCallback` and `SamlPost` -> `SamlCallback` on the
  controller for symmetry; the `[HttpGet]`/`[HttpPost]` route attribute
  literals are untouched, so the wire contract does not change. Also
  renamed the SAML flow service's `Post` -> `Callback` to match, since the
  OIDC sibling flow service already reads `CallbackAsync` after #501 (the
  extraction hadn't carried the naming fix down into the new tier).
- **Dead SAML accessor**, still applicable, and not touched by the flow
  extraction (`Saml.cs` is the hand-rolled core, untouched by #500/#501). A
  whole-solution grep confirmed `SamlResponse.GetCustomAttribute` (singular)
  had zero production callers, only the two test assertions the issue
  called out. Deleted it and retargeted those two tests onto the plural
  `GetCustomAttributes` (already the only form production code uses,
  including one call the retired accessor's plural sibling always returns
  a list, never null, so the retargeted "unknown attribute" test now
  asserts an empty list rather than null). Also hoisted the four repeated
  `"Role"` attribute-name literals in `SamlLoginService` (where the RBAC
  check moved to in #501) into one `RoleAttributeName` constant.
- **IP-classifier home**, still applicable. #503 moved the rate-limit
  *gate orchestration* into `Api/Shared/SsoRateLimitGate`, but the
  classifier predicates (`IsBlockedAddress`, `TryExtractEmbeddedIPv4`)
  stayed in the avatar-specific `AvatarUrlValidator.cs` even though
  `SsoRateLimiter.NormalizeClientKey` depends on them too, tuning the
  avatar SSRF policy could still silently change which clients the login
  rate limiter exempts. Extracted both into a new, neutrally named
  `IpAddressClassifier`; the method bodies moved verbatim (diffed against
  origin/main with comments stripped to confirm byte-identical logic), so
  this is a pure relocation. Split `AvatarUrlValidatorTests` accordingly
  into `AvatarUrlValidatorTests` (the SSRF URL gate) and a new
  `IpAddressClassifierTests` (the address-range classification).
- **Asset casing**, still applicable. Renamed `Views/apiClient.js` to
  `Views/ApiClient.js` so the on-disk file matches the name it is already
  published and imported under (`SSOPlugin.GetViews`'s `"ApiClient.js"`
  registration, `linking.html`'s `import("./ApiClient.js")`) instead of
  only agreeing case-insensitively on Windows. Updated the csproj
  `EmbeddedResource`/`None Remove` entries and the manifest-pinning test
  (`SSOPluginPageManifestTests`) to the corrected resource path. Left the
  dual publication of `Config.style.css` under two different names
  (`SSO-Auth.css` via `GetPages`, `style.css` via `GetViews`) exactly as it
  was, that is two unrelated consumers (the admin config page vs. the
  public linking page), not a casing bug, and added a comment on the
  `GetPages`/`GetViews` tables documenting them as the plugin's public URL
  contract, per the issue's ask.

Also recorded the `Oidc`-over-`Oid` naming convention for new OpenID types
in `ARCHITECTURE.md` (the issue's "worth adopting" aside): no mass rename
of existing serialized config field names or routes, just a convention
note for new types going forward.

## Type of change

## Security checklist

This touches the login path (`SSOController`, the SAML/OIDC flow services,
`AvatarUrlValidator`, `SsoRateLimiter`) but every change is either a pure
identifier rename or a verified byte-identical code relocation - nothing
here changes runtime behavior, so the full four-lens adversarial gate
wasn't run; each item was instead verified directly:

- [x] Fail-closed preserved: no validation, signature, time-bound, or
      audience logic changed anywhere in this diff.
- [x] No secrets logged; nothing here touches logging or config export.
- [x] Light self-review performed in place of the full gate, per the brief:
      route attribute literals diffed unchanged; the extracted
      `IpAddressClassifier` body diffed comment-stripped against
      origin/main and found byte-identical; the deleted accessor confirmed
      dead via a whole-solution grep before removal.
- [x] N/A, no new log call sites.

## Quality checklist

- [x] No duplicated logic; the `Role` literal and the IP classifier each
      now have exactly one definition.
- [x] The change adds no more code than the problem requires (net +302/-247
      across 23 files, mostly test-file mechanical renames).
- [x] The code is self-documenting; the misleading "called `Post` for
      consistency" comment is gone because the name is now accurate.
- [x] Architecture-conformance tests pass unchanged; no new structural
      property is introduced (the extracted classifier doesn't match any
      `HelperSuffixes` entry, so no existing rule applies to it, and none
      needed adding).

## Verification

- [x] `dotnet build -c Debug --warnaserror` is green (0 warnings).
- [x] `dotnet build -c Release --warnaserror` is green (0 warnings).
- [x] `dotnet test` is green: 1040/1040 passed.
- [x] Prettier: no new failures. The touched web asset (`Views/ApiClient.js`)
      and `ARCHITECTURE.md`/`CONTRIBUTING.md` are part of a pre-existing
      baseline `--check` warning set (confirmed via `git stash` against
      origin/main — the same file set, modulo the renamed file, already
      warns before this change) unrelated to this PR; not introduced here.

## Notes

Scope was kept to the four items plus the explicitly-optional naming-
convention note; no other identifiers were renamed (e.g. the
`SSOControllerOidPostTests`/`SSOControllerSamlPostTests` test class and
method names deliberately keep their historical `...Post...` spelling to
keep this diff focused on the production action-method rename itself, a
follow-up could rename them too if desired).


iderex (commented):
My sign-off — contributor-clarity sweep (#318 step 7, Closes #370), behaviour-preserving. Delivered by a delegated agent. All four items applied (none obviated by the #500/#501/#503/#504 refactors): (1) callback naming — OidPost→OidCallback, SamlPost→SamlCallback, SamlLoginService.Post→Callback for symmetry (route attribute literals + wire contract untouched — identifier-only); (2) dead SAML accessor — SamlResponse.GetCustomAttribute (singular) confirmed zero production callers via whole-solution grep and deleted, tests retargeted onto the plural, the four Role literals hoisted to one RoleAttributeName const; (3) IP-classifier home — IsBlockedAddress/TryExtractEmbeddedIPv4 extracted from AvatarUrlValidator into a new IpAddressClassifier.cs as a verified byte-identical move (SsoRateLimiter.NormalizeClientKey depends on them), tests split; (4) asset casing — Views/apiClient.js→ApiClient.js to match the already-published/imported name, csproj + the SSOPluginPageManifestTests pin updated (the manifest test would fail on any path mismatch — it passes, so the embedded-resource load is intact). Oidc-over-Oid naming convention recorded in ARCHITECTURE.md. Nothing behavioural changed; 1040 tests; CodeQL + required checks green; mergeable clean. Ready to merge.

PR #513, opened 2026-07-18
Throttle the pending-legacy-link warning through a shared IntervalGate

### LINKED ISSUE #362: Bound the pending-legacy-link warning to a rate-limited signal in the #318 flow service
Raised by CodeRabbit on #358 (CWE-400, log-volume) and by the availability lens / F4 there. The terminal warnings for a pending legacy link (`CanonicalLinkService`, the `CreateNewAccount`-orphan and `RejectNameTaken`-migratable branches) are emitted once per matching login, not deduplicated.

Why it was NOT bounded in #358 (documented decline, Lead-Dev call): the reachability is already bounded, the warning fires only when a legacy link exists for the presented name (ceiling = count of un-migrated legacy users, not arbitrary attacker-chosen names), each attempt requires a redeemed one-time OIDC state behind the anonymous-endpoint rate limiter, and the whole condition is transient (it disappears once the admin migrates). Two reviewers accepted the bounded residual. The clean fix needs a process-wide, concurrency-safe throttle, but `CanonicalLinkService` is constructed per request by the controller, so bolting static/singleton state onto it now would leak across the xUnit process and make the log-assertion tests order-dependent.

The right home is the #318 migration, where link resolution becomes a **singleton flow service**. At that point a shared `IntervalGate` (the repo's sanctioned once-per-interval log-flood primitive, #246/#195), or a bounded-cardinality counter, can gate the terminal warnings for free, with a deterministic clock injected so the throttle itself is unit-testable.

Scope:
- When link resolution is a singleton, inject a shared `IntervalGate` (+ an injectable clock) and gate the two terminal legacy-link warnings through it.
- Preserve per-account identification as far as the primitive allows (prefer a bounded-cardinality "who is still pending" signal over a single global once-per-interval line, so triage keeps naming users).
- Add a throttle test with a fake clock.

Refs #354, #358, #318

iderex:
# What & why

The two terminal pending-legacy-link warnings in `CanonicalLinkService` - the
`CreateNewAccount`-orphan branch and the `RejectNameTaken`-migratable branch - 
were emitted once per matching login. During a migration/upgrade window a hot
login loop for a not-yet-migrated user re-emits the same line on every attempt,
so the log volume is bounded only by the anonymous-endpoint rate limiter
(CWE-400, log-volume; raised by CodeRabbit and the availability lens on #358).

We route both warnings through the repo's sanctioned once-per-interval log-flood
primitive (`IntervalGate`, #246/#195), so a login loop is bounded to one line per
interval instead of one per attempt. The gate is a **process-wide static** field
with a one-minute interval matching the sibling cap-warn gates
(`OidcStateStore` / `SamlRequestCache`): the service is constructed per request
by the controller, so an instance gate would reset every login and throttle
nothing. Both the gate and the clock are injectable, so the throttle is
unit-testable with a fake clock and stays isolated per test case.

Only the warning FREQUENCY changes. The refusal throw and the fresh-account
creation still run on every login, the migration/adoption resolution is
untouched, and the inline log-forging sanitizer (`ReplaceLineEndings`) stays at
each call site.

Closes #362

## Type of change

- [ ] Bug fix
- [ ] Feature

(Hardening of a login-path log signal; no behaviour change beyond log frequency.)

## Security checklist

- [x] Fail-closed preserved: the `RejectNameTaken` throw and the account
      resolution are unconditional; the gate wraps only the `LogWarning` call.
- [x] No secrets logged; secrets are redacted on config export. (Unchanged - the
      warnings log only username/provider, already sanitized.)
- [x] A security review was performed - the four-lens adversarial gate, results
      under Notes.
      (`username?.ReplaceLineEndings(...)`, `provider?.ReplaceLineEndings(...)`
      unchanged, inside the gated block).

## Quality checklist

- [x] No duplicated logic; the throttle reuses the existing `IntervalGate`
      primitive.
- [x] Architecture-conformance tests pass; the new static-gate property is locked
      in as a rule in `ArchitectureConformanceTests`.

## Verification

- [x] `dotnet test` is green (1042 passed).
- [x] `scripts/check-jellyfin-compat.sh` consistent.

## Notes

**Scope.** Only `CanonicalLinkService.cs` (+ its tests + one conformance rule).
The ordinary #95 name-collision warning (the non-legacy `else` arm of
`RejectNameTaken`) is deliberately left unthrottled, it is a different signal
and out of this issue's scope.

**Adversarial self-review (login-path linking), four refute-by-default lenses:**

- **Availability / mass-lockout, PASS.** The gate changes log frequency only; no
  login is newly blocked or allowed, so there is no lockout. The first occurrence
  always enters (fresh cursor), and a second line surfaces each interval
  (pinned by the test), so detection of the first event and periodic triage are
  preserved. Documented residual: with a single shared gate, when several
  distinct users are pending in the same minute only one is named per interval - 
  a triage-completeness reduction the issue explicitly accepts ("as far as the
  primitive allows") and the task specifies ("emitted at most once per
  interval"). The bounded-cardinality "who is still pending" counter the issue
  floats as the richer alternative can be a later opportunistic follow-up.
- **Security-bypass / fail-open, PASS.** The gate wraps only `LogWarning`. The
  `throw new AccountLinkForbiddenException()` and the account provisioning sit
  outside the gated block and run on every login; the `&&` short-circuit in the
  orphan branch skips only the log, never a side effect. Migration/adoption,
  config locking, and the admin/adoption refusal audit lines are untouched. The
  throttle strictly improves posture (removes a log-flood DoS) without weakening
  the 403.
- **Correctness, PASS.** `IntervalGate.TryEnter` enters on the first call
  (cursor `MinValue`), throttles within the interval, and re-anchors past it. The
  clock is consulted only when `legacyLink.HasValue`, so no gate slot is consumed
  on the ordinary-collision path. The new test pins: three refused logins within
  one minute emit one warning; a fourth just past the interval emits the second.
- **Integration, PASS.** The static gate is process-wide, thread-safe
  (Interlocked CAS), and shared across per-request service instances via the
  4-arg production constructor (SSOController's call site is unchanged, new
  params are optional). Tests inject a fresh gate + fake clock, so the static
  cursor never leaks across cases. The inline log-forging sanitizers are intact.
  A conformance rule pins the gate as static so a refactor cannot silently demote
  it to an instance field (which would compile and pass the unit tests, since
  those inject a fresh gate) and reopen the flood.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -185,6 +185,24 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void CanonicalLinkService_ThrottlesTheLegacyLinkWarningThroughAStaticIntervalGate()
+    {
+        // Locked in by #362 (CWE-400, log-volume): the terminal pending-legacy-link warnings live in a
+        // service the controller constructs PER REQUEST, so the once-per-interval throttle must be a
+        // PROCESS-WIDE (static) IntervalGate — an instance field would reset every login and throttle
+        // nothing, letting a hot login loop for a not-yet-migrated user flood the log. Pin the static gate
+        // so a later refactor cannot silently demote it to an instance field (which compiles and passes the
+        // unit tests, because those inject a fresh gate) and reopen the flood.
+        var staticGate = typeof(CanonicalLinkService)
+            .GetFields(BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.DeclaredOnly)
+            .Any(f => f.FieldType == typeof(IntervalGate));
+
+        Assert.True(
+            staticGate,
+            "CanonicalLinkService must throttle its pending-legacy-link warning through a static IntervalGate (#362), because the service is constructed per request and an instance gate would throttle nothing.");
+    }
+
+    [Fact]
     public void AuthorizeStates_AreImmutableVariants()
     {
         // Locked in by #341: the in-flight OpenID authorize state is a CLOSED, IMMUTABLE sum — an
--- a/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
@@ -40,7 +40,12 @@ public class CanonicalLinkServiceTests
         var store = new ProviderConfigStore(() => cfg, _ => { }, new CapturingLogger());
         var users = Substitute.For<IUserManager>();
         var log = new CapturingLogger();
-        var service = new CanonicalLinkService(users, new FakeCryptoProvider(), store, log);
+
+        // Inject a FRESH legacy-link-warning gate per service so the process-wide static gate (#362) never
+        // leaks its cursor across cases: a fresh gate starts at MinValue, so the single warning each of the
+        // legacy-link tests provokes always enters. The dedicated throttle test below drives its own gate +
+        // fake clock to pin the once-per-interval bound.
+        var service = new CanonicalLinkService(users, new FakeCryptoProvider(), store, log, new IntervalGate(TimeSpan.FromMinutes(1)));
         return (service, cfg, users, log);
     }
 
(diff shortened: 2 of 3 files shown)


iderex (commented):
My sign-off — login-path linking hardening (Closes #362), behaviour-preserving except the log throttle. Delivered by a delegated agent; four-lens review all PASS. A process-wide static readonly IntervalGate (1-minute) on CanonicalLinkService now gates both terminal pending-legacy-link warnings (the CreateNewAccount-orphan branch and the RejectNameTaken-migratable branch) so a hot login loop for a not-yet-migrated user no longer floods the logs; the gate + a Func<DateTime> clock are optional ctor params (default = shared static gate + UtcNow) so the SSOController call site is untouched and tests inject a fresh gate + fake clock. Preserved: the AccountLinkForbiddenException throw and fresh-account provisioning are unconditional (outside the gated block), migration/adoption/config-locking unchanged, the inline ReplaceLineEndings sanitizer stays at each call site; the && short-circuit gates only LogWarning. The ordinary #95 name-collision warning is intentionally left unthrottled (out of scope). New throttle test (3 refusals in a minute → 1 warning; a 4th past the interval → a 2nd) + conformance rule CanonicalLinkService_ThrottlesTheLegacyLinkWarningThroughAStaticIntervalGate (pins the gate static so a refactor can't demote it to an instance field and reopen the flood). 1042 tests; CodeQL + required checks green; mergeable clean. Documented residual (single gate names one pending user per interval) accepted per the issue. Ready to merge.

PR #514, opened 2026-07-18
Rate-limit the authenticated link/unlink admin endpoints

### LINKED ISSUE #382: Rate-limit the authenticated link/unlink admin endpoints

iderex:
# What & why

The authenticated link/unlink admin write surface, `AddCanonicalLink` (POST
`{mode}/Link/{provider}/{jellyfinUserId}`) and `DeleteCanonicalLink` (DELETE
`{mode}/Link/{provider}/{jellyfinUserId}/{canonicalName}`), was
`AssertCanUpdateUser`-gated but had no rate limit. An authorized caller could
force serialized config-XML disk writes at will: each name-miss DELETE still
runs a full persist under the global config lock. It was the one write surface
with no throttle (raised by review on #373).

We front both endpoints with the shared per-client gate `SsoRateLimitGate.Check`
under a distinct `"link"` endpoint class, placed **after** the
`AssertCanUpdateUser` 403 guard so the auth decision stays first (no rate-limit
enumeration oracle) and the `"link"` budget is independent of the anonymous
login classes (`challenge`/`callback`/`auth`), which are byte-for-byte
unaffected. Over-limit renders the same `LoginOutcome.Throttled` 429 +
`Retry-After` the login path uses (#474). Rate limiting stays opt-in (default
off); the default 30/60s budget does not throttle real interactive admin use,
and the fail-open IP classifier still exempts private/proxied peers, so a
reverse-proxy deployment cannot be mass-locked. The gate's log line and doc
comments (and the ARCHITECTURE.md store table) were generalized from "anonymous
login endpoints" to cover the admin link surface too.

Closes #382

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the auth check (403) runs before the limiter, and
      the limiter's fail-open is availability-only, after authorization — no auth
      is ever default-accepted. Over-limit yields 429; the mapper throws rather
      than default-accepting on a wiring fault.
- [x] No secrets logged; the throttle signal carries only an aggregate count, no
      client key.
- [x] A security review was performed (four adversarial lenses, see Notes).
- [x] Log inputs from the IdP are sanitized inline at the log call. (No new IdP
      input is logged; the added log line carries only an integer count.)

## Quality checklist

- [x] No duplicated logic; the existing shared gate is reused, no new limiter.
- [x] The code is self-documenting and matches the target architecture (#318):
      the controller keeps no rate-limit state; the gate owns the whole check.
- [x] Architecture-conformance tests pass. No new structural property is
      established (the gate and its no-mutable-static-on-controller rule already
      exist); the reuse conforms to them.

## Verification

- [x] `dotnet test` is green (1044 passed, +4 new).
- [x] Prettier is clean for the `.md` change.

## Notes

Adversarial-review gate, four refute-by-default lenses, all resolved:

- **Availability / mass-lockout, PASS.** Default-off; `"link"` is a distinctly
  keyed bucket that cannot steal the login budgets; the fail-open classifier
  (private/proxy/CGNAT → null) keeps reverse-proxy deployments un-throttled; the
  403 guard runs first so only authorized callers are ever throttled.
- **Security-bypass / fail-open, PASS.** The `AssertCanUpdateUser` 403 fires
  before the limiter on both endpoints (unauthorized → 403, never 429), the
  `"link"` bucket is key-isolated from the login classes, and every limiter
  fail-open path is availability-only after authorization, no auth bypass, no
  enumeration oracle, no cross-bucket exhaustion.
- **Correctness, PASS.** The `is { } throttled` null-pattern proceeds on null
  and returns the 429 otherwise; over-limit → 429, under-limit → normal; the new
  tests use genuinely public (throttleable) IPs unique to the file and fail for
  the right reason under a regression.
- **Integration, NEEDS_IMPROVEMENT, resolved.** The lens flagged an incomplete
  reword: the gate's class summary + field comment and the ARCHITECTURE.md store
  row still said "anonymous". Fixed in this PR (all three generalized). The
  shared gate is reused with an unchanged signature; the login buckets are
  byte-for-byte unaffected; both endpoints are covered.

Out of scope (one line each, not addressed here):

- The `Unregister` endpoint (a heavier, admin-elevation-only config write) is
  also un-throttled; worth a follow-up hardening pass, outside #382's link/unlink
  scope.
- The throttled body reuses the shared "Too many login attempts…" wording on a
  non-login admin surface, cosmetic, inherited from the single mapper (#474),
  not a leak; a copy tweak could follow.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -92,7 +92,7 @@ only way to get one process-wide instance):
 | `SamlReplayCache`    | `SamlLoginService` | one-time-use SAML assertion IDs (replay protection)                                         |
 | `SamlRequestCache`   | `SamlLoginService` | outstanding SAML `AuthnRequest` IDs for `InResponseTo` correlation                          |
 | `OidcDiscoveryCache` | `OidcLoginService` | per-discovery-URL PKCE-S256 / RFC 9207 `iss` facts, 15-minute TTL; owns the fetch/parse too |
-| `SsoRateLimiter`     | `SsoRateLimitGate` | opt-in per-client rate limiting on the anonymous login endpoints                            |
+| `SsoRateLimiter`     | `SsoRateLimitGate` | opt-in per-client rate limiting on the login endpoints and the link/unlink admin surface    |
 
 The controller itself now holds **no mutable static state** — every store above
 lives in a flow service or the Shared tier, pinned by
--- a/SSO-Auth.Tests/SSOControllerLinkTests.cs
+++ b/SSO-Auth.Tests/SSOControllerLinkTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Net;
 using System.Threading.Tasks;
 using Duende.IdentityModel.OidcClient;
 using Jellyfin.Data;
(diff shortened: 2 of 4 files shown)


iderex:
My sign-off, security hardening (Closes #382). Rate-limits the admin link/unlink endpoints (AddCanonicalLink POST, DeleteCanonicalLink DELETE) via the shared SsoRateLimitGate under a new independent 'link' bucket, AFTER the AssertCanUpdateUser 403 (auth first, no enumeration oracle); over-limit maps to the login path's 429 + Retry-After; default-off, the 30/60s budget does not throttle real interactive admin use, and the fail-open IP classifier still exempts private/proxied peers (no reverse-proxy mass-lockout); the anonymous login classes (challenge/callback/auth) are byte-for-byte unaffected. Four-lens review all PASS (integration NEEDS_IMPROVEMENT for an incomplete 'anonymous' reword in the gate summary/comment/ARCHITECTURE.md, all generalized, resolved). 1044 tests; build/prettier/dependency-review/CodeQL green; mergeable clean.

PR #515, opened 2026-07-18
Support ECDSA service-provider signing keys for outgoing SAML requests

### LINKED ISSUE #493: Support ECDSA SP signing keys for outgoing SAML requests (currently RSA-only)
Noticed while delivering #167 (outgoing SAML AuthnRequest signing, PR #492). The new `SamlRedirectSigner` is **RSA-only**: a non-RSA (e.g. ECDSA) SP signing key provided as a PKCS#12 is rejected as invalid rather than used to sign. This is a documented limitation, not a defect, RSA-SHA256 is the common, safe default and what the inbound allowlist centers on.

## Proposal
If SP operators want ECDSA signing keys, extend the signer to detect the key type and select the matching SigAlg (e.g. ecdsa-sha256), but ONLY if it is added to the signature-method allowlist (`SamlSignatureAlgorithms`) that both the inbound and outbound paths enforce. No SHA-1, no weakening the allowlist.

## Acceptance
- [ ] Decision recorded: support ECDSA SP keys, or document RSA-only as intentional.
- [ ] If supported: ECDSA SigAlg added to the allowlist; signer selects it by key type; a signed-request round-trip test with an ECDSA key; RSA path unchanged.

Follow-up to #167. Low priority unless an operator needs it.

iderex:
# What & why

The outgoing SAML AuthnRequest redirect-binding signer (#167) was RSA-only: an
operator who supplied an ECDSA PKCS#12 as the service-provider signing key had
it rejected as invalid at the admin write path rather than used to sign. This
extends the signer to detect the SP key type and sign with the matching,
already-allowlisted algorithm, an additive change that leaves the RSA path
byte-identical.

- `SamlRedirectSigner.BuildSignedRedirectUrl` now takes the SP private key as
  `AsymmetricAlgorithm` and derives the `SigAlg` from the key's own runtime
  type: RSA → `rsa-sha256` (the exact prior call, byte-identical), ECDSA →
  `ecdsa-sha256`. A key that is neither is rejected with an
  `InvalidOperationException`, never silently left unsigned or signed weakly.
- The emitted `SigAlg` is still re-checked against the shared inbound/outbound
  allowlist (`SamlSignatureAlgorithms.IsSignatureMethodAllowed`) before signing,
  so no SHA-1 can slip in, the check now validates the *actual* emitted value.
- ECDSA signs with the IEEE P1363 (r‖s) signature format XML-DSig mandates for
  the redirect binding, not .NET's default DER sequence, so the identity
  provider verifies over the exact octets transmitted.
- `SamlSigningKey.GetSigningKey` surfaces the RSA-or-ECDSA private key from the
  certificate (RSA-preferred, ECDSA fallback, null otherwise); `IsInvalid` now
  accepts an ECDSA key so an operator can save one.
- `ecdsa-sha256` was already on the `SamlSignatureAlgorithms` allowlist (added
  with the inbound hardening), so no allowlist source change was needed.

Closes #493

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. (Signing-on with an unusable or
      unsupported key still throws; the only unsigned path is the unchanged
      opt-out `SignAuthnRequests = false`.)
- [x] No secrets logged; secrets are redacted on config export. (The signer has
      no logging; the sole diagnostic string emits a key *type* name, never key
      material.)
      config persistence, or the release pipeline. (Four-lens adversarial review
      below — all PASS.)
      logging added.)

## Quality checklist

      testable unit. (Key-type → SigAlg and key-type → signing are two small
      switch expressions sharing one `Unsupported` factory.)
      (No new structural property; existing rules pass.)

## Verification

      0 warnings.)
- [x] `dotnet test` is green. (1050/1050 passing, including a full ECDSA
      PKCS#12 load → key-detect → IEEE-P1363 sign → verify round-trip.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (N/A - 
      only `.cs` files changed.)

## Notes

**Scope.** The core change is in `SamlRedirectSigner.cs` and `SamlSigningKey.cs`.
Two thin plumbing edits were unavoidable for the feature to reach production
rather than be dead code: the outbound wrapper `SamlAuthnRequest.GetSignedRedirectUrl`
(`Saml.cs`) and the challenge builder `SamlLoginService.BuildChallengeRedirectUrl`
widen the key parameter from `RSA` to `AsymmetricAlgorithm` and select the key
via `GetSigningKey`. Neither is the inbound SAML validator; both are outbound-only
signing code. `SamlSignatureAlgorithms.cs` needed no change (the ECDSA entries
already exist).

**RSA byte-identical.** For an RSA key the resolved `SigAlg` is the same constant,
the signed octets are the same, and the signing call is `rsa.SignData(data,
SHA256, Pkcs1)` as before. The existing #167 RSA tests are untouched and pass.

**Adversarial review, four refute-by-default lenses, all PASS:**

- **Availability / mass-lockout, PASS.** Purely additive, monotonic widening.
  RSA path byte-identical; signing stays strictly opt-in; the only behavior
  change is *accepting* ECDSA keys that were previously unpersistable, so no
  existing RSA deployment can regress or be locked out.
- **Security-bypass / fail-open, PASS.** `SigAlg` is derived from the key type
  (two hard-coded constants), re-checked against the allowlist; no code path
  yields a `*-sha1` URI; unsupported/keyless input throws (never silent
  degrade); signed octets equal transmitted bytes; ECDSA uses IEEE P1363; no key
  material leaked.
- **Correctness, PASS.** ECDSA format is IEEE P1363 (empirically pinned, the
  verifier uses the same format, so a DER signer would fail the positive
  assert); RSA branch statement-for-statement identical; `GetSigningKey`
  null-safe RSA-then-ECDSA fallback with correct caller-owned disposal
  (key-before-cert); exhaustive fail-closed switches; tests are non-vacuous.
- **Integration, PASS.** `ECDsa.SignData(…, DSASignatureFormat)` and
  `IeeeP1363FixedFieldConcatenation` are BCL APIs present on net9.0 (TFM-bound,
  unrelated to the 10.11 ABI floor); the widened `AsymmetricAlgorithm` signature
  is source-compatible for all existing RSA callers; the admin-write path
  accepts ECDSA with no change to the forbidden `SSOController`; build + full
  suite pass empirically on Windows.

**Out-of-scope follow-up.** The operator-facing validation error strings in
`ProviderConfigValidator.cs` and `SSOController.cs` still say "RSA private key"
though an ECDSA key is now accepted. Cosmetic; one lives in `SSOController` (owned
by a parallel PR), so both are left for a small follow-up rather than partially
fixed here.

### DIFF
--- a/SSO-Auth.Tests/SamlAuthnRequestTests.cs
+++ b/SSO-Auth.Tests/SamlAuthnRequestTests.cs
@@ -47,6 +47,20 @@ public class SamlAuthnRequestTests
     }
 
     [Fact]
+    public void GetSignedRedirectUrl_EcdsaKey_CarriesAVerifiableEcdsaSignature()
+    {
+        var request = new SamlAuthnRequest("jellyfin-sp", "https://jellyfin.example.com/sso/SAML/p/adfs");
+        using var certificate = SamlSigningKeyFactory.CreateEcdsaCertificate();
+        using var privateKey = certificate.GetECDsaPrivateKey()!;
+
+        var url = request.GetSignedRedirectUrl(Endpoint, relayState: null, privateKey);
+
+        Assert.Contains("SigAlg=" + Uri.EscapeDataString("http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256"), url);
+        using var publicKey = certificate.GetECDsaPublicKey()!;
+        Assert.True(EcdsaSignatureVerifies(url, publicKey));
+    }
+
+    [Fact]
     public void GetSignedRedirectUrl_NullEndpoint_Throws()
     {
         var request = new SamlAuthnRequest("jellyfin-sp", "https://jellyfin.example.com/sso/SAML/p/adfs");
@@ -64,6 +78,16 @@ public class SamlAuthnRequestTests
 
         var signedQuery = "SAMLRequest=" + Uri.EscapeDataString(samlRequest) + "&SigAlg=" + Uri.EscapeDataString(sigAlg);
         return publicKey.VerifyData(Encoding.UTF8.GetBytes(signedQuery), signature, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+    }
+
+    private static bool EcdsaSignatureVerifies(string url, ECDsa publicKey)
+    {
+        var samlRequest = QueryValue(url, "SAMLRequest");
+        var sigAlg = QueryValue(url, "SigAlg");
+        var signature = Convert.FromBase64String(QueryValue(url, "Signature"));
+
+        var signedQuery = "SAMLRequest=" + Uri.EscapeDataString(samlRequest) + "&SigAlg=" + Uri.EscapeDataString(sigAlg);
+        return publicKey.VerifyData(Encoding.UTF8.GetBytes(signedQuery), signature, HashAlgorithmName.SHA256, DSASignatureFormat.IeeeP1363FixedFieldConcatenation);
     }
 
     private static string QueryValue(string url, string name)
--- a/SSO-Auth.Tests/SamlRedirectSignerTests.cs
+++ b/SSO-Auth.Tests/SamlRedirectSignerTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using Jellyfin.Plugin.SSO_Auth;
 using Jellyfin.Plugin.SSO_Auth.Api;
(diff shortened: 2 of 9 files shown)


iderex:
My sign-off, SAML crypto feature (Closes #493). Adds ECDSA SP signing-key support to the outgoing AuthnRequest signer: key-type detection (RSA preferred, ECDSA fallback, else null), SigAlg selected by the loaded key's runtime type (RSA to rsa-sha256 unchanged, ECDSA to ecdsa-sha256), and ECDSA signed in the IEEE P1363 r||s form XML-DSig mandates (not .NET's default DER) so the IdP verifies over the transmitted octets. ecdsa-sha256 was already on the shared SamlSignatureAlgorithms allowlist and the emitted SigAlg is re-checked against it before signing (no SHA-1 path). RSA path byte-identical; an unsupported key type throws (never silent-unsigned); opt-in (SignAuthnRequests) unchanged. Tests include a full PKCS#12 load-detect-sign-verify round-trip with an ECDSA key, a foreign-key negative, DSA-rejected, and the allowlist. Four-lens review all PASS. 1050 tests; build/prettier/dependency-review/CodeQL green; mergeable clean.

PR #521, opened 2026-07-18
Extract a dedicated SAML assertion-validator type

### LINKED ISSUE #496: Extract a dedicated SAML assertion-validator type (home for FromValidatedSaml)
Noticed while delivering #473 (VerifiedIdentity). The SAML side of the new `VerifiedIdentity` keystone is built by a `FromValidatedSaml` factory called at the SAML mint endpoint after the full inbound validation (signature / time / audience / recipient / InResponseTo / allowlist / replay / non-empty NameID). Today that validation is spread across the SAML callback path rather than living in a single dedicated type.

## Proposal
As part of the #318 SAML partition, extract a dedicated SAML **assertion-validator** type (internal sealed) that owns the full inbound validation and is the single place that yields a validated SAML result, the natural home for the `FromValidatedSaml` factory call and its conformance allow-list. This makes the "VerifiedIdentity is only produced after complete validation" invariant local and testable, and pairs with #251 (the assertion is currently round-tripped/validated in a way worth consolidating).

Part of #318 (domain-partitioned helpers). Behaviour-preserving extraction; adversarial gate applies (login/crypto path).

iderex:
# What & why

Consolidate the inbound SAML validation, signature / time bounds / audience / recipient / algorithm-allowlist / one-time replay / non-empty NameID, which was spread across the SAML callback path in `SamlLoginService`, into one dedicated internal sealed type, `SamlAssertionValidator`. The validator owns the full inbound validation and is the single place that yields a validated SAML result, so it is the natural home for the `VerifiedIdentity.FromValidatedSaml` factory call. This makes the "a `VerifiedIdentity` is produced only after complete validation" invariant local and testable.

This is a behaviour-preserving relocation, not a logic change: every check runs in the same order with the same reject/accept outcome and the same error surface in all three legs (callback, authenticate, link). `Saml.cs` (the XSW/algorithm-allowlist/audience-AND/recipient/time core) is untouched.

Closes #496. Part of #318 (domain-partitioned helpers).

**What moved into `SamlAssertionValidator`:**
- `ValidateSaml` (signature + time bounds + AudienceRestriction), unchanged, still `internal static`.
- The recipient binding and the algorithm-diagnostic warning log (the old `IsSamlResponseValid` / `ExpectedAcsUrls`), now `TryValidate` / `IsResponseValid`.
- The one-time replay consume (`TryConsumeReplay`) together with the process-wide `SamlReplayCache` static.
- The non-empty-NameID guard and the sole SAML `VerifiedIdentity.FromValidatedSaml` construction, in `TryProduceVerifiedIdentity`, downstream of every gate.
- The role-attribute name is one private const; the flow service reads roles through `GetAssertionRoles` so every read agrees on the attribute.

**What deliberately stayed in `SamlLoginService`:**
- The InResponseTo correlation and browser binding (the outstanding-request cache), request-issued session-fixation correlation, not assertion validation.
- The login allow-list (`SamlLoginPolicy`), a policy gate enforced at BOTH the assertion-consumer page and the session-minting endpoint.

**Check order + outcomes (authenticate leg), identical to before:** config → parse + signature/time/audience/recipient → InResponseTo/binding → login allow-list → replay consume → non-empty NameID → `FromValidatedSaml` → shared completion. `Rejected(SamlResponseInvalid)` for malformed/invalid/replay, `Denied` for role-fail and empty-NameID, unchanged.

**Conformance rules re-homed:** `VerifiedIdentity_IsConstructedOnlyByProtocolValidators` now pins the `FromValidatedSaml` call to `SamlAssertionValidator`; `Controller_DelegatesSamlFlowToTheFlowService`'s liveness scan now covers both flow-tier files (`SamlLoginService` + `SamlAssertionValidator`).

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green.

## Notes

Built Debug + Release with `--warnaserror` (0 warnings) and ran the full suite (1059 tests, all pass) on the rebased tree (branch rebased cleanly onto current `origin/main` @ eddfd26; the four files touched here are disjoint from #519/#520).

**Adversarial self-review, four refute-by-default lenses, all PASS:**
- **Availability / mass-lockout, PASS.** No new mass-lockout: the replay cache stays a single process-wide `static readonly` (survives the two-step post→authenticate legs); no validation reordered to reject a previously-passing login; the base URL passed into validation is identical to the old in-validation computation.
- **Security-bypass / fail-open, PASS.** No check dropped, weakened, or reordered-to-change-outcome; `FromValidatedSaml` is reachable only after the full gate chain; the one-time replay consume is intact and still shared by authenticate + link; `Saml.cs` (algorithm allowlist, XSW, audience-AND, recipient) untouched.
- **Correctness / check-for-check equivalence, PASS.** Line-by-line equivalence confirmed across all three legs; derived-state build, role reads, and warning logs preserved; `requestBase` computed once and reused.
- **Integration, PASS.** `SamlLoginService` constructor signature unchanged (controller wiring and test `Build` untouched); both conformance rules correctly re-homed and green; `ValidateSamlTests` updated to the new type with the now-unused `using` dropped; the new `*Validator` type satisfies the internal-and-sealed helper rules.

**Non-blocking behavioural note (availability + correctness lenses):** `GetRequestBase` is now computed eagerly before validation, so on an already-broken (unpersistable) `baseUrlOverride` config an invalid SAML response would surface as a 500 instead of a 400. This is strictly fail-closed (never fail-open) and unreachable through supported config, the config-save validation blocks a `baseUrlOverride` that would make base-URL resolution throw. Recorded here for transparency; no fix warranted.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -279,9 +279,9 @@ public class ArchitectureConformanceTests
         // Sentinel + call-site pins. The factory NAMES are the contract; require both to exist (a rename
         // must consciously update this rule), then confine each factory's invocation to the file(s) that own
         // its protocol's validation. AuthorizeSession is where the OpenID identity is built (from the
-        // role-gate result); the SAML factory is invoked from the SAML flow service's session-minting leg
-        // (SamlLoginService), where it moved with the SAML flow extraction (#160, #318 step 13) — off the
-        // controller, as the earlier comment anticipated.
+        // role-gate result); the SAML factory is invoked from the dedicated SamlAssertionValidator, the
+        // single home the SAML inbound validation moved into (#496) — downstream of every gate, so the
+        // "constructed only after complete validation" invariant is local to the validator.
         const string oidcFactory = "FromOidcRedemption";
         const string samlFactory = "FromValidatedSaml";
         var factoryMethods = typeof(VerifiedIdentity)
@@ -294,9 +294,9 @@ public class ArchitectureConformanceTests
             $"VerifiedIdentity must expose the two named validation factories ({oidcFactory}, {samlFactory}); one was renamed, so update this rule and the source-scan allow-list with it (#473).");
 
         var oidcHome = SourceFilesDeclaring(new[] { typeof(AuthorizeSession) });
-        var samlHome = SourceFilesDeclaring(new[] { typeof(SamlLoginService) });
+        var samlHome = SourceFilesDeclaring(new[] { typeof(SamlAssertionValidator) });
         AssertFactoryInvocationsConfinedTo("VerifiedIdentity." + oidcFactory + "(", oidcHome, "the OpenID redeem path (AuthorizeSession.Ready)");
-        AssertFactoryInvocationsConfinedTo("VerifiedIdentity." + samlFactory + "(", samlHome, "the SAML session-minting leg (SamlLoginService)");
+        AssertFactoryInvocationsConfinedTo("VerifiedIdentity." + samlFactory + "(", samlHome, "the SAML assertion validator (SamlAssertionValidator)");
     }
 
     // Fails if the given factory-invocation token appears in any SSO-Auth source file outside the allowed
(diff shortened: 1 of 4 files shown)


iderex:
My sign-off, SAML crypto/validation refactor (Closes #496), behaviour-preserving. Delivered by a delegated agent; four-lens review all PASS, and the delivery-hygiene issue the lenses flagged was resolved (committed + cleanly rebased onto current origin/main @ eddfd26, commit beef03d, and re-tested on the rebased tree, 1059 pass). The full inbound SAML validation is extracted into a new internal sealed SamlAssertionValidator: ValidateSaml (signature/time/AudienceRestriction), the recipient binding + algorithm-diagnostic log, the one-time replay consume (the SamlReplayCache stays a single process-wide static), the non-empty-NameID guard, and the sole VerifiedIdentity.FromValidatedSaml construction, so the 'produced only after complete validation' invariant is now local to the validator (conformance rule re-homed). InResponseTo correlation + the login allow-list deliberately stay in SamlLoginService. Check order + fail-closed outcomes identical in all three legs (Rejected for malformed/invalid/replay, Denied for role-fail/empty-NameID); Saml.cs (allowlist/XSW/audience AND-semantics) untouched; controller wiring + tests unchanged. Documented non-blocking note: GetRequestBase now eager, surfacing an invalid SAML response as 500 vs 400 only on an unpersistable broken baseUrlOverride config (strictly fail-closed, unreachable via supported config). CodeQL + required checks green; mergeable clean.

PR #522, opened 2026-07-18
Add opt-in verified-email gate for OpenID logins

### LINKED ISSUE #166: Verified-email gate for OIDC logins
Optionally require email_verified=true (and a present email claim) before provisioning or linking an account, so an IdP that allows unverified emails cannot be used to hijack an account by email. Feature parity with V2.

Acceptance criteria:
- [ ] Opt-in per provider; when on, a login without email_verified=true is rejected (fail closed).
- [ ] The check is explicit and does not treat a missing claim as verified.
- [ ] Tests cover verified/unverified/missing.

iderex:
# What & why

Closes #166.

Adds an opt-in, per-provider verified-email gate for OpenID **logins**, feature
parity with the sibling `jellyfin-plugin-sso-V2 (private)`. Some identity
providers let a principal hold an account with an unverified email address; a
deployment that wants to keep such accounts out entirely could not express that
before this change (the existing `RequireVerifiedEmailForAdoption`, #218, only
guards the one moment a login _adopts_ a same-named account).

**Design**

- **New `OidConfig` flag `RequireVerifiedEmailForLogin`, default `false`.** Default
  off is the availability-critical property: an operator who never sets it, and
  an IdP that never emits `email_verified`, sees **zero** change on upgrade. The
  guard short-circuits on the flag, so with it off no new code path is taken.
- **Fail-closed when on.** In `OidcLoginService.AuthenticateAsync`, right after the
  one-time state redeem and **before** the shared completion tail (`CompleteAsync`,
  which resolves/creates the account and mints), a login whose `email_verified` is
  not exactly `true`, absent, `false`, or unparseable all map to `!= true`, is
  refused. Because the check precedes `CompleteAsync`, a refusal provisions no
  account and mints no session.
- **Gates the login itself, not just adoption.** It applies to every OpenID login
  for the provider (adopt, create, or sign in to an already-linked account) and is
  independent of `AllowExistingAccountLink` / `RequireVerifiedEmailForAdoption`. It
  deliberately does **not** touch the manual admin-link redeem (`Link`), which is a
  separately authorized action.
- **`email_verified` is reused, not re-parsed.** The check reads the single value
  `OidcAuthorizeStateBuilder.ResolveEmailVerified` already parsed and carried on the
  redeemed `VerifiedIdentity.EmailVerified`, so the login gate and the adoption gate
  can never disagree about the same claim. The value originates from the
  signature-validated id_token claims (#134), so there is no unvalidated source.
- **Refusal surface.** A new `PublicReason.EmailNotVerified` maps to `403` with a
  fixed, operator-facing body ("A verified email is required to log in."), mirroring
  the PKCE reasons rather than the uniform role denial, the failing condition is a
  provider/IdP configuration property, not an account-enumeration oracle. Not a 500.
- **Scope requirement documented.** Enabling it needs the `email` scope in
  `OidScopes`, otherwise the claim is absent and every login is refused. The flag is
  XML-only for now (mirrors how `RequireVerifiedEmailForAdoption` shipped before it
  was surfaced in the admin form by #484/#488); `providers.md` documents it.

**Also in this PR (small, in-scope doc cleanup):** the XML-doc comments on
`AllowExistingAccountLink` and `RequireVerifiedEmailForAdoption` still described those
flags as "XML-only", which went stale when #484/#488 surfaced them in the admin
provider form. Both are corrected to note they are settable in the admin form as well
as `config.xml`.

## Type of change

- [x] Feature
- [ ] Bug fix

(Feature with a security posture, a login-path gate, so the security checklist and
the adversarial-review gate below both apply.)

## Security checklist

- [x] Fail-closed preserved: with the flag on, a missing/`false`/unparseable
      `email_verified` is rejected (`!= true`), never default-accepted; the check is
      ordered before any account resolution or session mint.
- [x] No secrets logged; the refusal log line carries only the provider name,
      sanitized inline with `provider?.ReplaceLineEndings(string.Empty)`.
- [x] A security review was performed - the four-lens adversarial gate ran on the
      committed diff (verdicts under Notes).

## Quality checklist

- [x] No duplicated logic; the gate reuses the already-parsed `email_verified` value
      rather than re-parsing the claim.
- [x] The change adds no more code than the problem requires (one guard, one config
      flag, one `PublicReason` + mapper row).
- [x] The code is self-documenting; the comments explain _why_ (fail-closed, reuse,
      distinct from adoption), not _what_.
- [x] Architecture-conformance tests pass. The new flag is deliberately XML-only, so
      it is not added to the `securityCritical` form-field roster; the roster comment
      is clarified to say it covers form-surfaced toggles (an XML-only toggle stays out
      until it is surfaced, as `RequireVerifiedEmailForAdoption` was).

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green (Debug **and** Release, 0
      warnings), on the tree rebased onto current `main`.
- [x] `dotnet test` is green (1063 passed, 0 failed).
- [x] Jellyfin compatibility metadata check (`scripts/check-jellyfin-compat.sh`) is
      consistent.

## Notes

**Tests** (`SSO-Auth.Tests/SSOControllerOidAuthTests.cs`, driven end-to-end through the
thin `OidAuth` endpoint):

- flag **off** (default) + `email_verified` absent → login **succeeds** (the
  behaviour-preserving pin for the whole userbase);
- flag **on** + `email_verified == true` → **succeeds**;
- flag **on** + `email_verified` `false` **and** absent (Theory) → **rejected** with
  403 "A verified email is required to log in.", and **no account provisioned**.
- `LoginStatusMapperTests` totality table extended with the `EmailNotVerified` → 403
  row (the mapper totality invariant fails a member without an entry).

**Adversarial-review gate, four refute-by-default lenses on the committed diff:**

- **Availability / mass-lockout**, VERDICT: the gate is inert by default and
  upgrade-safe, a new `bool` XML property defaults `false`, so no existing provider
  flips on; when set, the lockout is scoped to the single opting-in OpenID provider,
  with a predicate/value byte-identical to the adoption gate. No availability defect., **PASS**
- **Security-bypass / fail-open**, VERDICT: default-off takes no new path; when on,
  `!= true` rejects absent/`false`/unparseable and only exact `true` passes; the value
  is the signature-validated id_token claim reused from the verified identity (no
  divergent parse, no unvalidated source); the check precedes the mint so nothing is
  provisioned on refusal; no alternate login endpoint bypasses it., **PASS**
- **Correctness**, VERDICT: the full on/off × true/false/absent matrix behaves as
  specified and is pinned by tests; `bool?` short-circuit correct; log line sanitized;
  the XML-only flag correctly stays out of the conformance roster., **PASS**
- **Integration**, VERDICT: the bool round-trips via the same XML/JSON path as the
  sibling flag and is not wiped by config-preservation; the new `PublicReason`
  satisfies the mapper totality invariant; the adoption gate (#218) and the SAML path
  are unaffected; no fitness function requires the flag to be a form field., **PASS**

All lens findings resolved (none required a code change). Out-of-scope follow-up:
surface `RequireVerifiedEmailForLogin` in the admin OpenID provider form (and add it to
the `securityCritical` roster then), mirroring #484/#488 for the adoption flag, a
separate P4 issue.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -871,7 +871,9 @@ public class ArchitectureConformanceTests
         // stop a field persisting while leaving the forward check green. For a security setting that is
         // fail-open (the server keeps the stored value; the admin can no longer harden it), so pin the
         // security-critical settings: each MUST remain a marked, correctly-typed persisting field. Extend
-        // this roster in the same PR that adds a new security toggle.
+        // this roster in the same PR that surfaces a new security toggle in the admin form; a deliberately
+        // XML-only toggle (e.g. RequireVerifiedEmailForLogin #166) is not a form field and so stays out of
+        // this roster until it is surfaced (as RequireVerifiedEmailForAdoption was, #484/#488).
         var securityCritical = new[]
         {
             "EnableAuthorization", "OidSecret", "DisableHttps", "DisablePushedAuthorization",
--- a/SSO-Auth.Tests/LoginStatusMapperTests.cs
+++ b/SSO-Auth.Tests/LoginStatusMapperTests.cs
@@ -29,6 +29,7 @@ public class LoginStatusMapperTests
             (PublicReason.SamlResponseInvalid, 400, "SAML response validation failed"),
             (PublicReason.PkceNotSupported, 400, "The identity provider does not advertise the required PKCE (S256) support."),
             (PublicReason.PkceUnverifiable, 400, "The identity provider's PKCE (S256) support could not be verified."),
+            (PublicReason.EmailNotVerified, 403, "A verified email is required to log in."),
         };
 
         // The table itself must stay total: a new member without a row fails here.
--- a/SSO-Auth.Tests/SSOControllerOidAuthTests.cs
+++ b/SSO-Auth.Tests/SSOControllerOidAuthTests.cs
@@ -143,16 +143,94 @@ public class SSOControllerOidAuthTests
         await harness.UserManager.Received(1).CreateUserAsync("alice");
     }
 
+    [Fact]
+    public async Task OidAuth_RequireVerifiedEmailForLoginOff_Succeeds_RegardlessOfEmailVerified()
+    {
+        // The availability-critical property (#166): with the flag off (the default), the login succeeds
+        // even when email_verified is absent (null) — behaviour is byte-identical to before the gate, so no
+        // existing deployment or claim-omitting IdP is affected on upgrade.
+        var harness = new SsoControllerHarness(c => c.OidConfigs["kc"] = new OidConfig
+        {
+            Enabled = true,
+            EnableAuthorization = false,
+            AllowExistingAccountLink = false,
+            RequireVerifiedEmailForLogin = false,
+        });
+        SeedValidState(harness, "state-token", emailVerified: null);
+        SetBindingCookie(harness, Binding);
+
+        var result = await harness.Controller.OidAuth("kc", Redeem("state-token"));
+
+        Assert.IsType<OkObjectResult>(result);
+        await harness.UserManager.Received(1).CreateUserAsync("alice");
+    }
+
+    [Fact]
+    public async Task OidAuth_RequireVerifiedEmailForLoginOn_EmailVerifiedTrue_Succeeds()
+    {
+        var harness = new SsoControllerHarness(c => c.OidConfigs["kc"] = new OidConfig
+        {
+            Enabled = true,
+            EnableAuthorization = false,
+            AllowExistingAccountLink = false,
+            RequireVerifiedEmailForLogin = true,
+        });
+        SeedValidState(harness, "state-token", emailVerified: true);
+        SetBindingCookie(harness, Binding);
+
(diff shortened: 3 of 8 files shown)


iderex:
My sign-off, OIDC login-path security feature (Closes #166). Delivered by a delegated agent; four-lens review all PASS, rebased cleanly onto current main (60f19d3) and re-tested (1063 pass). New opt-in OidConfig.RequireVerifiedEmailForLogin (XML-only), default false = the availability-critical property: an operator who never sets it (and an IdP omitting email_verified) takes no new code path and sees zero change on upgrade. When on: in OidcLoginService.AuthenticateAsync, after the one-time state redeem and BEFORE the shared mint tail, a login whose EmailVerified != true (absent/false/unparseable all reject; only exact true passes) is refused, so a refusal provisions no account and mints no session. Gate is on the login/session-mint leg only, not the manual admin-link redeem (distinct from the #218 adoption gate); it reuses the single signature-validated email_verified value (no divergent parse). Refusal → new PublicReason.EmailNotVerified → 403 (not 500), mapper totality satisfied. Also fixes the two stale OidConfig XML-doc comments (AllowExistingAccountLink/RequireVerifiedEmailForAdoption now note they are admin-form-settable since #484/#488). CodeQL + required checks green; mergeable clean.

PR #523, opened 2026-07-18
Derive a real device name for the SSO linking view

### LINKED ISSUE #310: Derive a real device name for the SSO linking view instead of "DUMMY"
Split out from #63.

The admin linking view (`SSO-Auth/Config/linking.html` -> `SSO-Auth/Views/apiClient.js`) constructs its `jellyfinApiclient.ApiClient` / `ConnectionManager` with a fixed placeholder device name `"DUMMY"` (`const deviceName = "DUMMY";`). As a result the session that view registers with the Jellyfin API shows up as device "DUMMY".

#63 deduplicated the device-name derivation onto the single source of truth in `WebResponse.cs` (the server-rendered auth-completion page) without changing behavior, and deliberately left this placeholder byte-identical. Actually deriving a real device name here is a behavior change (the session name flips from "DUMMY" to a real value), so it is tracked separately.

Options to weigh:
- Reuse the already-present `window.ApiClient` device name when the linking view runs inside the Jellyfin dashboard (it does for the normal config-page load), rather than re-deriving.
- Or share the browser-detection derivation from `WebResponse.cs` (larger change; that logic is currently inlined into the auth page's script string and not otherwise reusable).

Low priority / admin-only surface. Requires deciding the desired session-name value before it ships, because it changes what device sessions register under.

iderex:
# What & why

The SSO self-service linking view (`/SSOViews/linking`) registers its
Jellyfin session with a hard-coded placeholder device name, literally
`"DUMMY"` (`SSO-Auth/Views/ApiClient.js`), so the resulting entry in
Jellyfin's device list was meaningless. Closes #310.

Replaces the placeholder with the fixed, descriptive label
`"SSO Account Linking"`. We considered deriving the name from the
User-Agent (the way the server-rendered auth-completion page does in
`WebResponse.cs`), but that derivation is ~150 lines of inlined
browser sniffing that isn't reusable from this asset without a much
larger refactor, the issue itself flags that as the bigger option.
Given the linking view is an admin/user-management surface that opens
briefly rather than a regular login client, a fixed, stable, non-PII
label reads more clearly in the device list than a per-browser guess,
and it keeps this asset free of the extra dependency. We're flagging
the exact wording as a product call in case a different label is
preferred, happy to adjust.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

Not applicable, this touches only a display label on an
already-authenticated linking session; it does not touch signature
validation, config persistence, or the release pipeline.

## Quality checklist

      testable unit.
      change establishes is locked in as a rule in `ArchitectureConformanceTests`
      (no new structural property introduced — one constant value changed).

## Verification

- [x] `dotnet test` is green (1059/1059 passing).
- [x] Prettier is clean for the changed file (`ApiClient.js` unchanged by
      `--write`); the repo-wide `--check` warnings on other files are a
      pre-existing CRLF/baseline issue on `origin/main`, confirmed unrelated
      to this change (reproduced identically before this commit).

## Notes

This is browser-JS that CI cannot exercise at runtime, so review is the
primary verification: the linking flow is otherwise byte-identical - 
only the `deviceName` constant value passed into
`jellyfinApiclient.ApiClient` / `ConnectionManager` changed, from
`"DUMMY"` to `"SSO Account Linking"`. No other file touches or asserts
on that placeholder string.

### DIFF
--- a/SSO-Auth/Views/ApiClient.js
+++ b/SSO-Auth/Views/ApiClient.js
@@ -90,11 +90,11 @@ export async function serverAddress({ basePath = "/web" }) {
 }
 
 // The browser-to-device-name derivation is defined once, in the server-rendered auth-completion
-// page (WebResponse.cs). This linking view is not a login client, so it does not re-derive that
-// value; it registers under a fixed placeholder identifier. This is kept byte-identical to the
-// prior behavior on purpose - deriving a real device name here would change the name SSO-view
-// sessions register under, which is a separate, deliberate behavior change.
-const deviceName = "DUMMY";
+// page (WebResponse.cs), and is ~150 lines of inlined browser sniffing that is not reusable from
+// here without a much larger refactor (issue #310). This linking view only opens briefly to manage
+// provider links, so a fixed, descriptive, non-PII label is clearer in Jellyfin's device list than a
+// per-browser guess would be, and it keeps this asset dependency-free.
+const deviceName = "SSO Account Linking";
 
 function getDeviceId() {
   return localStorage.getItem("_deviceId2");


iderex:
My sign-off, web-UI behaviour-preserving (Closes #310). Delivered by a delegated agent. Replaces the hard-coded deviceName 'DUMMY' with a fixed descriptive label 'SSO Account Linking' in Views/ApiClient.js (the copy loaded exclusively by the linking view's import chain, verified no other caller), so the SSO self-service linking session shows a meaningful entry in Jellyfin's device list. A stable non-PII label was chosen over User-Agent derivation (that ~150-line browser-sniffing path from WebResponse.cs isn't reusable here without a larger refactor, the issue itself flagged it as a separate option); the exact wording is a product call I can revise. The linking flow is otherwise unchanged (same ApiClient/ConnectionManager construction, credentials, device-ID). 1059 tests green; required checks green. Ready to merge.

Issue #489, opened 2026-07-17
Correct stale 'XML-only' provider-doc claims for the two adoption flags now in the admin form


iderex:
## What
\`providers.md\` still says \`AllowExistingAccountLink\` and \`RequireVerifiedEmailForAdoption\` are edited in the provider's \`config.xml\`, not on the config page. That went stale once PR #484 and PR #488 added both toggles to the admin OpenID provider form (\`SSO-Auth/Config/configPage.html\`).

## Why it matters
The doc actively steers admins away from the UI control that now exists, adding needless config.xml friction for two adoption/security flags.

## Fix
Update the affected passage(s) in \`providers.md\` to say each flag is settable in the admin OpenID provider form, or by editing \`config.xml\` directly. Docs-only change; no code touched.

PR #525, opened 2026-07-18
Surface RequireVerifiedEmailForLogin in the admin provider form

### LINKED ISSUE #524: Surface RequireVerifiedEmailForLogin in the admin provider form
Follow-up to #166 (PR #522). `RequireVerifiedEmailForLogin` is the login-time analog of `RequireVerifiedEmailForAdoption`, but like the adoption flags *before* #484/#488 it is currently config-XML-only. Surface it in the admin OpenID provider form the same way.

## Proposal
Add one `.sso-toggle` checkbox to the OID provider form in `SSO-Auth/Config/configPage.html` with `id="RequireVerifiedEmailForLogin"` (rides the generic #365 save contract → no config.js change), default unchecked = `false` = current behaviour, `fieldDescription` noting it requires `email_verified == true` for *every* OIDC login for the provider (needs the `email` scope) and points to the security implication. Extend the `ArchitectureConformanceTests` security-critical form-field reverse-pin roster, exactly as #484/#488 did for the adoption flags.

## Acceptance
- [ ] Toggle renders + round-trips (unchecked→false=default; ServerManagedFields untouched).
- [ ] Form-contract reverse-pin added to the conformance roster.

iderex:
# What & why

Closes #524

`RequireVerifiedEmailForLogin` (`OidConfig`, #166, PR #522) requires every
OpenID login for a provider to carry a provider-verified `email_verified ==
true` claim, not only the one moment an adoption happens, but every login:
adopt, create, or sign-in to an already-linked account. It is a real,
security-relevant toggle, but it was XML-only: `configPage.html` /
`config.js` never referenced it, so `listArgumentsByType` never persisted it
and the only way to set it was hand-editing `SSO-Auth.xml` per provider.
#484/#488 surfaced its siblings, `AllowExistingAccountLink` and
`RequireVerifiedEmailForAdoption`, the same way; this closes the same gap for
the login-time flag.

We add an `sso-toggle` checkbox for it to `#sso-new-oidc-provider`, directly
below the two adoption-related toggles, wired through the existing generic
save path.

**How the flag is rendered + saved (mirrors #484/#488):** the checkbox
carries `class="sso-toggle"` and `id="RequireVerifiedEmailForLogin"`, spelled
exactly like the `OidConfig` property. That is the entire #365 save contract
 - no `config.js` change is needed because the save/load path is generic over
the marker classes:

  `#RequireVerifiedEmailForLogin.checked = Boolean(provider["RequireVerifiedEmailForLogin"])`
  `current_config["RequireVerifiedEmailForLogin"] = checkbox.checked` onto the
  existing provider object and PUTs the whole config, it WRITES the value on
  save. The server keeps the JSON member because it is a real `OidConfig`
  property.

default (`false`, fail closed) and its semantics are unchanged. An unchecked
box persists `false`, which equals the property default, so a provider with
no explicit value stays `false` after a save. The `fieldDescription` states
the security implication, notes it needs the `email` scope for the provider
to actually return `email_verified`, and that it is independent of
`AllowExistingAccountLink` / `RequireVerifiedEmailForAdoption` (run either,
both, or neither).

persists it -> `loadProvider` re-reads and re-checks the box. Other
server-managed/XML-only fields are untouched: `ServerManagedFields.Preserve`
only re-injects `CanonicalLinks` and `OidSecret`, and this flag is neither, so
the overlay and server-managed preservation are intact.

The **SAML side needs no change**: `RequireVerifiedEmailForLogin` has no
`email_verified` equivalent on SAML (documented in `providers.md`) and there
is no SAML provider form in the admin UI at all (SAML config is entirely
XML-only), so there is nothing to add the affordance to there.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. (No change to auth/login logic; the
      flag's `false` default and semantics are untouched. An unchecked box
      persists `false`, the fail-closed value.)
      or secret handling touched.)
      config persistence, or the release pipeline. (Four-lens adversarial
      self-review below; config-persistence surface.)
      logging.)

## Quality checklist

- [x] Architecture-conformance tests pass, and any new structural property
      this change establishes is locked in as a rule in
      `ArchitectureConformanceTests`. (Added `RequireVerifiedEmailForLogin`
      to the security-critical reverse-pin roster, and updated the roster's
      comment, as that test's comment instructs.)

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. (Debug, 0
      warnings.)
- [x] `dotnet test` is green. (1063 passed, incl. the extended
      (`configPage.html` formatted with `prettier --write`, confirmed clean
      with `prettier --check`; verified the committed blob is LF-only, not
      the CRLF working-tree false positive.)

## Notes

Adversarial self-review, four refute-by-default lenses (config-persistence +
a security-relevant flag):

  the admin page renders one more toggle; `listArgumentsByType` simply gains
  one `check_fields` entry, no iteration/limit concern. A provider with no
  explicit value loads unchecked and, if saved, persists `false`, the
  existing default, so no login is silently gated on an unmet
  `email_verified` claim for a deployment that was not already relying on it.
  stored flag without an explicit admin click is the pre-existing "save
  without Load Provider first" behavior shared by every toggle, and here
  that direction is `true -> false`, which loosens a defense-in-depth check
  but does not touch the primary auth gates (signature/issuer/audience/expiry
  validation is untouched by this flag). Before this change a stale XML
  `true` could never be turned off in the UI; now it can. No server-managed
  field is touched: `ServerManagedFields` preserves only `CanonicalLinks` and
  `OidSecret`.
- **Correctness (read-on-load / write-on-save round-trip), PASS.** `id`
  matches the `OidConfig` property exactly and `class="sso-toggle"` puts it
  in `check_fields`; `loadProvider` sets `checked = Boolean(provider[id])` and
  `saveProvider` writes `current_config[id] = checked`. Default unchecked ->
  `false` = property default. The conformance test now asserts it persists.
- **Integration (the `saveProvider` overlay + `ServerManagedFields` intact,
  and the field sits with the flags it is independent of), PASS.** The
  checkbox is inside `#sso-new-oidc-provider`, directly below
  `RequireVerifiedEmailForAdoption`, so the form-scoped scan picks it up and
  the UI groups the three email/adoption-related flags together. The overlay
  (seed from existing provider, overlay form fields, PUT whole object) is
  unchanged, and the added field is orthogonal to server-managed
  re-injection.

Out of scope: `providers.md`'s closing line for this flag ("This flag is
currently set by editing the provider's `config.xml` directly (it is not yet
in the admin form)") and the `RequireVerifiedEmailForLogin` XML-doc comment
on `OidConfig` ("XML-only") are now stale, the same way #484 made the
`providers.md` claim for `AllowExistingAccountLink` stale until #489 (a
separate follow-up PR) corrected it. Both are prose/doc-only, touch no
runtime behavior, and are out of the "touch only configPage.html + the
conformance test" scope given for this delivery; flagging as a follow-up doc
fix rather than widening this diff.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -872,14 +872,14 @@ public class ArchitectureConformanceTests
         // fail-open (the server keeps the stored value; the admin can no longer harden it), so pin the
         // security-critical settings: each MUST remain a marked, correctly-typed persisting field. Extend
         // this roster in the same PR that surfaces a new security toggle in the admin form; a deliberately
-        // XML-only toggle (e.g. RequireVerifiedEmailForLogin #166) is not a form field and so stays out of
-        // this roster until it is surfaced (as RequireVerifiedEmailForAdoption was, #484/#488).
+        // XML-only toggle is not a form field and so stays out of this roster until it is surfaced (as
+        // RequireVerifiedEmailForAdoption was, #484/#488, and RequireVerifiedEmailForLogin was, #524).
         var securityCritical = new[]
         {
             "EnableAuthorization", "OidSecret", "DisableHttps", "DisablePushedAuthorization",
             "DoNotValidateEndpoints", "DoNotValidateIssuerName", "DoNotValidateResponseIssuer",
             "DoNotLoadProfile", "RequirePkce", "AllowExistingAccountLink",
-            "RequireVerifiedEmailForAdoption",
+            "RequireVerifiedEmailForAdoption", "RequireVerifiedEmailForLogin",
         };
         var unsaved = securityCritical.Where(p => !matchedIds.Contains(p)).ToList();
         Assert.True(
--- a/SSO-Auth/Config/configPage.html
+++ b/SSO-Auth/Config/configPage.html
@@ -797,6 +797,35 @@
                   </div>
                 </div>
 
+                <div
+                  class="checkboxContainer checkboxContainer-withDescription"
+                >
+                  <label>
+                    <input
+                      is="emby-checkbox"
+                      id="RequireVerifiedEmailForLogin"
+                      name="RequireVerifiedEmailForLogin"
+                      type="checkbox"
+                      class="sso-toggle"
+                    />
+                    <span>Require Verified Email For Login (Sensitive)</span>
+                  </label>
+                  <div class="fieldDescription checkboxFieldDescription">
+                    ⚠ Sensitive: requires every OpenID login for this provider
+                    to carry a provider-verified email (email_verified == true)
+                    — whether it adopts, creates, or signs in to an
+                    already-linked account — not only the adoption moment above.
+                    A login whose email_verified is not exactly true (absent,
+                    false, or unparseable) is refused. Needs the email scope so
+                    the provider actually returns email_verified; without it,
+                    every login is refused. Off by default so a deployment that
+                    does not set it, or an IdP that never emits the claim, is
+                    unaffected. Independent of Allow Adopting Existing Accounts
+                    and Require Verified Email For Adoption above (run either,
+                    both, or neither).
+                  </div>
+                </div>
+
                 <div class="inputContainer">
                   <label
                     class="inputLabel inputLabelUnfocused"


iderex:
My sign-off, config-persistence, surfaces the #166 RequireVerifiedEmailForLogin flag in the admin UI (Closes #524), no semantics change. Delivered by a delegated agent; four-lens review all PASS. Exact clone of the merged #484/#488 pattern: one .sso-toggle checkbox id=RequireVerifiedEmailForLogin in configPage.html (grouped under RequireVerifiedEmailForAdoption) riding the generic #365 save contract, no config.js change; unchecked persists false = the property default; ServerManagedFields (CanonicalLinks/OidSecret) untouched; the extended ProviderFormFieldIds_MatchOidConfigProperties conformance test mechanically proves the id↔property round-trip. Two files only (configPage.html + ArchitectureConformanceTests). 1063 tests; required checks green. Ready to merge.

Issue #385, opened 2026-07-15
Make AvatarService's fetch path unit-testable (ReadCappedAsync + a handler seam)


TheRealStroopwafel:
From the #375 review (iderex MEDIUM + LOW; rekamer/shippingToken coverage notes). The SSRF-safe avatar fetch is now in `AvatarService`, but `TrySetAsync` constructs its `SocketsHttpHandler`/`HttpClient` inline, so the content-type gate, Content-Length pre-check, streamed size cap, happy path, and timeout stay unreachable by unit tests, only the two early returns are covered (the bulk of the #375 SonarCloud new_coverage gap, accepted as coverage-only).

1. **Make `ReadCappedAsync` internal and pin #220 directly**, needs only an `HttpResponseMessage` over `StreamContent`, no live HTTP: a body over 10 MiB with a lying/absent Content-Length throws; a body at the cap round-trips with `Position == 0`. iderex: "the single most security-relevant untested branch; no reason to defer."
2. **Add an internal ctor overload accepting an `HttpMessageHandler`** (production passes the hardened handler, tests a stub), ~5 lines, unlocks the content-type-rejection and happy-path tests. Land before the pattern is copied to the OIDC region.

While there: rename the self-named `ConnectCallback = ConnectCallback` object-initializer assignment (all three reviewers) to a distinct name (e.g. `ConnectToAllowedAddressAsync`). Ties into #192.

Refs #375, #192, #220

PR #530, opened 2026-07-18
Force-refresh the SSO avatar when the on-disk file is missing

### LINKED ISSUE #480: Avatar 304 skips self-heal when on-disk profile file is deleted out-of-band
Noticed while delivering #248 (avatar fetch became conditional via `If-Modified-Since`). Best-effort self-heal edge: if the on-disk `profile.*` file is deleted out-of-band while the DB `user.ProfileImage` (`ImageInfo`) record survives, the next login now gets a **304 Not Modified** and skips the re-download, so the avatar no longer self-heals from the DB record as it did under the old unconditional-download behaviour.

Low severity (requires out-of-band deletion of just the file but not the DB record). Follow-up only if self-heal on missing-on-disk is desired.

## Options
- Detect a missing on-disk `profile.*` (despite a live `ImageInfo`) and force an unconditional fetch that login.
- Or accept the trade-off (document it), the conditional fetch is the decided cadence (#248).

## Acceptance
- [ ] A decision recorded: force-refresh-on-missing-file, or document-accept.

iderex:
# What & why

We fix a self-heal regression in the login-time SSO avatar sync (#480). Since
the avatar fetch became conditional via `If-Modified-Since` (#248), a login
whose on-disk `profile.*` file was deleted out-of-band, while the DB
`user.ProfileImage` (`ImageInfo`) record survives, receives a **304 Not
Modified** and skips the re-download, so the avatar can no longer be restored
from the live record.

We chose **force-refresh-on-missing-file**: when the `ImageInfo` record is live
but its `profile.*` file is absent on disk, we omit `If-Modified-Since` and
fetch unconditionally so the avatar self-heals; the conditional cadence (#248)
is preserved for the normal file-present case. The on-disk probe is an injected
`Func<string,bool>` defaulting to `File.Exists`, matching the existing
test-seam idiom (the #385 `HttpClient` and #400 `KeyedLockStore` seams), so the
missing-file branch is unit-testable without touching the filesystem.

Issue #480 acceptance, a decision is recorded: **force-refresh-on-missing-file**
(not document-accept).

Closes #480

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

This touches the login-time avatar sync path. The IdP-supplied avatar URL is
attacker-influenced, but the fetch path itself (the SSRF `ConnectCallback`, the
content-type allow-list, the size cap, and the 10s deadline) is unchanged, we
only gate whether the `If-Modified-Since` header is sent.

- [x] Fail-closed preserved: the change only decides whether to send
      `If-Modified-Since`; every outbound fetch still runs the unchanged
      SSRF / content-type / size guards, and the new branch fails **toward**
      re-download, never toward skipping a guard.
      changed; no secrets are involved.)
- [x] A security review was performed: the adversarial correctness and
      availability lenses read the full touched files plus the caller
      (`SessionMinter.MintAsync` / `SSOController`) — results under Notes.
- [x] Log inputs from the IdP are sanitized inline at the log call. (Unchanged;
      this change adds no new log call — the existing disallowed-URL and
      content-type warnings already strip line endings inline.)

## Quality checklist

      testable unit (a one-clause extension of the existing conditional plus a
      `Func<string,bool>` seam).
- [x] Architecture-conformance tests pass. No new structural property was
      established (this is a behavioral gate, not a new architectural boundary),
      so no new `ArchitectureConformanceTests` rule was required.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green, Debug **and** Release.
- [x] `dotnet test` is green, 1066 tests (1063 baseline + 3 new), including
      `ArchitectureConformanceTests`.
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change - N/A,
      no such files touched (only `.cs`).

## Notes

**Adversarial review**, two refute-by-default lenses ran on the full touched
files plus the caller:

- **correctness-lens → PASS.** Ran an empirical mutation probe: reverting the
  guard fails exactly the two core new tests (so they are non-vacuous) while the
  #248 cadence tests keep passing (so the normal case is not regressed).
  Confirmed the guard reads `Path` and `LastModified` from one `ImageInfo`
  capture, short-circuits safely (no NRE, `File.Exists` never throws), and that
  the stored path equals the path the store writes, so the existence check is
  meaningful; any TOCTOU only costs one extra login of self-heal latency and
  never fails open. Flagged one optional in-diff mutation survivor on the
  pre-existing `lastStored > DateTime.MinValue` clause; we closed it with a
  third test (a `MinValue`-anchor record fetches unconditionally even with the
  file present).
- **availability-lens → PASS.** Confirmed no fetch-amplification: the
  unconditional path self-limits to a single fetch, a successful store rewrites
  the file to exactly the recorded path, so the next login's existence check
  passes and the conditional 304 cadence resumes. The path is best-effort
  (wrapped in the existing catch, so it never blocks or locks out login),
  bounded by the unchanged 10s deadline and 10MB cap, and the trigger (deleting
  the on-disk file while the DB record survives) requires host filesystem
  access, not an attacker capability. A non-compliant origin that answers 304
  to an unconditional request is handled by the early return without a throw.

**N/A:** crypto (SAML/OIDC), config persistence, and the release pipeline are
untouched.

**Out of scope:** none identified.

### DIFF
--- a/SSO-Auth.Tests/AvatarServiceTests.cs
+++ b/SSO-Auth.Tests/AvatarServiceTests.cs
@@ -447,8 +447,10 @@ public class AvatarServiceTests
 
     // Builds a service over a StubHandler the test keeps a handle to, so it can assert the conditional
     // fetch header and the handler reuse count (#248). The logger is returned too, so a test can assert the
-    // 304 path took the early return rather than a swallowed error.
-    private static (AvatarService Service, IProviderManager Providers, IUserManager Users, StubHandler Handler, CapturingLogger Log) BuildWithHandler(HttpResponseMessage response)
+    // 304 path took the early return rather than a swallowed error. The on-disk probe (#480) defaults to
+    // "present" so the conditional-cadence tests exercise the normal file-present path; the missing-file
+    // self-heal tests pass their own predicate returning false.
+    private static (AvatarService Service, IProviderManager Providers, IUserManager Users, StubHandler Handler, CapturingLogger Log) BuildWithHandler(HttpResponseMessage response, Func<string, bool>? fileExists = null)
     {
         var users = Substitute.For<IUserManager>();
         var providers = Substitute.For<IProviderManager>();
@@ -456,7 +458,7 @@ public class AvatarServiceTests
         serverConfig.ApplicationPaths.UserConfigurationDirectoryPath.Returns(UserDataRoot);
         var handler = new StubHandler(response);
         var log = new CapturingLogger();
-        var service = new AvatarService(users, providers, serverConfig, log, "test-agent/1.0", new HttpClient(handler));
+        var service = new AvatarService(users, providers, serverConfig, log, "test-agent/1.0", new HttpClient(handler), fileExists: fileExists ?? (_ => true));
         return (service, providers, users, handler, log);
     }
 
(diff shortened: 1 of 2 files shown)


iderex:
## Pre-merge review

Gates below, all green; ready for approval.

**Adversarial multi-lens review** (refute-by-default, each lens reading the full
touched files plus the caller, `SessionMinter.MintAsync` / `SSOController`):

- **correctness-lens → PASS.** Empirical mutation probe: reverting the guard
  fails exactly the two core new tests (so they are non-vacuous), while the
  #248 cadence tests keep passing (so the normal file-present case is not
  regressed). Confirmed `Path` and `LastModified` are read from one `ImageInfo`
  capture, the guard short-circuits safely (no NRE; `File.Exists` never throws),
  and the probed path equals the path the store writes, so the existence check
  is meaningful and any TOCTOU only ever fails toward re-download.
- **availability-lens → PASS.** No fetch-amplification: the unconditional path
  self-limits to a single fetch, a successful store rewrites the file to
  exactly the recorded path, so the next login's existence check passes and the
  conditional 304 cadence resumes. It is best-effort (never blocks or locks out
  login), bounded by the unchanged 10s deadline and 10MB cap, and the trigger
  (deleting the on-disk file while the DB record survives) requires host
  filesystem access, not an attacker capability.

**Findings and resolution.** No blocking findings. The one optional item, an
in-diff mutation survivor on the pre-existing `lastStored > DateTime.MinValue`
clause pulled into the changed hunk, was FIXED in the diff (a third test pins a
`MinValue`-anchor record fetching unconditionally even with the file present),
not declined.

**CI, all checks green:** build, ABI floor build, Package (JPRM), CodeQL
(csharp / actions / javascript-typescript), dependency-review, prettier, zizmor,
Audit workflows, Enforce greppable invariants, Reject Trojan Source Unicode.

**Local gate:** `dotnet build --no-restore --warnaserror` green (Debug **and**
Release); `dotnet test` green, 1066 tests (1063 baseline + 3 new), including
`ArchitectureConformanceTests`.

**N/A:** crypto (SAML/OIDC), config persistence, and the release pipeline are
untouched.

**Overall verdict: PASS, ready for approval.**

iderex:
My sign-off, login-time avatar self-heal fix (Closes #480; delivered by a scheduled maintenance routine, reviewed here). Verified against the diff: an injectable fileExists probe is added to AvatarService; when the on-disk profile.* file is missing out-of-band (while the DB ImageInfo record survives), the login now fetches the avatar UNCONDITIONALLY (If-Modified-Since deliberately withheld, no advertising a stale last-store timestamp) and re-stores it, restoring the self-heal that #248's conditional cadence had removed for the missing-file case. Fail-closed/best-effort preserved: a non-compliant origin answering 304 to the unconditional request leaves the existing record untouched and NEVER throws into the login; a sentinel DateTime.MinValue LastModified also forces an unconditional fetch (guards a >/>= slip). Three new tests pin all three cases. build/prettier/dependency-review/CodeQL green; mergeable clean. Ready to merge.

iderex:
Independent pre-merge review (more-eyes gate)

We re-read AvatarService.cs and the tests independently of the implementation pass and confirmed:

- The fix gates the #248 If-Modified-Since conditional on the on-disk file still being present (`storedImage.Path is { } storedPath && _fileExists(storedPath)`). File present with a real timestamp keeps the conditional cadence unchanged; a file deleted out-of-band while the ImageInfo record survives now omits the header, fetches unconditionally, and self-heals. The normal path is behavior-preserving.
- Fail-closed edges hold: a null Path or the DateTime.MinValue sentinel both fall through to an unconditional fetch (never a bogus year-0001 If-Modified-Since); a non-compliant origin that answers 304 to an unconditional request leaves the record untouched and never throws into login (no Error log, no SaveImage, no ClearProfileImageAsync).
- No fetch amplification: the trigger (deleting the on-disk profile file) requires host filesystem access, not an attacker capability; the fetch is once per login, bounded by the unchanged 10s/10MB limits.
- The `fileExists` seam defaults to File.Exists and mirrors the existing handler/lock seams; the three new tests pin the self-heal, the suppressed-conditional fail-safe, and the `> DateTime.MinValue` guard (a `>`->`>=` slip is caught).

Debug + Release build clean with analyzers-as-errors; full suite green; all CI checks pass. crypto/config-persistence/release untouched.

Gate: PASS. Merging.

Issue #528, opened 2026-07-18
Drop the deprecated assertion-embedded SAML/Auth path in a later release


iderex:
## Context

#251 replaced the SAML assertion browser round-trip with a one-time server-side outcome token: the ACS callback (`SAML/post`) validates the assertion once and hands the intermediate auth page only an opaque token; `SAML/Auth` redeems the token and completes the login without re-parsing.

To keep logins that are already in flight when an admin upgrades from breaking, `SamlLoginService.AuthenticateAsync` still accepts and FULLY validates the **pre-#251 assertion-embedded shape** (a full base64 assertion posted to `SAML/Auth`) for one release, the deprecation window. The plugin itself renders only tokens going forward.

## What to do (a later release)

Once the deprecation window has elapsed (at least one release after #251 ships), remove the deprecation branch in `SamlLoginService.AuthenticateAsync`:

- Delete the fallback that runs `_validator.TryValidate` + `CorrelateAndBind` + the login allow-list + `TryProduceVerifiedIdentity` on a raw assertion posted to `SAML/Auth`; keep only the one-time outcome-token redeem.
- After the removal, `SAML/Auth` accepts ONLY the opaque outcome token. A scripted client that POSTs a raw assertion straight to `SAML/Auth` (bypassing the rendered page) will then be rejected, this is the wire-contract change #251 flagged as `breaking`, so it must land on a feature/breaking release with a changelog note.
- The linking flow still renders the assertion to its own `SAML/Link` leg; re-check whether the follow-on `SAML/Auth` post (which currently lands on the deprecation branch as a no-op after the link consumed the one-time id) needs any adjustment once the branch is gone.
- Update the deprecation-window test in `SSOControllerSamlTokenTests` (the full-assertion-to-`SAML/Auth` case) to assert the post is now rejected, and drop the deprecation comments.

Refs #251.

iderex:
Implemented on the 4.2 branch via #610 (merged). Deprecated assertion-embedded SAML/Auth path removed; fail-closed preserved (guard clause + negative test); all four adversarial review lenses PASS; 1272 tests green on net9.0 + net10.0. Reaches main at the 4.2.0 release forward-merge. (Closed manually, a PR to 4.2 does not auto-close per BRANCHING.md.) Linking-UX quirk surfaced in review filed as #614.

Issue #415, opened 2026-07-16
Bind the SAML login flow to the initiating browser (RelayState/InResponseTo), as OIDC now is


iderex:
Follow-up from #326 (acceptance criterion 4: "Decide the SAML analogue").

#326 bound the OpenID authorize state to the initiating browser to close a forced-login / session-fixation vector. The SAML path (`SamlPost` -> `SamlAuth`) likewise mints a session from a presented credential with no browser binding. The classic SAML login-CSRF defenses, `RelayState` echo + `InResponseTo` correlation, are currently opt-in (`ValidateInResponseTo`, off by default, and it disables IdP-initiated SSO).

Decide and implement the SAML analogue of the OIDC browser binding: either a binding cookie mirroring #326 (works for SP-initiated flows regardless of `InResponseTo`), or make the RelayState/`InResponseTo` correlation the browser tie. IdP-initiated (unsolicited) SAML has no request to correlate, so the design must state its posture for that case explicitly (fail-closed vs. an opt-in for IdP-initiated deployments).

Filed separately per #326's acceptance criteria so the SAML change gets its own review and its own decision on the IdP-initiated tradeoff.

Refs #326, #238.

iderex:
Design (SP-initiated browser binding, the SAML analogue of #326):

**Threat.** SP-initiated forced-login / session fixation: an attacker obtains a valid, signed SAMLResponse for their own identity (solicited, `InResponseTo=R_a`) and lures the victim's browser to submit it, silently signing the victim in as the attacker.

**Where to check, `SamlAuth`, not the ACS `SamlPost`.** The response reaches `SamlPost` as a **cross-site POST** from the IdP origin, on which a `SameSite=Lax` cookie is NOT sent, a binding check there would fail every login. `SamlAuth` (the session-mint boundary reached by the auth page's **same-origin** fetch) does carry the Lax cookie, and it is already where `InResponseTo` and replay are consumed one-time. So the check lives there, mirroring #326's Lax choice.

**Mechanism.**
1. `SamlChallenge` mints a 256-bit binding id, sets `__Host-sso_saml_state_binding` (HttpOnly, Secure, SameSite=Lax, Path=/), and registers it in the outstanding-request store keyed by the AuthnRequest id, now done **always** for login flows (not only when `ValidateInResponseTo` is on), because the binding is a separate, always-on protection.
[...]

PR #531, opened 2026-07-18
Replace the SAML assertion browser round-trip with a one-time outcome token

### LINKED ISSUE #251: SAML flow validates the assertion twice by round-tripping it through the browser
## Problem

The SAML browser flow validates the same assertion twice because the full signed XML round-trips through the client.

`SamlPost` parses and validates the response (`SSOController.cs:666-667`), re-serializes the whole document (`Response.Xml` = `OuterXml`, `Saml.cs:78`), and embeds it base64 in the rendered auth page (`:678`); the page JS then POSTs it to `SamlAuth`, which parses and fully validates it again (`:842-843`). Per login, the base64 decode / DOM parse / C14N + RSA verify / certificate load pipeline runs twice, the document is copied ~4 times to embed it, and the auth page is inflated by the base64 assertion. The second validation is required as long as raw XML crosses the untrusted client, it must not be removed under the current handoff.

The OIDC flow already uses the better shape: only a state ID crosses the page (`WebResponse.cs:411`).

## Fix (design change, needs a deprecation window)

Validate once in `SamlPost`, store the extracted login outcome server-side in a bounded one-time store keyed by a CSPRNG token (the existing `AuthStateStore` / replay-cache patterns), and hand the page only the token; `SamlAuth` consumes the token the way `OidAuth` consumes state. Halves per-login crypto, deletes the re-serialization and the embedded assertion, and thins `SamlAuth` (aligns with #160).

Caveat: this changes the `SAML/Auth` API contract for any client that POSTs raw assertions directly today. The plugin serves the page JS itself, so browser flows migrate atomically, but scripted clients need a documented deprecation window, schedule, don't hot-fix.

iderex:
# What & why

The SAML login flow validated the same signed assertion **twice**. The ACS
callback (`SAML/post`) parsed and validated the response, re-serialized the whole
document, and embedded it base64 in the intermediate auth page; the page then
posted that assertion back to `SAML/Auth`, which parsed and fully validated it
again. Per login the base64-decode / DOM-parse / C14N + RSA-verify pipeline ran
twice, the document was copied ~4 times to embed it, and the signed assertion
crossed the untrusted client. The OpenID flow already uses the better shape - 
only a state token crosses the page.

This validates the assertion **once** at the callback and stores the verified
outcome server-side in a one-time, CSPRNG-keyed store, mirroring the OpenID
authorize-state redeem. The page carries only an opaque token; the assertion
never crosses the browser again.

Closes #251.

## Type of change

- [ ] Bug fix
- [ ] Feature

(Security-relevant: it touches the SAML login path, so the adversarial gate was run, see Notes.)

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green.

### Design

- **One-time outcome store + atomic redeem.** New `SamlOutcomeStore` (a `*Store`
  under the flow tier) keyed by a 256-bit CSPRNG token: cap-bounded registration
  with a per-client sub-cap (1/100), an `IntervalGate`-throttled expired-entry
  sweep, provider- and lifetime-scoping on redeem, and a single atomic
  `TryRemove` claim, so a token mints at most one session even under concurrent
  posts. `SamlLoginOutcome` is an immutable record, so the redeem is never a torn
  read; the store faithfully mirrors `OidcStateStore` (cap / release / prune /
  negative-age guard). A new fitness function `SamlLoginOutcome_IsImmutable`
  locks in that immutability.
- **Once-only validation + replay consume.** The callback runs the full
  validation exactly once - signature/time/audience/recipient, the role gate, the
  one-time replay consume, and the sole SAML `VerifiedIdentity` construction - 
  then stores the verified outcome and hands the page only the token. `SAML/Auth`
  redeems the token and completes the login through the shared completion tail
  with no re-parse. The `InResponseTo` correlation + browser binding (#415) stay
  at the same-origin mint leg, where the `SameSite=Lax` cookie is sent, which
  the cross-site ACS POST could not carry, via a shared `CorrelateAndBind` that
  operates on the outcome's stored `InResponseTo`. The one-time replay id is
  consumed exactly once (at the callback on the token path, at `SAML/Auth` on the
  deprecation path); never double-consumed, never skipped.
- **Deprecation-window dual-accept.** For one release, `SAML/Auth` also accepts
  and FULLY validates the pre-#251 assertion-embedded shape (byte-equivalent to
  the old full validation), so a login already in flight when an admin upgrades
  does not break. Linking keeps the assertion page (its own leg consumes the
  one-time id; the follow-on `SAML/Auth` post lands on the deprecation branch),
  so linking is unchanged. Follow-up #528 tracks dropping the deprecated path in
  a later (breaking) release.
- **Fail-closed token handling.** An unknown / expired / provider-mismatched /
  replayed token misses the store and falls through to the deprecation
  validation, which rejects it (its decoded bytes are not a SAML response), a
  clean 400, never a mint.

### Adversarial self-review, all four lenses PASS

- **Availability**, PASS. In-flight logins survive the upgrade via the
  deprecation window; the store refuses-at-cap (never evicts) with a per-client
  sub-cap, and the callback is reached only after full signature validation (not
  anonymously floodable); no mass-lockout.
- **Security-bypass**, PASS. Token is 256-bit CSPRNG, single-use via the atomic
  `TryRemove` winner, provider- and lifetime-scoped; the deprecation path is
  byte-equivalent full validation (not a bypass); replay is consumed exactly
  once; the browser binding / lost-correlation fail-closed is preserved.
- **Correctness**, PASS. Assertion validated exactly once; outcome identical to
  today (same `VerifiedIdentity`, `AdoptionGate.None`); token consumed once; the
  re-ordering across the two legs opens no observable edge case; store
  release/prune shapes match the siblings.
- **Integration**, PASS. The auth-page JS posts `data` opaquely and is unchanged
  (token for login, assertion for linking); the store reuses
  `PerClientBudgetLimiter` / `IntervalGate` / `ProviderScopedKey`; the harness
  resets the new process-wide statics; conformance rules still hold; the
  controller / admin UI / config persistence are untouched.

### E2E checklist

- SAML login, including IdP-initiated, now requires session affinity in
  multi-instance deployments (extends the existing OIDC `OidcStateStore`
  constraint, the outcome, like the OpenID authorize state, lives in-process
  between the callback and the mint leg).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -242,6 +242,39 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void SamlLoginOutcome_IsImmutable()
+    {
+        // Locked in by the SAML one-time outcome token (#251): the login outcome stored between the ACS
+        // callback and the mint leg is redeemed by an atomic remove of the WHOLE record, so a redeemer never
+        // observes a torn outcome. A settable property or writable instance field would reopen an in-place
+        // mutation window on a value that IS the proof the assertion passed every gate, so pin it structurally
+        // exactly as AuthorizeSession's variants are (#341). A get-only auto-property / positional record
+        // parameter compiles to a readonly (initonly) backing field and passes; a `{ get; set; }` or a plain
+        // writable field would be flagged.
+        var outcome = typeof(SamlLoginOutcome);
+        Assert.True(outcome.IsSealed, "SamlLoginOutcome must be a sealed leaf.");
+
+        const BindingFlags members = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly;
+        var mutable = outcome.GetProperties(members)
+            .Where(p => p.SetMethod is not null && !IsInitOnlySetter(p.SetMethod))
+            .Select(p => $"{SimpleName(outcome)}.{p.Name} (settable property)")
+            .Concat(outcome.GetFields(members)
+                .Where(f => !f.IsInitOnly && !f.IsLiteral && !f.Name.Contains('<', StringComparison.Ordinal))
+                .Select(f => $"{SimpleName(outcome)}.{f.Name} (writable field)"))
+            .ToList();
+
+        Assert.True(
+            mutable.Count == 0,
+            "SamlLoginOutcome must be immutable (no settable property or writable instance field) so the store's one-time redeem stays torn-read-free (#251): " + string.Join(", ", mutable));
+    }
+
+    // A record's positional properties expose an `init` setter (SetMethod is non-null) but are immutable
+    // after construction; treat an init-only setter as read-only so a record's own positional members are
+    // not mis-flagged as mutable. An init-only setter carries the IsExternalInit modreq on its return type.
+    private static bool IsInitOnlySetter(MethodInfo setter) =>
+        setter.ReturnParameter.GetRequiredCustomModifiers().Any(m => m.FullName == "System.Runtime.CompilerServices.IsExternalInit");
+
+    [Fact]
     public void VerifiedIdentity_IsConstructedOnlyByProtocolValidators()
     {
         // Locked in by #473: VerifiedIdentity is the keystone the session-minting path is keyed on, and it
--- a/SSO-Auth.Tests/SSOControllerSamlTokenTests.cs
+++ b/SSO-Auth.Tests/SSOControllerSamlTokenTests.cs
@@ -1,0 +1,193 @@
+using System;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Jellyfin.Plugin.SSO_Auth.Api.Flows;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Microsoft.AspNetCore.Mvc;
+using NSubstitute;
+using Xunit;
+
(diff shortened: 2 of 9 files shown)


iderex:
My sign-off, SAML login-flow security refactor (Closes #251), approved by me with a deprecation window (so this ships backward-compatible; the actual wire break is deferred to #528). Delivered by a delegated agent; four-lens review all PASS, rebased cleanly onto current main (0ad2b55) and re-tested (1084 pass). Replaces the assertion-embedded browser round-trip (which caused a SECOND full validation) with a one-time server-side outcome token: the ACS callback validates the assertion EXACTLY ONCE (signature/time/audience/recipient, role gate, the one-time replay consume, the sole SAML VerifiedIdentity build), stores the outcome keyed by a 256-bit CSPRNG token, and hands the page only the token; SAML/Auth redeems it atomically (TryRemove single-winner) and completes with no re-parse. Fail-closed: unknown/expired/provider-mismatched/replayed tokens miss the store and fall through to the deprecation branch which rejects them (decoded bytes aren't a SAML response), clean 400, never a mint. SamlOutcomeStore mirrors OidcStateStore (cap/per-client sub-cap/IntervalGate sweep/negative-age); SamlLoginOutcome immutable (new SamlLoginOutcome_IsImmutable fitness function). InResponseTo correlation + browser binding stay at the same-origin mint leg; replay id consumed exactly once. Deprecation window: SAML/Auth still fully validates the pre-#251 assertion-embedded shape for one release so in-flight logins survive an upgrade (drop scheduled #528); linking unchanged. Two documented non-blocking notes: privilege snapshot now at callback (≤15min, symmetric with OIDC #473, live-revocation gate #232 still at mint); SAML now needs multi-instance session affinity incl. IdP-initiated (extends the OIDC constraint, added to the E2E checklist). CodeQL + required checks green; mergeable clean.

PR #534, opened 2026-07-18
Offer only enabled providers on the linking page and surface failures

### LINKED ISSUE #344: Linking page offers disabled providers and swallows the rejection
Surfaced by the #343 review. `GetNames` lists providers regardless of `Enabled`, so the linking page renders an active add button for a disabled provider; since #343 the link attempt is correctly rejected (400, uniform body), but the page's XHR discards the response, so the user sees a silent no-op instead of feedback.

Two candidate improvements, either or both: filter disabled providers out of the linking page's provider list (`GetNames` is also used elsewhere, check consumers before changing it), and surface a generic failure message in the linking UI when the XHR returns non-2xx. No security impact, the server-side rejection is in place; this is UX honesty.

iderex:
# What & why

`OID/GetNames` and `SAML/GetNames` listed providers regardless of `Enabled`, so
the self-service linking page rendered an active "add" button for a disabled
provider. Since #343 the server rejects a link/challenge through a disabled
provider with a uniform 400, so that button dead-ends in a rejection, and the
page's XHRs discarded non-2xx responses, so the user saw a silent no-op instead
of feedback. This is a UX-honesty fix; the server-side rejection is unchanged.

- `GetNames` (the sole consumer is the linking page) now returns only **enabled**
  provider names. A disabled provider is already rejected by both link twins
  (#343), so listing it only produced a dead-ending add button. A null-valued
  entry (reachable via a pre-#350 null-body add) is skipped rather than
  dereferenced, matching `LinksByUser`, the anonymous endpoint cannot NRE into a
  500.
- The elevation-gated admin surfaces `OID/Get` / `SAML/Get` are unchanged (they
  read the full map through `SnapshotConfigs`), so an administrator still sees and
  can re-enable a disabled provider.
- `linking.js` offers the add button only for enabled providers, but still renders
  the existing links to a disabled provider, without an add button, so the
  deliberate disable-then-clean-up unlink workflow (#380, pinned by
  `TryRemoveLink_ProviderDisabled_StillRemoves`) keeps working. A disabled
  provider the user has no links to is not shown.
- `linking.js` / `linking.html` surface a single generic failure banner when any
  of the page's requests (provider list, existing links, unlink) returns a
  non-2xx / rejects. The banner carries no status code or server detail, and the
  unlink flow no longer reloads on a rejected delete (which previously presented a
  failure as success).

Closes #344.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. _The #343 server-side disabled/unknown
      provider rejection (challenge and link twins) is untouched; hiding a
      provider from the display list is not relied on as the control — a
      hand-crafted add URL still hits the uniform 400._
- [x] No secrets logged; secrets are redacted on config export. _No logging or
      secret surface touched._
      config persistence, or the release pipeline. _Adversarial multi-lens review
      (bypass, correctness, integration) ran; results under Notes._
- [x] Log inputs from the IdP are sanitized inline at the log call. _No new log
      call sites._

## Quality checklist

      testable unit. _`EnabledProviderNames` mirrors `SnapshotConfigs`; the JS row
      builder is extracted into `createProviderRow`._
      _No new structural rule; existing conformance tests pass._

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. _Debug and Release, 0
      warnings / 0 errors._
- [x] `dotnet test` is green. _1064 passed, 0 failed._
      _`linking.js`, `linking.html`._

## Notes

The browser JS (`linking.js` / `linking.html`) is not exercised by CI, so the
review is its primary verification, alongside `node --check` for syntax.

Adversarial multi-lens review (refute-by-default), reading the full touched files
plus callers:

- **bypass-lens, PASS.** Confirmed the enabled-only filter cannot hide an enabled
  provider, the #343 server-side rejection is intact at every challenge/callback/
  authenticate/link sink, the generic banner leaks no internal detail, and no
  fail-open path was introduced (the delete flow is now fail-closed for display).
- **correctness-lens, first pass NEEDS_IMPROVEMENT, resolved → re-review pending.**
  Flagged that `EnabledProviderNames` dereferenced a provider value that a
  reachable null-valued entry could make null (NRE → 500 on an anonymous
  endpoint). Fixed with the `is { Enabled: true }` property pattern plus
  `*_NullValuedEntry_SkipsItWithoutThrowing` tests for both protocols.
- **integration-lens, first pass FAIL, resolved → re-review pending.** Caught that
  the enabled-only list broke a second consumer: the existing-links overlay
  null-dereferenced on disabled providers, spuriously firing the banner and
  removing the disable-then-clean-up unlink capability. Fixed by rendering
  disabled-but-linked providers without an add button and only when the user has
  links to remove.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerAdminTests.cs
+++ b/SSO-Auth.Tests/SSOControllerAdminTests.cs
@@ -18,7 +18,9 @@ public class SSOControllerAdminTests
     [Fact]
     public void OidDel_RemovesTheProvider()
     {
-        var harness = new SsoControllerHarness(c => c.OidConfigs["keycloak"] = new OidConfig());
+        // Seed it enabled so it would otherwise appear in the names list — the assertion then proves the
+        // delete removed it, not the enabled-only linking filter (#344) hiding a still-present provider.
+        var harness = new SsoControllerHarness(c => c.OidConfigs["keycloak"] = new OidConfig { Enabled = true });
 
         harness.Controller.OidDel("keycloak");
 
@@ -39,7 +41,7 @@ public class SSOControllerAdminTests
     [Fact]
     public void SamlDel_RemovesTheProvider_ReturnsOk()
     {
-        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig());
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig { Enabled = true });
 
         Assert.IsType<OkResult>(harness.Controller.SamlDel("adfs"));
 
--- a/SSO-Auth.Tests/SSOControllerEndpointTests.cs
+++ b/SSO-Auth.Tests/SSOControllerEndpointTests.cs
@@ -125,12 +125,15 @@ public class SSOControllerEndpointTests
     }
 
     [Fact]
-    public void OidProviderNames_ReturnsSeededNames()
+    public void OidProviderNames_ReturnsEnabledProviders_ExcludesDisabled()
     {
+        // The linking page is the sole consumer (#344): enabled providers are offered, a disabled one is
+        // withheld because linking through it is rejected (#343), so an add button for it would dead-end.
         var harness = new SsoControllerHarness(c =>
         {
-            c.OidConfigs["keycloak"] = new OidConfig();
-            c.OidConfigs["authelia"] = new OidConfig();
+            c.OidConfigs["keycloak"] = new OidConfig { Enabled = true };
+            c.OidConfigs["authelia"] = new OidConfig { Enabled = true };
+            c.OidConfigs["retired"] = new OidConfig { Enabled = false };
         });
 
         var result = Assert.IsType<OkObjectResult>(harness.Controller.OidProviderNames());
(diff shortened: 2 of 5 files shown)


iderex:

## Adversarial multi-lens review

Refute-by-default panel, each lens reading the full touched files plus their callers (the flow services, the config model, `CanonicalLinkService`, and the tests). Because the change touches the account-linking flow and the provider listing, we ran bypass, correctness, and integration lenses; the last two each opened a real finding that we fixed and then re-ran to a clean verdict.

- **bypass-lens, PASS.** The enabled-only filter cannot hide an enabled provider; the #343 server-side disabled/unknown-provider rejection is intact at every challenge/callback/authenticate/link sink (hiding a provider from the list is not relied on as the control, a hand-crafted add URL still meets the uniform 400); the generic banner leaks no internal detail; and no fail-open path was introduced (the unlink flow is now fail-closed for display).
- **correctness-lens, NEEDS_IMPROVEMENT, then PASS.** It flagged that `EnabledProviderNames` dereferenced a provider value that a reachable null-valued entry (a pre-#350 null-body add) could make null, an unguarded NRE→500 on an anonymous endpoint, against the null-safety convention the rest of the codebase follows. Resolved: the helper now uses the `is { Enabled: true }` property pattern (skipping a null entry as `LinksByUser` does), backed by two non-vacuous regression tests (`OidProviderNames_NullValuedEntry_SkipsItWithoutThrowing` and the SAML twin). Re-review: PASS.
- **integration-lens, FAIL, then PASS.** It caught that the enabled-only list broke a second consumer inside `linking.js`: the existing-links overlay iterates every provider returned by `sso/{mode}/links/{userId}` (enabled and disabled), so removing the disabled containers null-dereferenced on each disabled provider, spuriously firing the banner and stripping the server-supported disable-then-clean-up unlink capability (#380, pinned by `TryRemoveLink_ProviderDisabled_StillRemoves`). Resolved: `linking.js` now renders a disabled provider that the user has links to as a row without an add button, skips a disabled provider with no user links, and always populates a non-null container. Re-review: PASS.

Consumer enumeration was verified independently: an exhaustive grep across the `.cs`, `.js`, `.html`, `.md` sources and the vendored `jellyfin-apiClient.esm.min.js` confirms the linking page is the sole consumer of `GetNames`; the admin config page and the elevation-gated `OID/Get`/`SAML/Get` read the full map through a different path and are unaffected (a test pins that they still list disabled providers).

## CI and local gates

- CI: all required checks green, `build`, `prettier`, `dependency-review`, plus CodeQL (`Analyze` csharp/actions/javascript), the greppable-invariants/Trojan-Source/zizmor guards, and the JPRM package build.
- Local: `dotnet build --no-restore --warnaserror` clean in Debug and Release (0 warnings / 0 errors); `dotnet test` green at 1064 passed / 0 failed; Prettier clean on `linking.js` and `linking.html`.
- The browser JS is not exercised by CI, so the adversarial review is its primary verification, alongside `node --check` for syntax.

## Verdict

All findings fixed and re-reviewed to PASS; every gate green. We consider this ready for approval.

iderex:
Superseded by #535, which is merged. #535 branched fresh off origin/main after this branch went stale (this PR is now dirty/conflicted against the merged linking refactor). Closing in favour of the merged work; issue #344 is closed by #535.

PR #535, opened 2026-07-18
Stop offering disabled providers on the linking page and surface link rejections

### LINKED ISSUE #344: Linking page offers disabled providers and swallows the rejection

iderex:
# What & why

Closes #344.

The self-service linking page (`/SSOViews/linking`) offered providers a user
could not actually link through, and swallowed the resulting rejection.

`OID/GetNames` and `SAML/GetNames`, the only feed for the linking page's add
buttons, listed every configured provider regardless of `Enabled`, so a
disabled provider got an active add button. Linking through it is already
rejected server-side (#343), but the served auth page threw that rejection
away: its `link()` XHR resolved the response without reading the status, so the
page fell through to the auth leg and showed a misleading "Login failed" (or
nothing), never telling the user the link did not take.

Two fixes, both UX honesty layered on top of the unchanged server-side gate:

- **Filter at the source.** `OID/GetNames` and `SAML/GetNames` now return only
  enabled providers, via a shared `EnabledProviderNames` helper materialized
  under the config read lock. The disabled provider is never offered, so no dead
  add button is drawn. `linking.js` is the only consumer of `GetNames` (the
  admin config page reads the full `OidConfigs`/`SamlConfigs` from the plugin
  config, so it still shows disabled providers for management), so filtering at
  the endpoint is safe.
- **Surface the rejection.** `WebResponse.link()` now resolves the HTTP status,
  and the auth page stops on a non-2xx link with a clear message (a distinct one
  for 429) instead of proceeding. A missing status (no stored credentials, or a
  network error) keeps the prior proceed-anyway behavior, since it cannot be
  told apart from success.

Defense in depth is unchanged: a disabled provider is still rejected by the link
leg (#343) even though the UI no longer offers it, hiding it is not the gate.

The follow-up commit keeps existing links to a since-disabled provider visible
and removable. The `links/{userId}` feed deliberately returns links for all
providers regardless of `Enabled`, and `TryRemoveLink` deletes them with
`requireEnabled:false` (disable-then-clean-up is the intended workflow). With
the add-button list now enabled-only, the existing-links display is driven off
the links feed instead: a held link whose provider is no longer offered renders
as a removable, `(disabled)`-marked row (a shared `appendProviderContainer`
builder), rather than a null container that would throw. Add buttons stay
enabled-only.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (No validation touched; the disabled-provider
      link rejection in the link legs, #343, is unchanged — the UI filter is additive.)
      secret handling touched.)
      config persistence, or the release pipeline. (Four-lens adversarial review — see Notes.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (No new log calls.)

## Quality checklist

      testable unit. (`EnabledProviderNames` helper; `appendProviderContainer` builder.)
      (No new structural property; existing conformance tests pass.)

## Verification

- [x] `dotnet test` is green. (1087 passed.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (`linking.js`.)

## Notes

Scope note: the swallowed link rejection lives in the served auth page
(`WebResponse.cs` `link()`), not `linking.js`, `linking.js` renders the list
and handles unlink; the add button is a full navigation to the challenge, and
the actual link POST is issued by the auth page after the IdP round-trip. The
fix is applied where the swallow actually is. `WebResponse.cs` is C# (embedded
JS string), so Prettier does not apply to it. The recently-merged #251 only
touched a doc comment in that file, no collision.

Four-lens adversarial review (login/link surface):

- **security-bypass, PASS.** The server-side `Enabled` gate on both link legs
  (`OidLink`/`SamlLink`, #343) is unchanged; the UI filter is additive
  defense-in-depth and reduces the enumeration surface. No secret leak from
  surfacing the status (generic text; only 429 distinguished).
- **availability, PASS.** No legitimate login/link is blocked; the enabled-only
  filter runs under the config read lock and is materialized to a `List`; the
  `undefined`-status path preserves proceed-anyway so a network hiccup never
  false-blocks.
- **correctness, PASS.** `EnabledProviderNames` is provably equivalent to the
  old `Keys.ToList()` apart from the intended filter; the JS 2xx boundary
  (`< 200 || >= 300`) and the `undefined`-proceed path are correct.
- **integration, NEEDS_IMPROVEMENT → PASS after the follow-up.** The first
  review caught a real regression: with `GetNames` enabled-only, a held link to
  a since-disabled provider (still returned by `links/{userId}` and still
  deletable via `requireEnabled:false`) hit a null container and threw, aborting
  existing-links rendering and removing the stale-link delete path. The
  follow-up commit drives the existing-links display off the `links/{userId}`
  feed, rendering a removable `(disabled)` row for such links, and adds a service
  test (`DisabledProvider_LinksStayListedAndRemovable`) pinning the contract.
  Re-review returned PASS: no null-container throw remains, held disabled-provider
  links stay visible and removable, add buttons stay enabled-only, and the
  XSS-safety invariants (`createElement`/`textContent`/`CSS.escape`/
  `encodeURIComponent`) are intact.

Tests: OID and SAML `*_ExcludesDisabledProviders` (a disabled provider is not in
the offered names, an enabled one is); the provider-delete tests now seed
`Enabled` providers so the names-list observation still proves removal; and the
disabled-link list/remove contract test above.

### DIFF
--- a/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
@@ -606,6 +606,28 @@ public class CanonicalLinkServiceTests
     }
 
     [Fact]
+    public void DisabledProvider_LinksStayListedAndRemovable()
+    {
+        // The linking page no longer offers a disabled provider for new links (#344), but a link the
+        // user already holds to one must stay visible and removable — disable-then-clean-up is the
+        // intended workflow. This pins the server contract the page now relies on: LinksByUser returns
+        // the disabled provider's links, and TryRemoveLink removes them (requireEnabled:false), so the
+        // enabled-only GetNames filter never strands a stale link.
+        var (service, cfg, _, _) = Build(c => c.OidConfigs["kc"] = new OidConfig
+        {
+            Enabled = false,
+            CanonicalLinks = new SerializableDictionary<string, Guid> { ["sub-1"] = Existing },
+        });
+
+        var listed = service.LinksByUser(ProviderMode.Oid, Existing);
+        Assert.Equal(new[] { "sub-1" }, listed["kc"]);
+
+        var result = service.TryRemoveLink(ProviderMode.Oid, "kc", "sub-1", Existing);
+        Assert.Equal(CanonicalLinkRemoveResult.Removed, result);
+        Assert.False(cfg.OidConfigs["kc"].CanonicalLinks.ContainsKey("sub-1"));
+    }
+
+    [Fact]
     public void TryRemoveLink_OwnLink_RemovesItAndReturnsRemoved()
     {
         var (service, cfg, _, _) = Build(c => c.OidConfigs["kc"] = new OidConfig
--- a/SSO-Auth.Tests/SSOControllerAdminTests.cs
+++ b/SSO-Auth.Tests/SSOControllerAdminTests.cs
@@ -18,7 +18,9 @@ public class SSOControllerAdminTests
     [Fact]
     public void OidDel_RemovesTheProvider()
     {
-        var harness = new SsoControllerHarness(c => c.OidConfigs["keycloak"] = new OidConfig());
+        // Enabled so the provider would appear in the enabled-only names list if it were not deleted
+        // (#344), keeping the DoesNotContain assertion a real proof of removal.
+        var harness = new SsoControllerHarness(c => c.OidConfigs["keycloak"] = new OidConfig { Enabled = true });
 
         harness.Controller.OidDel("keycloak");
 
@@ -39,7 +41,8 @@ public class SSOControllerAdminTests
     [Fact]
     public void SamlDel_RemovesTheProvider_ReturnsOk()
     {
-        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig());
+        // Enabled so it would appear in the enabled-only names list if not deleted (#344).
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig { Enabled = true });
 
         Assert.IsType<OkResult>(harness.Controller.SamlDel("adfs"));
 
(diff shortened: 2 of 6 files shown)


iderex:

Four-lens adversarial gate: security-bypass PASS, availability PASS, correctness PASS, integration NEEDS_IMPROVEMENT → PASS after the follow-up. The integration lens caught a regression in the first cut, filtering disabled providers out of GetNames stranded held links to a since-disabled provider (the links/{userId} feed still returns them and TryRemoveLink still deletes them with requireEnabled:false, but linking.js threw on the missing container and lost the delete path). The follow-up commit drives the existing-links display off the links/{userId} feed, so a held link to a disabled provider now renders as a removable "(disabled)" row; add buttons stay enabled-only. Re-review confirmed no null-container throw remains, disabled links stay visible+removable, XSS invariants (createElement/textContent/CSS.escape/encodeURIComponent) intact.

Server-side Enabled gate on both link legs (#343) is untouched, the UI filter is additive defense-in-depth, not the gate.

CI: build, prettier, dependency-review (required) green; CodeQL csharp/js/actions, zizmor, ABI floor, JPRM package, unicode/greppable-invariants all green. Full suite 1087 passing; Debug+Release --warnaserror 0 warnings; Prettier clean.

Out-of-scope follow-up noted: linking.js load/unlink XHRs still swallow non-2xx, tracked separately.

PR #529, opened 2026-07-18
Surface a throttled capacity warning from SamlReplayCache

iderex:
# What & why

The 100k cap on `SamlReplayCache` (#452) refuses a new SAML assertion fail-closed at the cap, but did so **silently**, unlike its sibling login-path caches (`OidcStateStore`, `SamlRequestCache`) it exposed no capacity signal. A full replay cache is reached only under extreme volume or a compromised identity provider replaying signed assertions, and until now that condition was invisible to operators.

This mirrors the sibling pattern to make the refusal observable. It is pure observability parity, the replay and fail-closed semantics are untouched.

- `SamlReplayCache.TryConsume` surfaces a throttled `shouldWarnCapacity` out of its cap-refusal path, gated by its **own** cap-warn `IntervalGate` (separate from the prune gate) so a refusal burst cannot amplify into log volume (CWE-400).
- `SamlAssertionValidator` threads the signal out through `TryConsumeReplay` and `TryProduceVerifiedIdentity`.
- `SamlLoginService` logs it **once per interval at the flow boundary** on both the session-mint (`AuthenticateAsync`) and manual-link (`Link`) legs, with the provider inline-sanitized (`provider?.ReplaceLineEndings(string.Empty)`) so the log-forging sanitizer never crosses a helper boundary.

The reject behaviour is byte-identical: `TryConsume`'s boolean outcomes and the never-evict-a-live-entry rule (#32) are unchanged; the signal only decides whether to log. A replay or a missing assertion id fails closed and does **not** warn, only genuine cap pressure is surfaced, so the signal cannot be read as a per-assertion oracle.

Closes #470

## Type of change

- [ ] Security hardening / vulnerability fix
- [ ] Bug fix
- [ ] Feature
- [x] Refactor / code quality / docs

Observability parity on the SAML login path; no behaviour change. Quality milestone P3.

## Security checklist

Fill in for any change touching the login path, crypto (SAML/OIDC), config persistence, or the release pipeline.

- [x] Fail-closed preserved: a missing signature, time-bound, or audience is rejected, never default-accepted. `TryConsume` still refuses a new assertion at the cap and rejects a replay/missing-id, byte-identical to #452 (confirmed by the review gate and the retained cap/replay tests).
- [x] No secrets logged; secrets are redacted on config export. The new warning logs only the route-derived, already-validated provider name, no assertion id, NameID, or signature material.
[...]

PR #537, opened 2026-07-18
Surface a throttled cap-refusal capacity warning from SamlReplayCache

### LINKED ISSUE #470: Surface a throttled cap-refusal capacity warning from SamlReplayCache
## Context

#452 (PR #469) gave `SamlReplayCache` the hard 100k cap and the shared `IntervalGate`-throttled prune its sibling caches use. To keep `TryConsume`'s signature byte-identical (and avoid touching `SSOController`, owned by a parallel PR at the time), it deliberately did NOT wire a capacity-warning signal out of the cache.

`OidcStateStore` exposes a throttled `shouldWarnCapacity` signal that the controller logs when the cache refuses at the cap, giving operators visibility that the login-path cache is under genuine pressure. `SamlReplayCache` now refuses fail-closed at the cap but does so silently.

## Proposal

Mirror `OidcStateStore`'s pattern: surface a throttled (its own cap-warn `IntervalGate`) capacity-warning signal from `SamlReplayCache`'s cap-refusal path and log it once per interval at the controller boundary, so a full replay cache (which only happens under extreme/compromised-IdP conditions) is observable.

Pure observability parity, the security property (fail-closed, no live-entry eviction) does not depend on it.

## Acceptance

- A throttled capacity warning is emitted when `SamlReplayCache` refuses at the cap, consistent with `OidcStateStore`'s mechanism (its own gate, not the prune gate).
- No change to the replay/fail-closed semantics; `TryConsume`'s reject behavior is unchanged.
- Covered by a test asserting the warning is throttled (not per-refusal).

Follow-up to #452.

iderex:
# What & why

Closes #470. Follow-up to #452.

`SamlReplayCache` (the SAML assertion one-time-replay cache) hard-caps consumed
IDs at 100k and refuses a new assertion fail-closed at the cap, but did so
**silently**. Its sibling login-path caches (`OidcStateStore`,
`SamlRequestCache`, `SamlOutcomeStore`) each surface a throttled
capacity-warning signal so an operator can see the cache is under genuine
pressure; the replay cache did not, so a full replay cache - only reachable
under extreme login volume or a compromised IdP replaying signed assertions - 
was invisible.

This mirrors the siblings' pattern. It is a pure observability addition:

- `SamlReplayCache` gains its **own** `_capWarnGate` `IntervalGate` (distinct
  from the prune gate), and `TryConsume` now yields `out bool
  shouldWarnCapacity`, set true for at most one caller per interval **only** on
  the fail-closed cap-refusal branch. The cap refusal, the one-time replay
  consume, and all validation are byte-for-byte unchanged.
- `SamlAssertionValidator.TryConsumeReplay`, the single consume site shared by
  the login mint (`TryProduceVerifiedIdentity`), the #251 deprecation-window
  mint leg, and the manual link redeem, logs one throttled `LogWarning` when
  the signal is set, with the IdP-routed `provider` value sanitized inline at
  the log call.

This is built on current `main` (post-#251 one-time-outcome-token flow and
post-#496 `SamlAssertionValidator` extraction). It supersedes the earlier
attempt (PR #529, closed) which predated the #251 SAML refactor.

## Type of change

- [ ] Bug fix
- [ ] Feature

(Hardening/quality: an observability signal on a security-critical login-path
cache; no behavioural change to the security controls.)

## Security checklist

- [x] Fail-closed preserved: the cap refusal still returns false (login fails
      closed); the replay reject and every validation gate are unchanged. The
      new signal has no effect on any return value.
- [x] No secrets logged; the warning logs only the provider name and a static
      message — no assertion, key, or token.
- [x] A security review was performed (four adversarial lenses; see Notes).
      (`provider?.ReplaceLineEndings(string.Empty)`), not via a helper.

## Quality checklist

- [x] No duplicated logic; the cap-warn signal reuses the shared `IntervalGate`
      primitive and the caller-logs / cache-decides split the siblings use.
- [x] The code is self-documenting (comments explain the why - the fail-closed
      invariant and the throttle rationale) and matches the sibling caches.
- [x] Architecture-conformance tests pass; the new structural property (every
      login-path cache carries a distinct `_capWarnGate`) is locked in, and the
      two cache lists were hoisted into one shared canonical list so they cannot
      drift.

## Verification

- [x] `dotnet build --warnaserror` is green (Debug and Release, 0 warnings).
- [x] `dotnet test` is green (1093/1093).
- [ ] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (N/A - 
      only `.cs` files changed.)

## Notes

**Adversarial self-review, four refute-by-default lenses:**

- **Availability / mass-lockout: PASS.** The signal is throttled by its own
  gate (cannot amplify a flood into log volume), cannot suppress a real signal,
  and the login return path is unchanged (no new lockout).
- **Security-bypass / fail-open: PASS.** `shouldWarnCapacity` is a pure
  side-signal; every `TryConsume` return is behaviourally identical to before,
  the replay guard and cap are not weakened, and no secret is logged.
- **Correctness: PASS.** The warning fires only on a genuine cap refusal (not a
  first use, a replay rejection, or a missing ID), is throttled once per
  interval via an atomic CAS gate, and the tests pin the throttle and the
  gate-independence.
- **Integration: NEEDS_IMPROVEMENT → resolved.** The lens found the new
  conformance rule listed only three caches and omitted a fourth login-path
  cap-warn cache, `SamlOutcomeStore` (#251), which carries the identical
  `_pruneGate`/`_capWarnGate` pattern, so a future refactor could drop its
  `_capWarnGate` uncaught. Resolved by (a) adding `SamlOutcomeStore` to the
  guarded set, and (b) hoisting a single shared canonical `LoginPathCapWarnCaches`
  list consumed by both the prune-gate rule (#452) and the cap-warn rule (#470),
  so a cache can never fall out of one rule's list but not the other's. Both
  rules pass with the widened set.

**Out of scope / follow-ups:** none. Deliberately untouched: the linking view /
`SSOViewsController` / `OidcLoginService` (owned by parallel PRs).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -51,6 +51,17 @@ public class ArchitectureConformanceTests
         "Validator", "Cache", "Builder", "Mapper", "Policy", "Probe", "Store", "Revoker", "Extractor", "Gate", "State", "Resolver",
     };
 
+    // The login-path caches that converged on the shared bounding pattern — a hard global cap plus TWO
+    // distinct IntervalGates: "_pruneGate" (throttled expired-entry sweep, #452) and "_capWarnGate"
+    // (throttled cap-refusal capacity warning, #246/#327/#470). ONE canonical list, consumed by both the
+    // prune-gate rule and the cap-warn rule below, so a cache can never fall out of one rule's list but not
+    // the other's (which is exactly how SamlOutcomeStore was missed on first draft). A new cap-bounded
+    // login-path cache is added here once, and both fitness functions guard it.
+    private static readonly Type[] LoginPathCapWarnCaches =
+    {
+        typeof(SamlReplayCache), typeof(SamlRequestCache), typeof(OidcStateStore), typeof(SamlOutcomeStore),
+    };
+
     // Every production type, compiler-generated ones excluded — the base sequence for structural rules
     // that must cover interfaces/enums/structs/delegates too (e.g. the namespace boundary).
     private static IEnumerable<Type> AllPluginTypes =>
@@ -171,9 +182,10 @@ public class ArchitectureConformanceTests
         // "_pruneGate"), not merely some IntervalGate: the siblings also carry a "_capWarnGate", so keying
         // on the field type alone would miss a sibling that dropped its prune gate but kept cap-warn.
         // SamlRequestCache and OidcStateStore already had it (#246, #327); SamlReplayCache adopted it (#452).
+        // The cache set is the shared LoginPathCapWarnCaches list, so this rule and the cap-warn rule below
+        // can never disagree on which caches are login-path caches.
         const string pruneGateField = "_pruneGate";
-        var caches = new[] { typeof(SamlReplayCache), typeof(SamlRequestCache), typeof(OidcStateStore) };
-        var missing = caches
+        var missing = LoginPathCapWarnCaches
             .Where(t => !t.GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.DeclaredOnly)
                 .Any(f => f.FieldType == typeof(IntervalGate) && f.Name == pruneGateField))
             .Select(SimpleName)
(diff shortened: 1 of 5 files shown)


iderex:
Pre-merge review, gates below, all green; ready.

Four-lens adversarial gate: availability PASS, security-bypass PASS, correctness PASS, integration NEEDS_IMPROVEMENT → resolved. The integration lens caught that the new conformance rule guarded only three login-path cap-warn caches and omitted a fourth carrying the identical _pruneGate/_capWarnGate pattern, SamlOutcomeStore (#251). Resolved by adding SamlOutcomeStore to the guarded set and hoisting a single shared canonical LoginPathCapWarnCaches list consumed by both this rule and the existing #452 prune-gate rule, so the two lists can never drift apart again. Both rules pass with the widened set.

Behaviour-preserving: the cap refusal, the one-time replay consume, and all validation are byte-for-byte unchanged; the only new observable effect is a throttled (at-most-once-per-interval) capacity warning logged at the single shared consume site (SamlAssertionValidator.TryConsumeReplay), IdP-routed provider sanitized inline. Correctness lens hand-verified the warn fires only on the genuine fail-closed cap-refusal branch and the throttle/gate-independence math.

CI: build, prettier, dependency-review (required) green; CodeQL csharp/js/actions, zizmor, ABI floor, JPRM, unicode/greppable-invariants all green. 1093/1093 tests; Debug+Release --warnaserror 0 warnings.

Issue #542, opened 2026-07-18
ArchitectureConformanceTests.SourceFilesDeclaring does not match record/struct declarations


iderex:
## What & why

While reviewing #509 (extracting a shared \`RouteSuffix\` reader), the integration lens
noted that \`ArchitectureConformanceTests\`' \`SourceFilesDeclaring\` helper matches source
files on the regex \`\bclass\s+{Name}\b\`. That pattern never matches a \`record struct\` or
plain \`struct\` declaration (e.g. \`internal readonly record struct RouteSuffix(...)\`).

No current fitness-function rule calls \`SourceFilesDeclaring\` with a struct/record-struct
type, so nothing is broken today, this is a latent gap, not a live bug. We already have
several small internal record-struct helpers in \`SSO-Auth/Api\` (\`DiscoveryFacts\`,
\`RouteSuffix\`, and others), and if a future PR ever needs a rule that pins one of them by
declaring-file lookup, \`SourceFilesDeclaring\` will silently return an empty file list
rather than failing loudly.

## Change

Extend the regex (or add a small alternate matcher) so \`SourceFilesDeclaring\` also matches
\`record struct\`, \`readonly record struct\`, and plain \`struct\` declarations, not just
\`class\`. Low priority / low risk, pure test-infrastructure hardening, not a production
change.

iderex:
Delivered in #558, merged into the 4.2 feature-dev branch. (Auto-close does not fire for PRs merged into a non-default branch, so closing manually.)

PR #543, opened 2026-07-18
Share the route-suffix reader between ChallengePath and OidcCallbackPath

### LINKED ISSUE #509: Share the route-suffix reader between ChallengePath and OidcCallbackPath
## What & why

`ChallengePath.IsNewPath` (`SSO-Auth/Api/ChallengePath.cs:32-42`) and
`OidcCallbackPath.RedirectSegment` (`SSO-Auth/Api/OidcCallbackPath.cs:33-44`) contain an identical
route-suffix extraction:

```csharp
var segments = path?.Trim('/').Split('/');
if (segments is null || segments.Length < 3) return <default>;
var protocol = segments[^3];
var pathKind = segments[^2];
```

Only the terminal comparison differs, ChallengePath: `protocol in {OID,SAML}` and
`pathKind == "start"` returning a bool; OidcCallbackPath: `protocol == "OID"` and
`pathKind == "redirect"` returning `"redirect"`/`"r"`. The two files already document that they
share this "suffix-anchored form (#411)" and cross-reference each other.

## Change

Extract one shared reader, e.g. `bool TryReadRouteSuffix(string? path, out string protocol, out string pathKind)`,
that performs the trim/split/length-guard and yields the last-three suffix. Each classifier keeps
only its own 2-line terminal comparison and default. ~6-8 lines. The #411 boundary-slash rationale
(trim only boundary slashes so a doubled internal slash shifts the suffix and fails to match rather
than being silently collapsed) moves onto the shared reader once.

## Behaviour / security

Behaviour-preserving: the split/guard is mechanically identical at both sites and the distinguishing
comparison stays in each classifier, so both the challenge redirect_uri spelling (RFC 6749 §4.1.3)
and the callback redirect segment are unchanged; `ChallengePathTests` and `OidcCallbackPathTests`
continue to pass. The reverse-proxy-prefix defense (#337 / #411) is preserved, the suffix anchoring
is exactly what moves into the shared reader.

iderex:
# What & why

`ChallengePath.IsNewPath` and `OidcCallbackPath.RedirectSegment` carried an identical
trim/split/anchor extraction of the route's trailing `{protocol}/{path-kind}/{provider}`
suffix, the fix for #337 (a provider literally named `start` on the legacy route must not
flip the challenge's redirect_uri spelling) and #98/#411 (a protocol-like reverse-proxy
prefix earlier in the path must not decide the spelling). Only the terminal comparison and
default differed between the two. We pulled the shared mechanics into one small internal
reader, `RouteSuffix.TryRead`, that both callers now use; each keeps only its own terminal
comparison and default. Closes #509.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

This touches the login challenge/callback routing path, so we filled this in even though
the change is behavior-preserving.

      rejected, never default-accepted. (Unaffected — the extraction is a byte-for-byte
      move of the trim/split/guard/index expressions; each caller's own fail-safe default
      stayed local.)
- [x] No secrets logged; secrets are redacted on config export. (Not applicable - no
      logging or secrets in this code path.)
      config persistence, or the release pipeline. (4-lens adversarial self-review below.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (Not applicable - no
      logging in this code.)

## Quality checklist

      testable unit. (The whole point of #509 — the duplicated trim/split/anchor now
      lives once, in `RouteSuffix`.)
- [x] The change adds no more code than the problem requires. (~30 lines for the reader
      + its doc comment; net diff removes more lines than it adds across the two callers.)
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. (No new
      rule added — a 2-caller trim/split helper isn't a load-bearing structural invariant;
      the integration lens confirmed this matches how the other small internal record-struct
      helpers in this codebase, e.g. `DiscoveryFacts`, are already exempted, and that adding
      a fitness function here would be over-engineering.)

## Verification

- [x] `dotnet test` is green (1105/1105, Debug and Release).
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (Not applicable - 
      only `.cs` files touched.)

## Notes

Adversarial 4-lens self-review (this PR touches the login challenge/callback routing path):

- **Availability lens, PASS.** The extraction is index-for-index identical; both callers
  retain their own terminal comparison and fail-safe default, the `out` param is only ever
  read after a checked `true` return, and the full 1105-test suite passes, so no mass-lockout
  or availability regression is introduced.
- **Security-bypass lens, PASS.** Byte-for-byte mechanical extraction of the trim/split/index
  logic with unchanged terminal comparisons at both call sites; the null-default `out` value
  is never dereferenced on the false path; the #337/#98/#411 regression tests were carried
  forward unchanged, no suffix-anchoring bypass, spoofing, or shift.
- **Correctness lens, PASS.** Verified byte-identical trim/split/guard/index expressions
  against `origin/main`, no index shift, no null/empty divergence, no nullable-warning
  regression (`--warnaserror` clean), unchanged call sites, full 1105-test green run
  including all three touched test files.
- **Integration lens, initially NEEDS_IMPROVEMENT**, flagging that the diff under review
  was still uncommitted working-tree state (nothing on the branch yet). Not a code defect - 
  we staged and committed the four touched files (`fa980c4`) and re-ran build+test green
  against the committed state. All other integration findings (no other caller needed a
  change, `OidcLoginService.cs:430`, `SamlLoginService.cs:523`, `SsoUrlBuilder.cs:31` are
  untouched; folder placement in `Api/` not `Api/Shared/` matches precedent; no
  architecture-conformance rule needed) held and are addressed above.

Out of scope: the integration lens flagged that `ArchitectureConformanceTests`'
`SourceFilesDeclaring` helper matches on `\bclass\s+{Name}\b` and would silently return no
files for a `record struct`/`struct` type if a future rule ever tried to pin one. No current
rule does this, so nothing is broken today, filing a separate low-priority issue for
whoever eventually needs that regex to cover struct declarations.

### DIFF
--- a/SSO-Auth.Tests/RouteSuffixTests.cs
+++ b/SSO-Auth.Tests/RouteSuffixTests.cs
@@ -1,0 +1,69 @@
+using Jellyfin.Plugin.SSO_Auth.Api;
+using Xunit;
+
+#nullable enable
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Tests for <see cref="RouteSuffix.TryRead"/> — the trim/split/anchor mechanics shared by
+/// <see cref="ChallengePath.IsNewPath"/> and <see cref="OidcCallbackPath.RedirectSegment"/> (#509).
+/// <see cref="ChallengePathTests"/> and <see cref="OidcCallbackPathTests"/> already cover the two
+/// callers' terminal comparisons end to end; these tests cover the reader itself in isolation.
+/// </summary>
+public class RouteSuffixTests
+{
+    [Theory]
+    [InlineData("")]
+    [InlineData(null)]
+    public void TryRead_EmptyOrNullPath_ReturnsFalse(string? path)
+    {
+        Assert.False(RouteSuffix.TryRead(path, out _));
+    }
+
+    [Theory]
+    [InlineData("/sso")]
+    [InlineData("/sso/OID")]
+    [InlineData("sso/OID")]
+    public void TryRead_FewerThanThreeSegments_ReturnsFalse(string path)
+    {
+        Assert.False(RouteSuffix.TryRead(path, out _));
+    }
+
+    [Fact]
+    public void TryRead_ThreeOrMoreSegments_ReadsTheLastThreeAsTheSuffix()
+    {
+        Assert.True(RouteSuffix.TryRead("/sso/OID/start/keycloak", out var suffix));
+        Assert.Equal("OID", suffix.Protocol);
+        Assert.Equal("start", suffix.PathKind);
+    }
+
+    [Fact]
+    public void TryRead_TrimsOnlyBoundarySlashes()
+    {
+        // Leading/trailing slashes are trimmed, but the segment count is otherwise untouched.
+        Assert.True(RouteSuffix.TryRead("sso/OID/start/keycloak/", out var suffix));
+        Assert.Equal("OID", suffix.Protocol);
+        Assert.Equal("start", suffix.PathKind);
+    }
+
+    [Fact]
+    public void TryRead_InternalDoubledSlash_ShiftsTheSuffixInsteadOfCollapsing()
+    {
+        // The doubled slash inserts an empty segment, so the suffix read off the END of the path no
+        // longer names the real protocol/path-kind pair — it must not be silently collapsed away.
+        Assert.True(RouteSuffix.TryRead("/sso/OID//start/keycloak", out var suffix));
+        Assert.Equal(string.Empty, suffix.Protocol);
+        Assert.Equal("start", suffix.PathKind);
+    }
+
+    [Fact]
+    public void TryRead_IgnoresPrefixSegmentsBeforeTheSuffix()
+    {
+        // A protocol-like reverse-proxy prefix earlier in the path must not be picked up: only the
+        // last three segments count.
+        Assert.True(RouteSuffix.TryRead("/OID/start/proxy/sso/SAML/redirect/keycloak", out var suffix));
+        Assert.Equal("SAML", suffix.Protocol);
+        Assert.Equal("redirect", suffix.PathKind);
+    }
+}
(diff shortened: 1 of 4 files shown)


iderex:

Change: extracted the identical route-suffix trim/split/anchor logic shared by ChallengePath.IsNewPath and OidcCallbackPath.RedirectSegment into one internal reader, RouteSuffix.TryRead; each caller keeps only its own terminal comparison and default. Behaviour-preserving, closes #509.

Adversarial 4-lens self-review (login routing path):
- Availability: PASS
- Security-bypass: PASS
- Correctness: PASS
- Integration: initially flagged the diff was still uncommitted (procedural, not a code defect); resolved by committing fa980c4 and re-verifying build+test green against the committed state. All other integration findings held.

CI: build, prettier, dependency-review, CodeQL, ABI floor, JPRM package, zizmor, dependency-review, greppable invariants, all green. Local dotnet build --no-restore --warnaserror (Debug and Release) 0 warnings/0 errors; dotnet test 1105/1105 passing in both configs.

Out of scope: filed #542 (ArchitectureConformanceTests.SourceFilesDeclaring regex does not match record/struct declarations, latent gap, nothing broken today).

Not merging, leaving this for approval per the standing process.

iderex:
Pre-merge review, all gates green; ready. Four-lens gate all PASS (integration NEEDS_IMPROVEMENT was only an uncommitted-diff artifact, resolved by committing fa980c4 and re-verifying). Extracted the identical trim/split/anchor route-suffix logic from ChallengePath.IsNewPath and OidcCallbackPath.RedirectSegment into a new internal RouteSuffix record + TryRead reader; both callers keep only their own terminal comparison. Behaviour preserved. build/prettier/dependency-review + CodeQL + all others green; 1105 tests; Debug+Release --warnaserror 0 warnings. Filed follow-up #542 (conformance helper regex only matches class, not record struct, latent, nothing broken).

Issue #541, opened 2026-07-18
Avatar fetch+store worst case stacks to ~40s on the synchronous login path


iderex:
Surfaced by the availability lens review of PR for #448 (LOW, non-blocking).

`SessionMinter.MintAsync` awaits `AvatarService.TrySetAsync` synchronously before minting the session, so it is fully on the login request/response critical path. `TrySetAsync` bounds the HTTP fetch at 10s; `StoreAsync`'s per-user lock acquire is now bounded at 30s (#448). Stacked, a single login can legitimately take up to ~40s wall-clock under lock contention before the server responds.

Many reverse-proxy defaults sit at or below 30-60s, so a client/proxy could time out around the 30-40s mark and show the user a failed login while the server-side login actually completes moments later without an avatar.

## Improvement
The store-lock bound doesn't need to be as generous as the fetch bound, the store step is purely best-effort cosmetic work (disk write + occasional DB clear), not a network call. Consider shortening `AvatarService.StoreLockAcquireTimeout` well below the fetch's 10s (e.g. a few seconds), so the avatar's worst case never meaningfully extends the login response time, while still bounding a genuinely stalled holder.

Refs #448, #400, #445.

PR #544, opened 2026-07-18
Bound the avatar store-lock acquire wait with a timeout

### LINKED ISSUE #448: Bound the per-user avatar store-lock acquire with a cancellation token

iderex:
# What & why

`AvatarService.StoreAsync` acquired the per-user `KeyedLockStore` gate with
`CancellationToken.None`, so the same-user store-lock acquire wait was
unbounded. If one user's store step stalled while holding the gate (e.g.
`IProviderManager.SaveImage` blocking on disk I/O, or `ClearProfileImageAsync`
on the host DB), concurrent logins for that same user would queue behind it
forever.

We bound the wait with a 30s `CancellationTokenSource` instead. A timed-out
acquire never reaches the guarded write, it logs a warning naming the user
and returns, exactly like this class's other best-effort fail-closed branches
(disallowed URL, disallowed content type, unsafe username). The timeout is
injectable via a new optional constructor parameter, mirroring the existing
`userStoreLocks`/`fileExists` test seams, so the abort path is reachable in
tests without a real 30s wait.

`KeyedLockStore.AcquireAsync` already honored cancellation correctly (proven
by the existing `KeyedLockStoreTests`) and leaked no waiter or permit on it - 
this change only threads a bounded token into the one call site that was
passing `None`.

Closes #448.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: a timed-out/cancelled acquire never falls through
      to an unguarded store — it aborts (skips `StoreLockedAsync` entirely),
      the same as the other fail-closed branches in this class.
- [x] No secrets logged; secrets are redacted on config export. (Not touched - 
      this path logs a username, already logged elsewhere in this file.)
- [x] A security review was performed for this change (see the four-lens
      results below — this is an avatar-store path that runs on every SSO
      login).
- [x] Log inputs from the IdP are sanitized inline at the log call (the new
      warning uses `user.Username?.ReplaceLineEndings(string.Empty)`, matching
      every other IdP-controlled value logged in this file).

## Quality checklist

- [x] No duplicated logic; the new logic is a single guarded `try`/`catch`
      around the existing `AcquireAsync` call, plus one test seam.
- [x] The change adds no more code than the problem requires (one bounded
      `CancellationTokenSource`, one catch branch, one constructor parameter).
- [x] The code is self-documenting and matches the target architecture (#318);
      the load-bearing "why" comments explain the bound and the fail-closed
      contract.
- [x] Architecture-conformance tests pass. No new structural property is
      established here — `_userStoreLocks.AcquireAsync` has exactly one call
      site in the whole codebase, so a dedicated fitness function would be
      premature generalization; the behavior is locked in by two direct unit
      tests instead (see below).

## Verification

- [x] `dotnet test` is green, 1098/1098, both Debug and Release.
- [x] Prettier, N/A, this PR only touches `.cs` files.

New/extended tests (`SSO-Auth.Tests/AvatarServiceTests.cs`):

- `StoreAsync_LockAcquireTimesOut_AbortsWithoutStoringUnguarded`, with the
  per-user lock held for the whole call (standing in for a stalled
  concurrent store) and a short injected timeout, proves `SaveImage` and
  `ClearProfileImageAsync` are never called, the existing profile-image
  record is left untouched, a "Timed out" warning is logged, and the
  `KeyedLockStore` map is left at zero tracked keys (no leaked waiter).
- `StoreAsync_LockAcquiredBeforeTimeout_StoresNormally`, the counterpart:
  with the lock free, a bounded wait still lets a normal store through
  exactly as before.

## Notes

We ran the full four-lens adversarial review (availability/mass-lockout,
security-bypass/fail-open, correctness, integration) before opening this PR,
since the avatar store runs on the login path:

- **Availability/mass-lockout, PASS.** Blast radius stays strictly per-user
  (`KeyedLockStore` keys by username; unrelated users never share a
  semaphore). 30s comfortably covers normal same-user store queuing. No new
  leak: the `CancellationTokenSource` is a method-scoped `using var`, disposed
  on every exit path. One residual, non-blocking observation: `StoreAsync`'s
  new 30s bound stacks with `TrySetAsync`'s existing 10s fetch deadline on the
  synchronous login path, for a worst case of ~40s before the login response
  returns. We filed #541 to consider shortening the store-lock bound (the
  store step is local disk/DB work, not network I/O, so it doesn't need the
  fetch's own generous bound).
- **Security-bypass/fail-open, PASS.** Traced every exit from the
  `try`/`catch`: the only way to reach the guarded write is through a
  successfully assigned `storeLock`; the timeout catch always returns first.
  Verified `SemaphoreSlim.WaitAsync`'s "takes the permit or throws having
  taken none" contract empirically (a 200k-iteration cancel-vs-release stress
  probe, zero leaks/double-grants) rather than trusting the code comment
  alone, and confirmed `KeyedLockStore.AcquireAsync`'s existing
  `catch { Unclaim(...); throw; }` never abandons a held permit.
- **Correctness, NEEDS_IMPROVEMENT → FIXED.** Flagged that `StoreAsync`'s XML
  doc comment still read "Throws on save failure" without noting the new
  silent-return-on-timeout path. Fixed: the doc now documents both outcomes.
  (Definite-assignment, `CancellationTokenSource` disposal on every exit path,
  test determinism, and call-site compatibility were all independently
  verified and passed.)
- **Integration, NEEDS_IMPROVEMENT → FIXED.** Flagged that the new timeout
  warning omitted `user.Username`, unlike every comparable warning in this
  file (disallowed URL, disallowed content type, unsafe username all log a
  sanitized IdP-controlled value). Fixed: the warning now names the user,
  sanitized inline with `ReplaceLineEndings(string.Empty)`. Confirmed the
  fetch's 10s timeout and the store's new 30s timeout are fully independent
  tokens, and that the bound applies only to the lock *acquire*, the guarded
  `SaveImage`/`ClearProfileImageAsync` calls still receive no cancellation
  token, exactly matching the issue's scope. No architecture-conformance gap:
  the single call site doesn't warrant a new fitness function.

Gates: build (Debug + Release) green, `dotnet test` green (1098/1098), four-
lens adversarial review complete with both findings fixed in this PR before
opening it. Follow-up filed as #541 (non-blocking, LOW priority).

### DIFF
--- a/SSO-Auth.Tests/AvatarServiceTests.cs
+++ b/SSO-Auth.Tests/AvatarServiceTests.cs
@@ -66,6 +66,29 @@ public class AvatarServiceTests
         return (service, providers, users, log);
     }
 
+    // Builds a service whose store-lock ACQUIRE WAIT is bounded by the given (short, test-only) timeout
+    // rather than the production 30s (#448), so a test can drive the abort-on-timeout branch deterministically
+    // and quickly instead of waiting out the real bound.
+    private static (AvatarService Service, IProviderManager Providers, IUserManager Users, CapturingLogger Log) Build(KeyedLockStore userStoreLocks, TimeSpan storeLockAcquireTimeout)
+    {
+        var users = Substitute.For<IUserManager>();
+        var providers = Substitute.For<IProviderManager>();
+        var serverConfig = Substitute.For<IServerConfigurationManager>();
+        serverConfig.ApplicationPaths.UserConfigurationDirectoryPath.Returns(UserDataRoot);
+        var log = new CapturingLogger();
+        var service = new AvatarService(
+            users,
+            providers,
+            serverConfig,
+            log,
+            "test-agent/1.0",
+            new HttpClient(new StubHandler(new HttpResponseMessage())),
+            userStoreLocks,
+            fileExists: null,
+            storeLockAcquireTimeout: storeLockAcquireTimeout);
+        return (service, providers, users, log);
+    }
+
     // An allowed public URL: the stub handler answers before any connection, so the URL only needs to
     // clear the allow-list — the SSRF transport guard (ConnectCallback) is never reached here.
     private const string AllowedUrl = "https://cdn.example.com/avatar.png";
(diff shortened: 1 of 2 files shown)


iderex:
Pre-merge review.

Local gates: dotnet build --no-restore --warnaserror clean on Debug and Release; dotnet test green at 1098/1098 on both configs. Prettier not applicable (no .js/.html/.md/.css touched).

Four-lens adversarial review (avatar store runs on the login path):
- availability/mass-lockout: PASS. Blast radius confirmed strictly per-user; 30s bound covers normal same-user queuing; no CancellationTokenSource leak. Filed a non-blocking follow-up, #541, for the ~40s worst-case stack with the existing 10s fetch deadline.
- security-bypass/fail-open: PASS. Traced every exit path; a timed-out acquire never reaches the guarded write. Empirically stress-tested SemaphoreSlim's cancel-vs-acquire atomicity (200k iterations, zero leaks) rather than trusting the comment alone.
- correctness: NEEDS_IMPROVEMENT, fixed in this PR before opening it (stale XML doc on StoreAsync not mentioning the new silent-return-on-timeout path).
- integration: NEEDS_IMPROVEMENT, fixed in this PR before opening it (the new timeout warning was missing the username every comparable warning in this file logs).

CI is still running as of this note (build/CodeQL/prettier in progress; dependency-review and the opengrep invariant lint already green). Gates below, will confirm all green before this is ready for approval. Not merging, leaving this for review.

iderex:
CI is fully green (build, ABI floor build, package, CodeQL, prettier, dependency-review, zizmor, unicode-guard, opengrep invariants). All gates pass; ready for approval.

iderex:
Pre-merge review, all gates green; ready. Four-lens gate all PASS (correctness + integration NEEDS_IMPROVEMENT fixed in-PR: stale XML doc updated, warning now names the user sanitized inline). AvatarService.StoreAsync now bounds the per-user KeyedLockStore acquire with a 30s timeout; a timed-out acquire logs and returns WITHOUT reaching the guarded write (fail-closed, no leaked lock), verified incl. a 200k-iteration SemaphoreSlim cancel-vs-acquire atomicity stress test. Timeout injectable for tests. build/prettier/dependency-review + CodeQL + all others green; 1098 tests; Debug+Release --warnaserror 0 warnings. Filed follow-up #541 (30s store bound stacks with 10s fetch deadline ~40s worst case, consider shortening).

PR #546, opened 2026-07-18
Revoke tokens on last-link unlink; document per-provider disable non-revoke

### LINKED ISSUE #343: OidLink accepts a redeemable state for a disabled provider

### LINKED ISSUE #468: Extend token revocation to single-link unlink and per-provider disable
## Context

#440 (PR #467) added active-session/token revocation to the hard lock-down path only: the admin-elevated `POST /sso/Unregister/{username}` now calls `ISessionManager.RevokeUserTokens(userId, null)` after removing the user's canonical links, so a fully-unregistered user's already-issued Jellyfin tokens are killed.

Two adjacent paths that also sever an SSO identity still leave live tokens valid:

- **`DeleteCanonicalLink` (single-link unlink):** removing one provider link for a user does not revoke that user's active sessions/tokens. If the unlink is a response to a compromised IdP identity, the attacker's already-minted Jellyfin token keeps working until it expires.
- **Per-provider disable:** disabling a provider stops new SSO logins through it, but sessions established through it before the change stay live.

## Proposal

Evaluate extending token revocation to these paths, weighing security against availability:

- **Single-link unlink:** revoking is defensible when the unlink is a security action, but a user who links/unlinks a secondary provider for convenience would be surprised to be logged out everywhere. Options: revoke only when the unlink removes the user's LAST/primary SSO link (they can no longer SSO in at all), or make it an explicit opt-in/flag. Decide fail-closed vs. least-surprise here and document the rationale.
- **Per-provider disable:** decide whether disabling a provider should terminate sessions that were established through it. This needs a way to attribute a live session to the originating provider (Jellyfin may not expose this), investigate empirically before committing.

## Acceptance

- A documented decision (with rationale) for each path: revoke, revoke-conditionally, or intentionally-not, tied to the security-vs-availability trade-off.
- Any revocation added is strictly scoped to the affected user id, fail-closed (never a false success), runs after the durable state change, and is covered by tests mirroring #440's (revoked for the target only; other users untouched; the operation still completes if revoke is a no-op).
- The E2E checklist gains a note for any new behavior.

Follow-up to #440. Security-sensitive (login/session path) → the adversarial multi-lens review gate applies.

iderex:
# What & why

Follow-up to #440 (PR #467), which added `ISessionManager.RevokeUserTokens(userId, null)` to the admin hard-lockdown path (`POST /sso/Unregister`) so a fully-unregistered user's already-issued Jellyfin tokens are killed. Two adjacent paths that also sever an SSO identity still left live tokens valid; we evaluated both against the security-vs-availability trade-off the issue asks us to resolve and reached a documented decision for each. Closes #468.

**Single-link unlink (`DELETE /sso/{mode}/Link/...`), revoke-conditionally.** We revoke the target user's tokens ONLY when the unlink removes their LAST canonical SSO link across all providers, the terminal "can no longer SSO in at all" state that matches Unregister's posture. A user who unlinks a secondary provider while still holding another link keeps a working SSO identity, so revoking there would be a self-inflicted mass-logout with no security gain. The "does the user retain any link" check is evaluated atomically inside the same config-lock transaction as the removal, so a concurrent link add/remove cannot interleave between the remove and the check. Scoped strictly to the one user id; runs after the durable removal is persisted (a revoke that throws leaves the unlink already complete).

**Per-provider disable, intentionally NOT revoked (documented).** `ISessionManager.RevokeUserTokens(Guid userId, string currentAccessToken)` is the only revocation primitive, and its second argument is the token to KEEP, not a provider filter, revocation is per user id. Jellyfin attributes no live session to the SSO provider that minted it (an SSO login yields an ordinary user token). Revoking on disable would mean killing every linked user's ALL tokens, including their password and other-provider sessions, an unscoped mass-logout. New logins through a disabled provider are already fail-closed (#343/#380 write guards + the #232 in-flight re-check). An operator who needs to hard-sever specific users uses Unregister (#440) or removes their last link (this change).

Changes: `CanonicalLinkService.TryRemoveLink` now returns `CanonicalLinkRemoval(CanonicalLinkRemoveResult Result, bool UserRetainsAnyLink)`, the remainder computed via a new static `UserHasAnyLink` in the same `Mutate`; `SSOController.DeleteCanonicalLink` revokes when `removal is { Result: Removed, UserRetainsAnyLink: false }`, with the Ok/404/409/400 response mapping unchanged.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (The revoke runs after the durable config change; the last-link check is atomic with the removal; scoped strictly to one user id. Only a real removal can trigger it — no-op outcomes revoke nothing.)
- [x] No secrets logged; secrets are redacted on config export. (Only the user id is logged.)
      config persistence, or the release pipeline. (Adversarial 4-lens review below.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (The new log line interpolates a Guid only, no IdP-controlled strings.)

## Quality checklist

      testable unit. (`UserHasAnyLink`, `CanonicalLinkRemoval`.)
      why, not what) and matches the target architecture (#318). (Thin controller decision; the pure resolution stays in the service under the lock.)
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. (No new structural property; existing conformance tests pass.)

## Verification

- [x] `dotnet test` is green. (1123 passed, 0 failed.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (None touched; N/A.)

## Notes

Adversarial multi-lens review (refute-by-default), all four PASS:

- Availability / mass-lockout, VERDICT: UserHasAnyLink is evaluated in the same Mutate after the removal, so a multi-link user unlinking a secondary provider (or a second same-provider identity) reports UserRetainsAnyLink=true and is not revoked; disable calls no session API; the remainder is atomic with the removal so no concurrent add/remove over-revokes another user; tests pin these, no mass-logout regression. PASS
- Security-bypass / under-revoke, VERDICT: last-link removal fires RevokeUserTokens(userId, null); ContainsValue(userId) is scoped to the unlinked user, so another user's link cannot suppress the revoke; RevokeUserTokens has no provider filter, so per-provider disable is genuinely unscopable and documented (new logins stay fail-closed #343/#380/#232); revoke runs after the durable persist, no under-revoke of a severed identity. PASS
- Correctness, VERDICT: the trigger is exactly "last link removed" (the removed entry is already gone when the remainder is computed); the default-struct footgun is unreachable and fail-safe; the switch bodies are byte-identical; live-user link presence is the right signal; dangling links do not skew it. PASS
- Integration, VERDICT: RevokeUserTokens(Guid, string) confirmed in Jellyfin.Controller 10.11.0 (2nd arg is the token to keep; null revokes all), matching #440's existing call; no per-provider revocation primitive exists; Mutate<T> returns the record struct fine; the response and rate-limit contracts are unchanged. PASS

Out of the PR (local, git-ignored process doc): the E2E checklist gains two entries, last-link unlink terminates sessions (#468), and per-provider disable does NOT (documented decision, with Unregister / last-link-unlink as the hard-sever paths).

### DIFF
--- a/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
@@ -585,7 +585,7 @@ public class CanonicalLinkServiceTests
 
         var result = service.TryRemoveLink(ProviderMode.Oid, "broken", "sub-1", Existing);
 
-        Assert.Equal(CanonicalLinkRemoveResult.UnknownProvider, result);
+        Assert.Equal(CanonicalLinkRemoveResult.UnknownProvider, result.Result);
     }
 
     [Fact]
@@ -623,7 +623,7 @@ public class CanonicalLinkServiceTests
         Assert.Equal(new[] { "sub-1" }, listed["kc"]);
 
         var result = service.TryRemoveLink(ProviderMode.Oid, "kc", "sub-1", Existing);
-        Assert.Equal(CanonicalLinkRemoveResult.Removed, result);
+        Assert.Equal(CanonicalLinkRemoveResult.Removed, result.Result);
         Assert.False(cfg.OidConfigs["kc"].CanonicalLinks.ContainsKey("sub-1"));
     }
 
@@ -638,7 +638,7 @@ public class CanonicalLinkServiceTests
 
         var result = service.TryRemoveLink(ProviderMode.Oid, "kc", "sub-1", Existing);
 
-        Assert.Equal(CanonicalLinkRemoveResult.Removed, result);
+        Assert.Equal(CanonicalLinkRemoveResult.Removed, result.Result);
         Assert.False(cfg.OidConfigs["kc"].CanonicalLinks.ContainsKey("sub-1"));
     }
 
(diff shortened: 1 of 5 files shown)


iderex:

Scope: extend the #440 session-revocation posture to the single-link unlink path (revoke only when the unlink removes the user's LAST canonical SSO link), and record the reasoned decision that per-provider disable does NOT revoke (Jellyfin has no provider-scoped revocation primitive, so it would be an unscoped mass-logout).

Gates:
- Build: dotnet build -c Debug AND -c Release, --warnaserror, 0 warnings / 0 errors.
- Tests: dotnet test green, 1123 passed / 0 failed (new coverage for last-link revoke, multi-link no-revoke, no-op-outcome no-revoke, target-scoping, disable no-revoke, and the service-level remainder flag incl. different-user scoping).
- Adversarial multi-lens review (refute-by-default), all four PASS:
  - Availability / mass-lockout: the remainder is computed atomically inside the same config-lock Mutate as the removal; a multi-link user unlinking a secondary provider (or a second same-provider identity) is not revoked; disable/delete paths do not revoke. No mass-logout regression.
  - Security-bypass / under-revoke: last-link removal reliably fires RevokeUserTokens(userId, null); ContainsValue(userId) is scoped to the unlinked user so another user's link cannot suppress the revoke; the disable non-revoke rationale is verified accurate (no ignored scopable primitive). No fail-open.
  - Correctness: trigger is exactly last-link-removed; the removed entry is gone before the remainder scan; the default-struct footgun is unreachable and fail-safe; switch bodies byte-identical.
  - Integration: RevokeUserTokens(Guid, string) confirmed identical in both pinned Jellyfin.Controller 10.11.0 and 10.11.11 (2nd arg = token to keep; null revokes all), matching the shipped #440 call; struct returns cleanly through Mutate<T>; response and rate-limit contracts unchanged.

Fail-closed posture: the revoke runs after the durable config change, scoped strictly to one user id, only on a genuine removal.

Not in this PR: the local (git-ignored) E2E checklist gains two entries, last-link unlink terminates sessions, and per-provider disable intentionally does not, noted so it is not lost.

iderex:
Pre-merge review, gates below, all green; ready. P2 Security-parity item.

Four-lens gate all PASS (each run inline and via an independent lens):
- availability/mass-lockout: revoke fires ONLY when the unlink removes the user last remaining SSO link across all providers, computed atomically in the same config-lock Mutate as the removal and scoped to the one user id, a multi-link user unlinking a secondary provider is NOT logged out. No mass-logout.
- security-bypass/under-revoke: last-link revoke is reliable; the retains-any-link check (ContainsValue over the user id) is correctly scoped; per-provider disable intentionally does not revoke, with a documented rationale (RevokeUserTokens second arg is the token-to-keep, not a provider filter; Jellyfin attributes no session to the originating provider, so revoking on disable would be an unscoped mass-logout), new logins already fail-closed via #343/#380/#232, so no severed-identity token survives a real severance.
- correctness: trigger is exactly last-link-removed; default-struct footgun unreachable and fail-safe.
- integration: RevokeUserTokens(Guid, string) confirmed identical across pinned 10.11.0 and 10.11.11, matching #440; the new CanonicalLinkRemoval struct returns cleanly through Mutate<T>.

CI: build, prettier, dependency-review (required) green; CodeQL csharp/js/actions + all others green. 1123 tests; Debug+Release --warnaserror 0 warnings.

E2E-CHECKLIST updated locally with the two flagged manual items (last-link unlink terminates sessions; per-provider disable intentionally does not).

PR #548, opened 2026-07-18
Source OIDC discovery facts from the login's own discovery response

### LINKED ISSUE #450: Source the OIDC discovery facts from PrepareLoginAsync's response, not a separate best-effort probe
Widened after the #210 review (PR #446): CodeRabbit (CWE-345) + the independent bypass lens both flagged that the RFC 9207 `ResponseIssuerRequired` flag, and the #141 PKCE-S256 fact, are read from the plugin-side best-effort probe `ReadDiscoveryFactsAsync`, a SEPARATE discovery fetch from the one OidcClient's `PrepareLoginAsync` performs internally to populate `ProviderInformation`.

Two consequences of the separate best-effort probe:
1. **Perf:** two discovery fetches per challenge (OidcClient's + the plugin probe). Unifying to one cuts a round-trip on the hot login path.
2. **Robustness / CWE-345 (LOW):** a probe failure or a stale `DiscoveryFactsCache` entry leaves `ResponseIssuerRequired=false`, so a callback omitting `iss` is accepted even when the AS advertises `authorization_response_iss_parameter_supported`, the RFC 9207 requirement is silently downgraded to the tolerant default. **Security impact is LOW**: mix-up is already structurally mitigated (route-fixed token endpoint + id_token validated against the bound JWKS), so the `iss` requirement is defense-in-depth and the downgrade opens no real bypass. But the enforcement should not silently depend on a best-effort probe.
3. **RequireHttps:** the probe fetch does not itself re-assert `RequireHttps`; under an admin `DisableHttps` config a network attacker on the probe could strip the advertised flag (relax to tolerant), same weak channel, same fix.

## Fix
Source both discovery facts (PKCE S256 #141 + RFC 9207 response-iss #210) from the SAME discovery response `PrepareLoginAsync` uses to build `ProviderInformation`, and persist them together on the authorize state, eliminating the separate probe. The obstacle: Duende's `ProviderInformation` does not expose the raw discovery JSON flags, so this needs to intercept/reuse the library's internal discovery fetch (or replace `PrepareLoginAsync`'s discovery with a plugin-owned one fed to both). Investigate the cleanest seam. Honor `RequireHttps` on whatever fetch remains.

Touches the OIDC login path → adversarial-review gate. Refs #247, #141, #210, #449.

iderex:
# What & why

The RFC 9207 authorization-response `iss` requirement (#210) and the PKCE-S256
fact (#141) were read from a **separate best-effort probe**
(`OidcDiscoveryCache.ReadFactsAsync`), a distinct discovery fetch from the one
`OidcClient.PrepareLoginAsync` performs internally to populate
`ProviderInformation`. Two consequences: the two fetches could disagree, and a
failed or omitted probe left `ResponseIssuerRequired = false`, silently
downgrading the RFC 9207 `iss` requirement to the tolerant default. The probe
also issued a raw, unvalidated GET, so it re-asserted no `RequireHttps`, under
an admin `DisableHttps`, a network attacker on the plaintext discovery channel
could strip the advertised flags.

This sources **both facts and the login's provider metadata from one discovery
response**, so enforcement is consistent and cannot be silently weakened by a
diverging or failed second probe:

- `OidcDiscoveryReader` does a single IdentityModel `GetDiscoveryDocumentAsync`
  under the provider's `DiscoveryPolicy` (`RequireHttps` / `ValidateIssuerName`
  / `ValidateEndpoints` + the additional base addresses), the exact call and
  policy object `OidcClient` uses internally, so the plugin-owned read honours
  the same channel and endpoint validation rather than a bespoke GET (closing
  the probe's `RequireHttps` gap).
- The facts come from that response's raw body via the existing pure parsers
  (`PkceDiscovery.SupportsS256`, `OidcResponseIssuer.DiscoveryAdvertisesResponseIssuer`);
  the `ProviderInformation` is built from the same response with the field-for-field
  mapping `OidcClient.EnsureProviderInformationAsync` performs internally, then
  fed to `PrepareLoginAsync` via `OidcClientOptions.ProviderInformation` (which
  sets the library's internal use-discovery flag false, suppressing its second
  fetch) and reused again at the callback (#247).
- **Fail closed:** an unreadable document returns `Unavailable` and the
  challenge is refused, never proceeding on a tolerant default. This is not a
  new lockout, without discovery, `PrepareLoginAsync` could not build the
  authorize redirect either. Because the facts are now definite whenever the
  document is readable, the `PkceUnverifiable` reason is removed.
- Removes `OidcDiscoveryCache` (a stale entry could still diverge from the
  login). Discovery is now read once per challenge and cached nowhere, so the
  JWKS the callback validates against is never held beyond a single authorize
  state's lifetime (#247), the same or a narrower key-rotation window than
  before.

Closes #450

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. An unreadable discovery document fails
      the challenge closed; the RFC 9207 `iss` requirement can no longer be
      silently downgraded by a failed probe.
- [x] No secrets logged; secrets are redacted on config export. (No secret
      handling changed; the discovery URL and library error are the only logged
      values.)
      config persistence, or the release pipeline. (Four-lens adversarial gate,
      verdicts under Notes.)
      (`?.ReplaceLineEndings(string.Empty)` on the provider name and the
      library error string).

## Quality checklist

      testable unit (`OidcDiscoveryReader` + `OidcDiscoveryResult`). The
      fact-parsing helpers are reused unchanged.
- [x] The change adds no more code than the problem requires; it removes the
      `OidcDiscoveryCache` and the now-dead `PkceUnverifiable` reason.
- [x] Architecture-conformance tests pass; the delegated-OIDC-flow rule's
      `nameof` marker was updated to the renamed reader, and the
      mutable-keyed-state rule's retired-cache comment updated.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green (Debug and Release, 0
      warnings).
- [x] `dotnet test` is green (1113 tests).
- [x] Prettier is clean for the `.md` change (`ARCHITECTURE.md`).

## Notes

**Adversarial-review gate, 4 lenses, all PASS** (refute-by-default; the
diff touches the OIDC login path, so this gate is the primary verification).
Each independently decompiled the pinned Duende libraries to check the claims.

- **bypass / fail-open**, VERDICT: sourcing both facts and the login metadata
  from one discovery response removes the probe/login divergence, and a failed
  read fails closed instead of silently downgrading `iss`; could not construct
  any refutation where enforcement is weakened relative to `origin/main`, the
  change strengthens it. **PASS**
- **availability / mass-lockout**, VERDICT: a correct fail-closed
  single-discovery refactor with no config-migration, mass-lockout,
  cache-growth, or rate-limit regression; the sole behavioural narrowing is the
  login-critical discovery/JWKS timeout dropping from the ~100s library default
  to 10s for pathologically slow IdPs, which is transient, self-healing, and a
  defensible DoS trade-off (comment updated to state it). **PASS**
- **correctness (RFC 9207 / logic)**, VERDICT: a behaviour-preserving,
  security-improving refactor, the reader's `ProviderInformation` is
  field-identical to the library's internal build, RFC 9207 anchoring and the
  `bool?`→`bool` narrowing are correct and strictly fail-closed, with only a
  benign 400→500 status shift on a null-endpoint + `RequirePkce`
  misconfiguration (both fail closed). **PASS**
- **integration (library contract)** - VERDICT: the `ProviderInformation`
  mapping is field-for-field identical to `OidcClient.EnsureProviderInformationAsync`
  in 7.1.0, pre-assignment correctly forces the internal use-discovery flag
  false into the same validation, the discovery call and policy object are the
  exact ones the library uses (and stronger than the old raw probe), `Raw` is
  reliably populated on success, and the HTTP wiring is unchanged, no contract
  mismatch. **PASS**

**Out of scope / follow-ups.**

- A minor pre-existing edge (an empty-body 200 with admin `DoNotValidateIssuerName`,
  or a null `OidEndpoint` with `RequirePkce`) surfaces as a 500 rather than a
  clean 400. Both still fail closed (no login, no `iss` downgrade); left as-is
  to keep the diff scoped, since the library threw identically before #450.
- **E2E checklist:** the library cannot be CI-exercised against real IdPs.
  Before release, confirm one real login each for (a) a provider whose
  `OidEndpoint` is the full `.well-known` URL (PocketID-style), (b) Google /
  off-authority endpoints via `AdditionalEndpointBaseAddresses`, and (c) a
  templated / multi-tenant issuer with `DoNotValidateIssuerName`. All use the
  same policy object, so they should behave identically to before, but they
  exercise the reader's now-policy-validated fetch on the login path.


iderex:

Scope: sources the RFC 9207 response-iss requirement (#210) and the PKCE-S256 fact (#141) from the single discovery response the login itself uses, replacing the separate best-effort probe. Fail-closed: an unreadable discovery document refuses the challenge rather than silently downgrading enforcement. Removes the OidcDiscoveryCache and the now-dead PkceUnverifiable reason.

Gates:
- Local build: Debug and Release, --no-restore --warnaserror, 0 warnings.
- Tests: 1113 passing (added OidcDiscoveryReaderTests; added challenge single-fetch and fail-closed-without-RequirePkce tests; updated the conformance marker to the renamed reader).
- Prettier: clean (ARCHITECTURE.md).
- CI: all 14 required/non-required checks green (build, CodeQL analyze x3, dependency-review, prettier, zizmor, greppable invariants, unicode-guard, JPRM package).
- Adversarial four-lens gate (primary verification for this login-path change), all PASS; each lens decompiled the pinned Duende libraries to verify the ProviderInformation mapping is field-for-field identical to OidcClient internal discovery and that pre-assigning it suppresses the library second fetch. Bypass: enforcement cannot be silently weakened; strictly stronger than main. Availability: no mass-lockout; the only narrowing is the login-critical discovery timeout dropping to 10s for pathologically slow IdPs (transient, self-healing, documented). Correctness: bool? to bool narrowing and RFC 9207 anchoring correct and fail-closed. Integration: no library contract mismatch.

Out of scope: a pre-existing empty-body / null-endpoint edge surfaces as a 500 instead of a 400 (still fail-closed). E2E: before release, confirm one real login each for a full-.well-known authority (PocketID), off-authority endpoints (Google), and a templated issuer (DoNotValidateIssuerName).

iderex:
Pre-merge review, gates below, all green; ready. Security refactor on the OIDC login path.

Four-lens gate all PASS (each decompiled the pinned Duende libs to verify):
- bypass/fail-open: enforcement cannot be silently weakened, the RFC 9207 iss-required and PKCE-S256 facts are now inseparable from the login own discovery (single GetDiscoveryDocumentAsync under the provider DiscoveryPolicy, the same call OidcClient uses internally), and any read failure fails closed (challenge refused), strictly stronger than the prior separate best-effort probe that silently left ResponseIssuerRequired=false. Also closes the old raw-GET probe RequireHttps gap.
- availability/mass-lockout: no config-migration/mass-lockout/cache-growth regression; sole narrowing is the discovery timeout dropping to 10s for pathologically slow IdPs (transient, self-healing, documented).
- correctness/RFC 9207: ProviderInformation field-identical to the library internal build; bool?→bool and iss anchoring correct and fail-closed.
- integration/library-contract: mapping field-for-field identical to EnsureProviderInformationAsync; pre-assigning options.ProviderInformation forces the same validation and suppresses the second fetch; no contract mismatch.

Two pre-existing benign edge cases documented (empty-body 200 / null OidEndpoint → 500 instead of 400, both fail-closed), judged not worth fixing. Discovery cached nowhere, so the callback JWKS window is the same or narrower.

CI: all 14 checks green (build, CodeQL x3, dependency-review, prettier, zizmor, greppable invariants, unicode-guard, JPRM, ABI floor). 1113 tests; Debug+Release --warnaserror 0 warnings. Real-IdP E2E items in the PR Notes.

Issue #550, opened 2026-07-18
Harden SecretStore key handling: orphan-prevention + atomic 0600 temp key (follow-up to #158)


iderex:
Follow-up hardening from the adversarial review of #158 (secrets encrypted at rest). Both are narrow, fail-closed-adjacent refinements the 4-lens gate rated as non-blocking notes, not defects â€” the current behavior is safe, these make it safer.

## 1. Refuse to mint a new data-encryption key when the config already holds envelopes

`SecretStore.GetOrCreateKey` silently generates a fresh key when `sso-secret.key` is missing. In the steady state this is only reached on a genuine first-run (no envelopes exist yet), so it is correct. But in the narrow double-fault where the key file was lost AND a still-plaintext secret is present, the next persist calls `GetOrCreateKey` -> creates a NEW key -> encrypts the plaintext under it, while any secrets already encrypted under the OLD (lost) key become permanently un-decryptable (orphaned) rather than surfacing the loss.

Proposal: when encrypting, if the configuration already contains at least one well-formed `ssoenc:` envelope, treat a missing key file as fail-closed (throw, same as `Reveal` does) rather than regenerating â€” the presence of envelopes proves a key existed and was lost. This needs the encrypt path to know whether any envelope is present (pass a flag from `ConfigSecretProtection.ProtectAll`, which already iterates every value). Keep the true first-run (no envelopes anywhere) creating a key as today.

## 2. Create the temp key file with owner-only mode atomically

`SecretStore.CreateKeyLocked` does `File.WriteAllBytes(tempPath, key)` and then `TryRestrictToOwner(tempPath)` (chmod 0600 on Unix). Between the two there is a brief window where the temp key file exists at the umask default (potentially group/world-readable) inside the plugin data folder. Exploiting it requires pre-existing filesystem read of that folder, which already defeats the whole at-rest scheme, so this is low severity â€” but it is worth closing by creating the file with `UserRead | UserWrite` at creation time (e.g. open a `FileStream` with a `UnixFileMode` on creation) rather than tightening after the write.

Both were surfaced by the security-bypass / availability lenses on #158; neither blocks that PR. Files: `SSO-Auth/Api/SecretStore.cs`, `SSO-Auth/Api/ConfigSecretProtection.cs`.

Issue #158, opened 2026-07-12
Encrypt secrets at rest (AES-256-GCM envelope, transparent migration)


iderex:
Provider secrets (OidSecret, signing key material) are stored in plaintext in the plugin configuration XML. Encrypt them at rest with an AES-256-GCM envelope (prefix `ssoenc:v1:`), a SecretStore key file with restrictive permissions outside the config, and transparent read-time migration of existing plaintext values. This is a breaking change to the on-disk format and ships in its own late release with a documented rollback note.

Acceptance criteria:
- [ ] Secrets are written as `ssoenc:v1:` envelopes; plaintext secrets are migrated transparently on first load and rewritten encrypted.
- [ ] The key file lives outside the config, is created with least-privilege permissions, and its absence fails closed (no silent plaintext fallback).
- [ ] Decryption failure is surfaced, never silently ignored.
- [ ] Downgrade path documented (config N-1 compatibility); shipped as its own release.

BREAKING - own release, late. Depends on the config export redaction work.

iderex:
Unblocked and scheduled for the P2 close-out. We are delivering this now (AES-256-GCM envelope, ssoenc:v1: prefix, SecretStore key file outside the config with restrictive permissions, transparent read-time migration of existing plaintext, redaction on export). It is a breaking on-disk-format change, so it ships with a documented rollback note (a downgrade cannot read ssoenc: values). Design-first with a full four-lens adversarial review (fail-closed on a missing/corrupt key, never fall back to plaintext; fresh 96-bit nonce per encryption). Porting the proven envelope/migration pattern from the private jellyfin-plugin-sso-V2 reference. Completing this brings P2 to 100% and gates the 4.0.0.5 security patch.

PR #554, opened 2026-07-18
Bind OpenID canonical links to the discovered issuer

### LINKED ISSUE #186: Bind OpenID canonical links to the discovered issuer
Follow-up from the #155 review (OIDC domain reviewer, R1). OpenID canonical links are keyed on `sub` within a per-provider dictionary. That composite is equivalent to the OIDC-recommended `(iss, sub)` pair only while the provider entry keeps pointing at the same IdP. If an admin repoints an existing provider entry (same provider name) at a different IdP via `OID/Add`, the accumulated `sub`-keyed links persist across the issuer change: a user at the new IdP whose `sub` collides with an old link (realistic for IdPs minting short numeric subjects like "1") is silently mapped onto the old user's Jellyfin account.

Requires admin action plus a subject collision, and the exposure is no worse than the pre-#155 model or the SAML side, a narrow residual, not a regression. Two candidate fixes:

- Store the discovered issuer alongside the links (or a per-link issuer) and refuse a link whose recorded issuer does not match the current discovery issuer (fail closed, self-healing on repoint), or
- Clear/invalidate `CanonicalLinks` when `OidEndpoint` changes in `OID/Add` (simpler, destructive but explicit).

The `DoNotValidateIssuerName` escape hatch slightly widens the surface (sub-issuer binding relaxes to "signed by the discovered keyset"); consider whether the chosen fix should warn when that toggle is set.

Acceptance criteria:
- [ ] A provider entry repointed at a different IdP cannot resolve links minted under the previous issuer (fail closed), with a test.
- [ ] The chosen mechanism (issuer-stamped links or clear-on-endpoint-change) is documented in providers.md.
- [ ] Migration/compat: existing links keep working while the provider endpoint is unchanged.

Refs #155, #134, #140.

iderex:
# What & why

Closes #186.

OpenID canonical links were keyed on `sub` within a per-provider dictionary
(`OidConfig.CanonicalLinks`). That composite equals the OIDC-recommended
`(iss, sub)` pair only while the provider entry keeps pointing at the same
identity provider. If an admin repoints an existing provider entry (same
provider name) at a **different** IdP, the accumulated `sub`-keyed links persist
across the issuer change, so a user at the new IdP whose `sub` collides with an
old link (realistic for IdPs minting short numeric subjects like `1`) is
silently mapped onto the old user's Jellyfin account.

We bind each link to the issuer it was minted under and enforce it, with two
complementary guards plus transparent migration:

- **Per-link issuer store.** A new server-managed `OidConfig.CanonicalLinkIssuers`
  (`sub -> issuer`), parallel to `CanonicalLinks`: `[JsonIgnore]` (withheld from
  every JSON response, cannot be set via a config PUT) and persisted in the
  config XML. The issuer is the **validated id_token `iss`**, read from the raw
  token, OidcClient filters the standard protocol claims (`iss`, `aud`, `exp`, …)
  out of the redeemed principal, so we read it from `result.IdentityToken` the
  same way the RFC 9207 mix-up check already does, and threaded through the
  verified identity into the link workflow.
- **Login-time issuer check (non-inert).** A resolved subject link whose stored
  issuer does not match the login's issuer **refuses the login** (fail closed).
  This catches a swapped IdP even behind an **unchanged** discovery URL (a
  DNS/backend repoint), which the endpoint-change belt cannot see.
- **Clear-on-endpoint-change belt.** Changing a provider's `OidEndpoint` (admin
  form or `OID/Add`) drops that provider's links and issuer bindings, treating
  the URL change as a provider re-identification, the same way the stored client
  secret is dropped on an endpoint change. This protects even un-stamped legacy
  links against a repoint done through the endpoint.
- **Trust-on-first-use migration.** Links minted before this store carry no
  issuer; they keep working while the endpoint is unchanged and are stamped with
  the current issuer on the next successful login. No userbase lockout on upgrade.

`DoNotValidateIssuerName` relaxes the binding to the token's self-declared issuer
(still audited). Documented in `providers.md`, including an operator note that an
endpoint edit is destructive to links and to plan a re-link/adoption window.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

## Quality checklist

      testable unit.

## Verification

- [x] `dotnet test` is green (1173 passed, Debug and Release).

### Design summary

- **Issuer captured** from the validated id_token's `iss` via
  `OidcResponseIssuer.IdTokenIssuer(result.IdentityToken)` at the OpenID callback
  (not from `result.User`, which OidcClient strips protocol claims from), carried
  on `OidcAuthorizeState.Issuer` → `VerifiedIdentity.Issuer` →
  `LoginCompletionService` → `CanonicalLinkService.ResolveOrCreateAsync(issuer)`.
- **Stored** in `OidConfig.CanonicalLinkIssuers` (`sub -> issuer`), server-managed
  exactly like `CanonicalLinks`.
- **`ClassifyIssuer` states:** `NotBound` (SAML / no subject link), `Match`
  (stored == login, ordinal), `Absent` (no stored issuer → trust-on-first-use),
  `Mismatch` (differs, including a login with no issuer against a stamped link).
- **Refuse path:** a `Mismatch` on a resolved subject link throws
  `AccountLinkForbiddenException` before any create/adopt/migrate, the identity
  is not mapped onto the old account.
- **Clear-on-repoint belt:** `ServerManagedFields.Preserve(OidConfig,…)` clears
  both maps when `OidEndpoint` changes, preserves both when unchanged.
- **`DoNotValidateIssuerName` warning:** documented in `providers.md`; the toggle
  is already recorded via `SsoAudit.InsecureOptionsEnabled` at save/`OID/Add`.

### How this resolves the prior review rejection (inert + mass-lockout)

A prior take on this issue was rejected as "inert at runtime + mass-lockout risk".
Both are resolved, and proven empirically, not asserted:

- **Non-inert.** The issuer check fires at runtime and refuses a repointed
  provider's old-issuer links. `CanonicalLinkServiceTests.ResolveOrCreateAsync_SubjectLinkStampedWithADifferentIssuer_RefusesFailClosed`
  pins the rejection, and the end-to-end
  `SSOControllerOidPostTests.OidPost_FullLogin_StampsTheRealIdTokenIssuerOntoTheCanonicalLink`
  drives a **real signed id_token** through the full callback + authenticate legs
  and asserts the token's actual `iss` reaches the stamp. That end-to-end test
  caught a genuine bug mid-implementation: the issuer was first sourced from
  `result.User.Claims`, which OidcClient strips `iss` from, so it was switched to
  the raw `result.IdentityToken`. Without the empirical test this would have
  shipped inert.
- **No mass-lockout on upgrade.** At upgrade every link is un-stamped, so
  `ClassifyIssuer` returns `Absent` → the login is stamped (trust-on-first-use)
  and proceeds; the refuse path is unreachable until a link is stamped **and** the
  issuer later changes. Pinned by
  `ResolveOrCreateAsync_UnstampedSubjectLink_TrustOnFirstUse_StampsAndProceeds`.
  The hot path (stamped, match) writes nothing (0-persist test).

### Accepted narrow residual (per the issue itself, not a regression)

A repoint that happened **before** upgrading to this version leaves an un-stamped
legacy link; on that user's first post-upgrade login (endpoint unchanged),
trust-on-first-use stamps the **new** issuer onto the old link rather than
catching it. The clear-on-endpoint-change belt covers every repoint from the
upgrade onward. This is strictly better than the current behaviour (where every
collision succeeded) and documented in `providers.md`.

### Adversarial review gate, four lenses (refute-by-default)

- **Availability, PASS.** No mass-lockout on upgrade (all links `Absent` →
  TOFU + proceed); the belt is admin-triggered and bounded, consistent with the
  existing secret-drop-on-endpoint-change; hot path writes nothing. It flagged an
  operator-doc gap (endpoint edits are destructive to links), added to
  `providers.md` (re-link / adoption-window guidance, `AllowExistingAccountLink`
  on vs off).
- **Security-bypass, PASS.** The check is non-inert; `iss` is inside the signed
  payload (not forgeable past the signature); null-issuer login against a stamped
  link is refused; create/adopt/migrate cannot reach a mismatched link; the #133
  race loser never overwrites a live stamped link; SAML unaffected. Notes the
  same-URL-swap-of-an-un-stamped-legacy-link window is the documented TOFU
  boundary above, not a regression.
- **Correctness, PASS.** Issuer capture is verbatim/ordinal-stable (same
  extraction stamps and compares); TOFU is lock-guarded and idempotent and fires
  only for a used subject link; write-path stamps only when the write happened;
  cleanup prunes only orphans (`ToList` before removal); belt handles null-vs-null.
  Two comment-accuracy nits it raised were fixed in this PR.
- **Integration, PASS (after fix).** XML round-trip, whole-config `Preserve` +
  belt, id_token issuer sourcing, discovery reuse (#450 undisturbed), and
  admin-form isolation all hold. It returned NEEDS_IMPROVEMENT for one concrete
  gap: `Controller_NeverTouchesProviderLinkMaps` scanned only `.CanonicalLinks`,
  not the new `.CanonicalLinkIssuers` map. **Fixed in this PR**, the fitness
  function now pins and scans both maps.

Out of scope: SAML (this issue is OpenID-only).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -154,12 +154,14 @@ public class ArchitectureConformanceTests
         // scattered cap/lifetime/sweep conventions. The former SSOController.DiscoveryFactsCache moved into a
         // *Cache type in #449 and was then removed entirely in #450 (discovery is now read once per challenge
         // and fed to the login, with nothing cached), so no discovery-facts dictionary remains to exempt.
-        // One documented exemption remains:
+        // Two documented exemptions remain, both persisted account-link config state:
         // - ProviderConfigBase._canonicalLinks: the persisted account-link map — serialized plugin
         //   configuration mutated only under the config lock, so a runtime store type would be the
         //   wrong home; it is config state, not in-flight state.
+        // - OidConfig._canonicalLinkIssuers: the per-link issuer binding (#186), the exact parallel of
+        //   _canonicalLinks — serialized config mutated only under the config lock, same rationale.
         var storeLike = new[] { "Store", "Cache", "Limiter" };
-        var exemptions = new[] { "ProviderConfigBase._canonicalLinks" };
+        var exemptions = new[] { "ProviderConfigBase._canonicalLinks", "OidConfig._canonicalLinkIssuers" };
 
         var offenders = PluginClasses
             .Where(t => !storeLike.Any(s => SimpleName(t).EndsWith(s, StringComparison.Ordinal)))
@@ -410,25 +412,37 @@ public class ArchitectureConformanceTests
         // the "call-level invariants stay with CodeQL" note in the class summary).
         //
         // Sentinel against a vacuous pass (#388): a zero-occurrence scan only means something while its
-        // target token still names the link map. A property rename (CanonicalLinks -> anything) would make
-        // the scan match nothing and pass for the wrong reason, so pin the property by reflection — a rename
-        // fails HERE and forces a conscious update of `linkMapProperty` (and the scanned token with it).
-        const string linkMapProperty = "CanonicalLinks";
-        Assert.True(
-            typeof(ProviderConfigBase).GetProperty(linkMapProperty, BindingFlags.Public | BindingFlags.Instance) is not null,
-            $"ProviderConfigBase.{linkMapProperty} was renamed or removed; point this rule at the new provider link-map property so the scan keeps guarding it (#388).");
+        // target token still names a link map. A property rename (CanonicalLinks -> anything) would make
+        // the scan match nothing and pass for the wrong reason, so pin each property by reflection — a
+        // rename fails HERE and forces a conscious update of the roster (and the scanned token with it).
+        // BOTH server-managed link maps are guarded: the account-link map (ProviderConfigBase.CanonicalLinks,
+        // #157) and its per-link issuer binding (OidConfig.CanonicalLinkIssuers, #186). Both are owned by
+        // CanonicalLinkService and ServerManagedFields.Preserve; a controller must touch neither directly.
+        var linkMapProperties = new[]
+        {
+            (Declaring: typeof(ProviderConfigBase), Name: "CanonicalLinks"),
+            (Declaring: typeof(OidConfig), Name: "CanonicalLinkIssuers"),
+        };
+        foreach (var (declaring, name) in linkMapProperties)
+        {
+            Assert.True(
+                declaring.GetProperty(name, BindingFlags.Public | BindingFlags.Instance) is not null,
+                $"{declaring.Name}.{name} was renamed or removed; point this rule at the new provider link-map property so the scan keeps guarding it (#388).");
+        }
 
-        var token = "." + linkMapProperty;
+        // The two tokens are disjoint substrings (".CanonicalLinkIssuers" does not contain ".CanonicalLinks"
+        // — the char after "Link" is "I", not "s"), so scanning for both cannot cross-match.
+        var tokens = linkMapProperties.Select(p => "." + p.Name).ToList();
         var linkMapLines = ControllerSourceFiles()
             .SelectMany(path => File.ReadAllLines(path)
                 .Select((line, index) => (File: Path.GetFileName(path), Text: line.Trim(), Number: index + 1)))
-            .Where(l => l.Text.Contains(token, StringComparison.Ordinal))
+            .Where(l => tokens.Any(t => l.Text.Contains(t, StringComparison.Ordinal)))
             .Select(l => $"{l.File} line {l.Number}: {l.Text}")
             .ToList();
 
         Assert.True(
             linkMapLines.Count == 0,
-            "A controller must not access a provider CanonicalLinks map directly; route link-map access through CanonicalLinkService and server-managed re-injection through ServerManagedFields.Preserve. Found: " + string.Join(" | ", linkMapLines));
+            "A controller must not access a provider link map (CanonicalLinks / CanonicalLinkIssuers) directly; route link-map access through CanonicalLinkService and server-managed re-injection through ServerManagedFields.Preserve. Found: " + string.Join(" | ", linkMapLines));
     }
 
     [Fact]
--- a/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
+++ b/SSO-Auth.Tests/CanonicalLinkServiceTests.cs
@@ -1137,6 +1137,263 @@ public class CanonicalLinkServiceTests
         await users.DidNotReceive().CreateUserAsync(Arg.Any<string>());
     }
 
+    // --- OpenID canonical-link issuer binding (#186) ---
+
+    private const string OldIssuer = "https://old-idp.example";
+    private const string NewIssuer = "https://new-idp.example";
+
+    [Fact]
+    public async Task ResolveOrCreateAsync_SubjectLinkStampedWithADifferentIssuer_RefusesFailClosed()
+    {
+        // The non-inert proof (#186, acceptance criterion 1): a subject-keyed link minted under one issuer
+        // must NOT resolve when the login presents a different one (the provider was repointed at another
+        // IdP behind the same discovery URL). The check FIRES at runtime and refuses — the prior review
+        // rejected an inert take, so this is the load-bearing test. No takeover: the old account is not
+        // handed to the new-IdP identity, and the stale link/issuer are left untouched for the admin.
+        var (service, cfg, users, log) = Build(c => c.OidConfigs["kc"] = new OidConfig
+        {
+            Enabled = true,
+            CanonicalLinks = new SerializableDictionary<string, Guid> { ["1"] = Existing },
+            CanonicalLinkIssuers = new SerializableDictionary<string, string> { ["1"] = OldIssuer },
+        });
+        users.GetUserById(Existing).Returns(UserNamed("alice", Existing));
+
+        await Assert.ThrowsAsync<AccountLinkForbiddenException>(() =>
+            service.ResolveOrCreateAsync(ProviderMode.Oid, "kc", "1", "mallory", allowExistingAccountLink: false, issuer: NewIssuer));
+
+        await users.DidNotReceive().CreateUserAsync(Arg.Any<string>());
+        Assert.Equal(Existing, cfg.OidConfigs["kc"].CanonicalLinks["1"]); // stale link untouched
+        Assert.Equal(OldIssuer, cfg.OidConfigs["kc"].CanonicalLinkIssuers["1"]); // stored issuer untouched
+
+        var warnings = log.Entries.FindAll(e => e.Level == Microsoft.Extensions.Logging.LogLevel.Warning);
+        Assert.Single(warnings);
+        Assert.Contains("issuer", warnings[0].Message, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public async Task ResolveOrCreateAsync_SubjectLinkStampedWithTheSameIssuer_ReusesItWithoutWriting()
+    {
+        // The hot path stays pure (#186): a login whose issuer matches the stored one reuses the link and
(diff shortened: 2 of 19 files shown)


iderex:

This is the last P2 code item before the 4.0.0.5 release and it is migration-sensitive, so the write-up is thorough.

## The problem

OpenID canonical links were keyed on sub within a per-provider dictionary. That composite is the OIDC-recommended (iss, sub) pair only while the provider entry keeps pointing at the same IdP. Repointing an entry (same provider name) at a different IdP left the accumulated sub-keyed links in place, so a new-IdP user whose sub collided with an old link (realistic for IdPs minting short numeric subjects like 1) was silently mapped onto the old user's account.

## The design

- Per-link issuer store: new server-managed OidConfig.CanonicalLinkIssuers (sub to issuer), parallel to CanonicalLinks, JsonIgnore + XML-persisted. The issuer is the validated id_token iss, read from the raw result.IdentityToken (OidcClient filters protocol claims out of result.User, the same reason the RFC 9207 mix-up check re-reads it there), threaded via OidcAuthorizeState.Issuer to VerifiedIdentity.Issuer to LoginCompletionService to CanonicalLinkService.ResolveOrCreateAsync.
- ClassifyIssuer states: NotBound (SAML / no subject link), Match (stored equals login, ordinal), Absent (no stored issuer, eligible for trust-on-first-use), Mismatch (differs, including a login with no issuer against a stamped link).
- Refuse path: a Mismatch on a resolved subject link throws AccountLinkForbiddenException BEFORE any create/adopt/migrate, so the new identity is not mapped onto the old account. This is the check that fires at runtime.
- Clear-on-repoint belt: ServerManagedFields.Preserve(OidConfig) clears both maps when OidEndpoint changes, preserves both when unchanged. Used by both the config-page save and OID/Add.
- Trust-on-first-use migration: un-stamped legacy links keep working while the endpoint is unchanged and are stamped with the current issuer on the next successful login.
- DoNotValidateIssuerName relaxes the binding to the token's self-declared issuer (still audited); documented in providers.md.

## How this resolves the prior inert + mass-lockout rejection

A prior take was rejected as inert at runtime + mass-lockout risk. Both are resolved and proven empirically, not asserted:

- Non-inert: the refuse test (CanonicalLinkServiceTests.ResolveOrCreateAsync_SubjectLinkStampedWithADifferentIssuer_RefusesFailClosed) pins the rejection, and the end-to-end test (SSOControllerOidPostTests.OidPost_FullLogin_StampsTheRealIdTokenIssuerOntoTheCanonicalLink) drives a REAL signed id_token through the full callback + authenticate legs and asserts the token's actual iss reaches the stamp. That end-to-end test caught a genuine bug mid-implementation: the issuer was first sourced from result.User.Claims, which OidcClient strips iss from, so it was switched to the raw id_token. Without the empirical test this would have shipped inert.
- No mass-lockout on upgrade: at upgrade every link is un-stamped, so ClassifyIssuer returns Absent to trust-on-first-use to proceed; the refuse path is unreachable until a link is stamped AND the issuer later changes. Pinned by ResolveOrCreateAsync_UnstampedSubjectLink_TrustOnFirstUse_StampsAndProceeds. The hot path (stamped, match) writes nothing (0-persist test).

## Accepted narrow residual (per the issue itself, not a regression)

A repoint that happened BEFORE upgrading leaves an un-stamped legacy link; on that user's first post-upgrade login (endpoint unchanged), trust-on-first-use stamps the new issuer onto the old link rather than catching it. The clear-on-endpoint-change belt covers every repoint from the upgrade onward. Strictly better than main (where every collision succeeded); documented in providers.md.

## Adversarial review gate, four lenses (refute-by-default)

- Availability: PASS. No upgrade mass-lockout (all links Absent to TOFU to proceed); belt is admin-triggered and bounded, consistent with the existing secret-drop-on-endpoint-change; hot path writes nothing. It flagged an operator-doc gap (endpoint edits are destructive to links) which is now in providers.md with re-link / adoption-window guidance for AllowExistingAccountLink on vs off.
- Security-bypass: PASS. Non-inert; iss is inside the signed payload (not forgeable past the signature); null-issuer login against a stamped link is refused; create/adopt/migrate cannot reach a mismatched link; the concurrency race loser never overwrites a live stamped link; SAML unaffected. Notes the same-URL-swap-of-an-un-stamped-legacy-link window as the documented TOFU boundary above, not a regression.
- Correctness: PASS. Issuer capture is verbatim/ordinal-stable (same extraction stamps and compares); TOFU is lock-guarded and idempotent and fires only for a used subject link; write-path stamps only when the write happened; cleanup prunes only orphans (ToList before removal); belt handles null-vs-null. Two comment-accuracy nits it raised were fixed in this PR.
- Integration: PASS after fix. XML round-trip, whole-config Preserve + belt, id_token issuer sourcing, discovery reuse (#450 undisturbed), and admin-form isolation all hold. It returned NEEDS_IMPROVEMENT for one concrete gap: Controller_NeverTouchesProviderLinkMaps scanned only .CanonicalLinks, not the new .CanonicalLinkIssuers map. Fixed in this PR: the fitness function now pins and scans BOTH maps.

## Gates

- Debug build --no-restore --warnaserror: 0 warnings, 0 errors.
- Release build --no-restore --warnaserror: 0 warnings, 0 errors.
- dotnet test: 1173 passed, 0 failed (Debug and Release), rebased onto current origin/main (which includes #158 secrets-at-rest and #550 SecretStore hardening; those files are disjoint from this change).
- Prettier clean for providers.md.

Out of scope: SAML (OpenID-only issue). Do not merge without the iderex sign-off.

iderex:
Pre-merge review, gates below, all green; ready. P2 Security-parity item (closes the sub-collision account-takeover on OIDC provider repoint). This is the migration-sensitive last P2 code item, so beyond the delivering agent 4-lens gate I ran an independent bypass lens over the COMMITTED diff.

Delivering-agent four-lens: availability PASS (flagged the endpoint-edit operator-doc gap → added the providers.md re-link/adoption-window note), security-bypass PASS, correctness PASS (2 comment nits fixed), integration NEEDS_IMPROVEMENT → PASS (the Controller_NeverTouchesProviderLinkMaps fitness function now pins BOTH CanonicalLinks and the new CanonicalLinkIssuers map, locking in the structural property).

Independent SECURITY-BYPASS lens (refute-by-default, committed diff): PASS. Confirmed:
- The stamped/compared iss is sourced from the RAW validated id_token (OidcResponseIssuer.IdTokenIssuer over result.IdentityToken, the same signed blob OidcClient already validated: signature vs discovery JWKS, ValidIssuer = discovery IssuerName, asymmetric-only algorithms so no alg downgrade). Not forgeable without controlling the current provider signing keys (the IdP itself). CanonicalLinkIssuers is [JsonIgnore] so an attacker cannot set the stored side via a config PUT.
- The refuse FIRES on the attack critical path: a colliding sub necessarily resolves bySubject, issuerVerdict is computed in the SAME locked read, and a Mismatch refuses BEFORE any migrate/adopt/create. Non-inert, and ClassifyIssuer states cannot be steered from Mismatch to Match/Absent/NotBound by attacker input.
- TOFU residual is documented, minimized by universal stamping + the endpoint-change belt, and strictly better than origin/main (which had no binding at all); the residual needs an external same-discovery-URL IdP swap (a real IdP change moves OidEndpoint → belt fires), not an admin config action.
- The endpoint-change belt is fail-closed (clears links + issuers), re-establishment runs the unchanged name-adoption policy; a benign save (endpoint unchanged, ordinal compare) preserves both maps. DoNotValidateIssuerName introduces no new bypass and is surfaced in the insecure-toggle + API-mutation audits.

Design: each OIDC link records the id_token iss it was minted under (new server-managed OidConfig.CanonicalLinkIssuers, [JsonIgnore]+XML). Mismatch on a resolved subject link refuses the login; un-stamped legacy links keep working and are trust-on-first-use stamped; endpoint change clears links+issuers. Non-inert + no-upgrade-mass-lockout proven by a rejection unit test AND an end-to-end real-signed-id_token test (which caught a real mid-build inert bug, iss first sourced from the claim-stripped principal).

CI: build/prettier/dependency-review (required) + CodeQL x3 + all others green. 1173 tests; Debug+Release --warnaserror 0 warnings. providers.md documents the mechanism, the accepted residual, the endpoint-edit note, and the DoNotValidateIssuerName caveat.

PR #559, opened 2026-07-18
Make EnabledProviderNames predicate null-safe

### LINKED ISSUE #538: GetNames EnabledProviderNames uses a non-null-safe predicate (NRE/500 on a null-valued config entry)
Surfaced while reviewing the #344 linking-page work. PR #535 (merged) introduced the shared `EnabledProviderNames<TConfig>` helper behind the anonymously reachable `OID/GetNames` / `SAML/GetNames` endpoints:

```csharp
configs.Where(kvp => kvp.Value.Enabled).Select(kvp => kvp.Key).ToList();
```

`kvp.Value.Enabled` dereferences the entry value directly. A null-valued entry in `OidConfigs` / `SamlConfigs` therefore throws an `NullReferenceException`, which surfaces as a 500 on the `GetNames` endpoints. This diverges from the fail-closed null-skip convention the codebase already applies to the same maps elsewhere, e.g. `CanonicalLinkService.LinksByUser` skips a null-valued entry rather than dereferencing it.

The empirically-established fact is that the predicate NREs on a null-valued entry (a review mutation probe seeding `OidConfigs["broken"] = null` fails against `kvp.Value.Enabled`). What we have not confirmed is whether a null-valued entry is actually reachable through the current persistence path (post-#350) or is only a defensive concern; either way the predicate should match the fail-closed convention.

## Fix

Use a null-safe property pattern so a null entry is skipped rather than dereferenced (this exact fix was written in the now-closed duplicate PR #534):

```csharp
configs.Where(kvp => kvp.Value is { Enabled: true }).Select(kvp => kvp.Key).ToList();
```

## Acceptance

- `EnabledProviderNames` skips a null-valued entry instead of throwing.
- A regression test seeds a null-valued config entry alongside an enabled provider and asserts the endpoint returns 200 with only the enabled name, for both the OID and SAML twins.
- We record whether a null-valued entry is reachable in the current persistence path or is defensive-only.

iderex:
# What & why

`EnabledProviderNames<TConfig>` (shared by the anonymously reachable `OID/GetNames`
and `SAML/GetNames` endpoints, introduced in PR #535) dereferenced `kvp.Value.Enabled`
directly:

```csharp
configs.Where(kvp => kvp.Value.Enabled).Select(kvp => kvp.Key).ToList();
```

A null-valued entry in `OidConfigs` / `SamlConfigs` (a plain `Dictionary<TKey,TValue>`,
so a null reference-type value is legal) throws a `NullReferenceException` there, which
surfaces as a 500 on an anonymous endpoint. This diverged from the fail-closed
null-skip convention the codebase already applies to the same maps elsewhere (see
`CanonicalLinkService`, which treats a null-valued entry as unusable rather than
dereferencing it).

We switch to a null-safe property pattern so a null entry is skipped instead of
dereferenced:

```csharp
configs.Where(kvp => kvp.Value is { Enabled: true }).Select(kvp => kvp.Key).ToList();
```

Closes #538

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (A null-valued entry is now skipped —
      the same conservative outcome the endpoint already applies to a
      disabled entry — rather than crashing the request.)
- [x] No secrets logged; secrets are redacted on config export. (Unchanged - 
      this endpoint returns provider names only, never secrets.)
      config persistence, or the release pipeline. (Light correctness +
      bypass self-review run given the small, well-scoped diff on an
      already-anonymous, read-only listing endpoint; both lenses returned
      PASS — see Notes.)
- [ ] Log inputs from the IdP are sanitized inline at the log call. (N/A - no
      logging in this diff.)

## Quality checklist

      testable unit.
      (No new structural property introduced — this is a one-line predicate
      fix inside an existing private helper.)

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. (Both Debug and Release.)
- [x] `dotnet test` is green. (1175/1175, including two new regression tests.)
- [ ] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (N/A - 
      C# only.)

## Notes

We ran a light correctness + bypass self-review (not the full four-lens gate)
given the diff is a one-line predicate change confined to a read-only,
already-anonymous listing helper with no bearing on the login/callback path:

- **Correctness lens, PASS.** Confirmed `is { Enabled: true }` is a
  non-throwing null check (verified empirically), eliminates the only
  null-dereference path in this method and its two callers, and is
  semantically equivalent to the existing fail-closed convention already
  used in `CanonicalLinkService`. The two new tests reproduce the real
  crash trigger and would catch a regression.
- **Bypass lens, PASS.** Omitting a null-valued entry is strictly
  fail-closed (a null entry was never a usable, enabled provider), adds no
  new information disclosure (a clean 200 with a shorter list discloses
  less than an unhandled 500 would), does not touch any
  authentication/authorization decision (`GetNames` is read-only), and a
  null-valued entry cannot itself be attacker-injected (only reachable via
  a malformed config file or the elevation-gated `Set` endpoints).

We record, per the issue's acceptance criteria: whether a null-valued entry
is reachable through the current persistence path is not re-derived here - 
`CanonicalLinkService`'s existing null-skip comments already document it as
a real, reachable state (not merely defensive), which this fix now handles
consistently on the `GetNames` read path too.

This targets `main` (the released line, the bug shipped in the 4.1.0.0
candidate). After merge it needs a forward-merge from `main` into `4.2`.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerEndpointTests.cs
+++ b/SSO-Auth.Tests/SSOControllerEndpointTests.cs
@@ -159,6 +159,22 @@ public class SSOControllerEndpointTests
     }
 
     [Fact]
+    public void OidProviderNames_SkipsNullValuedEntry()
+    {
+        // A null-valued entry must not NRE the anonymous GetNames endpoint into a 500 (#538) — it is
+        // skipped, the same fail-closed treatment CanonicalLinkService already applies to these maps.
+        var harness = new SsoControllerHarness(c =>
+        {
+            c.OidConfigs["keycloak"] = new OidConfig { Enabled = true };
+            c.OidConfigs["broken"] = null;
+        });
+
+        var result = Assert.IsType<OkObjectResult>(harness.Controller.OidProviderNames());
+        var names = Assert.IsType<List<string>>(result.Value);
+        Assert.Equal(new[] { "keycloak" }, names);
+    }
+
+    [Fact]
     public void SamlProviderNames_ReturnsEnabledNames()
     {
         var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig { Enabled = true });
@@ -176,6 +192,21 @@ public class SSOControllerEndpointTests
         {
             c.SamlConfigs["adfs"] = new SamlConfig { Enabled = true };
             c.SamlConfigs["disabled-idp"] = new SamlConfig { Enabled = false };
+        });
+
+        var result = Assert.IsType<OkObjectResult>(harness.Controller.SamlProviderNames());
+        var names = Assert.IsType<List<string>>(result.Value);
+        Assert.Equal(new[] { "adfs" }, names);
+    }
+
+    [Fact]
+    public void SamlProviderNames_SkipsNullValuedEntry()
+    {
+        // SAML twin of OidProviderNames_SkipsNullValuedEntry (#538).
+        var harness = new SsoControllerHarness(c =>
+        {
+            c.SamlConfigs["adfs"] = new SamlConfig { Enabled = true };
+            c.SamlConfigs["broken"] = null;
         });
 
         var result = Assert.IsType<OkObjectResult>(harness.Controller.SamlProviderNames());
--- a/SSO-Auth/Api/SSOController.cs
+++ b/SSO-Auth/Api/SSOController.cs
@@ -330,10 +330,12 @@ public class SSOController : ControllerBase
     }
 
     // Names of the enabled providers in a config map, materialized to a detached list (the caller holds
-    // the config lock). Shared by both GetNames twins so the enabled-only rule lives in one place.
+    // the config lock). Shared by both GetNames twins so the enabled-only rule lives in one place. A
+    // null-valued entry is skipped rather than dereferenced (#538) — the same fail-closed convention
+    // CanonicalLinkService already applies to these maps.
     private static List<string> EnabledProviderNames<TConfig>(SerializableDictionary<string, TConfig> configs)
         where TConfig : ProviderConfigBase =>
-        configs.Where(kvp => kvp.Value.Enabled).Select(kvp => kvp.Key).ToList();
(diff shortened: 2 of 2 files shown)


iderex:

Fix: EnabledProviderNames<TConfig> now skips a null-valued config-map entry (kvp.Value is { Enabled: true }) instead of dereferencing it, closing the NRE-to-500 on the anonymous OID/GetNames and SAML/GetNames endpoints. Matches the fail-closed null-skip convention CanonicalLinkService already applies to the same maps.

Gates:
- Local build (Debug + Release, --warnaserror): green, 0 warnings.
- dotnet test: 1175/1175 green, including the two new regression tests (OidProviderNames_SkipsNullValuedEntry, SamlProviderNames_SkipsNullValuedEntry).
- Light correctness + bypass self-review (not the full four-lens gate, given the one-line predicate scope): both PASS. Correctness confirmed the pattern is a non-throwing null check with no residual dereference path; bypass confirmed the omission is strictly fail-closed, discloses no more than before, and touches no auth/authz decision (GetNames is read-only, not on the login/callback path).
- CI: build, prettier, dependency-review, CodeQL, zizmor, and the rest all green.

Base is main (the released line, the bug shipped in the 4.1.0.0 candidate). Needs a forward-merge from main into 4.2 after this lands.

Not merging, leaving for approval per the standing process.

iderex:
Pre-merge review, all gates green; ready. One-line null-safety fix in the shared EnabledProviderNames helper (kvp.Value is { Enabled: true }), matching the fail-closed null-skip convention CanonicalLinkService already applies to the same maps; a null-valued config entry no longer 500s the anonymous GetNames endpoints and is simply omitted. Two regression tests. Correctness + bypass self-review both PASS (non-throwing check, strictly fail-closed, no new exposure, read-only listing path). CI all green; 1175 tests; Debug+Release --warnaserror 0 warnings. Folds into 4.1.0.0 (tag pending); forward-merge main→4.2 follows with the batch.

PR #563, opened 2026-07-18
Reserve the SAML outcome-store slot before the replay consume

### LINKED ISSUE #539: SAML: an assertion consumed at the ACS callback cannot be retried if SamlOutcomeStore then refuses at its cap
Surfaced by the adversarial review while rebasing the #470 SAML replay-cache work onto the just-merged #531 / #251 (`Replace the SAML assertion browser round-trip with a one-time outcome token`).

In the new outcome-token flow the ACS callback consumes the assertion ID against the replay cache (one-time use) and then stores the login outcome in `SamlOutcomeStore`. If `SamlOutcomeStore.TryAdd` refuses at its cap **after** the replay cache has already recorded the assertion, that login cannot be retried with the same assertion, the replay cache now rejects it, so the user is fail-closed out of a completed authentication even though nothing was malicious. The blast radius is bounded to periods of genuine outcome-store capacity pressure.

This is a property of #531's design, not a regression introduced by #470.

## Options to weigh

- Reserve the outcome-store slot before the replay consume, so a cap refusal does not burn the assertion.
- Or make the cap refusal at that point surface a retryable outcome.

Fail-closed is correct on the security axis; this is an availability-edge refinement.

## Acceptance

- A decision recorded: accept the edge (documented) or reorder so a completed authentication is not lost to an outcome-store cap refusal.
- If reordered, a test covers the consume-then-cap-refusal path.

iderex:
# What & why

Closes #539.

In the one-time outcome-token SAML flow (#251) the ACS callback consumed the assertion's one-time replay-cache id and only *then* stored the login outcome in `SamlOutcomeStore`. When the store refused at its hard cap **after** the replay consume, the assertion was already burned (one-time use), so the completed authentication could never be retried, the legitimate user was locked out under outcome-store capacity pressure even though nothing malicious happened. The issue offered two options; we took the reorder (reserve the store slot before the consume) because it fixes the availability edge without touching the replay semantics.

What we changed:

- Split `SamlOutcomeStore.TryAdd` into three primitives, `TryReserve` (per-client sub-cap + global cap, no token needed), `CommitReserved`, and `ReleaseReservation`. `TryAdd` now composes them, preserving its exact contract for the tests and any atomic caller.
- The login-path ACS callback now reserves an outcome-store slot **first**; only on a successful reserve does it consume the replay cache, build the verified identity, and commit the outcome. A capacity refusal now fails closed *before* the assertion is burned, so the assertion stays valid and the user can retry once the store drains.
- A `try/finally` around the reserve→commit region releases the reservation unless it was handed to a committed outcome, so "exactly one release per reservation" is structural, it covers every early return (a genuine replay, a NameID-less assertion) and any unexpected throw, so no per-client slot can leak.

Replay protection is unchanged: the one-time consume is still the atomic claim and now runs only once a slot is secured. A committed login always consumed the assertion, and a consumed assertion always completed, a genuine replay still fails closed and mints nothing. Under capacity pressure a replayed assertion is refused at the cap (500-retry) before the consume and, once the store drains, is rejected by the consume itself (400), fail-closed either way, no session minted.

## Type of change

- [x] Bug fix
- [ ] Feature

(Bug fix on the SAML login path with a hardening character, an availability edge, not an exploitable vulnerability.)

## Security checklist

## Quality checklist

      testable unit.

(No new structural property is established, the reserve/commit split is a store-internal refinement of the existing keyed-state store type; the existing conformance rules still hold and pass.)

## Verification

- [x] `dotnet test` is green.

Debug and Release both build clean under `--warnaserror` (0 warnings); the full suite is 1177/1177 green in both configurations, including the new store-level reserve/commit/release tests and the end-to-end `CallbackRefusedAtOutcomeStoreCap_DoesNotBurnTheAssertion_RetrySucceeds_ReplayStillRejected`. Prettier is N/A (C# only).

## Notes

Adversarial review, the four-lens gate ran on the SAML login path + replay protection, all four PASS:

- **Availability**, PASS. A cap refusal leaves the assertion unconsumed and the login retryable; every reserve→commit exit path releases or commits exactly once; no new mass-lockout. It raised the reserve→commit throw window, which the `try/finally` now closes.
- **Security-bypass** (the critical lens), PASS. The one-time consume still gates every committed outcome; a replayed assertion mints nothing under capacity pressure and is rejected by the consume once the store drains; two concurrent presentations of the same fresh assertion resolve to a single winner (the loser releases and mints nothing); cross-provider/token-scoping guards unchanged.
- **Correctness**, PASS. The composed `TryAdd` preserves the old contract exactly; the dropped `!ContainsKey` sub-condition was inert for fresh CSPRNG tokens; per-client capping stays exact via CAS; the reserve↔commit global-cap TOCTOU only relaxes the documented approximate memory ceiling.
- **Integration**, PASS. The deprecation branch and the Link path are unaffected; the harness reset installs a fresh store so no cap/state leaks between tests; the 500-refusal keeps parity with the challenge/OpenID refusal.

All four independently flagged the same non-blocking robustness note (a per-client slot could leak if a call between reserve and commit threw); we addressed it with the `try/finally` that makes the release structural, so the note is resolved rather than accepted.

Out of scope: the pre-#251 deprecation branch in `AuthenticateAsync` (legacy full-assertion POST) and the manual-link redeem consume replay directly and never touch the outcome store, so they are unchanged.

After merge this needs the standing forward-merge `main` → `4.2`.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerSamlTokenTests.cs
+++ b/SSO-Auth.Tests/SSOControllerSamlTokenTests.cs
@@ -152,6 +152,45 @@ public class SSOControllerSamlTokenTests
         await harness.UserManager.Received(1).CreateUserAsync("alice");
     }
 
+    [Fact]
+    public async Task CallbackRefusedAtOutcomeStoreCap_DoesNotBurnTheAssertion_RetrySucceeds_ReplayStillRejected()
+    {
+        // #539: in the token flow the ACS callback consumes the assertion's one-time replay id and then stores
+        // the outcome. If the store refused AFTER the consume, a cap refusal would burn the assertion for a
+        // login that never completed and permanently lock out the legitimate user. Reserving the store slot
+        // BEFORE the consume fixes that: a cap refusal leaves the assertion untouched and the login retryable,
+        // while a genuine replay is still rejected.
+        var fixture = SamlTestFactory.Create(nameId: "alice");
+        var harness = ProvisioningHarness(fixture, out _);
+        var assertion = fixture.EncodeResponse();
+
+        // Install a cap-1 outcome store (the production 100k ceiling is unreachable in a unit test) and fill
+        // its single global slot with an unrelated in-flight outcome, so the next callback's reservation is
+        // refused. The filler uses a null (exempt) client key so it occupies the GLOBAL slot.
+        var store = new SamlOutcomeStore(1, TimeSpan.FromMinutes(15), TimeSpan.FromMinutes(1));
+        SamlLoginService.SetSamlOutcomeStoreForTests(store);
+        var filler = SamlOutcomeStore.NewToken();
+        var fillerIdentity = VerifiedIdentity.FromValidatedSaml("adfs", "filler", SamlAuthorizeStateBuilder.Build(new System.Collections.Generic.List<string>(), new SamlConfig()));
+        Assert.True(store.TryAdd(new SamlLoginOutcome(filler, "adfs", fillerIdentity, string.Empty, null, DateTime.UtcNow), out _));
+
+        // The callback is refused at the store cap — a fail-closed 500 — and the assertion's one-time replay
+        // id is NOT consumed, because the reservation is checked ahead of the consume.
+        var refused = Assert.IsType<ContentResult>(harness.Controller.SamlCallback("adfs", formSamlResponse: assertion));
+        Assert.Equal(500, refused.StatusCode);
+
+        // Drain the store (redeem the filler), then re-POST the SAME assertion. Because it was never burned,
+        // the retry now validates, consumes the id, and renders a real token — the login was not lost.
+        Assert.NotNull(store.TryRedeem(filler, "adfs", DateTime.UtcNow));
+        var token = ExtractToken(harness.Controller.SamlCallback("adfs", formSamlResponse: assertion));
+        Assert.IsType<OkObjectResult>(await harness.Controller.SamlAuth("adfs", new AuthResponse { Data = token }));
+        await harness.UserManager.Received(1).CreateUserAsync("alice");
+
+        // Replay protection is intact: re-POSTing the same assertion a third time (the store now has room) is
+        // refused by the one-time replay guard, since the retry above consumed the id.
+        var replay = Assert.IsType<ContentResult>(harness.Controller.SamlCallback("adfs", formSamlResponse: assertion));
+        Assert.Equal(400, replay.StatusCode);
+    }
+
     private const string AuthnRequestId = "_authnreq-251";
     private const string Binding = "251251251251251251251251251251251251251251251251251251251251251A";
 
--- a/SSO-Auth.Tests/SamlOutcomeStoreTests.cs
+++ b/SSO-Auth.Tests/SamlOutcomeStoreTests.cs
@@ -151,6 +151,47 @@ public class SamlOutcomeStoreTests
     }
 
     [Fact]
+    public void Reserve_ThenCommit_StoresARedeemableOutcome()
+    {
+        // The two-phase path the ACS callback uses (#539): reserve capacity first, then commit the built
+        // outcome once the assertion is consumed. The committed outcome redeems exactly like a TryAdd one.
+        var store = SmallStore();
+        Assert.True(store.TryReserve(clientKey: "A", Now, out var warn));
+        Assert.False(warn);
+        Assert.True(store.CommitReserved(Outcome("tok-1", clientKey: "A")));
+
+        Assert.NotNull(store.TryRedeem("tok-1", "adfs", Now));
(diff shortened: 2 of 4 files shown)


iderex:

The ordering fix. In the one-time outcome-token login flow the ACS callback did two things in this order: consume the assertion's one-time replay-cache id (inside TryProduceVerifiedIdentity), then register the login outcome in SamlOutcomeStore. The replay consume is one-time by design, an assertion recorded as used can never be presented again, so when the store refused at its hard cap on the second step, the assertion was already burned for a login that never completed. The legitimate user could not retry, and there is no malicious actor in that scenario: it is a pure availability loss under outcome-store capacity pressure.

We reordered so the capacity check happens before the assertion is burned. SamlOutcomeStore.TryAdd is split into TryReserve (the per-client sub-cap reservation plus the global-cap check, which needs no token), CommitReserved (the atomic insert of the built outcome), and ReleaseReservation. The callback now reserves a slot first; only on a successful reserve does it consume the replay cache, build the identity, and commit. A capacity refusal therefore fails closed at the reserve step, before the replay cache is touched, so the assertion is untouched and the login is retryable once the store drains. TryAdd is kept as a thin composition of the three primitives so its exact contract, and the store's own test suite, are unchanged.

Why replay protection is preserved. The reorder does not move, weaken, or duplicate the one-time consume; it only gates when the consume runs. Three properties hold after the change:

- The consume still runs for every completed login. TryReserve touches only the capacity counters, never the replay cache. The single path to CommitReserved runs through TryProduceVerifiedIdentity, whose first act is the atomic replay consume. So a redeemable, committed outcome always has its assertion id already consumed, there is no window where an outcome is redeemable without the consume having happened.

- A replayed assertion still mints nothing, on every branch. If the reserve succeeds and the assertion is a genuine replay, the consume fails, we release the reservation, and reject, nothing minted. If the store is at its cap, the reserve fails first and returns a fail-closed 500 before the consume; that mints nothing either, and when the store later drains the same assertion is rejected by the consume itself. The only observable change for a replay under capacity pressure is a 500-retry instead of a 400, which carries no session and no usable oracle. Fail-closed is preserved end to end.

- Concurrency is unchanged. Two concurrent presentations of the same fresh assertion may both reserve a slot, but the replay cache TryConsume is the single-winner atomic claim, so exactly one commits an outcome; the loser releases its reservation and mints nothing. The cross-provider and lifetime guards on redeem are untouched.

Release accounting is now structural. Everything between the reserve and the commit runs under a try/finally that releases the reservation unless it was handed to a committed outcome (which owns the slot until its redeem or prune). This makes exactly-one-release-per-reservation a property of the control flow rather than of remembering to release on each branch, and it closes the one robustness gap the review raised: an unexpected throw between reserve and commit can no longer leak a per-client slot.

Gates.

- Build: Debug and Release both clean under --warnaserror, 0 warnings.
- Tests: 1177/1177 green in both configurations. New coverage: store-level reserve-then-commit, global-cap reserve refusal via an exempt key (isolating the global branch from the sub-cap), and release-frees-the-slot; plus an end-to-end callback test that fills a cap-1 store, confirms the refused callback returns 500 without consuming the assertion, drains the store and proves the retry mints, then proves a third post of the same assertion is rejected 400 by the replay guard.
- Adversarial four-lens gate on the SAML login path and replay protection: availability PASS, security-bypass PASS (the critical lens, refuted hard, no replay hole), correctness PASS, integration PASS. All four independently flagged the reserve-to-commit throw window; we resolved it with the try/finally rather than accepting it.
- Prettier: N/A (C# only).

Scope. The pre-#251 deprecation branch in AuthenticateAsync (legacy full-assertion post) and the manual-link redeem consume replay directly and never touch the outcome store, so both are unchanged. After merge this needs the standing forward-merge from main into 4.2.

iderex:
Pre-merge review, all gates green; ready. SAML availability fix with replay protection provably intact. Design: SamlOutcomeStore.TryAdd split into TryReserve/CommitReserved/ReleaseReservation (TryAdd composes them, contract unchanged); the ACS callback reserves a slot BEFORE consuming the one-time replay id, so a capacity refusal fails closed before the assertion is burned and the login stays retryable once the store drains. No replay hole: every committed outcome still ran the atomic consume first; a replayed assertion mints nothing on any branch (cap refusal → fail-closed 500 pre-consume; post-drain → consume rejects 400); concurrent duplicates still resolve to one winner via the cache's atomic claim; a try/finally makes exactly-one-release structural (all four lenses independently flagged that throw window, resolved, not accepted). Four-lens gate all PASS. CI all green incl. CodeQL; 1177 tests incl. the end-to-end retry-after-cap + replay-still-rejected test; Debug+Release --warnaserror 0 warnings. Folds into 4.1.0.0; the standing forward-merge main→4.2 (#562) carries it.

Issue #564, opened 2026-07-18
linking.js also swallows non-2xx on the GetNames provider-list fetch


iderex:
While delivering #536 (which scoped its fix to exactly two XHRs, the links/{userId} existing-links load and the unlink DELETE flow), our correctness-lens review surfaced a third, adjacent instance of the same bug class in the same file.

loadProviders() in SSO-Auth/Config/linking.js issues a bare native `fetch(new Request(...GetNames))` for both provider dropdowns (SAML/OID), with no `.catch` and no status check. Native `fetch` does not reject on 4xx/5xx, so a failing GetNames call (auth expiry, a 500) either throws unhandled on `resp.json()` of a non-JSON error body, or silently renders an empty provider list, the same silent-failure problem #536 just fixed for the other two paths.

We deliberately left this out of #536: that issue's body explicitly scoped itself to "two of its own client XHRs" with a smaller, separate acceptance criteria set than #344's original fix, and pulling this third path in would have widened that PR's diff for a call site the issue didn't name.

Fix direction: switch loadProviders() to ApiClient.fetch (or check resp.ok on the native fetch) and reuse the ssoConfigLinking.showError() banner added by #536, so a failed provider-list load surfaces the same generic error instead of an empty section.

Acceptance:
- A failed GetNames load for either provider mode shows the error banner, not a silently empty provider-list section.
- No regression to the #344/#536 behavior (enabled-only add buttons, disabled-provider rows still render and remain removable).

Issue #412, opened 2026-07-16
Record the challenge NewPath through MutateConfiguration, not a post-lock callback


iderex:
Follow-up from PR #410 (CodeRabbit), pre-existing (not introduced by #337).

`ResolveChallengeNewPath` (SSO-Auth/Api/SSOController.cs) records the derived spelling via a callback `value => config.NewPath = value`, but `config` was read under `ReadConfiguration` and the callback runs after that lock is released. Concurrent OID/SAML challenges can race on the shared server-managed `NewPath` route state, and the mutation bypasses the `MutateConfiguration` write path every other config write goes through.

Worst case today is a fail-closed login failure (a wrong redirect_uri spelling the IdP rejects), not a bypass, hence low priority, but the shared-mutable-state pattern should be hardened: perform the route derivation and the provider lookup/update inside a single `MutateConfiguration` operation and use the returned value to build the redirect URI. Part of the #318 thin-controller direction.

Refs #337, #318.

iderex:
Overlaps #303, which tracks the same underlying design point from PR #302: the challenge records `NewPath` on the provider object outside the config lock, deliberately (a scalar flip that cannot tear serialization, readers accept either spelling) because routing it through `MutateConfiguration` would serialize and disk-write the whole config on every login challenge. This issue proposes exactly the `MutateConfiguration` side of that trade-off, so the two should be resolved together: the decision in #303 governs, and this one becomes its implementation (or is closed with the same rationale if we keep the documented in-place flip).

PR #565, opened 2026-07-18
Surface failed loads and unlink deletes on the linking page

### LINKED ISSUE #536: linking.js swallows non-2xx on its own load/unlink XHRs
linking.js still swallows non-2xx responses on two of its own client XHRs, so a server-side failure there shows no error to the user:

1. The `links/{userId}` existing-links fetch, a non-2xx (e.g. auth expiry, 5xx) leaves the existing-links list silently empty rather than surfacing that the load failed.
2. The unlink flow, `Promise.all(delete_requests).then(reload)` reloads unconditionally, so a failed DELETE (e.g. 4xx/5xx) is indistinguishable from success; the row just reappears after reload with no message.

This is a UX-honesty gap surfaced while fixing #344 (which covered the two defects in scope: disabled providers offered + the link-POST rejection swallowed). It is out of #344 scope and smaller.

Fix direction: check the XHR status on both paths and surface a clear message on failure (mirror the pattern #344 added to WebResponse.link()/main()); do not reload-as-success when a DELETE failed. Keep enabled-only add buttons and the disabled-provider removable-row behaviour from #344 intact.

Acceptance:
- A failed existing-links load shows an error state, not a silently empty list.
- A failed unlink DELETE shows an error and does not present as success.
- No regression to the #344 disabled-provider list/remove behaviour.

iderex:
# What & why

linking.js discarded a non-2xx response on two of its own client XHRs, so a server-side failure there showed no error to the user, a UX-honesty gap surfaced while fixing #344 but out of that issue's scope:

1. The `links/{userId}` existing-links fetch, a non-2xx (auth expiry, a 5xx) left the existing-links list silently empty rather than surfacing that the load failed.
2. The unlink flow, `Promise.all(delete_requests).then(reload)` reloaded unconditionally, so a failed DELETE (4xx/5xx) was indistinguishable from success.

We fixed both: a small static `#sso-linking-error` banner is added to `linking.html`/`style.css`, and a `showError()` helper unhides it. Both promise chains now wire a `.catch()` to it, the existing-links load surfaces the banner instead of leaving the list empty, and the unlink flow no longer reloads unless every DELETE actually succeeded. The banner carries no status code or server detail, only a static message, so a rejection cannot leak anything into the admin UI.

This page runs without the Jellyfin dashboard shell (`linking.html`'s own head comment says "Polyfill styles that are missing when serving without dashboard"), so `Dashboard.alert(...)`, the mechanism `config.js` uses elsewhere for the dashboard-embedded config page, is not available here; a local static banner is the correct mechanism for this page specifically.

The enabled-only add-button and disabled-provider removable-row behavior from #344 (PR #535) is untouched, only the promise-chain shape around the two named XHRs changed, plus the added banner markup/CSS.

Closes #536

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (N/A — no signature/time-bound/audience
      check in this diff; this is the client-side self-service linking admin
      page, not the login/callback path.)
- [x] No secrets logged; secrets are redacted on config export. (N/A - no
      secrets touched; the error banner is static text, never interpolated
      from a server response.)
      config persistence, or the release pipeline. (The adversarial-review
      gate ran because `linking.js`/`linking.html` sit on the flagged
      security-sensitive surface — see Notes for the 4-lens result; this
      diff touches none of SAML/OIDC validation, config persistence, or the
      release pipeline itself.)
      logging in this diff.)

## Quality checklist

      testable unit. (`showError()` is one small helper reused by both fixed
      call sites.)
- [x] The change adds no more code than the problem requires. (42 insertions
      across the two named call sites, the banner element, and its styling;
      the adjacent GetNames swallow found by review is filed separately as
      #564, not folded in here.)
      (No new structural property — this is a client-JS-only change; the C#
      architecture-conformance suite is unaffected and still green.)

## Verification

      0 warnings, 0 errors.)
- [x] `dotnet test` is green. (1179/1179 passed - a client-JS-only change, so
      this confirms no regression rather than exercising the fix itself.)
      (`npx prettier --check` reports the three touched files unchanged /
      clean; the pre-existing warnings on unrelated files predate this PR.)

## Notes

This is browser JS that CI cannot exercise end to end, so the adversarial-review
gate was the primary verification here (in addition to a manual trace of both
fixed call sites). We ran all 4 lenses:

- **Availability**, PASS. The pre-change unlink code had no `.catch` at all,
  so a failed DELETE already never reloaded before this PR; this change only
  makes that existing non-reload outcome visible instead of silent. No new
  mass-lockout or hang vector; `showError()` is idempotent and guarded against
  a missing banner element.
- **Security bypass**, PASS. The banner is strictly static text (no
  interpolation of status/body/provider name). Not reloading on a failed
  DELETE only causes client-side display staleness, every DELETE is still
  independently authorized and enforced server-side per request, so stale DOM
  state cannot grant anything the server wouldn't re-check. URL encoding
  (`encodeURIComponent`) and the disabled-provider gating from #344/#535 are
  untouched.
- **Correctness**, NEEDS_IMPROVEMENT, disposition: the two named call sites
  are verified correct (including tracing the vendored
  `jellyfin-apiClient.esm.min.js` to confirm `ApiClient.fetch(..., true)`
  resolves with a real `Response` object here, so `.then(resp => resp.json())`
  is valid and a malformed-JSON rejection now correctly reaches the trailing
  `.catch()`). The lens separately flagged that `loadProviders()`'s bare
  native `fetch()` call to `GetNames` also swallows non-2xx with no `.catch` - 
  a third instance of the same bug class, but one issue #536's own acceptance
  criteria explicitly did not name (it scoped itself to "two of its own client
  XHRs"). We filed this as #564 rather than widening this diff.
- **Integration**, PASS. No `.cs` file changed; the local banner (rather than
  `Dashboard.alert`) is correct because this page is confirmed to run without
  the dashboard shell; no id/class collisions; Prettier-clean; no inline
  script added.

Follow-up filed: #564 (GetNames provider-list fetch on this same page also
needs the same treatment, deliberately out of scope here).

Forward-merge reminder: this is a fix branched off `main`; it needs to
forward-merge into `4.2` after merge, per our branching model.

### DIFF
--- a/SSO-Auth/Config/linking.html
+++ b/SSO-Auth/Config/linking.html
@@ -46,6 +46,15 @@
               >${Help}</a
             >
           </div>
+          <div
+            id="sso-linking-error"
+            class="sso-linking-error"
+            role="alert"
+            hidden
+          >
+            Something went wrong loading or updating your links. Please reload
+            the page and try again.
+          </div>
           <p>
             Use the
             <span class="fab" aria-hidden="true">
--- a/SSO-Auth/Config/linking.js
+++ b/SSO-Auth/Config/linking.js
@@ -1,5 +1,15 @@
 const ssoConfigLinking = {
   pluginUniqueId: "505ce9d1-d916-42fa-86ca-673ef241d7df",
+
+  // A single generic banner for a failed request on this page (existing-links load or unlink, #536).
+  // It never carries a status code or server message, so a rejection cannot leak an internal detail
+  // into the admin UI; it just tells the user the page is no longer trustworthy as shown.
+  showError: () => {
+    const banner = document.querySelector("#sso-linking-error");
+    if (banner) {
+      banner.hidden = false;
+    }
+  },
   loadProviders: (view) => {
     ["oid", "saml"].forEach((provider_mode) => {
       const container = view.querySelector(
@@ -44,8 +54,9 @@ const ssoConfigLinking = {
           url: ApiClient.getUrl(`sso/${provider_mode}/links/${currentUserId}`),
         },
         true,
-      ).then((resp) => {
-        resp.json().then((provider_map) => {
+      )
+        .then((resp) => resp.json())
+        .then((provider_map) => {
           Object.keys(provider_map).forEach((provider_name) => {
             let existing_links = container.querySelector(
               `.sso-provider-existing-links-container[data-provider="${CSS.escape(provider_name)}"]`,
@@ -77,8 +88,10 @@ const ssoConfigLinking = {
               provider_map[provider_name],
             );
           });
-        });
-      });
+        })
+        // ApiClient.fetch rejects on a non-2xx status (auth expiry, a server error, ...), so a failed
+        // existing-links load surfaces the banner instead of silently leaving the list empty (#536).
+        .catch(() => ssoConfigLinking.showError());
     }
   },
 
@@ -223,9 +236,14 @@ const ssoConfigLinking = {
         });
       });
 
-    Promise.all(delete_requests).then((values) => {
-      window.location.reload();
-    });
+    Promise.all(delete_requests)
+      .then(() => {
+        window.location.reload();
+      })
+      // ApiClient.fetch rejects on a non-2xx status, so a rejected DELETE must not fall through to
+      // the unconditional reload below it: that would show the exact same page a successful removal
+      // shows, with no indication that the link the user asked to remove is still present (#536).
+      .catch(() => ssoConfigLinking.showError());
   },
 };
 
--- a/SSO-Auth/Config/style.css
+++ b/SSO-Auth/Config/style.css
@@ -35,6 +35,14 @@
   margin-top: 0.25em;
 }
 
+.sso-linking-error {
+  color: #ff8a80;
+  border: 1px solid #ff8a80;
+  border-radius: 0.25em;
+  padding: 0.75em 1em;
+  margin-bottom: 1em;
+}
+
 .sso-role-mapping-container,
 .sso-bordered-list {
   border-color: rgba(255, 255, 255, 0.135);


iderex:

Scope: SSO-Auth/Config/linking.js, linking.html, style.css only. No .cs changes.

Build/test: dotnet build --warnaserror clean on both Debug and Release (0 warnings, 0 errors). dotnet test: 1179/1179 passed (client-JS-only change, so this confirms no regression rather than exercising the fix directly). Prettier: the three touched files are clean; pre-existing warnings on unrelated files predate this branch.

Adversarial-review gate (4 lenses, since linking.js/linking.html sit on the flagged security-sensitive surface):
- Availability: PASS. The pre-change unlink path had no catch at all, so a failed DELETE already never reloaded before this fix, the change only makes that outcome visible instead of silent.
- Security bypass: PASS. The banner is static text only, no interpolation of status/body/provider name. Not reloading on a failed DELETE is display staleness only, every DELETE stays independently authorized and enforced server-side per request. Encoding and the disabled-provider gating from #344/#535 are untouched.
- Correctness: NEEDS_IMPROVEMENT. The two named call sites were verified correct, including tracing the vendored jellyfin-apiClient library to confirm ApiClient.fetch resolves with a real Response here. The lens separately found a third, adjacent swallow in loadProviders' GetNames fetch that #536's own acceptance criteria did not name. Disposition: filed as #564, not folded into this PR.
- Integration: PASS. No server contract touched; the local banner is correct given this page runs without the dashboard shell (confirmed Dashboard is not available here); no id/class collisions; CSP-safe, no inline script added.

Overall verdict: green, ready to merge. Forward-merge into 4.2 after merge, per our branching model. Not merging this ourselves per process, leaving it for approval.

iderex:
Pre-merge review, all gates green; ready. Both swallowed XHR paths in linking.js now surface a failure: the existing-links load .catch(showError) instead of rendering a silently-empty list, and the unlink flow no longer reloads-as-success on a rejected DELETE (ApiClient.fetch rejects on status >= 400, verified against the vendored client). A static #sso-linking-error banner carries no server detail. The #344/#535 enabled-only add-button + disabled-provider removable-row behavior is untouched. Four-lens gate: availability/bypass/integration PASS; correctness NEEDS_IMPROVEMENT flagged the adjacent GetNames XHR (out of #536's two-XHR scope) → filed follow-up #564, not widened here. CI all green incl. CodeQL; 1179 tests; Debug+Release --warnaserror 0 warnings. Forward-merge main→4.2 will batch with the other in-flight main-line fixes (#412/#505).

Issue #139, opened 2026-07-12
Add a canonical external base-URL override: redirect_uri and SAML base URL derive from the Host header


iderex:
`GetRequestBase` offers scheme and port overrides but no host override: `Request.Host.Host` flows directly into the OIDC redirect_uri (challenge and callback, SSOController.cs lines 227/291) and the rendered auth-page baseUrl. RFC 9700 sect. 4.1 (2025, https://datatracker.ietf.org/doc/rfc9700/) and the OAuth 2.1 draft rely on exact redirect-URI matching at the AS as the backstop, but the self-hosting audience frequently registers wildcard redirect URIs (Keycloak/authentik allow them), and proxies forwarding X-Forwarded-Host unfiltered make the host attacker-influenceable, a poisoned Host header then redirects the authorization response, code included, to an attacker origin. Add an optional per-provider canonical base-URL override (mirroring the existing scheme/port overrides), prefer it everywhere GetRequestBase is used, and document exact redirect-URI registration as required in providers.md.

Acceptance criteria:
- [ ] Per-provider canonical base-URL override exists and takes precedence in every GetRequestBase consumer (OIDC and SAML).
- [ ] Test: with the override set, a spoofed Host header does not alter redirect_uri or baseUrl.
- [ ] Admin UI warns when no override is configured.
- [ ] providers.md documents exact redirect-URI registration as required.

PR #567, opened 2026-07-18
Add SAML service-provider metadata endpoint

### LINKED ISSUE #162: SAML SP metadata endpoint
Expose an SP metadata endpoint so administrators can register this service provider at the IdP by URL instead of hand-configuring entityID, ACS URL and certificate. Feature parity with V2.

Acceptance criteria:
- [ ] Endpoint serves valid SAML 2.0 SP metadata (entityID, ACS binding/URL, SP certificate if signing is enabled).
- [ ] Metadata reflects the configured canonical base URL (see the base-URL override issue), not a spoofable Host header.
- [ ] Documented in providers.md.

iderex:
# What & why

Expose standard SAML 2.0 service-provider metadata so administrators can register this service provider at their identity provider by URL, instead of hand-configuring the entity id, assertion-consumer URL and signing certificate. Feature parity with the reference implementation. Closes #162.

New endpoint (anonymous, like a real IdP fetch of SP metadata):

```
GET /sso/SAML/metadata/<ProviderName>   ->   application/samlmetadata+xml
```

It emits an `EntityDescriptor`/`SPSSODescriptor` with the `entityID` (the provider's `SamlClientId`, the same value we send as the AuthnRequest `Issuer`), one HTTP-POST `AssertionConsumerService`, and, only when request signing (`SignAuthnRequests`) is enabled, a `KeyDescriptor use="signing"` carrying the SP's public signing certificate.

Design (thin controller per #318):

- `SamlSpMetadataBuilder`, a pure, request-free, `internal static` builder that serializes exactly its inputs into correctly-namespaced SAML 2.0 metadata (KeyDescriptor before AssertionConsumerService, per the metadata XSD), via `XmlWriter` (so entity id / ACS are XML-escaped). It only ever receives the public certificate (`certificate.RawData`); the private key never reaches it.
- `SamlLoginService.Metadata`, resolves the entity id and ACS **only** from the provider's configured canonical Base URL (`BaseUrlOverride`, #139), never the request `Host`, and advertises the signing certificate through `TryResolveSigningCertificate` (reveal-at-rest, load, export public DER, dispose).
- `SSOController.SamlMetadata`, a thin, anonymous endpoint behind the shared rate-limit gate that delegates to the flow service.

## The anti-spoofing decision (base URL source)

Published SP metadata is consumed by the identity provider to decide **where it POSTs assertions**. If the ACS/entityID were derived from the request `Host` (which a reverse proxy forwarding an unfiltered `X-Forwarded-Host` can influence), an attacker could serve metadata pointing the IdP at an attacker-controlled ACS. So this endpoint uses the same authoritative source as the login flow's ACS, the configured canonical Base URL (`BaseUrlOverride`, landed in #139), and, unlike the login flow, **never falls back to the request host**. When `BaseUrlOverride` is unset (or the client id is missing, or signing is on but the key cannot be loaded) it **fails closed with 409** and publishes nothing, rather than bake a spoofable value into IdP-consumed metadata. The ACS it publishes is byte-for-byte the one the login flow produces for a provider that has `BaseUrlOverride` set.

## Type of change

- [x] Feature
- [ ] Bug fix

## Security checklist

      rejected, never default-accepted. (The endpoint fails closed with 409 when the canonical base URL / client id is unset or the signing key is unloadable; it never emits spoofable or signing-implying-but-keyless metadata.)
- [x] No secrets logged; secrets are redacted on config export. (Only the public certificate `RawData` is published; the private key / stored PFX / `OidSecret` never appear. `Metadata` logs nothing.)
      config persistence, or the release pipeline. (4-lens adversarial gate below.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (N/A, the endpoint logs nothing.)

## Quality checklist

      testable unit. (Pure `SamlSpMetadataBuilder`; base-URL/ACS derivation reuses `CanonicalBaseUrl` + `SsoUrlBuilder`.)
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. (No new structural property; the change conforms to the existing thin-controller / internal-helper rules, which pass.)

## Verification

- [x] `dotnet test` is green. (1193 passed, 0 failed, 13 new.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (`providers.md` unchanged by `--write`.)

## Notes

Targets the `4.2` feature branch (features branch off `4.2`, fixes off `main`). Because `4.2` is not the default branch, GitHub auto-close does not fire from a merge off it, **the issue will be closed manually on merge**.

Adversarial-review gate (4 refute-by-default lenses): availability PASS, security-bypass PASS, correctness PASS, integration PASS. The bypass lens refuted both named threats (private-key/secret leakage and Host-header spoofing of the ACS); the correctness lens confirmed the SPSSODescriptor element order (KeyDescriptor before AssertionConsumerService) is XSD-conformant.

Explicitly out of scope (candidates for follow-up, not blocking):
- The metadata advertises a single index-0 ACS in the canonical new-path (`post`) spelling. A login started on the deprecated legacy `SAML/p/...` route sends the legacy-spelled ACS; a strict IdP that validates the incoming `AssertionConsumerServiceURL` against the registered metadata could reject that legacy-route login. The canonical login matches, and our own return leg accepts both spellings. If we want to fully cover the legacy route, we would add a second `AssertionConsumerService` entry, a small follow-up if it proves needed.
- No admin-UI button for the metadata URL yet; the endpoint is fetched by URL, consistent with the other SAML options being API/config-driven today.

### DIFF
--- a/SSO-Auth.Tests/SSOControllerSamlMetadataTests.cs
+++ b/SSO-Auth.Tests/SSOControllerSamlMetadataTests.cs
@@ -1,0 +1,176 @@
+using System;
+using System.Security.Cryptography.X509Certificates;
+using System.Xml.Linq;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Microsoft.AspNetCore.Mvc;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// In-process tests of the SAML SP metadata endpoint (<c>SamlMetadata</c>) via
+/// <see cref="SsoControllerHarness"/> (#162). They cover the guard branches (unknown/disabled provider),
+/// the anti-spoofing invariant — the published entity id and assertion-consumer URL come from the
+/// configured canonical Base URL and NEVER the request host, and the endpoint fails closed when that base
+/// URL is unset rather than emitting a spoofable ACS — and the signing-conditional key descriptor, which
+/// advertises only the PUBLIC certificate.
+/// </summary>
+[Collection("SSOController")]
+public class SSOControllerSamlMetadataTests
+{
+    private static readonly XNamespace Md = "urn:oasis:names:tc:SAML:2.0:metadata";
+    private static readonly XNamespace Ds = "http://www.w3.org/2000/09/xmldsig#";
+
+    // The harness always serves from this request host; the anti-spoofing tests assert it never appears in
+    // the published metadata (only the configured canonical Base URL does).
+    private const string RequestHost = "jf.example.com";
+    private const string CanonicalBaseUrl = "https://canonical.example.com";
+
+    [Fact]
+    public void SamlMetadata_UnknownProvider_RejectsAsUnknownProvider()
+    {
+        var harness = new SsoControllerHarness();
+
+        var result = Assert.IsType<ContentResult>(harness.Controller.SamlMetadata("nope"));
+
+        Assert.Equal(400, result.StatusCode);
+        Assert.Equal("No matching provider found", result.Content);
+    }
+
+    [Fact]
+    public void SamlMetadata_DisabledProvider_RejectsAsUnknownProvider()
+    {
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig
+        {
+            Enabled = false,
+            SamlClientId = "https://sp",
+            BaseUrlOverride = CanonicalBaseUrl,
+        });
+
+        var result = Assert.IsType<ContentResult>(harness.Controller.SamlMetadata("adfs"));
+
+        // A disabled provider is refused with the uniform unknown-provider body, so this anonymous surface
+        // does not expose its entity id / signing certificate either.
+        Assert.Equal(400, result.StatusCode);
+        Assert.Equal("No matching provider found", result.Content);
+    }
+
+    [Fact]
+    public void SamlMetadata_NoCanonicalBaseUrl_FailsClosed_NeverDerivingFromTheRequestHost()
+    {
+        // The anti-spoofing invariant: with no canonical Base URL configured, the endpoint refuses rather
+        // than bake the request host (which a forwarded X-Forwarded-Host could influence) into metadata an
+        // identity provider consumes. So it never emits an ACS pointing at the request host.
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig
+        {
+            Enabled = true,
+            SamlClientId = "https://sp",
+            BaseUrlOverride = null,
+        });
+
+        var result = Assert.IsType<ContentResult>(harness.Controller.SamlMetadata("adfs"));
+
+        Assert.Equal(409, result.StatusCode);
+        Assert.DoesNotContain(RequestHost, result.Content ?? string.Empty, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public void SamlMetadata_BlankClientId_FailsClosed()
+    {
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig
+        {
+            Enabled = true,
+            SamlClientId = "   ",
+            BaseUrlOverride = CanonicalBaseUrl,
+        });
+
+        var result = Assert.IsType<ContentResult>(harness.Controller.SamlMetadata("adfs"));
+
+        Assert.Equal(409, result.StatusCode);
+    }
+
+    [Fact]
+    public void SamlMetadata_SigningOff_UsesTheCanonicalBaseUrl_NotTheRequestHost()
+    {
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig
+        {
+            Enabled = true,
+            SamlClientId = "https://sp.example.com/entity",
+            BaseUrlOverride = CanonicalBaseUrl,
+            SignAuthnRequests = false,
+        });
+
+        var result = Assert.IsType<ContentResult>(harness.Controller.SamlMetadata("adfs"));
+
+        Assert.Equal(200, result.StatusCode);
+        Assert.Equal("application/samlmetadata+xml", result.ContentType);
+
+        var document = XDocument.Parse(result.Content!);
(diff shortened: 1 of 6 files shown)


iderex:

## The base-URL finding (anti-spoofing) and the approach we took

We investigated first, before writing any metadata code, exactly how the plugin already derives its externally-visible URLs. The plugin ALREADY has a canonical base-URL config: the per-provider BaseUrlOverride field (#139), normalized by CanonicalBaseUrl and composed into the SAML ACS by SsoUrlBuilder.SamlAcsUrl. The login flow resolves its ACS via CanonicalBaseUrl.Resolve, which prefers BaseUrlOverride but, when it is blank, FALLS BACK to the request Host.

That request-host fallback is an accepted trade-off for a per-request login redirect, but it is NOT acceptable in published metadata an identity provider consumes to decide where it POSTs assertions, a forwarded X-Forwarded-Host could then point the IdP at an attacker-controlled ACS. So, per the two options we considered, we took option (b): the metadata endpoint is scoped to the explicitly-configured canonical base URL and FAILS CLOSED if it is unset. Concretely, Metadata resolves the entity id and ACS with CanonicalBaseUrl.TryNormalize(BaseUrlOverride) only, it never reads the request at all, and returns 409 when BaseUrlOverride (or SamlClientId, or a loadable signing key when signing is on) is missing. No metadata is ever built from the request Host. There is no new dependency: the base-URL override (#139) is already landed on this line, so nothing had to be landed first. For a provider that HAS BaseUrlOverride set, the metadata ACS is byte-for-byte identical to the login flow's ACS (both go through the same CanonicalBaseUrl normalization and SsoUrlBuilder), so the IdP's registered ACS matches the login's Recipient/redirect target.

## Adversarial-review gate, 4 refute-by-default lenses

- Availability / mass-lockout: PASS. The endpoint is strictly cheaper than the already-anonymous signed challenge (it never opens a private-key handle), carries its own independent "metadata" rate-limit bucket so it cannot starve login budgets, disposes the certificate on every path with no key-store churn, and its fail-closed 409 branches touch no shared mutable state, adding it cannot regress or lock out any existing login.
- Security-bypass: PASS. Both named threats refuted. Private key / secret: only certificate.RawData (public DER) is published; the stored PFX, the private key, and OidSecret never appear, and Metadata logs nothing. A disabled/unknown provider is refused with the uniform rejection, so no cert/entity-id leak. Host spoofing: entity id and ACS come solely from BaseUrlOverride, never the request Host, and the endpoint fails closed rather than falling back. Provider-name-into-XML is XML-escaped by XmlWriter.
- Correctness (valid SAML 2.0): PASS. Correct metadata/dsig/protocol namespaces and HTTP-POST binding; the SPSSODescriptor sequence is XSD-conformant (KeyDescriptor before AssertionConsumerService); AuthnRequestsSigned and the signing KeyDescriptor are driven by the SAME predicate so they can never disagree; the XML declaration is utf-8 (via a UTF-8-reporting StringWriter) and the media type is application/samlmetadata+xml; entity id / ACS are escaped exactly once. WantAssertionsSigned="true" is honest and fail-closed (this SP rejects unsigned assertions).
- Integration: PASS. The route SAML/metadata/{provider} is pairwise-disjoint by literal segment from every existing SAML route (no ambiguous match); anonymous is intended and consistent with the other public SAML endpoints; the ACS is byte-for-byte consistent with the login flow; architecture-conformance fitness functions pass; no DI/ABI surface change.

## CI / verification status

- dotnet build --warnaserror: green in BOTH Debug and Release, 0 warnings.
- dotnet test: 1193 passed, 0 failed (13 new, the pure-builder tests and the endpoint tests, covering well-formed SAML 2.0 parsing, the anti-spoofing property, signing-conditional KeyDescriptor, no-private-key leak, and every fail-closed branch).
- Prettier: clean (providers.md unchanged by --write).
- Architecture conformance: green.

## Sample metadata (signing enabled)

  <?xml version="1.0" encoding="utf-8"?>
  <md:EntityDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" entityID="https://jellyfin.example.com/sso/SAML/adfs">
    <md:SPSSODescriptor AuthnRequestsSigned="true" WantAssertionsSigned="true" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
      <md:KeyDescriptor use="signing">
        <ds:KeyInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
          <ds:X509Data>
            <ds:X509Certificate>MIID... (public DER only; never the private key)</ds:X509Certificate>
          </ds:X509Data>
        </ds:KeyInfo>
      </md:KeyDescriptor>
      <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://jellyfin.example.com/sso/SAML/post/adfs" index="0" isDefault="true" />
    </md:SPSSODescriptor>
  </md:EntityDescriptor>

With signing off, the KeyDescriptor is omitted and AuthnRequestsSigned is "false".

## Merge note

This PR targets 4.2 (the non-default feature branch), so GitHub auto-close will not fire from the merge, issue #162 should be closed manually on merge. Two non-blocking follow-up candidates are recorded in the PR body's Notes (a second legacy-spelling ACS entry for strict IdPs on the deprecated route; an admin-UI link to the metadata URL).

iderex:
Pre-merge review, all gates green; ready. First SAML feature on the 4.2 line. Anti-spoofing handled correctly: the metadata endpoint derives entityID (= SamlClientId) and ACS ONLY from the per-provider BaseUrlOverride (#139) via CanonicalBaseUrl.TryNormalize, it never reads the request Host, and fails closed with 409 when BaseUrlOverride/SamlClientId is unset or the signing key is unloadable. For a provider with BaseUrlOverride set, the metadata ACS is byte-for-byte identical to the login-flow ACS. Signing KeyDescriptor present iff SignAuthnRequests, carrying only the public DER (certificate.RawData), never the private key/PFX. Anonymous by design (SP metadata is public). Thin controller endpoint delegating to a pure SamlSpMetadataBuilder. Four-lens gate all PASS (bypass refuted both private-key leak and Host-spoofing). CI all green; 1193 tests (13 new); Debug+Release --warnaserror 0 warnings; providers.md documented.

PR #568, opened 2026-07-18
Decompose the ~210-line ResolveOrCreateAsync account-resolution method

### LINKED ISSUE #505: Decompose the ~210-line ResolveOrCreateAsync account-resolution method
`CanonicalLinkService.ResolveOrCreateAsync` (`SSO-Auth/Api/Linking/CanonicalLinkService.cs:94-304`) is a single ~210-line method that is now the hardest thing in the codebase for a new contributor to read. It is the account-resolution brain of every SSO login, and it interleaves four distinct concerns in one body:

1. the under-lock candidate read (subject-keyed link vs legacy username-keyed link), `:119-144`;
2. resolving the account currently bearing the display name, `:152-154`;
3. the legacy-link re-key decision, admin refusal, the single-transaction migrate-and-resolve, and its logging, `:156-190`;
4. the terminal four-arm `switch` (`UseExistingLink` / `AdoptExistingAccount` / `CreateNewAccount` / `RejectNameTaken`), each arm carrying its own adoption gate and its own set of `LogWarning` messages, `:204-303`.

The security reasoning here is correct and the "why" comments are load-bearing, so this is a readability/method-size problem, not a behaviour one: the reader has to hold all four concerns and every fail-closed branch in their head at once to change any one of them safely. It is exactly the drift the readability pass under #318 targets ("method size", "obvious control flow"), and it is not covered by the S3776 list in #196 (which names only `SSOController` and `OidcIdTokenValidator`).

## Proposal (behaviour-preserving, net-neutral to net-negative)

Keep the pure decision helpers (`AccountLinkResolver`, `AdoptionEligibilityResolver`) exactly as they are and lift the two self-contained blocks out of the orchestration body:

- extract the legacy-migration block (`:156-190`) into a private `Task<Guid?> MigrateLegacyLinkIfEligible(...)` (or similar) that returns the re-keyed id, so the top-level method reads "read candidates -> maybe migrate -> resolve -> act";
- move each terminal arm's warning text into the arm's own small method (or a per-outcome logging helper), so the `switch` reads as the decision it is rather than a wall of interleaved log lines.

The result should let `ResolveOrCreateAsync` read as a short orchestration whose steps a reader can follow by name, with the detail one level down.

## Constraints

- Login path + config persistence: behaviour-preserving refactor, so it runs the adversarial-review gate. Every fail-closed guard, the exact ordering of the under-lock reads/writes, and each distinct log line must be preserved verbatim, no guard collapsed for brevity (security > minimal code > readability).
- No test changes expected beyond the extraction being green under the existing `CanonicalLinkService` / account-resolution suites; the extracted helpers stay pure/single-responsibility and lock in the method-size property if we want a conformance assertion.

Refs #318.

iderex:
# What & why

Closes #505. Refs #318.

`CanonicalLinkService.ResolveOrCreateAsync`, the account-resolution brain of every SSO login, was a ~210-line method interleaving four concerns in one body: the under-lock candidate read (subject-keyed vs legacy username-keyed link), the issuer-binding refusal (#186), the legacy-link re-key decision (#155/#218/#363), and the terminal four-arm create/adopt/reject switch (#95/#133/#218), each arm carrying its own fail-closed guard and its own log lines. A reader had to hold every branch in their head to change any one safely, the exact method-size/obvious-control-flow drift the #318 readability pass targets, and not covered by the S3776 list in #196.

We lifted the self-contained blocks out so the top-level method reads as its orchestration and the detail lives one level down:

- `ReadResolutionCandidates`, the one `_configStore.Read` pass that resolves both candidate links AND classifies the subject link's issuer binding, returned as a new private `readonly record struct ResolutionCandidates`. The whole verdict stays linearized in a single locked read, so the concurrent-migration tear that would bounce a legitimate user with a spurious 403 stays closed.
- `RefuseRepointedIssuer`, the #186 mismatch refusal (fail closed, still fires before `GetUserByName`).
- `MigrateLegacyLinkIfEligible`, the admin refusal plus the single-transaction migrate-and-resolve (`MigrateAndResolveCanonicalLink`, unchanged), returning the authoritative id the login binds to (`Guid?`).
- `AdoptExistingAccount`, `CreateNewAccountAsync`, `RejectNameTaken`, each terminal arm's guard and its verbatim log lines, so the switch reads as the decision it is rather than a wall of interleaved log text.

This is pure code motion: no control-flow change, no lock-scope change, every distinct log string byte-for-byte, every `ReplaceLineEndings` sanitizer still inline at its log call.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. Every fail-closed gate still fires on the same path with the same operands — the admin-takeover ceiling on both the adopt and the legacy re-key paths (#218), the #95 name-taken throw, the #186 issuer-mismatch refusal, the unresolved-identity belt, and the deleted/disabled-provider guard (`TryGetLinks requireEnabled: true`).
- [x] No secrets logged; secrets are redacted on config export. (No logging change; no secret touched.)
      config persistence, or the release pipeline. The adversarial 4-lens gate ran and returned PASS on all four lenses (results under Notes).
- [x] Log inputs from the IdP are sanitized inline at the log call. Every `username?.ReplaceLineEndings(string.Empty)` / `provider?.ReplaceLineEndings(string.Empty)` stays inline at each `Log*` call inside the extracted helpers, none moved behind a helper-arg boundary (CodeQL `cs/log-forging`).

## Quality checklist

      testable unit. The pure decision helpers (`AccountLinkResolver`, `AdoptionEligibilityResolver`) are untouched; the new methods are single-responsibility.
- [x] The change adds no more code than the problem requires. Net line delta is code motion plus a small orchestration comment; no logic added.
      change establishes is locked in as a rule in `ArchitectureConformanceTests`. The introduced `MigrateLegacyLinkIfEligible` is already guarded by the existing #363 `LegacyLinkMigration_ReturnsTheAuthoritativeMapping` rule (any `Migrate*` helper must return `Guid?`); the typed-mode (#369) and static-warn-gate (#362) rules stay green. We deliberately did NOT add a method-size rule — see Notes.

## Verification

- [x] `dotnet test` is green. (1179 passed, 0 failed, including the full `CanonicalLinkServiceTests` behaviour-preservation suite unchanged.)
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (N/A, only `.cs` touched.)

## Notes

Adversarial 4-lens gate (login path + config persistence, the review is primary verification here, since CI cannot exercise the concurrency/lock properties):

- Availability / mass-lockout: PASS. Lock count per path is identical to `origin/main`, the hot no-migration path stays ONE `_configStore.Read`, the migration path Read + one `Mutate`. The candidate read + issuer classification remain in one locked pass; the migrate-and-resolve remains one transaction; no new lock, deadlock, or contention. The once-per-interval warn throttle (`_legacyLinkWarnGate`) is still consulted at most once per login (one terminal arm executes), so the CWE-400 log-flood bound is intact. No success/throw inversion.
- Security-bypass / fail-open: PASS. Every fail-closed gate still fires with the same operands. The admin ceiling is read from the same non-null `existingAccount` on both the legacy re-key (`AdoptionGate.None`) and the same-name adopt (`adoptionGate`) paths, the #218 verified-email distinction is preserved, not flipped. `RejectNameTaken` returns the exception and the caller throws it unconditionally (only the warning is throttled). No throwing accessor was turned into a guarded fall-through.
- Correctness: PASS. Statement-by-statement equivalence against `origin/main`: `ResolutionCandidates` maps positionally with no field swap; evaluation order preserved (repoint refusal still precedes `GetUserByName`); `existingAccount!.HasPermission(...)` evaluates at the same guarded, non-null points; migrate-then-log order and the returned-id binding preserved; all six distinct log strings verbatim; `StampIssuer` else-if condition unchanged; async `ConfigureAwait(false)` preserved.
- Integration: PASS. `ProviderConfigStore` Read/Mutate semantics intact (no post-lock callback introduced); #186 stamping call sites unchanged; the new `using Jellyfin.Database.Implementations.Entities;` is required (the helpers name `User` explicitly) and correctly SA1210-ordered; the record struct trips none of the reflection fitness functions (they scan classes / skip nested types).

Method-size conformance rule, deliberately NOT added. The optional size assertion the issue mentions ("if we want a conformance assertion") is not worth adding: reflection can only measure IL size (a brittle proxy for source LOC), and a source-scan line-count gate needs an arbitrary threshold and is exactly the vacuous-prone kind of rule the conformance suite's own summary warns against. The load-bearing structural property this decomposition introduces, the legacy migration returns the authoritative mapping, is already pinned by the existing #363 rule, which now guards the newly introduced `MigrateLegacyLinkIfEligible` too. We would rather not trade a real invariant for a brittle one.

Out of scope / follow-up: none opened. The integration lens flagged the standard login-path E2E smoke from `docs/E2E-CHECKLIST.md` (one fresh-account login, one same-name adoption, one legacy-username-key migration) as manual confirmation before a release, consistent with any change on the live login path that CI can only partially cover; not a blocker for this behaviour-preserving refactor.

Forward-merge: this is a fix/refactor off `main`; forward-merge into `4.2` (the current feature dev line) after it lands, per the branching model.

### DIFF
--- a/SSO-Auth/Api/Linking/CanonicalLinkService.cs
+++ b/SSO-Auth/Api/Linking/CanonicalLinkService.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Security.Cryptography;
 using System.Threading.Tasks;
 using Jellyfin.Data;
+using Jellyfin.Database.Implementations.Entities;
 using Jellyfin.Database.Implementations.Enums;
 using Jellyfin.Plugin.SSO_Auth.Config;
 using MediaBrowser.Controller.Library;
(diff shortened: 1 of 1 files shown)


iderex:
Resolved the CodeQL `cs/exposure-of-private-information` (medium) alert on the account-resolution path (327d62b).

Determination: false positive, and not new behaviour. The `{Reason}` argument was `verdict`, an `AdoptionVerdict` enum, a refusal reason code (`Allow` / `RefusePrivileged` / `RefuseUnverifiedEmail`), never an email or any user data. CodeQL's heuristic tripped on the `Email` in the `RefuseUnverifiedEmail` member name. `main` logs the identical `verdict` through the same `{Reason}` template; extracting the log into the small `AdoptExistingAccount` helper only made the flow interprocedural, so the analysis surfaced a latent (benign) flow rather than a real new leak.

Fix (log-side, minimal, repo-convention-consistent): there is no CodeQL config or inline-suppression convention in this repo, so instead of introducing one we map the verdict to a fixed, non-PII reason phrase for the log via `DescribeAdoptionRefusal(verdict)`, a switch returning string literals. The refusal reason stays in the audit line (and reads clearer than the raw enum name), the message template is unchanged, and the switch carries no data flow from the enum constant into the log sink, so the heuristic no longer fires. The refusal `throw` and every fail-closed gate are untouched. Debug + Release `--warnaserror` are clean and all 1179 tests pass (none assert the reason text).

iderex:
Pre-merge review, all gates green (incl. CodeQL); ready. Decomposes the ~210-line ResolveOrCreateAsync (the account-resolution brain) into a thin orchestrator + single-responsibility private helpers (ReadResolutionCandidates, RefuseRepointedIssuer, MigrateLegacyLinkIfEligible, AdoptExistingAccount, CreateNewAccountAsync, RejectNameTaken). Pure code motion: no control-flow/lock-scope change, every log string byte-for-byte, every ReplaceLineEndings inline. Four-lens gate all PASS (every fail-closed gate fires on the same path, lock count per path identical to main, statement-by-statement equivalence). CodeQL flagged one MEDIUM cs/exposure-of-private-information at the verdict log, verified FALSE POSITIVE (verdict is an AdoptionVerdict enum reason code, logged identically on main; the heuristic tripped on the RefuseUnverifiedEmail member name once the flow became interprocedural). Resolved cleanly by mapping the verdict to a fixed non-PII reason phrase via DescribeAdoptionRefusal (control-flow literal selection, no data flow to the sink), reason still logged, reads clearer, message template unchanged. CodeQL now green, 0 open alerts. 1179 tests; Debug+Release --warnaserror 0 warnings. Forward-merge main→4.2 batches with #412.

PR #571, opened 2026-07-18
Record the challenge NewPath through MutateConfiguration

### LINKED ISSUE #412: Record the challenge NewPath through MutateConfiguration, not a post-lock callback

iderex:
# What & why

`ResolveChallengeNewPath` (`OidcLoginService.cs` / `SamlLoginService.cs`) recorded the
derived "NewPath" redirect-path spelling by writing straight into a config object that
was read under `ReadConfiguration`, after that read's lock was already released. Two
concurrent OID/SAML challenges for the same provider could race that unsynchronized
field write, and the write never went through `MutateConfiguration`, the persist path
every other config mutation uses. Worst case today is a fail-closed login failure (a
wrong `redirect_uri` spelling the IdP rejects), not a bypass - hence the low priority - 
but the shared-mutable-state pattern needed hardening.

Both resolvers now compare the derived spelling against what is currently stored and,
only when it differs, record it atomically through `MutateConfiguration`, re-resolving
the provider by name inside the lock rather than trusting the outer config reference (so
a provider deleted/disabled in that race window is not written into). The common case - 
the derived spelling already matches what is stored, true for every login after the
first on a given route, stays a plain locked comparison with no persist, so an ordinary
login does not pay a config-file write on every challenge. This mirrors
`CanonicalLinkService.ResolveOrCreateAsync`'s read-first, persist-only-on-change shape,
which already avoids a config persist on every already-linked login.

The derived-spelling logic itself (`ChallengePath.IsNewPath`) is unchanged, only where
and how the result is persisted.

Closes #412

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. (Unaffected by this change; the challenge's
      unknown/disabled-provider rejection is unchanged.)
- [x] No secrets logged; secrets are redacted on config export. (Unaffected.)
      config persistence, or the release pipeline. (Adversarial 4-lens gate run
      below — this touches the login challenge path.)
- [x] Log inputs from the IdP are sanitized inline at the log call. (Unaffected;
      no new log statements.)

## Quality checklist

      testable unit.
      (No new structural property — this stays inside the existing Flows/Config
      layering; the fix routes an existing write through the store's existing
      Mutate path.)

## Verification

- [x] `dotnet test` is green (1185/1185, both configurations).
- [x] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. (No such
      files touched — C#/tests only.)

## Notes

Adds coverage at two levels:
- The OID/SAML challenge tests (`SSOControllerOidChallengeTests.cs`,
  `SSOControllerChallengeTests.cs`) assert the write reaches the config store's
  persist delegate (`IXmlSerializer.SerializeToFile`, now exposed on
  `SsoControllerHarness` for this purpose) when the spelling actually changes, and
  that it is skipped when nothing changed.
- `ProviderConfigStoreTests.cs` adds two concurrency regression tests against the
  store's `Mutate` lock: a concurrent-increment test proving no update is lost under
  64-way contention, and a challenge-shaped concurrent read-then-maybe-write test
  (alternating derived spellings) proving the provider map survives concurrent
  access intact.

Out of scope, left for a separate issue if we decide to pursue it: `OidConfig` and
`SamlConfig` do share a common base (`ProviderConfigBase`) with a `NewPath` setter, so
the two near-identical `ResolveChallengeNewPath` resolvers in `OidcLoginService.cs` and
`SamlLoginService.cs` could in principle be collapsed into one generic helper. We did
not do that here to keep this change to WHERE/HOW the write is persisted, as scoped by
#412, a dedup pass is a #318-flavored follow-up, not a hardening fix.

Forward-merge reminder: `main` is the fix/security base; this needs forward-merging
into `4.1` once merged, per branching policy.

### DIFF
--- a/SSO-Auth.Tests/OidcLoginServiceTests.cs
+++ b/SSO-Auth.Tests/OidcLoginServiceTests.cs
@@ -97,6 +97,26 @@ public class OidcLoginServiceTests
         Assert.Empty(service.StateSummaries());
     }
 
+    [Fact]
+    public void ResolveChallengeNewPath_ProviderDisabledInTheRaceWindow_SkipsThePersist_ButStillReturnsThisRequestsDerivedSpelling()
+    {
+        // #412 review follow-up: exercises the Mutate delegate's own re-check inside ResolveChallengeNewPath.
+        // `config` mirrors the reference the real caller already captured under ReadConfiguration's lock
+        // (FindOidConfig) before its own Enabled check passed; something then disables the provider in the
+        // LIVE store before this write attempt runs — a race the outer check cannot see. The current
+        // challenge must still get its own freshly-derived spelling for its redirect, but the disabled
+        // provider's stored NewPath must be left untouched rather than written into.
+        var (_, context) = Build(c => c.OidConfigs["kc"] = new OidConfig { Enabled = true, NewPath = false });
+        var config = SSOPlugin.Instance.ReadConfiguration(c => c.OidConfigs["kc"]);
+        SSOPlugin.Instance.MutateConfiguration(c => c.OidConfigs["kc"].Enabled = false);
+        context.Request.Path = "/sso/OID/start/kc";
+
+        var result = OidcLoginService.ResolveChallengeNewPath("kc", config, isLinking: false, context.Request, Substitute.For<ILogger>());
+
+        Assert.True(result); // this request's own redirect still uses the freshly-derived spelling
+        Assert.False(SSOPlugin.Instance.ReadConfiguration(c => c.OidConfigs["kc"].NewPath)); // stored value untouched
+    }
+
     // Builds an OidcLoginService over the same collaborator graph the controller constructs, against a
     // freshly-bootstrapped SSOPlugin.Instance seeded via the configure callback and a DefaultHttpContext for
     // the request/response the challenge reads and writes. The harness resets the process-wide OpenID state,
--- a/SSO-Auth.Tests/ProviderConfigStoreTests.cs
+++ b/SSO-Auth.Tests/ProviderConfigStoreTests.cs
@@ -204,7 +204,8 @@ public class ProviderConfigStoreTests
         // store must serialize each read-then-maybe-write so every call completes cleanly and the
         // provider map is never left in a corrupted or partially-written state.
         var (store, live, _) = CreateStore();
-        live.OidConfigs["kc"] = new OidConfig { Enabled = true };
+        var seeded = new OidConfig { Enabled = true };
+        live.OidConfigs["kc"] = seeded;
         var ct = TestContext.Current.CancellationToken;
 
         const int concurrency = 50;
@@ -226,9 +227,13 @@ public class ProviderConfigStoreTests
 
         await Task.WhenAll(tasks); // throws if any callback threw or the store deadlocked
 
-        // Either spelling is a legitimate race outcome; what matters is that the provider entry survived
-        // concurrent access intact rather than a corrupted/missing entry.
-        Assert.True(store.Read(c => c.OidConfigs.ContainsKey("kc")));
+        // Either spelling is a legitimate race outcome, but the entry itself must have survived
+        // concurrent access intact: exactly one "kc" entry, the SAME object instance (never replaced or
+        // duplicated by an interleaved write), with every OTHER field untouched by the race — proof the
+        // concurrent NewPath writes never corrupted the surrounding provider record.
+        Assert.Equal(1, store.Read(c => c.OidConfigs.Count));
+        Assert.Same(seeded, store.Read(c => c.OidConfigs["kc"]));
+        Assert.True(store.Read(c => c.OidConfigs["kc"].Enabled));
     }
 
     [Fact]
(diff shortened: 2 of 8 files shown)


iderex:
Pre-merge review note, gates below, ready for approval.

Build/test: dotnet build --no-restore --warnaserror is green on Debug and Release; dotnet test is green on both (1187/1187), repeated three times on each configuration with no flakiness.

Adversarial review (login challenge path, two rounds):

Round 1 (commit 8205962, the base fix):
- availability lens: NEEDS_IMPROVEMENT, a realistic mixed old/new-route traffic pattern on one provider would turn the "write-once" optimization into a persist on every challenge under the process-wide config lock; also a new unhandled-exception surface on the persist call.
- bypass lens: PASS, derivation logic unchanged, no new write-target surface, no security-load-bearing regression (NewPath already accepts either spelling on validation; worst case stays a fail-closed redirect_uri mismatch).
- correctness lens: NEEDS_IMPROVEMENT, production logic judged correct, but two test-quality gaps: a weak assertion in the new concurrency test, and no direct coverage of the new provider-disabled-in-the-race-window branch.
- integration lens: NEEDS_IMPROVEMENT, independently confirmed the availability finding (SamlAssertionValidator's own comments show a NewPath flip is routine under mixed-client traffic, not the rare one-time event the PR's original justification claimed), plus the same exception-safety gap.

Fix (commit 58329be): both resolvers now throttle actual persist attempts to one per 5-second interval, process-wide per protocol, via the existing IntervalGate primitive (a losing throttle check never touches the config lock). The persist call is now wrapped so a failure (disk/permissions/secret-envelope) is logged and swallowed rather than failing the login, NewPath is best-effort bookkeeping for a later linking flow, never a requirement for the current one. Also fixes the two test-quality gaps: strengthens the concurrency test's assertions and adds direct tests for the disabled-in-race-window branch (making ResolveChallengeNewPath internal, matching this file's existing test-hook convention).

Commit 98fe642: a re-review of 58329be found the strengthened concurrency-test assertions still could not fail even without the store's lock, since the workload never performed a structural dictionary mutation. Rewrote it so half the concurrent tasks add/remove an unrelated provider entry while the rest race the NewPath read-then-maybe-write, a genuine read-during-write hazard the lock actually has to prevent.

Round 2 (commit 58329be, re-review):
- availability lens: PASS, confirmed the throttle bounds writes to at most one per 5s per protocol regardless of traffic shape, never contends the lock when throttled, and the try/catch fully removes the new unhandled-exception path.
- integration lens: PASS, independently confirmed the same, plus OID/SAML parity, test isolation, and no architecture-conformance regression, backed by a green full build+test run.
- correctness lens: NEEDS_IMPROVEMENT on this round only for the one concurrency-test assertion noted above (production logic, throttle math, fail-open persist, the race-window re-check, OID/SAML symmetry, was independently confirmed correct, including verifying the new race-window tests exercise the intended branch via the actual object-reference semantics of ProviderConfigStore.Read). Addressed in commit 98fe642 as described above.

Scope note called out in the PR body: OidConfig and SamlConfig do share a common base (ProviderConfigBase) with a NewPath setter, so the two near-identical ResolveChallengeNewPath resolvers could in principle be collapsed into one generic helper. Left out of this change to keep it to WHERE/HOW the write is persisted, as scoped by #412; a dedup pass is a separate #318-flavored follow-up if we decide to pursue it.

Forward-merge reminder: this is a fix on main; it needs forward-merging into 4.1 once merged, per branching policy.

iderex:
Pre-merge review, all gates green (CodeQL: no new alerts); ready. Closes the concurrent-challenge race + config-write-path bypass on the derived NewPath spelling: ResolveChallengeNewPath (now in OidcLoginService/SamlLoginService after the #318 move) no longer writes a bare config.NewPath outside the store lock, it compares to stored, and only on a real change atomically re-resolves the provider by name and writes through SSOPlugin.MutateConfiguration. The adversarial gate (2 rounds) added two hardenings it surfaced: (1) an IntervalGate throttle (one persist per 5s per protocol) since both route spellings stay live and flip routinely under mixed-client traffic, unthrottled this serialized every login behind disk I/O; (2) the persist is best-effort bookkeeping, so a write failure is logged+swallowed rather than failing the login. Four lenses PASS after the fixes, incl. a genuine structural-mutation concurrency regression test. CI all green; 1187 tests; Debug+Release --warnaserror 0 warnings.

PR #573, opened 2026-07-18
Add a SAML SP signing-key rollover (part 1 of #491)

iderex:
# What & why

Part 1 of #491. It lets the SAML service provider rotate its **own** signing
certificate with zero downtime by publishing both the old and the new public
certificate in the SP metadata during an overlap window, while still signing
`AuthnRequests` with a single (primary) key. Refs #491, this delivers only
capability 1 (outbound SP signing-key rollover); capability 2 (inbound IdP
verification-cert rotation) stays open for a follow-up, so this does **not**
close the issue.

Today `SamlConfig.SamlSigningKeyPfx` holds one SP signing key: it signs the
outgoing `AuthnRequest` (#167/#493) and is exposed as a single signing
`KeyDescriptor` in the SP metadata (#567). We add an optional second key,
`SamlRolloverSigningKeyPfx`:

- **Signing is unchanged.** Outgoing `AuthnRequests` are always signed with the
  **primary** key only. The rollover key is **publish-only**, it never signs.
- **Metadata advertises both.** When the rollover key is set and its
  certificate loads, the metadata publishes **two** `KeyDescriptor use="signing"`
  entries (primary then rollover, public certificates only), so the identity
  provider trusts either while the admin swaps. When the rollover key is unset,
  the output is byte-for-byte the previous single-`KeyDescriptor` document.
- **Clean end state.** When the rollover certificate equals the primary (the
  natural state once the admin promotes the rollover into the primary field),
  the redundant second descriptor is dropped, returning to a single key, no
  separate "clear the key" step is needed (the rollover key is write-only, so a
  blank save cannot clear it; the dedup makes that unnecessary).
- **The rollover key is a secret** (a private PFX) and is handled exactly like
  the primary: write-only JSON redaction (`WriteOnlySecretConverter` → `null` on
  read), **encrypted at rest** via `ConfigSecretProtection.ProtectAll` (and
  covered by `HasAnyEnvelope` so the missing-key orphan-prevention still fires),
  **preserved on a blank save** (`ServerManagedFields.ResolveUpdatedRolloverSigningKey`),
  and a garbage value is rejected at every admin write door (`SAML/Add` and the
  config-page save validator).
- **Fail-closed.** A set-but-unloadable rollover key returns `409` and publishes
  nothing (never a broken `KeyDescriptor`), mirroring the primary key. Only
  public certificate DER (`certificate.RawData`) ever reaches the builder; no
  private key leaves the server. #567's anti-spoofing (entity id/ACS from the
[...]

PR #581, opened 2026-07-18
Add admin config export/import (#161)

iderex:
# What & why

Closes #161.

Adds an admin-only **config export / import** surface so a plugin configuration can be exported from one instance and imported into another (replicating a setup, migrating a server). Two thin controller endpoints delegate to pure, testable Config helpers (per #318):

- `GET /sso/Config/Export` → `ConfigExport.Build` returns a **redacted** `ConfigExportDocument`.
- `POST /sso/Config/Import` → `ConfigImport.Apply` validates and **merges** the document into this instance.

Both are `[Authorize(Policy = Policies.RequiresElevation)]`, matching the other config endpoints. The admin config page gets **Export Configuration** / **Import Configuration** buttons.

**The redaction is reused, not duplicated.** The export serializes the `PluginConfiguration`, so the config's own JSON-boundary withholding applies unchanged: the three secret fields (`OidSecret`, `SamlSigningKeyPfx`, `SamlRolloverSigningKeyPfx`) are written as `null` by their `WriteOnlySecretConverter` (#189), and the server-managed link maps (`CanonicalLinks`, `CanonicalLinkIssuers`) are dropped by `[JsonIgnore]` (#157/#186). An exported document therefore carries no plaintext secret, no `ssoenc:` envelope, and no account-link data. The at-rest DEK (`sso-secret.key`) is a separate file, never part of the config object.

**The import preserves, it never wipes.** It is a fail-closed merge:

- Validated first through the same `ProviderConfigValidator` the config-page save uses (bad Base URL override #139, unloadable SAML cert/key #206/#167, reserved-character new provider name #336/#360 → rejected), and applied **atomically** via `MutateConfiguration`, a rejected document changes nothing.
- Each provider is merged through `ServerManagedFields.Preserve`, so a redacted (blank) secret keeps the target's stored secret (#189) and the server-managed links/issuers are re-injected from the target (#157/#186). `NewPath` (server-managed runtime state) is kept from the target for existing providers.
- Merge, not replace: a provider present only on the target is left untouched; a provider new to the target arrives with a blank secret and fails its login closed until an admin re-enters it.
- Imported non-blank secrets are encrypted at rest because persistence flows through `PersistBase → ConfigSecretProtection.ProtectAll`.

Round-trip documented in `providers.md`: export from A, import into B, then re-enter each provider's secret on B.

## Type of change

- [ ] Security hardening / vulnerability fix
- [ ] Bug fix
- [x] Feature
- [ ] Refactor / code quality / docs

[...]

PR #582, opened 2026-07-18
Accept a SAML response signed by an optional secondary IdP certificate (completes #491)

iderex:
# What & why

Completes #491 (part 2, the inbound half). Part 1, outbound SP signing-key
rollover, shipped in #573; this delivers the higher-value half: **inbound
verification-certificate rotation**, so an identity provider can roll its own
signing key over an overlap window instead of breaking every login at the
cutover.

A SAML response is now accepted when its signature verifies against **either**
the primary `SamlCertificate` **or** an optional new `SamlSecondaryCertificate`
(the IdP's PUBLIC signing certificate, mirroring the V2 field), and **expired
certificates are rejected**. When the secondary is blank, inbound validation is
byte-for-byte the primary-only behaviour.

How it works:

- `SamlResponse` loads the primary plus, when configured, the secondary
  certificate. A signature is accepted only when it verifies against at least
  one candidate that is **currently within its own `[NotBefore, NotAfter]`
  window**. The trial spans **only** the cryptographic key check, the
  single-assertion count, position-bound signature XPath, single-reference,
  reference-covers-{root|assertion}, enveloped-within binding, and the no-SHA-1
  algorithm/canonicalisation allowlist all still run once per signature,
  cert-independently, and the loop still requires **every** position-bound
  signature to validate. Trying a second certificate therefore only changes
  *which key* verifies the crypto, never *what* is signed or *which algorithm*
  is allowed.
- **Expiry gate.** `CheckSignature(cert, verifySignatureOnly: true)` does not
  itself check the certificate dates, so an explicit window check is the single
  place expiry is enforced, an expired (or not-yet-valid) candidate is skipped
  rather than trusted, for the primary as well as the secondary. This is what
  makes the overlap window terminate: an old key stops authenticating once it
  expires.
- **Public, not a secret.** `SamlSecondaryCertificate` is the IdP's public
  certificate, exactly like `SamlCertificate`, no `WriteOnlySecretConverter`,
  and left verbatim by `ConfigSecretProtection.ProtectAll` (contrast
  `SamlSigningKeyPfx` / `SamlRolloverSigningKeyPfx`, which hold a private key and
  are encrypted at rest). A set-but-unloadable value is rejected fail-closed at
  both admin write paths (`SAML/Add` and the whole-config save) and,
  defensively, at the login loader.
- `providers.md` documents the field and the inbound rotation procedure.

## Type of change

- [x] Feature
- [ ] Bug fix

## Security checklist

      rejected, never default-accepted. A response verifying against neither
      certificate is rejected; an expired certificate never verifies.
- [x] No secrets logged; secrets are redacted on config export. The new field is
      a public certificate, not a secret.
      config persistence, or the release pipeline. Four-lens adversarial gate
      plus the SAML domain reviewer — verdicts in Notes.
      sites; the existing algorithm-diagnostic log is unchanged.

## Quality checklist

      unit (`VerifiesAgainstCandidateCertificate` / `IsWithinValidityPeriod`,
      the certificate guards reuse `SamlCertificate.IsInvalid`).
- [x] Architecture-conformance tests pass. This change establishes no new
      structural property (the validation home is unchanged).

## Verification

- [x] `dotnet test` is green (1237 tests).

## Notes

Scope: inbound verification-cert rotation only; the outbound SP signing-key
rollover was #573. No admin-UI change, SAML providers are configured via the
`SAML/Add` API and config XML (the config page handles OpenID only), so the
field is reachable there, consistent with every other SAML option.

Adversarial-review gate (refute-by-default; SAML signature acceptance is the
primary verification for this change, CI cannot exercise a real IdP rotation,
so this review is the primary verification). Verdicts:

- Security-bypass / fail-open: PASS. Confirmed a response is accepted ONLY if it
  validly verifies against a real, in-validity primary or secondary certificate
 , never "a cert is configured". The multi-cert trial is a pure key-OR layered
  strictly beneath the unchanged per-signature checks; it opened no XSW /
  signature-wrapping and no algorithm-downgrade path.
- SAML domain review: PASS. Invariant-by-invariant: single-assertion count,
  position-bound signature XPath, single-reference/covers-{root|assertion}/
  enveloped-within, no-SHA-1 signature+digest+C14N allowlist all still run once
  per signature, cert-independently, and the loop still rejects the whole
  response if ANY position-bound signature fails against every valid candidate.
  XXE/DOCTYPE, time-bound, audience, recipient, replay/one-time-use, InResponseTo
  all untouched.
- Correctness: PASS. Empirically verified the local→UTC `NotBefore`/`NotAfter`
  conversion; expired/not-yet-valid certs cannot verify (the window gate short-
  circuits before `CheckSignature`, which with `verifySignatureOnly: true` does
  not check dates); tests prove a cryptographically-matching-but-expired cert is
  still rejected; exception ordering preserved for existing callers.
- Integration: PASS. `SamlSecondaryCertificate` is a PUBLIC value, no
  `WriteOnlySecretConverter`, left verbatim by `ConfigSecretProtection` (not
  encrypted), survives the OpenID-only config-page save, deserializes cleanly
  from legacy configs, rejected fail-closed on both admin write paths, no ABI
  break.
- Availability / mass-lockout: NEEDS_IMPROVEMENT on the first pass, then
  addressed and re-reviewed. The lens flagged that the newly-enforced
  certificate-expiry check now also rejects an ALREADY-EXPIRED primary (the
  pre-#491 path ignored certificate dates), a real upgrade behavior change. This
  is the issue's explicit "reject expired certs" requirement, so the behavior is
  kept (adding a fail-open "accept expired" toggle would contradict the mandate
  and the fail-closed posture); the upgrade risk is mitigated by applying the
  shared 5-minute `SamlAssertionTime.ClockSkew` to both edges of the window and a
  prominent providers.md upgrade note. The lens also asked to correct the
  misleading "byte-for-byte primary-only" wording (done) and the SAML reviewer
  asked for a doubly-signed-decoy-with-secondary test (added).

Standing decision for myself: rejecting an expired primary certificate
is required by #491 and is the correct fail-closed posture, but it is a
deployment-visible change for anyone running an expired-but-pinned IdP
certificate. We have documented it as an upgrade note and added clock-skew
tolerance rather than a fail-open override. This warrants a release-note callout
when 4.2 ships, and belongs on the manual E2E checklist (a real-IdP rotation is
not exercisable in CI).

### DIFF
--- a/SSO-Auth.Tests/ConfigPreservationTests.cs
+++ b/SSO-Auth.Tests/ConfigPreservationTests.cs
@@ -834,4 +834,36 @@ public class ConfigPreservationTests
     {
         ProviderConfigValidator.Validate(new PluginConfiguration { SamlConfigs = null }, new PluginConfiguration());
     }
+
+    [Fact]
+    public void ValidateSamlSecondaryCertificate_ValidOrBlank_DoesNotThrow()
+    {
+        // The inbound secondary verification certificate (#491) is validated on the whole-config save path
+        // exactly like the primary; a valid or blank value is accepted.
+        var incoming = new PluginConfiguration();
+        incoming.SamlConfigs["ok"] = new SamlConfig
+        {
+            SamlCertificate = SamlTestFactory.Create().CertificateBase64,
+            SamlSecondaryCertificate = SamlTestFactory.Create().CertificateBase64,
+        };
+        incoming.SamlConfigs["blank"] = new SamlConfig { SamlCertificate = null, SamlSecondaryCertificate = null };
+
+        ProviderConfigValidator.Validate(incoming, new PluginConfiguration());
+    }
+
+    [Fact]
+    public void ValidateSamlSecondaryCertificate_Garbage_Throws()
+    {
+        // A set-but-unloadable secondary certificate is rejected fail-closed before persistence, so it can
+        // never make every callback throw a CryptographicException 500.
+        var incoming = new PluginConfiguration();
+        incoming.SamlConfigs["idp"] = new SamlConfig
+        {
+            SamlCertificate = SamlTestFactory.Create().CertificateBase64,
+            SamlSecondaryCertificate = "QUJD", // valid base64, not a cert
+        };
+
+        var ex = Assert.Throws<ArgumentException>(() => ProviderConfigValidator.Validate(incoming, new PluginConfiguration()));
+        Assert.Contains("secondary", ex.Message, StringComparison.Ordinal);
+    }
 }
--- a/SSO-Auth.Tests/ConfigSecretProtectionTests.cs
+++ b/SSO-Auth.Tests/ConfigSecretProtectionTests.cs
@@ -60,6 +60,7 @@ public class ConfigSecretProtectionTests
             {
                 SamlSigningKeyPfx = "pfx-x",
                 SamlCertificate = "PUBLIC-IDP-CERT-X",
+                SamlSecondaryCertificate = "PUBLIC-IDP-CERT-SECONDARY-X",
             };
             config.SamlConfigs["y"] = new SamlConfig { SamlSigningKeyPfx = "pfx-y" };
 
@@ -75,6 +76,11 @@ public class ConfigSecretProtectionTests
             // it directly (never through Reveal), so encrypting it here would silently break every SAML
             // login. It - and non-secret ids - must be left verbatim.
             Assert.Equal("PUBLIC-IDP-CERT-X", config.SamlConfigs["x"].SamlCertificate);
+
+            // The inbound secondary verification certificate (#491) is likewise the identity provider's
+            // PUBLIC key, read directly by signature verification (never through Reveal); encrypting it would
+            // silently break rotation logins, so it too must be left verbatim.
+            Assert.Equal("PUBLIC-IDP-CERT-SECONDARY-X", config.SamlConfigs["x"].SamlSecondaryCertificate);
             Assert.Equal("client-a", config.OidConfigs["a"].OidClientId);
         });
     }
--- a/SSO-Auth.Tests/SSOControllerSamlAuthTests.cs
+++ b/SSO-Auth.Tests/SSOControllerSamlAuthTests.cs
@@ -99,6 +99,49 @@ public class SSOControllerSamlAuthTests
     }
 
     [Fact]
+    public async Task SamlAuth_SignedBySecondaryCertificate_DuringRotation_ReturnsOk()
+    {
+        // Inbound verification-cert rotation (#491): the identity provider has rolled its signing key, so
+        // the primary now holds an unrelated (old) certificate and the response is signed by the key whose
+        // certificate the administrator staged in the secondary field. The full callback must authenticate
+        // it via the secondary, end-to-end through the real controller and config.
+        var fixture = SamlTestFactory.Create(nameId: "alice");
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig
+        {
+            Enabled = true,
+            SamlCertificate = SamlFixture.ForeignCertificateBase64(),
+            SamlSecondaryCertificate = fixture.CertificateBase64,
+            DoNotValidateAudience = true,
+            EnableAuthorization = false,
+            AllowExistingAccountLink = false,
+        });
+        var user = new User("alice", "SSO-Auth", "Default") { Id = UserId };
+        harness.UserManager.CreateUserAsync("alice").Returns(user);
+        harness.UserManager.GetUserById(UserId).Returns(user);
+
+        Assert.IsType<OkObjectResult>(await harness.Controller.SamlAuth("adfs", new AuthResponse { Data = fixture.EncodeResponse() }));
+    }
+
+    [Fact]
+    public async Task SamlAuth_NeitherCertificateVerifies_Returns400()
+    {
+        // Fail closed: with a secondary configured, a response signed by neither the primary nor the
+        // secondary is still rejected — "a certificate is configured" is never acceptance.
+        var fixture = SamlTestFactory.Create();
+        var harness = new SsoControllerHarness(c => c.SamlConfigs["adfs"] = new SamlConfig
+        {
+            Enabled = true,
+            SamlCertificate = SamlFixture.ForeignCertificateBase64(),
+            SamlSecondaryCertificate = SamlFixture.ForeignCertificateBase64(),
+            DoNotValidateAudience = true,
+        });
+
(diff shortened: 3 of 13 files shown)


iderex:
Pre-merge review, gates below, all green; ready for approval. This changes SAML assertion signature acceptance (the highest-scrutiny crypto path), so the adversarial gate is the primary verification, CI cannot exercise a real IdP key rotation.

Design (either-cert, fail-closed). A SAML response is accepted only when its signature validly verifies against a real certificate that is currently within its validity window, the primary SamlCertificate OR, when configured, the optional SamlSecondaryCertificate. Never "a cert is configured": if the signature verifies against neither in-validity certificate, the response is rejected. When the secondary is blank the trial narrows to the primary alone. The multi-cert trial is a pure key-OR layered strictly beneath the existing per-signature checks, it changes only which key verifies the crypto, never what is signed or which algorithm is allowed. The secondary is the IdP's PUBLIC certificate, exactly like the primary: no write-only converter, not encrypted at rest, rejected fail-closed at both admin write paths and, defensively, at the login loader.

Expiry evidence. CheckSignature(cert, verifySignatureOnly: true) validates only the cryptography and does not check the certificate dates, so an explicit window gate (IsWithinValidityPeriod) is the single enforcement point: an expired or not-yet-valid candidate is skipped rather than trusted, for the primary as well as the secondary. Tests prove a cryptographically-matching-but-expired certificate is still rejected (not merely a non-matching one): expired-primary + valid-secondary accepts via the secondary; both-expired rejects; expired-primary-only rejects; not-yet-valid skipped. The window carries the shared five-minute SamlAssertionTime.ClockSkew on both edges so a cutover clock difference does not spuriously reject a valid login.

XSW / allowlist preservation evidence. The single-assertion count, the position-bound signature XPath (Response or Assertion direct child only), the exactly-one-reference and reference-covers-{root|assertion} and enveloped-within binding, and the no-SHA-1 signature/digest/canonicalization allowlist all still run once per signature, cert-independently, and the loop still requires every position-bound signature to validate, so a decoy cannot slip alongside a valid one. Covered by tests including SHA-1-rejected-when-secondary-matches, wrapping-second-assertion-rejected-with-secondary, doubly-signed-with-one-corrupted-signature-rejected-with-secondary, and DOCTYPE/XXE rejected at parse before any certificate is consulted. Replay/one-time-use, audience, recipient, NotOnOrAfter/time-bound, and InResponseTo are untouched.

Adversarial-review gate verdicts (refute-by-default):
- Security-bypass / fail-open: PASS, accepted only on a valid verify against a real in-validity cert; no XSW/wrapping and no algorithm-downgrade opened by the multi-cert trial.
- SAML domain review: PASS, every signature invariant confirmed to run per-signature, cert-independently; time-bound/audience/recipient/replay/InResponseTo preserved.
- Correctness: PASS, local-to-UTC NotBefore/NotAfter conversion empirically verified; expired/not-yet-valid cannot verify; exception ordering preserved for existing callers.
- Integration: PASS, SamlSecondaryCertificate is public (no WriteOnlySecretConverter, left verbatim by ConfigSecretProtection), survives the OpenID-only config-page save, deserializes from legacy configs, rejected fail-closed on both admin write paths, no ABI break.
- Availability / mass-lockout: NEEDS_IMPROVEMENT on the first pass, addressed and re-reviewed to PASS. The lens flagged that the new expiry check now also rejects an already-expired primary (the pre-#491 path ignored certificate dates). That is #491's explicit "reject expired certs" requirement, so the behavior is kept, adding a fail-open "accept expired" toggle would contradict the mandate and the fail-closed posture. The upgrade risk is mitigated by the clock-skew tolerance and a prominent providers.md upgrade note; the misleading "byte-for-byte primary-only" wording was corrected, and the doubly-signed-decoy-with-secondary test added.

Standing item for me: rejecting an expired primary certificate is a deployment-visible change for anyone running an expired-but-still-signing IdP certificate. It is correct and required, documented as an upgrade note, and warrants a release-note callout when 4.2 ships plus a real-IdP-rotation entry on the manual E2E checklist (not exercisable in CI).

Gate status: Debug and Release builds green with --warnaserror (0 warnings); dotnet test green (1279 tests, including the new secondary-certificate suite); Prettier clean for the providers.md change; rebased onto current origin/4.2 (through #581/#580) with no conflicts; the PR will also run the advanced CodeQL scan on 4.2.

Scope: inbound (IdP-side) verification-cert rotation only; the outbound SP signing-key rollover was #573. No admin-UI change, SAML providers are configured via the SAML/Add API and config XML (the config page handles OpenID only). This PR completes both halves of #491.

Merge housekeeping: 4.2 is a non-default branch, so merging this PR will not auto-close #491, close #491 manually on merge (both part 1, #573, and part 2, this PR, will then have landed).

iderex:
Pre-merge review, all gates green (incl. advanced CodeQL C# on 4.2); ready. Completes #491 (part 1 = #573). SAML inbound verification-cert rotation: accept a response whose signature verifies against the primary OR an optional public SamlSecondaryCertificate that is currently within its validity window.

Independent more-eyes verification (SAML signature acceptance = highest crypto scrutiny; beyond the agent five-lens I ran an independent bypass lens over the committed diff):
- Independent SECURITY-BYPASS lens: PASS. The multi-cert trial is confined to VerifiesAgainstCandidateCertificate (Saml.cs), a pure key-substitution at the innermost step. Everything above runs once per signature, cert-INDEPENDENT: single-direct-child-assertion count, position-bound SignatureXPath (/Response/Signature | /Response/Assertion/Signature), single-Reference/covers-root-or-assertion/enveloped binding, and the no-SHA-1 allowlist. So no XSW/wrapping is opened by the second cert (doubly-signed-decoy + wrapping-second-assertion tests pass). _certificates is built only from config (no response-KeyInfo key injection); either-cert is fail-closed (accept only when a configured, in-validity cert actually verifies). Expiry cannot be bypassed: CheckSignature(verifySignatureOnly:true) ignores dates, so IsWithinValidityPeriod is the enforced gate, short-circuited before the crypto, uniform on both certs with symmetric skew, net hardening (an expired primary, previously accepted, is now rejected). DOCTYPE/XXE reject, replay, audience, recipient, NotOnOrAfter, InResponseTo untouched.

Agent five-lens: security-bypass PASS, SAML-domain PASS, correctness PASS, integration PASS, availability NEEDS_IMPROVEMENT → PASS after fixes (shared clock skew, providers.md upgrade note, decoy test). The secondary is the IdP public cert, not encrypted at rest, rejected fail-closed at SAML/Add + whole-config save + the loader. CI all green; 1279 tests incl. the XSW/decoy/expiry/SHA-1/not-yet-valid suite; Debug+Release --warnaserror 0 warnings; providers.md documents the inbound rotation procedure. 4.2 release-note callout: an already-expired primary cert is now rejected on upgrade (fail-closed, per #491 mandate).

Issue #164, opened 2026-07-12
Full RBAC mapping (PermissionKind / Preference / scalar policy)


iderex:
Map IdP roles/claims to the full Jellyfin authorization surface - PermissionKind flags, Preference values and scalar policy fields - not just the admin/enabled subset handled today. Feature parity with V2.

Acceptance criteria:
- [ ] Claim-to-permission mapping covers PermissionKind, Preference and scalar policy fields, configurable per provider.
- [ ] Fail closed: an unmapped or malformed claim never grants a permission.
- [ ] Deterministic, ordinal claim matching (consistent with the auth-path comparison hardening).
- [ ] Tests cover grant and deny paths.

iderex:
Competitive/demand evidence for prioritization: K0lin/jellyfin-plugin-sso (a fork, v5.0.0.0, 2026-06-13, alpha) already ships claim/role mapping to per-folder and Live TV permissions. The archived upstream 9p4/jellyfin-plugin-sso also carried an unresolved request for claims-to-UserPolicy mapping (9p4/jellyfin-plugin-sso#295, opened 2025-09-07, never addressed before archival). Doesn't change the acceptance criteria, but supports moving this off the opportunistic tail once P2 closes out.

PR #665, opened 2026-07-19
SSO-only login (DisablePasswordLogin) with a break-glass admin safety net (#165)

### LINKED ISSUE #165: SSO-only login (DisablePasswordLogin) with a last-admin safety net
Optional SSO-only mode that disables native password login, with a mandatory safety net that refuses to lock out the last usable local administrator. Availability-sensitive: a naive implementation is a mass-lockout footgun.

Acceptance criteria:
- [ ] DisablePasswordLogin is opt-in per install.
- [ ] The setting cannot be enabled if it would leave no working admin login path (fail safe, blocks with a clear message).
- [ ] A documented break-glass recovery path exists (config toggle reversible without SSO).
- [ ] Tests cover the last-admin guard.

Availability gate: requires the availability lens in review.

iderex:
## What & why
Adds an **SSO-only login** mode (`DisablePasswordLogin`): while on, non-break-glass admin accounts are repointed to the SSO auth provider so they can only authenticate via the IdP. A designated **break-glass admin** stays on the built-in password provider as a guaranteed, IdP-independent recovery door. Closes #165.

## How it works
- **Enable sweep** (`SsoOnlyLoginService.SweepEnableAsync`): repoints eligible admins to the SSO provider and records each moved account in the server-managed `SsoOnlyRepointedUserIds` (track-first, so a mid-sweep crash never leaves a moved-but-untracked account). Never touches the break-glass admin or an account already off the password provider.
- **Per-login enforcement** (`SsoOnlyLoginService.ResolveLoginEnforcement`, called from `LoginCompletionService`): resolves the account **by id**, decides the login provider and break-glass status from the **resolved account** (not the mutable IdP username), track-firsts any first-time repoint, and threads `IsBreakGlassAdmin` into the mint.
- **Break-glass pin** (`SsoOnlyLoginGuard.ResolveLoginProvider`): the break-glass admin is pinned to the built-in password provider even when the operator's `DefaultProvider` points at SSO; and `SessionMinter` never demotes the break-glass admin's `IsAdministrator` while the mode is on.
- **Reversible off-switch**: `DisableAsync` + a boot `SsoOnlyReconciliationService : IHostedService` restore exactly the tracked accounts (gated on `IsSsoProvider`, so natively-SSO/plugin-created accounts never gain a password door), so editing `config.xml` to turn the mode off and restarting genuinely restores password login.
- **Un-forgeable config**: `DisablePasswordLogin`, `BreakGlassAdminUsername`, and `SsoOnlyRepointedUserIds` are server-managed (re-injected in `ServerManagedFields.Preserve`), so a config-page PUT can neither flip the mode nor move the break-glass designation. `IsDisabled` was added to `PermissionRolePolicy.DedicatedPermissions` so no SSO role mapping can ever disable an account.

## Integration
Forward-merged current `main` (brings the RBAC permission-mapping #164 + the docs→wiki migration). Zero textual conflicts, the SSO-only provider pin and the RBAC permission grants write different `SessionParameters`/`User` fields; both apply on every mint.

## Type of change
- [x] New feature (login-path / auth surface)

## Security & quality checklist (directive 1, auth surface)
- **Adversarial `/security-review` gate, all lenses PASS** after fixes (full journey in the review-gate comment below): availability, security-bypass, and correctness lenses each read the full touched files + callers, refute-by-default. The gate caught and drove out **five** real whole-org lockout / recovery-defeat defects before merge (two mass-lockout, one break-glass-demotion, one mutable-username lockout, one off-switch reversibility gap), this is the gate doing its job.
- **Fail-closed:** the break-glass password door survives every path; the recovery admin is never swept, tracked, demoted, or disabled. Missing/null resolution → configured default, nothing minted. Negative test for every fail-closed branch (incl. the two the final review named).
- **Tests:** 1510/1510 on both net9.0 and net10.0 (clean `--warnaserror` build). Break-glass survivability, mutable-username-rename, login-path track+restore, `IsDisabled` un-mappable, and the null/deleted-user branch are all pinned.
- One documented, availability-safe residual (a narrow login-vs-Disable interleaving affecting a single non-break-glass account's off-switch reversibility, self-healing; the restore-rework alternative was reasoned-DECLINED because it would strand legitimately-revoked ids forever) and one out-of-scope pre-existing follow-up (#690) are recorded.

## Review gate
Detailed lens verdicts in the comment below.


iderex:
Pre-merge note. Gate complete: three adversarial lenses PASS (availability, security-bypass, correctness) after driving out five whole-org lockout / recovery-defeat defects (two mass-lockout, break-glass demotion, mutable-username lockout, off-switch reversibility) plus a reflection-based ABI-floor fix and a fail-closed enumerate guard. CI green, build (net9+net10, 1511 tests), ABI floor (10.11.0), Package, CodeQL analyses. The one CodeQL high-severity finding was reviewed and dismissed as a false positive (an info log of a public provider type-name identifier, mis-tagged via the 'Password' substring, no secret logged). Follow-up #690 filed for the pre-existing third-party-provider tracking gap. Merging.

Issue #837, opened 2026-07-20
Pin the persisted SSO AuthenticationProviderId so it is decoupled from the controller's type location


iderex:
## Finding

`SsoAuthenticationProviders.SsoProviderId` is computed as `typeof(SSOController).FullName` (= `"Jellyfin.Plugin.SSO_Auth.Api.SSOController"`), and `CanonicalLinkService` stamps that same value on every SSO-created account's `User.AuthenticationProviderId`. That value is **persisted in Jellyfin's user database** and is the marker the plugin uses to detect SSO-managed / SSO-only accounts (`IsSsoProvider`).

The current design deliberately derives both the stamp and the detector from `typeof(SSOController).FullName` so they always agree **with each other**. But that couples a **persisted identity value to the controller's namespace**: any future namespace/rename of `SSOController` changes `FullName`, so:

- new stamps + the detector move to the new string and agree with each other, but
- **every already-persisted account keeps the old string**, so `IsSsoProvider(oldValue)` returns `false`, existing SSO-managed accounts become invisible to the SSO-only enforcement sweep, the break-glass/last-admin guard, and status reporting.

Password login stays disabled either way (any unregistered id routes to `InvalidAuthenticationProvider`), so this is not a lockout, but it silently breaks the plugin's own SSO-only bookkeeping for existing installs. It is a latent correctness/persistence bug independent of any refactor; the module move (#807) would trigger it.

## Fix

Pin `SsoProviderId` to the **literal legacy string** as a `const` (decoupled from `typeof`), used by both the detector and the `CanonicalLinkService` stamp:

```csharp
// SECURITY/persistence: this exact string is written to User.AuthenticationProviderId and
// persisted in Jellyfin's user DB. It must NEVER change or existing SSO accounts stop being
// recognized. Deliberately NOT typeof(SSOController).FullName — the controller may move.
internal const string SsoProviderId = "Jellyfin.Plugin.SSO_Auth.Api.SSOController";
```

- Both `SsoAuthenticationProviders` and `CanonicalLinkService` reference the one constant.
- Add a conformance test pinning the exact byte value (so it can never drift), mirroring the audit-log/keystone pinning pattern.
- This is strictly safer than the status quo (removes the rename-orphan hazard) and is the prerequisite that makes the `SSOController` → `Api/Http` move (#807) a behavior-preserving mechanical change.

Security-sensitive (persisted identity + the SSO-only guard) → passes the adversarial-review gate before merge.

Issue #799, opened 2026-07-19
Enable nullable reference types project-wide (16/97 files are opted in today)


iderex:
## Problem
Only **16 of 97** source files opt in to nullable reference types via a per-file `#nullable enable`; the project has no `<Nullable>enable</Nullable>`. For a login plugin, NRT is a correctness/hardening tool, the compiler proves null-flow on exactly the paths where a silent null becomes a fail-open or a 500 (`TreatWarningsAsErrors` is already on, so every NRT warning would be a hard error once enabled).

## Why it matters for the showcase bar
Modern .NET best practice is project-wide NRT; per-file opt-in is the transitional state. A showcase repo finishes the transition: every reference type's nullability is part of its contract, XML docs and annotations agree, and `!` (null-forgiving) appears only with a justifying comment (the existing `Subject!`/`Username!` pattern with upstream fail-closed reasoning is the model).

## Approach (gated, incremental)
1. Flip `<Nullable>enable</Nullable>` in `SSO-Auth.csproj` (and the test project) and remove the 16 per-file directives.
2. Fix resulting warnings **in small per-area PRs** (module by module, matching the new folder structure), never with blanket `!`, each `!` needs the justifying comment per the existing idiom.
3. Where annotation reveals a real null-flow bug, fix it fail-closed (and per the RCA rule, note it).
4. Keep both TFMs + ABI-floor green throughout.

## Acceptance criteria
- [ ] `<Nullable>enable</Nullable>` project-wide, zero per-file `#nullable` directives left.
- [ ] Zero NRT warnings (hard errors via TreatWarningsAsErrors), no unjustified `!`.
- [ ] Any real null-flow findings fixed fail-closed and documented.

Refs #790 (module structure the per-area PRs follow), the showcase polish loop.

iderex:
Reopening: the Saml PR #894 auto-closed this via 'Closes #799', but the Http module (SSOController, RequestHelpers, ProviderConnectionTester, SSOViewsController) was never completed (its scoping branch was paused at the rate-limit-null-IP fail-closed decision) and the three root files (SSOPlugin, SsoOnlyServiceRegistrator, AssemblyInfo) are not enabled. Completing project-wide now (set <Nullable>enable</Nullable> + remove the per-file pragmas).

PR #839, opened 2026-07-20
Enable nullable reference types across Provider, Linking and Session

iderex:
# What & why

Second step of the NRT rollout (#799): `#nullable enable` for every file in the **Provider, Linking, Session** modules, annotated to real nullability.

**Annotations only, no runtime behaviour change.** Suite 1589 both TFMs (the net10 run flaked on loopback-socket timeouts under local thermal load; clean on re-run, CI/Linux is unaffected).

Highlights:
- `AuthResponse` client DTO fields → `string?` (an incoming request may omit them).
- `issuer` is `string?` throughout `CanonicalLinkService` (#186); optional ctor gates `?`.
- SSO-only decision types state real nullability (`SsoOnlyLoginDecision.DefaultProvider`, `SsoOnlyEnableOutcome.BreakGlassAdmin`, `ResolveLoginEnforcement.configuredDefaultProvider` → `string?`), matching the already-`required string?` `SessionParameters.DefaultProvider`.
- Gracefully-null guards annotated (`IsIdentityStillLinked`, `RemoveUser`); `TryGetLinks` out map `[NotNullWhen(true)]`; `SessionMinter` looked-up user `User?`.
- Removed two redundant `?.` in a log call (non-null contract params; the `?.` poisoned flow to maybe-null).
- `typeof(SSOController).FullName!` annotated non-null; the persisted provider-id is properly pinned in #837.

Refs #799

## Type of change
- [x] Refactor / code quality

## Security checklist
- [x] **Touches Session (auth / SSO-only enforcement / session mint) and Linking (persisted canonical links)** → adversarial `/security-review` on this diff before merge.
- [x] Fail-closed preserved: the SSO-only guard, break-glass, last-admin, issuer-binding (#186) and account-provision paths are logic-identical; only annotations changed.
- [x] The persisted `AuthenticationProviderId` value is unchanged (`typeof(SSOController).FullName`, only annotated `!`); the pin is #837's job.

## Verification
- [x] `dotnet test` green (1589 both TFMs on clean run).

## Notes
Remaining NRT: Saml, Oidc, Shared/Flows, then the csproj flip + pragma removal.

### DIFF
--- a/SSO-Auth.Tests/Session/SsoOnlyLoginServiceTests.cs
+++ b/SSO-Auth.Tests/Session/SsoOnlyLoginServiceTests.cs
@@ -514,7 +514,8 @@ public class SsoOnlyLoginServiceTests
         });
 
         var decision = service.ResolveLoginEnforcement(AliceId, configuredDefaultProvider: null);
-        alice.AuthenticationProviderId = decision.DefaultProvider; // the mint repoints to what the decision returned
+        // The mint repoints to what the decision returned; non-null in this break-glass scenario (asserted next).
+        alice.AuthenticationProviderId = decision.DefaultProvider!;
         Assert.Equal(SsoAuthenticationProviders.SsoProviderId, alice.AuthenticationProviderId);
         Assert.Contains(AliceId, config.SsoOnlyRepointedUserIds);
 
--- a/SSO-Auth/Api/Linking/AccountLinkResolver.cs
+++ b/SSO-Auth/Api/Linking/AccountLinkResolver.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System;
 
 namespace Jellyfin.Plugin.SSO_Auth.Api.Linking;
--- a/SSO-Auth/Api/Linking/AdoptionEligibilityResolver.cs
+++ b/SSO-Auth/Api/Linking/AdoptionEligibilityResolver.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 namespace Jellyfin.Plugin.SSO_Auth.Api.Linking;
 
 /// <summary>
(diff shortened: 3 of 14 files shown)


iderex:
## Adversarial `/security-review`, 2 refute-by-default lenses on the diff, both PASS

Run because the diff touches Session (SSO-only enforcement / session mint) and Linking (persisted canonical links).

**Authz / bypass / fail-open lens, PASS.** No enforcement decision, break-glass/last-admin guard, or persisted provider-id value changes. Verified the two `?.`→`.` removals are safe by tracing the non-null contract: `ResolveOrCreateAsync` throws on `IsNullOrWhiteSpace(username)` (CanonicalLinkService.cs:164) before `CreateNewAccountAsync` runs, and `VerifiedIdentity.Provider` is a non-nullable string, so `username`/`provider` are provably non-null at the log sites. `typeof(SSOController).FullName!` assigns a byte-identical persisted string; `config?.Enabled == true` is behaviorally identical (guarded by `links != null`).

**Correctness / persistence-integrity lens, PASS.** Every `[NotNullWhen(true)]` is truthful (TryGetLinks returns true only when `links != null`; all six callers read `links` only on success). The `issuer` widening to `string?` across ~11 methods is signatures-only, a null issuer was already handled fail-closed (`ClassifyIssuer` → Mismatch; `StampIssuer` short-circuits on `IsNullOrWhiteSpace`). The AuthResponse DTO widening surfaces no latent NRE (its only consumer assigns, never dereferences; the fields could already be null from `[FromBody]` deserialization). The canonical-link WRITE path is unchanged beyond annotations.

Conclusion: a genuinely pure, behavior-preserving NRT-annotation refactor. Merging on green CI.

PR #841, opened 2026-07-20
Enable nullable reference types across the Shared module

iderex:
# What & why

Third NRT step (#799): `#nullable enable` for the six Shared (served-page / flow-response) files. Three annotations, no runtime behaviour change (suite 1589 both TFMs):

- `BrowserErrorPage.TryExtractPlainTextError` out `message` → `[NotNullWhen(true)] out string?`.
- `SsoRateLimitGate.Check` → `ActionResult?`, null already meant "allowed, no rejection" at runtime; this states it. The rate-limit / mass-lockout logic is untouched.
- Test helper mirrors the nullable `Check` return.

Refs #799

## Type of change
- [x] Refactor / code quality

## Security checklist
- [x] **Touches `SsoRateLimitGate`** (the per-client rate-limit / mass-lockout defense) → focused availability review on the diff before merge.
- [x] Fail-open-on-unattributable preserved: `Check` returning null = allowed is the pre-existing contract; `NormalizeClientKey`'s null-exemption and `IsAllowed` are unchanged.

## Verification
- [x] `dotnet test` green (1589 both TFMs).

## Notes
Remaining NRT (the crypto/auth core, Saml, Oidc, Flows) deferred to an unhurried session given the SAML validator's ~37 annotation sites and the stakes; then the csproj flip + pragma removal.

### DIFF
--- a/SSO-Auth.Tests/Shared/SsoRateLimitGateTests.cs
+++ b/SSO-Auth.Tests/Shared/SsoRateLimitGateTests.cs
@@ -48,7 +48,7 @@ public class SsoRateLimitGateTests
     // remoteIp is nullable here on purpose: the no-attributable-address case (a null connection peer) is one
     // of the fail-open inputs under test. The gate's own NormalizeClientKey handles null (returns no key), so
     // forwarding it is the behavior being exercised, not a defect — hence the deliberate null-forgiving pass.
-    private static ActionResult Check(string endpointClass, IPAddress? remoteIp, HttpResponse response) =>
+    private static ActionResult? Check(string endpointClass, IPAddress? remoteIp, HttpResponse response) =>
         SsoRateLimitGate.Check(endpointClass, remoteIp!, Substitute.For<ILogger>(), response);
 
     [Fact]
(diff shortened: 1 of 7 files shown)


iderex:
## Availability / mass-lockout review, PASS

Run because the diff touches `SsoRateLimitGate` (the per-client rate-limit / mass-lockout defense).

`Check`'s body is byte-identical; the annotation only makes its already-existing `null` returns explicit. All 11 callers go through the `RateLimitCheck` wrapper and use `if (RateLimitCheck(...) is { } throttled)`, `is { }` matches non-null only, so **null falls through (request proceeds) and non-null short-circuits to the 429**; SSOController isn't `#nullable enable`, so the annotation cannot change dispatch. The fail-open **null-key exemption** (`NormalizeClientKey` returns null for an unattributable client → `IsAllowed` treats null as always-allowed) is untouched, as are the threshold/window/key-composition/Retry-After emission. `TryExtractPlainTextError`'s `[NotNullWhen(true)]` is truthful (message non-null on both true arms, null on the false default; caller reads it only on true).

No behaviour change; the throttle-not-lock, fail-open-on-unattributable mass-lockout defense is fully intact. Merging on green CI.

Issue #807, opened 2026-07-20
Extract the Http boundary module and lock the flat Api kernel empty


iderex:
Kernel dissolution, final step (Closes #790 when merged). After the Routing leaf lands, the flat `Api/` holds only the web boundary: `SSOController`, `RequestHelpers`, `ProviderConnectionTester`. They become the `Http` module:

- `Api/Http`: the controller (transport boundary), its request helpers, and the admin Test-connection probe. DAG: `Http` is the composition boundary, its allowed-imports list is wide by design (it fronts every flow) and documented as such.
- The `OidcDiscoveryOptions` doc `cref` to `ProviderConnectionTester` becomes plain text (cross-module doc rule).
- **New fitness function: the flat `Api/` folder is EMPTY** (no `.cs` directly in `SSO-Auth/Api/`), locks the kernel dissolved for good.
- `docs/ARCHITECTURE.md`: replace the "shared kernel (flat Api)" section with the final module map (Http + Routing rows, updated DAG), close the #790 narrative.

Follow-ups already tracked elsewhere: SSOController size (1131 lines → area partials), `WebResponse` rename.

iderex:
Prerequisite found while scoping this: moving `SSOController` into `Api/Http` changes `typeof(SSOController).FullName`, which is a **persisted** `AuthenticationProviderId` value (stamped by `CanonicalLinkService`, detected by `SsoAuthenticationProviders.IsSsoProvider`). Moving it as-is would silently break SSO-only detection for existing accounts. #837 pins that identifier to its literal legacy string first (a standalone hardening); once it lands, the `SSOController`/`RequestHelpers`/`ProviderConnectionTester` → `Api/Http` move here becomes a behavior-preserving mechanical change. Sequencing: #837 → #807.

Issue #790, opened 2026-07-19
Evaluate a shared protocol abstraction for OIDC/SAML (composition-first OO), reduce duplication


iderex:
## Goal
Push the object-oriented design toward a clean shared abstraction where **OIDC and SAML are specializations of one common "login protocol / auth provider" contract**, polymorphism instead of two parallel code paths, no duplication, each protocol a well-named implementation of a shared shape. The #777 folder migration already makes OIDC and SAML **sibling modules** with an enforced dependency DAG; this is the logical next step: find where they should share a contract.

## Important design guardrail (the honest "fachlich" caveat)
The motivating mental model is an inheritance tree (`Factory ← AutoFactory ← TireFactory`). The **intent**, one shared abstraction, protocols as specializations, is exactly right. But the **mechanism** should be **composition and interfaces first, shallow inheritance only where it is genuinely `is-a`**, not a deep class-inheritance tower. This is a core OO principle ("favor composition over inheritance", GoF), and it matches this codebase's existing grain (sealed types, immutable records, typed values, sum types; the conformance suite enforces sealed/internal helpers and immutable state). Deep inheritance creates fragile base-class coupling and is hard to test and secure in isolation; a shared **interface / small abstract contract** + injected shared services gives the same de-duplication without that fragility. So: pursue the shared abstraction enthusiastically, realize it as an interface/abstract contract + composition.

## Existing seams to evaluate (evidence)
- **Flows:** `OidcLoginService` and `SamlLoginService` are parallel services with the same responsibility (validate → build authorize-state → complete login). Candidate: a common `ILoginService` / small abstract base for the shared skeleton, protocol-specific bits injected.
- **Config:** `OidConfig` and `SamlConfig` **already** derive from a shared `ProviderConfigBase` (#204), a good, shallow, genuine `is-a` example already in the tree. Confirm it carries everything common and nothing protocol-specific leaks up.
- **Authorize-state builders:** `OidcAuthorizeStateBuilder` and `SamlAuthorizeStateBuilder` produce equivalent shapes, candidate for a shared contract / shared assembly of the state.
- **Connection testing:** `ProviderConnectionTester` probes BOTH protocols in one type (surfaced as a cross-protocol coupling during #777, now parked in flat `Api`). Candidate: a per-protocol `IProviderProbe` with an OIDC and a SAML implementation, dispatched by protocol, a textbook place for the shared-abstraction pattern.
[...]

PR #842, opened 2026-07-20
Pin the persisted SSO AuthenticationProviderId, decoupled from the controller type

### LINKED ISSUE #837: Pin the persisted SSO AuthenticationProviderId so it is decoupled from the controller's type location

iderex:
# What & why

The value stamped on `User.AuthenticationProviderId` to mark an SSO-managed / SSO-only account, and compared against to detect one, was `typeof(SSOController).FullName` in two places. That coupled a value **persisted in Jellyfin's user database** to the controller's namespace: any future move of that type (e.g. into an `Api.Http` module, #807) would change `FullName`, so the stamp + detector would move to the new string while **every already-persisted account kept the old one**, silently making existing SSO-only accounts invisible to the enforcement sweep, the break-glass guard, and status reporting.

Fix: pin it. A new `SsoManagedProviderId.Value` const holds the exact legacy literal `"Jellyfin.Plugin.SSO_Auth.Api.SSOController"`; both `CanonicalLinkService` (the stamp) and `SsoAuthenticationProviders` (the detector) reference it. The const lives in the Linking module (the stamp's home); Session already depends on Linking (existing DAG edge), so no new coupling.

Closes #837

## Type of change
- [x] Security hardening (removes a persisted-value / type-location coupling hazard)

## Security checklist
- [x] **No behaviour change: the persisted string is byte-identical** to what every prior version wrote (`"Jellyfin.Plugin.SSO_Auth.Api.SSOController"`), a new conformance test pins the exact literal, asserts the detector resolves to it, and **source-scans both sites to forbid a regression** to `typeof(SSOController)` coupling.
- [x] Fail-closed preserved: the SSO-only enforcement, break-glass and last-admin logic are untouched; only the source of the (identical) id changed.
- [x] Stamp and detector are now provably identical (both reference the one const).

## Verification
- [x] `dotnet test` green (1590 both TFMs, +1 pin test).
- [x] ABI-floor build (10.11.0) green.
- [x] opengrep: no landmine referencing the moved token.

## Notes
This is the documented prerequisite that makes the `SSOController` → `Api/Http` move (#807, closing the #790 kernel-dissolution epic) a behaviour-preserving mechanical change.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -715,6 +715,32 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void SsoManagedProviderId_IsPinnedAndUsedByBothStampAndDetector()
+    {
+        // SECURITY / PERSISTENCE pin (#837). This exact string is written to User.AuthenticationProviderId
+        // and persisted in Jellyfin's user database: every SSO-managed account provisioned by any version
+        // carries it, the stamp (CanonicalLinkService) writes it, and the SSO-only detector
+        // (SsoAuthenticationProviders.IsSsoProvider) compares against it. It MUST NEVER change — a different
+        // value silently stops recognizing every existing SSO account — and it MUST stay decoupled from
+        // typeof(SSOController).FullName so a future move of that type (e.g. into an Api.Http module, #807)
+        // cannot orphan those accounts. The value equals the controller's historical full type name; that is
+        // a coincidence of history, not a live coupling.
+        Assert.Equal("Jellyfin.Plugin.SSO_Auth.Api.SSOController", SsoManagedProviderId.Value);
+
+        // The detector resolves to the same pinned value, so the stamp and the detector can never disagree.
+        Assert.Equal(SsoManagedProviderId.Value, SsoAuthenticationProviders.SsoProviderId);
+
+        // The stamp uses the pin, and neither the stamp nor the detector recomputes the id from the
+        // controller type. Source scans, so a regression to type-coupling fails here even if today's value
+        // still happens to match.
+        var stampSource = string.Join("\n", SourceFilesDeclaring(new[] { typeof(CanonicalLinkService) }).Select(File.ReadAllText));
+        var detectorSource = string.Join("\n", SourceFilesDeclaring(new[] { typeof(SsoAuthenticationProviders) }).Select(File.ReadAllText));
+        Assert.Contains("SsoManagedProviderId.Value", stampSource, StringComparison.Ordinal);
+        Assert.DoesNotContain("typeof(SSOController)", stampSource, StringComparison.Ordinal);
+        Assert.DoesNotContain("typeof(SSOController)", detectorSource, StringComparison.Ordinal);
+    }
+
+    [Fact]
     public void Controller_DelegatesLoginCompletionToTheFlowService()
     {
         // Locked in by the login-completion extraction (#160, #318 step 11): the one shared completion tail —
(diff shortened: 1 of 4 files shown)


iderex:
## Adversarial `/security-review`, 2 refute-by-default lenses on the persisted-identity change, both PASS

Run because this changes how a value **persisted in Jellyfin's user DB** (the SSO-managed `AuthenticationProviderId`) is sourced.

**Authz lens, PASS.** Byte-identity confirmed: `SSOController` is `namespace Jellyfin.Plugin.SSO_Auth.Api` / top-level non-generic `class SSOController`, so `typeof(SSOController).FullName` == `"Jellyfin.Plugin.SSO_Auth.Api.SSOController"` (FullName, no assembly suffix, no `+` nested-separator, no generic backtick), character-for-character equal to the pinned const. Stamp (`CanonicalLinkService.cs:461`) and detector (`SsoAuthenticationProviders.cs:41`) share the one compile-time const and cannot diverge. Every enforcement/break-glass/last-admin/revoke site routes through the helpers and the const; the only other `AuthenticationProviderId` writes assign a flow-resolved provider, untouched. `git grep` finds the literal in exactly three places (const, test expectation, one doc cref), no competing hardcoded comparison.

**Persistence-integrity lens, PASS.** Backward-compat holds (existing accounts carry the identical old value); no migration, no bulk rewrite, no new write path, a pure source-of-the-constant change; the toggle/revoke paths are consistent. The conformance test pins the exact literal (non-vacuous), cross-checks detector==stamp, and source-scans both sites to forbid a `typeof(SSOController)` regression.

Conclusion: the persisted string is byte-identical to every prior version's; the coupling hazard is removed. Merging on green CI. This unblocks the `SSOController` → `Api/Http` move (#807).

PR #848, opened 2026-07-20
Route the OpenID discovery/JWKS/token fetches through the SSRF-hardened transport

### LINKED ISSUE #755: Apply the private-IP/SSRF guard to the OIDC discovery fetch and the IdP-advertised JWKS/token/userinfo endpoints
## Problem
SSRF protection is applied to the user-influenced avatar fetch but not to the OIDC discovery/JWKS/token/userinfo fetches, even though those endpoint URLs come from the **IdP-controlled** discovery document. `AvatarUrlValidator` + `IpAddressClassifier` block loopback/private/link-local and revalidate at connect time (DNS-rebinding defense); `OidcDiscoveryReader.ReadAsync` fetches the admin Authority then trusts `discovery.KeySet`/`TokenEndpoint`/`UserInfoEndpoint` over `SsoHttp.CreateClient` with only `RequireHttps`, no address filter. A malicious/compromised IdP can set `jwks_uri` to an internal address, making the plugin a blind SSRF probe.

## Why it matters for RC
The IdP is a trust anchor, so this is defense-in-depth, but it is the identical threat class the avatar validator exists to stop, and the asymmetry is exactly what an external ASVS/RFC-9700 review flags. Closing it (or documenting a reasoned risk-acceptance) tightens the RC "zero open findings" posture.

## Evidence
- `SSO-Auth/Api/OidcDiscoveryReader.cs` `ReadAsync`; `SsoHttp.cs`; contrast `AvatarUrlValidator.cs`. Prior related: closed #378.

## Acceptance criteria
- [ ] Route the discovery fetch and the IdP-advertised JWKS/token/userinfo fetches through the shared `IpAddressClassifier` block-list, fail-closed with a sanitized log line.
- [ ] Revalidate the resolved address at connect time (reuse the rebinding defense), not just string-check the host.
- [ ] Admin opt-out toggle consistent with `OidcInsecureToggles` for the legitimate intranet-IdP case, **off by default** (secure by default).
- [ ] Negative test: a discovery document whose `jwks_uri` points at a loopback/private address is refused unless the opt-out is set.

iderex:
# What & why

Closes #755. The avatar fetch is transport-SSRF-guarded; the **OpenID backchannel (discovery / JWKS / token / userinfo) was not**, `OidcLoginService` and `OidcDiscoveryReader` built their client from `SsoHttp.CreateClient`, a plain factory client with no connect-time guard, so a provider endpoint (or a discovery redirect) resolving to a private/loopback address could be reached.

**Approach (the clean seam):** consolidate the SSRF-safe transport into `Net/SsoHttp`, the one hardened `SocketsHttpHandler` (`CreateHardenedHandler` + `ConnectToAllowedAddressAsync`, moved verbatim from `AvatarService`, using the `Net` `IpAddressClassifier`) and a named `"sso-outbound"` client. The composition root registers that named client with the hardened handler, so `SsoHttp.CreateClient`, **unchanged at its two OIDC call sites**, now returns the hardened client in production. The avatar drops its private copies and shares `SsoHttp.CreateHardenedHandler` (dedup: one SSRF implementation, two consumers).

**Test seam = the named client.** A stub `IHttpClientFactory` (the controller harness) returns its stub for any name; a loopback fixture returns a default (unhardened) client for `"sso-outbound"`, so an in-process test IdP stays reachable while production is fail-closed. No `OidcLoginService`/`OidcDiscoveryReader`/fixture code changed.

Closes #755

## Type of change
- [x] Security hardening

## Security checklist
- [x] Fail-closed: the OpenID backchannel now rejects a connection to a private/loopback/blocked address at the transport layer, incl. redirect hops and DNS rebinding, sharing the exact avatar guard (`IpAddressClassifier.IsBlockedAddress`).
- [x] No behaviour change for a correctly-configured public IdP; the stub/loopback test factories are unaffected (they own the handler for the named client).
- [x] Proxy disabled on the hardened handler (so the guard validates the real host, not a proxy); redirects bounded; new test pins this configuration against regression.
- [x] Touches the OIDC login-critical HTTP path → adversarial `/security-review` on this diff before merge.

## Verification
- [x] `dotnet test` green (1595 both TFMs; +2 SsoHttp tests, avatar + OIDC endpoint tests unchanged and green).
- [x] ABI-floor build (10.11.0) green.
- [x] opengrep `no-raw-outbound-httpclient` invariant preserved (its comment updated to reflect the shared hardened transport).

## Docs impact
- [ ] After merge: flip the [Security Conformance](https://github.com/iderex/jellyfin-plugin-sso/wiki/Security-Conformance) V12 row from **Partial** to **Met** and note the OIDC parity on the Security-Model SSRF section.

### DIFF
--- a/SSO-Auth.Tests/Net/SsoHttpTests.cs
+++ b/SSO-Auth.Tests/Net/SsoHttpTests.cs
@@ -1,5 +1,4 @@
 using System.Net.Http;
-using Jellyfin.Plugin.SSO_Auth.Api;
 using Jellyfin.Plugin.SSO_Auth.Api.Net;
 using NSubstitute;
 using Xunit;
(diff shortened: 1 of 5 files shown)


iderex:
## Adversarial `/security-review`, 2 refute-by-default lenses on the login-critical HTTP diff, both PASS

**SSRF / bypass lens, PASS.** The `"sso-outbound"` name is a single shared `const` (`SsoHttp.cs:31`) registered with the hardened handler (`SsoOnlyServiceRegistrator.cs`) and resolved by every OIDC fetch (`SsoHttp.cs`, `OidcLoginService.cs:558`, `OidcDiscoveryReader.cs:60`, `ProviderConnectionTester.cs:62`), no literal drift. Confirmed this is a **real net hardening**: before #755, `SsoHttp.CreateClient` returned the *default unhardened* client, so the token/userinfo backchannel was genuinely un-guarded. The moved ConnectCallback guard is byte-equivalent to the proven avatar version (connects to the validated IP, per-hop incl. redirects, `UseProxy=false`, bounded); no bypass found (OidcClient sub-requests resolve the injected client; the callback leg's pre-assigned ProviderInformation disables the library JWKS side-fetch). The stub/loopback test factories keep the in-process IdP reachable.

**Integration / correctness lens, PASS.** DI wiring correct (Microsoft.Extensions.Http via the ASP.NET FrameworkReference; the registrator runs into the same container that injects `IHttpClientFactory`; named-client options resolve the hardened handler). UA applied exactly once (fresh client per `CreateClient`), never doubled. All three OIDC callers reach the hardened client, the admin Test-connection probe now shares the login SSRF posture (comment accurate). Avatar dedup: no regression (`SharedClient` builds on `SsoHttp.CreateHardenedHandler`).

Applied both non-blocking cleanups the lenses flagged: removed the now-dead `using System.Net.Sockets;` in AvatarService, and added `SetHandlerLifetime(Timeout.InfiniteTimeSpan)` to the named client (the documented pattern when you own `PooledConnectionLifetime`, so one long-lived hardened handler serves instead of churning a new one every 2 min). Suite still 1595/1595. Merging on green CI.

Issue #206, opened 2026-07-12
SAML callback: map remaining non-response 500s (cert-config, non-form content-type) to clean errors


iderex:
Follow-up from the #199 adversarial-review gate (correctness + saml + integration lenses, all noted as pre-existing and out of #199 scope). #199 closed the malformed-SAMLResponse 500s; two adjacent SAML-callback paths can still surface an unhandled HTTP 500:

1. Certificate config: the Response constructor decodes config.SamlCertificate. A null/empty SamlCertificate -> Convert.FromBase64String(null) -> ArgumentNullException; valid-base64-but-not-a-cert -> X509CertificateLoader.LoadCertificate -> CryptographicException. Both 500 on every callback. Admin-config-controlled (not attacker/IdP input), low severity, but a misconfigured provider should surface a clear error, not a stack-trace 500.
2. Non-form content-type at the ACS: SamlPost reads Request.Form["SAMLResponse"] directly; a POST with a non-form content-type makes Request.Form throw InvalidOperationException -> 500. Unauthenticated-reachable, denies login (not a bypass).

Fix direction: validate the certificate at provider-save (or guard the decode) and surface a config error; guard the ACS with Request.HasFormContentType and return a clean 4xx. Add tests; endpoint-test layer (#192) + E2E. Refs #199, #137.

iderex:
Part 2 (non-form content-type at the ACS) delivered in PR #207 (merge b3e85f0): SAMLResponse is now bound via [FromForm], so a non-form POST binds null → clean 400 instead of Request.Form throwing a 500. **Remaining (this issue stays open):** part 1, a null/garbage `SamlCertificate` config still 500s on every callback; validate it at provider-save (or guard the decode) and surface a config error.

iderex:
## Implementation spec, part 1 (the remaining cert-config 500)

The `SamlCertificate` decode is the last non-response 500: `Response(string)` does `Convert.FromBase64String(cert)` (→ `FormatException`/`ArgumentNullException`) then `X509CertificateLoader.LoadCertificate(bytes)` (→ `CryptographicException` on garbage/empty bytes). Today `SamlResponseLoader.TryParse` catches only `FormatException`/`XmlException`, so a null/garbage/empty `SamlCertificate` throws `CryptographicException` on every callback → unhandled 500.

### Recommended approach, mirror the #139 BaseUrlOverride pattern exactly
[...]

PR #849, opened 2026-07-20
Enforce a minimum asymmetric signing-key strength for IdP signing keys [#733]

### LINKED ISSUE #733: Enforce a minimum asymmetric signing-key strength for IdP signing keys (OIDC JWKS + SAML certificate)
## Problem
Neither login path enforces a floor on the IdP's asymmetric signing-key size.
- **OIDC:** `OidcIdTokenValidator.ConvertRsaSigningKey` builds an `RsaSecurityKey` from any advertised modulus; Microsoft.IdentityModel's default minimum for RSA *verification* is only 1024 bits, so an RSA-1024 key from the discovery JWKS (or a compromised JWKS) validates.
- **SAML:** `.NET` `SignedXml.CheckSignature` enforces no minimum key size, so a weak IdP certificate is trusted.
We already fail-close on weak *algorithms* (SHA-1) but not weak *keys*.

## Why it matters for RC
The RC gate is "zero open security findings" against a named baseline. OWASP ASVS 5.0 V11 and NIST SP 800-131A mandate RSA ≥ 2048 bits. Without this floor an ASVS self-assessment cannot honestly tick the key-strength control; the audit-grade positioning would be overstated.

## Evidence
- `SSO-Auth/Api/OidcIdTokenValidator.cs` (`ConvertRsaSigningKey`/`ConvertEcSigningKey`/`BuildValidationParameters`, no minimum-key-size config).
- `SSO-Auth/Saml.cs` signature path, no key-size gate. Contrast `SamlSignatureAlgorithms.cs` (already fail-closes SHA-1).

## Acceptance criteria
- [ ] OIDC: skip a discovery JWKS RSA key with modulus < 2048 bits exactly like the existing malformed-key skip; if none usable remains, login fails closed via the existing key-not-found path. Confirm EC keys constrained to the already-allowed P-256/384/521 curves.
- [ ] SAML: reject an IdP cert whose public key is RSA < 2048 bits (or non-approved curve) before/at signature verification, fail-closed with a sanitized log line.
- [ ] The floor is a single shared constant referenced by both paths so they cannot drift.
- [ ] Negative tests: RSA-1024 id_token and RSA-1024 SAML cert both rejected; RSA-2048 + an approved EC curve still validate.
- [ ] README/wiki notes the minimum-key policy so a rare weak-IdP operator gets a clear failure reason, not a silent lockout.

iderex:
# What & why

The `id_token` and SAML assertion signatures are fully validated, but nothing constrained the *strength* of the IdP signing key behind them, a 1024-bit RSA key, or an EC key on an unapproved curve, would validate a signature just fine even though such a key is as forgeable as a weak hash. This adds a single shared minimum-strength floor gating both protocol paths so they cannot drift.

- **New `Api/Crypto` leaf, `SigningKeyStrength`:** `MinimumRsaKeyBits = 2048` (NIST SP 800-131A / OWASP ASVS 5.0 V11), the approved NIST P-256/384/521 curve OIDs, and the two predicates over them. Pure leaf; both `Saml` and `Oidc` now depend on it (new DAG edges + `InlineData` case + `ARCHITECTURE.md` row).
- **OIDC:** `ConvertRsaSigningKey` returns `null` for a JWKS RSA key below the floor, so it is skipped exactly like a malformed key, a token signed by an under-strength key then has no key to resolve against and fails closed via the existing key-not-found (`invalid_signature`) path. EC keys were already constrained to P-256/384/521 by `TryGetCurve`.
- **SAML:** `SamlCertificate.HasAcceptableSigningKey` checks the certificate's public key against the same floor; it gates both the admin write path (`IsInvalid`, so an operator gets a clear rejection at save, not a silent login failure later) and `SamlResponse` signature verification (an under-strength candidate is skipped before `CheckSignature`).

Closes #733

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: an under-strength key is skipped/rejected, never default-accepted; if every candidate is under-strength the signature check fails closed exactly as for a wrong key.
- [x] No secrets logged; secrets are redacted on config export. (No logging added; the SAML skip is silent and the OIDC path reuses the existing sanitized reject.)
- [x] A security review was performed (adversarial 2-lens review on the SAML/OIDC signature-validation surface).
- [x] Log inputs from the IdP are sanitized inline at the log call. (No new IdP-derived log lines.)

## Quality checklist

- [x] No duplicated logic; the floor lives in one small single-purpose leaf referenced by both paths.
- [x] The code is self-documenting; comments explain why (the forgeability rationale), not what.
- [x] Architecture-conformance tests pass; the new `Crypto` leaf and its `Saml`/`Oidc` edges are locked in as `InlineData` in `ApiModule_ImportsOnlyItsAllowedApiModules`.
- [x] Docs impact included here: README fail-closed list, Security-Model + Security-Conformance wiki (V11 key-strength row Partial → Met).

## Verification

- [x] `dotnet build --no-restore --warnaserror` green (net9.0 + net10.0, and the JF 10.11.0 ABI floor).
- [x] `dotnet test` green, 1620 tests, both TFMs.
- [x] Prettier clean for the `.md` change (README content-identical to Prettier output; the local CRLF `--check` warning is a Windows-working-tree artifact CI does not see).

## Notes

The wiki edits (Security-Model, Security-Conformance) live in the wiki repo and are pushed separately after merge. EC curve coverage is unchanged in intent, the floor simply names the same P-256/384/521 set the OIDC converter already enforced, now shared with SAML.


iderex:
Adversarial 2-lens security review (refute-by-default; bypass + correctness lenses read the full touched files and callers): both **PASS**. No signature-floor bypass, fail-open, or spurious rejection of a correctly-configured IdP found; curve OIDs verified (P-256 1.2.840.10045.3.1.7 / P-384 1.3.132.0.34 / P-521 1.3.132.0.35), null/explicit-curve OIDs fail closed, RSA/EC public-key handles disposed, and #206 IsInvalid behavior preserved.

Two non-blocking observations and their disposition:

- **FIX (done, commit c819ffb):** the SAML gate at `SamlResponse.VerifiesAgainstCandidateCertificate` was only covered indirectly via the unit predicate. Added an end-to-end negative, a response signed by an RSA-1024 cert now asserted to fail `IsValid()`, by parametrizing `SamlTestFactory`'s key size (default 2048 unchanged).
- **DECLINE (reasoned):** the OIDC path measures RSA strength via `RsaSecurityKey.KeySize` (= JWK modulus byte-length × 8), so a modulus padded with leading zero bytes could inflate the reported size. This is not a net-new bypass: injecting a padded modulus requires control of the discovery JWKS (fetched over TLS from the trusted issuer), and anyone with that control can simply advertise a genuine full-strength key and forge everything, the floor never defends a JWKS-integrity compromise, so normalizing the modulus adds code (against directive 4) for a threat already subsumed. A legitimate 2048-bit key is exactly 256 bytes (top bit set) and never under-reports, so there is no false-reject risk either.

PR #851, opened 2026-07-20
Support step-up / MFA passthrough for OpenID [#757]

### LINKED ISSUE #757: Support step-up / MFA passthrough (acr_values, prompt, max_age) for OIDC
## Problem
The OIDC authorization request cannot request a specific authentication context or force re-authentication, no `acr_values`, `prompt`, or `max_age`, and the plugin does not check the returned `acr`/`amr`. Admins cannot enforce 'MFA required' or 'always re-prompt' at Jellyfin login.

## Why it matters for RC
Step-up / forced-MFA is a standard enterprise expectation for an internet-exposed media server. Neither main nor V2 supports it, a real ecosystem gap (Authelia/authentik/Keycloak all speak these). P3: hardens advanced deployments, not a core-flow blocker.

## Evidence
- `grep -rniE 'acr_values|prompt|max_age|amr' SSO-Auth/*.cs` → no authorization-request usage. Single insertion point: `OidcLoginService.cs` / `OidcAuthorizeStateBuilder.cs`. OIDC Core §3.1.2.1.

## Acceptance criteria
- [ ] Optional per-provider `AcrValues` (space-separated), `Prompt` (e.g. `login`), `MaxAge` (seconds), sent only when set (empty = unchanged request, upgrade-safe).
- [ ] Optional fail-closed check that the returned `acr` is within the configured `acr_values`, gated behind an explicit opt-in so a mis-set value cannot lock out a userbase; default off for availability (negative test).

iderex:
# What & why

The OpenID authorization request could not request a specific authentication context or force re-authentication, and the returned `acr` was never checked, so an admin could not enforce "MFA required" or "always re-prompt" at Jellyfin login. This adds the standard request parameters plus an opt-in fail-closed `acr` check (#757). Step-up / forced-MFA is a standard enterprise expectation for an internet-exposed media server; Authelia/authentik/Keycloak all speak these.

- **Config (OidConfig):** `AcrValues` (space-separated), `Prompt`, `MaxAge` (int?), `RequireAcr` (bool), each off/empty by default.
- **Request (Part A):** `ChallengeAsync` passes the configured `acr_values` / `prompt` / `max_age` to `PrepareLoginAsync` as front-channel parameters, **each only when set**, an unconfigured provider's authorize request is byte-identical to before (upgrade-safe).
- **Enforcement (Part B):** when `RequireAcr` is on, the callback checks the **signature-verified** id_token's `acr` claim against the configured allow-list (ordinal), after the role gate and **before Promote**, so a login lacking the required context never becomes redeemable, covering both login and manual linking. New `PublicReason.AcrNotSatisfied` → 403. Off by default for availability; fail closed when on.
- **Footgun guard:** a save with `RequireAcr` on but no `AcrValues` is rejected by the config validator (an empty allow-list would refuse every login), so a mis-set cannot silently lock out a userbase or silently no-op.
- **Conformance:** the four fields join the provider-form reverse-pin (35 → 39); `RequireAcr`/`AcrValues` added to the security-critical set and `RequireAcr` pinned out of the auto-expand sensitive-relaxation trigger (it is fail-closed).

STRIDE threat model posted on the issue before implementation.

Closes #757

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] Fail-closed preserved: `RequireAcr` refuses a login whose verified `acr` is absent or not in the allow-list, before any Ready is promoted; no session is minted. Default off.
- [x] No secrets logged; the acr refusal logs only the provider name (sanitized), never the token/claim values.
- [x] A security review was performed (adversarial lenses on the OIDC login/authz surface, see below).
- [x] Log inputs from the IdP are sanitized inline at the log call (`ReplaceLineEndings`).

## Quality checklist

- [x] No duplicated logic; two small single-purpose helpers (`OidcFrontChannelParameters`, `AcrPolicy`), the acr read reuses the validated principal.
- [x] The change adds no more code than the problem requires (no identity-keystone threading, the gate reads the acr at the callback where it is freshest).
- [x] Self-documenting; comments explain the fail-closed "why" and the callback-placement rationale.
- [x] Architecture-conformance tests pass; the four form fields + the security-critical/trigger pins are updated.
- [x] Docs impact included: README + Provider-Setup wiki (pushed with merge).

## Verification

- [x] `dotnet test` green, 1656 tests (both TFMs), +26 new.
- [x] Prettier clean for the `.md`/`.html` change (content-identical to Prettier output; the local CRLF `--check` warning is a Windows-working-tree artifact CI does not see).

## Notes

The acr gate is placed at the callback (not the authenticate leg like the email gate) so it needs no identity-keystone threading and rejects before the auth page, consistent with the role gate. `AcrValues` doubles as both the requested `acr_values` and the accepted-`acr` allow-list. SAML has an equivalent (`RequestedAuthnContext`) not covered here.


iderex:
Adversarial 2-lens review (bypass + correctness, full touched files + callers): both initially **NEEDS_IMPROVEMENT**, two real findings, both fixed and re-reviewed **PASS**.

- **FIX (bypass):** the acr was read from `result.User.Claims`, which with `LoadProfile` on (the default) contains the merged **unsigned** UserInfo claims, so a UserInfo-supplied `acr` (a user's registered assurance level, not this session's) could satisfy a step-up requirement the session never met. Now read from the raw, signature-verified id_token (`OidcIdTokenAcr.Read(result.IdentityToken)`), mirroring how the RFC 9207 `iss` check re-reads `iss`; UserInfo cannot touch `result.IdentityToken`, so the guarantee holds structurally regardless of `LoadProfile`. Locked with a source-scan conformance test.
- **FIX (both lenses):** the `OID/Add` JSON API bypassed `ValidateAcrRequirement` (it persists via `MutateConfiguration`, skipping the save-time `Validate`), so `RequireAcr` with no `acr_values` could be persisted and then refuse every login for that provider (fail-closed reject-all, a silent single-provider lockout). The guard is now mirrored into `OidAdd` alongside its other door-guards; the config-page and import paths were already covered.

Tests added: `OidcIdTokenAcr.Read` reads acr from a raw token and null on a degenerate one; `OID/Add` rejects `RequireAcr` without `acr_values` and accepts it with them; a conformance test pins the acr source. Full gate green, build `--warnaserror` (net9 + net10 + JF 10.11.0 ABI floor), 1666 tests both TFMs.

PR #852, opened 2026-07-20
Map IdP groups to a parental-rating ceiling (MaxParentalRatingScore) [#736]

### LINKED ISSUE #736: Map IdP groups/claims to Jellyfin scalar policy (MaxParentalRatingScore): finish the piece left open by #164
## Problem
The plugin maps IdP roles/claims to boolean `PermissionKind` flags but cannot map them to **scalar policy fields**, chiefly `MaxParentalRatingScore`, or `Preference` values. An operator cannot say "members of the `kids` group get a content-rating ceiling" from SSO, the single most-requested parental control. Issue **#164** scoped this ("PermissionKind / Preference / scalar policy") and was CLOSED with the scalar/Preference boxes unchecked.

## Why it matters for RC
The audience is family/multi-user self-hosters; a content-rating mapping is a visible completeness gap and rounds out the RBAC story. Not RC-blocking; strong RC-era polish.

## Evidence
- Requests: 9p4 #296, #295, #95, #147, #111. Jellyfin demand: features.jellyfin.org/posts/563, discussions/10313. Boolean-only today: `PermissionRolePolicy.cs`, `ProviderConfigValidator.cs:156` (grep `MaxParentalRating` in `SSO-Auth` → nothing).

## Acceptance criteria (fail-closed)
- [ ] Per-provider optional scalar rule: group/claim → integer `MaxParentalRatingScore` ceiling.
- [ ] Multiple matching groups resolve to the **most restrictive** (minimum) ceiling, never the loosest.
- [ ] Unmapped/malformed claims never raise the ceiling (leave existing value or apply configured default); negative test for the malformed-claim branch.
- [ ] Config validation rejects non-integer/out-of-range ceilings at save time.
- [ ] Docs/wiki show a "kids group → max rating" example. `Preference` fields deferred as a follow-up to keep this minimal.

iderex:
# What & why

The plugin mapped IdP roles to boolean `PermissionKind` flags but not to **scalar** policy, chiefly `MaxParentalRatingScore`, so an operator could not say "the `kids` group gets a content-rating ceiling" from SSO, the single most-requested parental control. This finishes the scalar piece #164 left open, for both OpenID and SAML.

- **Config (`ProviderConfigBase`):** `EnableParentalRatingRoles` (bool) + a new `ParentalRatingRoleMappings` list of `{Score, Roles}`, both **XML / config-API only** (mirroring the #164 `PermissionRoleMappings`), default off, no change on upgrade.
- **Reducer:** `ParentalRatingPolicy.Resolve` reduces the login's roles to a single `int?` ceiling, the **minimum (most restrictive)** matched score, or null when the feature is off or no mapping matched. Minimum-wins, so several matched groups never end up looser than the strictest allows.
- **Threaded** as an `int?` through the identity keystone (`OidcAuthorizeState`/`SamlAuthorizeState` → `ValidatedLogin` → `VerifiedIdentity` → `SessionParameters`) and applied in `SessionMinter` under the `EnableAuthorization` master switch by direct assignment to `user.MaxParentalRatingScore`. A null leaves the account's existing ceiling untouched, so an unmapped or malformed claim **never raises** it.
- **Validation:** a negative score or a mapping with no roles is rejected at save, on the config-page save/import **and** the `OidAdd`/`SamlAdd` door-guards.

Closes #736

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] Fail-closed *toward the less-permissive outcome*: minimum-wins, and a null (no match) never raises the ceiling; the write is gated by `EnableAuthorization` like the other role-derived grants. The value is a restrictive content ceiling, never an access grant.
- [x] No secrets logged (no new logging).
- [x] A security review was performed (adversarial authz/correctness lenses, see below).
- [x] Log inputs from the IdP are sanitized inline at the log call (the validator control-strips the echoed provider name; no IdP claim is logged).

## Quality checklist

- [x] No duplicated logic; one pure reducer mirroring `PermissionRolePolicy`, threaded through the same six sites the boolean list uses (as init-only/optional properties, so no existing construction site or test broke).
- [x] Self-documenting; comments explain the minimum-wins and never-raise rationale.
- [x] Architecture-conformance tests pass (no form field added, XML/config-API only, like #164, so the provider-form rosters are unaffected).
- [x] Docs impact included: README RBAC line + a Provider-Setup wiki example (pushed with merge).

## Verification

- [x] `dotnet test` green, 1686 tests (both TFMs), +20 new.
- [x] Prettier clean for the `.md` change (the local CRLF `--check` warning is a Windows-working-tree artifact CI does not see).

## Notes

Modelled as a NEW list field rather than folding into `PermissionRoleMappings`, because that type's runtime `Map` returns boolean `PermissionGrant(Kind, bool)`, a scalar `int?` ceiling cannot ride it. `MaxParentalRatingSubScore` and `Preference` fields are deferred as follow-ups, per the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Sign in to Jellyfin with your existing identity provider — Keycloak, Authelia,
 ## Features
 
 - **OpenID Connect and SAML 2.0** — either or both, multiple providers side by side. See the [Provider Setup](https://github.com/iderex/jellyfin-plugin-sso/wiki/Provider-Setup).
-- **Role-based access control** — map identity-provider groups/roles to login, administrator, library folders, and Live TV.
+- **Role-based access control** — map identity-provider groups/roles to login, administrator, library folders, Live TV, generic permissions, and a per-group parental-rating (content) ceiling.
 - **Hardened, fail-closed login path** — identities bound to the stable `sub` / `NameID`, fail-closed SAML and `id_token` validation, and SSRF-guarded avatar fetches. Details on the [Security Model](https://github.com/iderex/jellyfin-plugin-sso/wiki/Security-Model) page.
 - **Optional SSO-only login** — disable password login for every account except a designated break-glass admin, behind a fail-closed last-admin guard so no configuration can strand a server without a working admin login. Runbook on the [Provider Setup](https://github.com/iderex/jellyfin-plugin-sso/wiki/Provider-Setup#sso-only-login-disable-password-login) page.
 - **Avatar sync, Quick Connect, and self-service account linking**.
--- a/SSO-Auth.Tests/Authz/ParentalRatingPolicyTests.cs
+++ b/SSO-Auth.Tests/Authz/ParentalRatingPolicyTests.cs
@@ -1,0 +1,67 @@
+using System.Collections.Generic;
+using Jellyfin.Plugin.SSO_Auth.Api.Authz;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Tests for <see cref="ParentalRatingPolicy"/> — the role → parental-rating-score reducer (#736). Fail
+/// closed toward the LESS permissive outcome: the feature off, no mapping matched, or an empty allow-list
+/// all yield null (leave the existing ceiling), and when several mappings match the MINIMUM (most
+/// restrictive) score wins, never the loosest.
+/// </summary>
+public class ParentalRatingPolicyTests
+{
+    private static OidConfig Config(bool enabled, params ParentalRatingRoleMap[] maps) => new OidConfig
+    {
+        EnableParentalRatingRoles = enabled,
+        ParentalRatingRoleMappings = new List<ParentalRatingRoleMap>(maps),
+    };
+
+    private static ParentalRatingRoleMap Map(int score, params string[] roles) => new ParentalRatingRoleMap { Score = score, Roles = roles };
+
+    [Fact]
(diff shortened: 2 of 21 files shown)


iderex:
Adversarial 2-lens review (authz + correctness, full touched files + threading callers):

- **Authz / fail-open: PASS**, minimum-wins is correct (never returns the looser score), a null (no match) never raises or clears the existing ceiling (HasValue guard), the apply is gated by `EnableAuthorization`, all three write paths validate (config save / import / `OidAdd`+`SamlAdd`), and even a hypothetical negative score fails *safe* (more restrictive). Two non-blocking doc notes addressed below.
- **Correctness: FIX applied.** Finding: the MIDDLE of the keystone thread (`ValidatedLogin → VerifiedIdentity → SessionParameters`) was exercised by nothing, every hop is optional/defaulted, so a dropped copy line would compile and pass the per-unit tests while the ceiling silently stopped reaching the mint. Fixed: `TestIdentities` (the shared helper that mirrors production destructuring) now threads the field on both protocols; a new end-to-end `LoginCompletionServiceTests` case asserts a resolved ceiling reaches the minted user's `MaxParentalRatingScore` (overwriting a looser pre-existing value); SAML builder-state tests added for symmetry. The ctor/record positional-arg order was verified correct (no transposition), and the XML attributes mirror `PermissionRoleMappings` for round-trip.
- **Doc clarifications (authz non-blocking notes):** the config comment + wiki now state a *match is authoritative* (can widen a hand-set value, like the other role-derived grants), and that the restriction is only as reliable as the IdP's role string.

Full gate green, build `--warnaserror` (net9 + net10 + JF 10.11.0 ABI floor), 1689 tests both TFMs, opengrep clean.

Issue #221, opened 2026-07-13
Escape media-folder Name/Id in _populateFolders to close the DOM-injection sink


iderex:
## Summary

`_populateFolders` builds each media-folder row by interpolating the Jellyfin library `Name`/`Id` into an HTML string assigned to `innerHTML`, with no escaping, the one remaining unescaped `innerHTML` sink in the admin UI.

## Impact, admin self-XSS

`SSO-Auth/Config/config.js:63` builds rows with `out.innerHTML = ... data-id="${folder.Id}" ... <span>${folder.Name}</span>`. A library named `<img src=x onerror=...>` (or an `Id` with a quote breaking out of `data-id`) executes script when the config page renders. The source is the Jellyfin core Library API (admin-controlled, not IdP-controlled), so this is admin self-XSS, not a user -> admin escalation, hence low. The sibling `linking.js` was already hardened to `createElement`/`textContent`; this is the last sink of that class.

## Affected code

- `SSO-Auth/Config/config.js:63-71`

## Fix sketch

Build the row with `document.createElement`, set `checkbox.dataset.id = folder.Id` and `span.textContent = folder.Name` instead of interpolated `innerHTML`, mirroring `linking.js` `populateExistingLinks`. Run Prettier on the changed file.

## V2 status

Closed in V2: `_populateFolders` builds rows via `createElement` + `span.textContent`, with a comment noting `textContent` over `innerHTML`. Ready-made port.

## Notes

Surfaced by the 2026-07-13 standing security audit. Pre-production, tracked publicly per the current triage policy.

PR #853, opened 2026-07-20
Show and copy the exact OIDC redirect URI on the provider form [#724]

### LINKED ISSUE #724: Auto-compute, display and copy the exact OIDC redirect URI on the provider form
## Problem
An admin setting up OIDC must register the plugin's redirect URI at their IdP, but the config page never shows that URL. It appears only as literal placeholder text `<this URL>/sso/OID/redirect/<ProviderName>` buried in the Base-URL-Override callout (`SSO-Auth/Config/configPage.html` ~L1028-1035). The admin must mentally concatenate base URL + fixed path + provider name; a wrong value is the most common OIDC setup failure (`redirect_uri` mismatch).

## Why it matters for RC
The RC bar includes a live functional pass and "a monkey can configure it." Registering the redirect URI is the highest-friction, highest-error step; removing the guesswork de-risks the functional pass and cuts the most likely support ticket.

## Evidence
- Missing on main: no computed-URL field/resolver in `config.js`. Route: `SSO-Auth/Api/SSOController.cs:136` `[HttpGet("OID/redirect/{provider}")]`.
- Proven in V2: `configPage.html` L186-201 (`input#OidRedirectUri` readonly, `data-url-template="/sso/OID/redirect/{name}"`) + `config.js updateEndpointUrls` L620-628.
- Security nuance: login builds the redirect from the **server-side canonical base** (`CanonicalBaseUrl.cs`/`BaseUrlOverride`), which behind a reverse proxy can differ from the browser's `ApiClient.serverAddress()`; the displayed value MUST match what login uses.

## Acceptance criteria
- [ ] Read-only, always-visible field in the OIDC Connection section shows the exact redirect URI for the provider being edited, updating live as the name is typed (placeholder when empty).
- [ ] Copy-to-clipboard button; success announced via aria-live (not colour-only).
- [ ] Displayed URL MUST equal the login-flow URL: derive from Base-URL-Override when set, else the server address; if the two can differ, state which is shown. (A mismatched URI makes the admin register a redirect the server rejects.)
- [ ] Field read-only, no `sso-*` marker class (stays out of the save contract; `ProviderFormFieldIds_MatchOidConfigProperties` green); rendered via textContent/value, never innerHTML (#221).
- [ ] Also surface the legacy `/sso/OID/r/<name>` spelling as a secondary note.

## Suggested approach
Mirror V2's `data-url-template` + resolver, preferring the Base-URL-Override value; optionally add a tiny elevation-gated endpoint returning the canonical URI from `CanonicalBaseUrl.cs`.

iderex:
# What & why

Registering the plugin's `redirect_uri` at the IdP is the highest-friction, highest-error OIDC setup step, the config page never showed it, so an admin had to mentally concatenate base URL + fixed path + provider name, and a wrong value (`redirect_uri` mismatch) is the most common failure. This shows it directly (#724).

- A **read-only, always-visible field** in the OIDC Connection section shows the exact redirect URI, **updating live** as the provider name is typed (placeholder when empty), with a **copy-to-clipboard** button whose success is announced through an **aria-live** region (not colour-only).
- The value derives from the **same canonical base the login uses**, Base URL Override when set, else this server's address, plus the fixed `/sso/OID/redirect/<name>` path, so what is shown equals what the login sends. The help text states which base is shown, notes the reverse-proxy caveat, and surfaces the legacy `/sso/OID/r/<name>` spelling.
- The field carries **no `sso-*` marker** and is read-only, so it never becomes a persisting field (not an `OidConfig` property; `ProviderFormFieldIds_MatchOidConfigProperties` stays green); its value is set via `.value`, **never `innerHTML`** (#221).
- A **conformance test** locks these structural properties (read-only + unmarked, derives from `BaseUrlOverride` + the fixed path, value-not-innerHTML, aria-live confirmation), since no JS runtime harness exists.

Closes #724

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] No new server surface (client-side display only; no new endpoint). The value is set via `.value`, never `innerHTML`, the provider name is not injected as markup (#221), pinned by the conformance test.
- [x] No secrets shown (the redirect URI is public).
- [x] A security review was performed (adversarial correctness/UX + injection lenses, see below).
- [x] N/A, no IdP-supplied input is logged.

## Quality checklist

- [x] No duplicated logic; one resolver mirrors the server-side `OidcRedirectUriBuilder` base + path.
- [x] The change adds no more code than the problem requires (no new endpoint; the existing `BaseUrlOverride` field + `ApiClient.serverAddress()` suffice).
- [x] Self-documenting; comments explain why it must match the server-side base.
- [x] Architecture-conformance: a new fitness test locks the read-only/unmarked/value-not-innerHTML/derives-from-override properties; the `ProviderFormFieldIds` contract is unaffected (the field is unmarked).
- [x] Docs impact included: Provider-Setup wiki notes the form now shows/copies the URI.

## Verification

- [x] `dotnet test` green, 1690 tests (both TFMs), +1 conformance test.
- [x] Prettier clean for the `.js` / `.html` / `.css` change (content-identical to Prettier output; the local CRLF `--check` warning is a Windows-working-tree artifact CI does not see).

## Notes

Client-side only, I deliberately did not add the optional elevation-gated endpoint the issue floated, because the browser already has both inputs the server uses (the `BaseUrlOverride` field and `ApiClient.serverAddress()`), so the resolver reproduces the server's base without a round-trip. Manual browser verification of the live-update + copy is not automatable here (no JS runtime harness), so the structural conformance test is the CI guard.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -714,6 +714,37 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void OidcRedirectUriField_IsReadOnlyAndUnmarked_AndDerivesFromTheSameBaseTheLoginUses()
+    {
+        // #724: the config page shows the exact redirect_uri the login uses so an admin registers it verbatim
+        // (a mismatch is the most common OIDC setup failure). Structural properties a JS runtime test cannot
+        // pin (no JS harness exists), locked as a source scan:
+        //  - the field is READ-ONLY and carries NO sso-* marker class, so it never becomes a persisting field
+        //    (it is not an OidConfig property; ProviderFormFieldIds_MatchOidConfigProperties stays green);
+        //  - its value is set via .value, never innerHTML (#221);
+        //  - it derives from the Base URL Override (the same canonical base the server's OidcRedirectUriBuilder
+        //    uses) plus the fixed /sso/OID/redirect/ path — deriving from anything else would display a URI the
+        //    login does not actually send;
+        //  - the copy confirmation is announced through an aria-live region (not colour-only).
+        var html = File.ReadAllText(Path.Combine(RepoRoot(), "SSO-Auth", "Config", "configPage.html"));
+        var js = File.ReadAllText(Path.Combine(RepoRoot(), "SSO-Auth", "Config", "config.js"));
+
+        var field = Regex.Match(html, "<input\\b[^>]*id=\"OidRedirectUri\"[^>]*>", RegexOptions.Singleline);
+        Assert.True(field.Success, "The read-only #OidRedirectUri field must exist in configPage.html (#724).");
+        Assert.Contains("readonly", field.Value, StringComparison.Ordinal);
+        Assert.DoesNotMatch(new Regex("class=\"[^\"]*sso-(text|line-list|toggle|folder-list|role-map)"), field.Value);
+
+        // The copy confirmation is a live region, not a colour-only signal.
+        Assert.Matches(new Regex("id=\"OidRedirectUri-copied\"[^>]*aria-live", RegexOptions.Singleline), html);
+
+        // config.js derives from the base-URL override + the fixed server path, and writes via .value.
+        Assert.Contains("/sso/OID/redirect/", js, StringComparison.Ordinal);
+        Assert.Matches(new Regex("computeRedirectUri[\\s\\S]{0,400}BaseUrlOverride", RegexOptions.Singleline), js);
+        Assert.Matches(new Regex("#OidRedirectUri\"\\)[\\s\\S]{0,200}\\.value\\s*=", RegexOptions.Singleline), js);
+        Assert.DoesNotMatch(new Regex("OidRedirectUri[\\s\\S]{0,200}innerHTML", RegexOptions.Singleline), js);
+    }
+
+    [Fact]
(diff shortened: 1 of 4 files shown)


iderex:
Adversarial 2-lens review (correctness + injection/DOM-safety):

- **Injection / XSS: PASS**, the redirect URI and the aria-live status are written only via `.value` / `.textContent`, never `innerHTML` (#221); an attacker-influenced provider name renders as inert text. The copy fallback restores `readonly` outside the try/catch (never left writable); listeners bind once to static elements with null-guards. Conformance test pins value-not-innerHTML.
- **Correctness: FIX applied.** Finding: the shown value did NOT match what the login sends, the client stripped only trailing slashes off the raw base, but the server's `CanonicalBaseUrl` (`System.Uri.GetLeftPart(UriPartial.Path)`) also lowercases scheme+host and elides the default port (443/80). So a valid-but-non-canonical override (`https://X.COM:443`, `HTTPS://…`) displayed a URI the login normalized away, the exact `redirect_uri` mismatch this feature prevents. Fixed by normalizing through the WHATWG URL parser (`new URL(raw).origin + pathname`), which matches `System.Uri` for every persistable override (Node-verified: default-port elision, scheme/host lowercasing, sub-path preservation, trailing-slash/query/fragment dropping). Re-reviewed: **PASS**. The conformance test now pins the origin+pathname normalization. The no-override case remains inherently best-effort (a future login's Host header is unknowable client-side), setting a Base URL Override is the documented remedy, which the field help already advises.

Full gate green, build `--warnaserror` (net9 + net10 + JF 10.11.0 ABI floor), 1690 tests both TFMs, prettier clean.

Issue #944, opened 2026-07-22
The JF12 publish legs run from branch 4.2, so main-only pipeline fixes never reach them


iderex:
Found by the adversarial review of #942.

`nightly-betas.yml` dispatches `publish-jf12-beta.yml` with `--ref 4.2`, and `publish-jf12-stable.yml` is triggered by a tag cut on the `4.2` lineage. GitHub executes a workflow file **from the ref it is dispatched against**, so the copies that actually run are the ones on `4.2`, not the ones on `main`.

Verified against the `4.2` branch:

- `.github/actions/` **does not exist there at all** (both `verify-manifest` and `cyclonedx-sbom` are 404)
- `4.2`'s `publish-jf12-beta.yml` still has the pre-SBOM checksum step
- `publish-jf12-stable.yml` does not exist on `4.2`, although its trigger is a JF12 stable tag on that lineage, so it would not run at all

## Consequences

1. **The #942 fix edits to both JF12 workflows on `main` are inert.** The fix covers the two JF10.11 legs in practice. That is not a correctness problem today, the JF12 artifact is still named `sso-authentication_*`, which sorts *after* `sbom.cyclonedx.md5`, so it picks the right checksum by luck, but it is luck, not design.
2. **The luck runs out on rename.** When the `4.2` line adopts the `community-sso-for-jellyfin` name (or the SBOM step), it must pick up the #942 fix **in the same merge**, or the JF12 channel breaks exactly the way the beta channel just did.
3. **A local action referenced from a branch that lacks it fails at step resolution**, after the release is sealed and the manifest pushed, i.e. at the worst possible moment.

Nothing in the repo enforces any of this ordering.

## What to do

- Decide and write down the rule: either forward-merge `main` → `4.2` before the rename lands there, or move the JF12 dispatch to a ref that carries the pipeline. The Releasing wiki page is the natural home.
- Consider a guard: a check that every `uses: ./.github/actions/...` referenced by a workflow resolves on the ref that workflow is dispatched against.
- Until then, treat "fixed in the publish pipeline" as meaning "fixed for the JF10.11 legs".

Refs #942, #135, #607.

iderex:
[...]

Issue #942, opened 2026-07-22
Beta channel: installing 4.3.0-beta.10 fails, the manifest publishes the SBOM's checksum, not the plugin's


iderex:
## Symptom

Installing the plugin from the **beta** repository URL on Jellyfin 10.11 fails. Jellyfin reports only a generic failure; nothing in its log points at the cause.

## Cause

Jellyfin verifies the downloaded zip against the `checksum` in the manifest and refuses to install on a mismatch. For `4.3.0.10` the manifest publishes a checksum that belongs to a **different file**:

| | |
| --- | --- |
| manifest `checksum` | `b702f1be7645682d87ea2c99066741f8` |
| actual zip MD5 | `efe81200364f18a3630011c0c1f504dd` |
| `sbom.cyclonedx.json` MD5 | `b702f1be7645682d87ea2c99066741f8` ← **this is what the manifest published** |

The download itself is fine (HTTP 200, 880 375 bytes) and both published checksum sidecars agree with the zip. Only the manifest is wrong.

### Why

The manifest generator, `Kevinjil/jellyfin-plugin-repo-action@v0.4.3`, selects a release's checksum **by name and keeps the last match** - no `break`, and no pairing against the zip:

```js
for (asset of release.assets) {
  if (asset.name.endsWith('.zip')) { sourceUrl = asset.browser_download_url }
  if (asset.name.endsWith('.md5'))  { checksum = ...  }   // no break — last one wins
}
```

GitHub returns assets alphabetically, and the publish legs emit **two** `.md5` assets, one for the plugin zip, one for the SBOM. Renaming the plugin to `community-sso-for-jellyfin` (#913) flipped which of them sorts last:

| release | `.md5` assets, in order | last match |
| --- | --- | --- |
| `4.3.0-beta.9` (old name) | `sbom.cyclonedx.md5`, `sso-authentication_4.3.0.9.md5` | **the plugin's** ✔ |
| `4.3.0-beta.10` (new name) | `community-sso-for-jellyfin_4.3.0.10.md5`, `sbom.cyclonedx.md5` | **the SBOM's** ✘ |

Confirmed: beta.9's manifest checksum matches its zip exactly; beta.10's matches the SBOM.

## Blast radius

- **Beta channel, Jellyfin 10.11: broken** since beta.10 (the first beta after the rename). Every install and update fails.
- **Stable channel: not yet broken, but will break identically** at the next stable release, 4.2.1 predates the rename.
[...]

PR #947, opened 2026-07-22
Generate the plugin manifest in-house: paginated, checksum-paired, per-ABI capped

### LINKED ISSUE #943: Beta manifest is capped at 30 releases: the JF12 line will silently disappear from it
Found by the adversarial review of #942, and it is a live countdown rather than a hypothetical.

## The cap

`Kevinjil/jellyfin-plugin-repo-action@v0.4.3` calls `octokit.rest.repos.listReleases(repo)` **unpaginated**, it sees only the first 30 releases. Measured right now:

- **32** non-draft releases exist
- `manifest-beta` holds exactly **30** versions, the cap is already reached and two releases have already fallen off
- of those 30: **12** are JF10.11 (`10.11.0.0`) and **18** are JF12 (`12.0.0.0`)

## Why it becomes an outage

Betas publish daily. Every new JF10.11 beta pushes the JF12 entries one place further down. Once all 18 JF12 entries fall off page 1, `targetAbi 12.0.0.0` **vanishes from `manifest-beta` entirely**, and because Jellyfin filters client-side by ABI, every Jellyfin 12.0 beta user's repository simply goes empty. No error, no stale-but-working entry; the plugin just stops being offered.

The JF12 beta leg publishes far less often than the JF10.11 one (it is dispatched against the `4.2` branch, which moves slowly), so the JF10.11 line will win this race.

## Why the new #942 gate does not catch it

The verification added in #942 iterates the ABIs **present in the fetched manifest**. A generation that has already vanished yields no iteration for that ABI, and `expect-version` only asserts the publishing leg's own version. So a JF10.11 beta publish that evicts the last JF12 entry passes green, the gate's description promises "the newest entry of EVERY targetAbi … the entry a fresh install of that generation is offered", which stops being true once a generation is gone.

## Options

1. **Assert an expected ABI set** in `.github/actions/verify-manifest`, the manifest must contain at least one entry for each ABI the project ships. Cheap, and closes the detection gap immediately even if the cap is not addressed.
2. **Address the cap itself**: prune old beta prereleases so the recent set of both generations stays inside the first 30, or replace/patch the generator with a paginated listing. Pruning is the smaller change but deletes immutable releases (which permanently burns their tags, see `publish.yml`'s header), so it needs care: prune only the *oldest* betas, never a tag that might be re-used.
3. Longer term, stop sharing one manifest branch between the generations. That was a deliberate design choice (one repository URL serves both), so it is the last resort.

Suggested order: (1) now, so the failure is at least loud; then (2).

Refs #942, #135.

iderex:
## What & why

Replaces `Kevinjil/jellyfin-plugin-repo-action@v0.4.3` with an in-house composite action (`.github/actions/generate-manifest`), called by all four publish legs. The third-party action caused two user-facing outages in one week and can be configured out of neither:

- **#942**, it selected `checksum` as the LAST release asset ending in `.md5`, with no pairing to the zip. The plugin rename moved the plugin's md5 ahead of `sbom.cyclonedx.md5` alphabetically, so the manifest published the SBOM's checksum and every install failed Jellyfin's MD5 verification.
- **#943**, it listed releases unpaginated (first 30 only). With 33 releases and one manifest branch shared by both Jellyfin generations, the JF12 entries were being squeezed off the end; ~18 further daily publishes from losing the plugin for every JF12 user.

## How the replacement closes the classes

- Checksum = the sidecar **belonging to** the selected zip, validated as a **single `md5sum` line naming exactly that zip**; an unresolvable or ambiguous pair on the release being published is fatal.
- Release listing is **paginated** (`per_page=100`); the manifest is capped **per targetAbi** (default 15), so no generation can be squeezed out by another's cadence.
- The plugin identity comes from the metadata JPRM wrote beside the zip **this run built** (new required `plugin-metadata` input), no nondeterministic discovery from the release list mid-rename.
- The one guarantee the action exists to give is asserted: **the run fails unless the build this run published survives into the committed manifest.** This covers the two stable legs, which pass no `expect-version` to `verify-manifest`.
- Fail-closed is **scoped to the release being published**: a defect on an immutable historical release (two zips, lost sidecar, unreadable metadata) is a loud warning-plus-drop, not a permanent freeze of both manifest branches. Dropping can never silently cost a generation: the zero-entry guard, the own-build assertion, and `verify-manifest`'s undated ABI-shrink check remain.
- Exactly-one cardinality is asserted for the zip **and** the md5 candidates; the changelog lookup carries the same draft filter/`first()` cutoff as the asset lookup.
- The four prepare steps write sidecars with `>` (truncate) instead of `>>`, removing the append hazard that made a multi-line sidecar reachable.
- `build-jf12.yaml`'s `overview`/`description` are reconciled word-identical with `build.yaml`'s (comment pins them): both beta legs write the shared `manifest-beta`, which holds one plugin object from the publishing leg, divergent prose alternates the catalogue entry nightly and defeats the unchanged-manifest commit skip.

## Empirical verification (not reasoning)

- Generated against the **live repository** (all 33 releases): plugin object **byte-identical** to the published manifest; 26/27 shared version entries byte-identical; the 27th is `4.3.0.10`, the entry the replaced action got wrong.
- **All 29 generated checksums verified by downloading each artifact and computing its MD5, 29/29 match**, including `4.3.0.10`, where the live manifest does not.
- All **34 live md5 sidecars** dumped; each is a single `md5sum` line naming its zip, so the strict whole-line validation is safe against the real history.
- **Five negative tests** against the real release payload: own-release-missing → fail; historical two-zip release → dropped, run green; own two-zip release → fail; historical lost sidecar → only that entry dropped; empty history → the written diagnostic (previously an unreachable branch behind a jq crash).
- Every touched YAML strict-parsed with **js-yaml** (not yq, which tolerates the duplicate-key class that caused the earlier startup failure).

## Adversarial review

4-lens gate (availability / supply-chain-bypass / correctness-shell-semantics / integration), refute-by-default, each lens executing the generator against the live payload. All four returned NEEDS_IMPROVEMENT; every finding above traces to one of them and is fixed in the second commit. Reasoned declines:

- *Serial sidecar fetches grow with history (~2/release).* `per_page=100` fixes the listing; the per-release fetches remain O(history). At current cadence the 15/20-minute job timeouts are years away; deferring fetches to post-cap entries would need version/ABI before the fetch that provides them. Tracked as follow-up rather than complicating this PR.
- *Three legs run the manifest step inline in the job that sealed the immutable release, so "re-run failed jobs" cannot reach it.* Pre-existing structure; a manifest-only regeneration dispatch is the right recovery path and is tracked as follow-up.

## Sequencing (#944), the remaining deployment step

`nightly-betas.yml` dispatches the JF12 beta leg with `--ref 4.2`, and branch `4.2` has **no `.github/actions/` at all**, its `publish-jf12-beta.yml` still runs the old action, and `publish-jf12-stable.yml` does not exist there. The 4.2 gate is currently dormant (branch HEAD equals the last beta's commit), so nothing regenerates from 4.2 tonight, but the **first commit landing on 4.2 wakes it** and the old generator would overwrite `manifest-beta` (restoring both #942 and #943, with no detector on that copy). Therefore the port of `.github/actions/*` + the two JF12 workflows to `4.2` follows **immediately after this merges**, before anything else lands on 4.2.

Closes #943. Refs #942, #944.

### DIFF
--- a/.github/actions/generate-manifest/action.yml
+++ b/.github/actions/generate-manifest/action.yml
@@ -1,0 +1,205 @@
+name: Generate the Jellyfin plugin manifest
+description: >-
+  Build and commit the Jellyfin plugin repository manifest from this
+  repository's releases.
+
+  This replaces Kevinjil/jellyfin-plugin-repo-action, which produced two
+  user-facing outages in a week and could not be configured out of either:
+
+  - It selected a release's checksum BY NAME, keeping the LAST asset ending in
+    `.md5` with no pairing against the zip. Renaming the plugin moved its md5
+    ahead of `sbom.cyclonedx.md5` alphabetically, so the manifest published the
+    SBOM's checksum and every install failed verification (#942). Here the
+    checksum is taken from the sidecar that BELONGS to the selected zip, and a
+    release whose pair cannot be resolved fails the run instead of guessing.
+  - It listed releases UNPAGINATED, so only the first 30 were ever considered.
+    With one manifest branch shared by both Jellyfin generations and the JF10.11
+    line publishing far more often, the JF12 entries were being pushed off the
+    end — after which that generation silently vanishes from the manifest and
+    every Jellyfin 12.0 user's repository goes empty (#943). Here the release
+    list is paginated, and the manifest is capped PER targetAbi rather than
+    globally, so no generation can be squeezed out by another's cadence however
+    long the release history grows.
+
+  The output is byte-compatible with what the previous action produced: the same
+  plugin and version fields, keys sorted, two-space indent, versions in
+  descending version order.
+inputs:
+  repository:
+    description: owner/repo whose releases the manifest is built from.
+    required: true
+  manifest-branch:
+    description: Branch the manifest is committed to (e.g. manifest-beta).
+    required: true
+  manifest-file:
+    description: Path of the manifest within that branch.
+    required: false
+    default: manifest.json
+  include-prereleases:
+    description: >-
+      "true" for the beta channel (prereleases are the point), "false" for
+      stable.
+    required: true
+  max-versions-per-abi:
+    description: >-
+      How many versions to keep for EACH targetAbi. Per-ABI rather than an
+      overall limit is the structural fix for #943: a generation that publishes
+      rarely keeps its entries no matter how often the other one publishes.
+    required: false
+    default: "15"
+  github-token:
+    description: Token used to read releases and commit the manifest.
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Generate and commit the manifest
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        REPO: ${{ inputs.repository }}
+        BRANCH: ${{ inputs.manifest-branch }}
+        MANIFEST_FILE: ${{ inputs.manifest-file }}
+        INCLUDE_PRERELEASES: ${{ inputs.include-prereleases }}
+        MAX_PER_ABI: ${{ inputs.max-versions-per-abi }}
+      run: |-
+        set -euo pipefail
+        work="$(mktemp -d)"
+
+        # Plugin-level metadata (guid, name, description, …) is taken from the NEWEST release
+        # that ships JPRM's metadata JSON — filled in by the release loop below, which walks
+        # releases newest-first. These are properties of the plugin rather than of a release,
+        # so any recent one carries the same values.
+        : > "$work/plugin.json"
+
+        # --paginate is the whole point: without it only the first 30 releases are seen.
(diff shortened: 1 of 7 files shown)


iderex:
Pre-merge summary: 4-lens adversarial gate ran twice (initial NEEDS_IMPROVEMENT x4, all findings fixed or reasoned-declined in the PR body; declines tracked as #948/#949). Empirical verification against the live repository: manifest byte-identical apart from the two entries this exists to fix, 29/29 checksums match the real artifacts, 34/34 sidecars satisfy the strict validation, five negative tests green. CI green. The 4.2 port ([#944]) follows immediately after this merge, before anything else lands on 4.2.

Issue #913, opened 2026-07-21
Internationalize all user-facing surfaces into every official EU language (Jellyfin-style catalogs)


iderex:
## Problem

Every user-facing string the plugin emits is **hardcoded English**: the login-page "Sign in with …" buttons, the OIDC/SAML browser error pages, the admin config-page labels and descriptions, and the account-linking page. There is **no culture/localization mechanism** at all, a repo-wide grep finds only invariant-culture number/date formatting, never a string lookup. Jellyfin itself is heavily internationalized; an English-only plugin is a real completeness gap for a global user base.

This supersedes the narrower #759 (which scoped only "login buttons / error pages / config UI" as a P3 polish item and named no target languages or mechanism). Fold #759 into this epic.

## Goal

1. Localize **all** user-facing surfaces into the official languages of the EU member states and of the EU’s neighbouring and associated countries. The committed set (BCP-47):

   > Albanian (`sq`) · Armenian (`hy`) · Basque (`eu`) · Belarusian (`be`) · Bosnian (`bs`) · Bulgarian (`bg`) · Catalan (`ca`) · Crimean Tatar (`crh`) · Croatian (`hr`) · Czech (`cs`) · Danish (`da`) · Dutch (`nl`) · English, British (`en-GB`) · Estonian (`et`) · Faroese (`fo`) · Finnish (`fi`) · French (`fr`) · Galician (`gl`) · Georgian (`ka`) · German (`de`) · Greek (`el`) · Hungarian (`hu`) · Icelandic (`is`) · Irish (`ga`) · Italian (`it`) · Kalaallisut / Greenlandic (`kl`) · Latvian (`lv`) · Lithuanian (`lt`) · Luxembourgish (`lb`) · Macedonian (`mk`) · Maltese (`mt`) · Norwegian Bokmål (`nb`) · Norwegian Nynorsk (`nn`) · Polish (`pl`) · Portuguese (`pt`) · Romanian (`ro`) · Slovak (`sk`) · Slovenian (`sl`) · Spanish (`es`) · Swedish (`sv`) · Turkish (`tr`) · Ukrainian (`uk`)

   English (`en-GB`) is the invariant fallback. Further languages can be added later behind the same mechanism.

[...]

PR #950, opened 2026-07-22
Port the in-house manifest generator to the 4.2 publish legs

iderex:
## What & why

Deploys the reviewed #947 manifest tooling to the branch that actually executes the JF12 legs. `nightly-betas.yml` dispatches the JF12 beta with `--ref 4.2`, so it runs **this branch's** workflow copy, which still ran the replaced third-party action against the shared `manifest-beta`, and `publish-jf12-stable.yml` did not exist here at all (a `5.0.0-JF12-stable` tag on a 4.2 commit would silently run nothing).

Without this port, the first commit landing on 4.2 wakes the nightly gate and the old generator overwrites the repaired `manifest-beta`: the SBOM checksum (#942) and the 30-release window (#943) restored, with no detector on this copy. This PR *is* that first commit, and it carries the fix with it.

## Contents (verbatim from the merged #947 state)

- `.github/actions/generate-manifest` + `verify-manifest`, plus `cyclonedx-sbom`, which the updated workflows call
- `publish-jf12-beta.yml` / `publish-jf12-stable.yml`, identical to main's copies
- `build-jf12.yaml`: the #913 rename (this line still built under the old name) and `overview`/`description` reconciled word-identical with `build.yaml`, pinned by comment, both beta legs write the shared `manifest-beta`, which holds one plugin object taken from the publishing leg, so divergent prose alternates the catalogue entry nightly

The next JF12 beta zip becomes `community-sso-for-jellyfin_5.0.0.<n>.zip`; the generator pairs checksums by name, so the rename cannot recreate #942's alphabetical-ordering failure. The generator itself was empirically verified on #947 (live-repository generation byte-compared, 29/29 artifact checksums, five negative tests); this PR adds no logic, the port's own risk is wiring, and the post-publish `verify-manifest` step (expect-version + ABI-shrink check) now guards this leg too.

Refs #943, #944, #942.

### DIFF
--- a/.github/actions/cyclonedx-sbom/action.yml
+++ b/.github/actions/cyclonedx-sbom/action.yml
@@ -1,0 +1,51 @@
+name: Generate CycloneDX SBOM
+description: >-
+  Emit a CycloneDX JSON SBOM for the SSO-Auth plugin from the locked restore
+  graph the caller has already materialised, so a release can carry the
+  dependency inventory CRA Annex I / OSPS-QA-02.02 expect. The single generator
+  for every publish leg (DRY): each leg calls this action, none re-implements it.
+  The SBOM covers the plugin's FULL multi-target dependency closure (both the
+  net9.0 / Jellyfin 10.11 and net10.0 / Jellyfin 12.0 lines) as pinned in the
+  committed packages.lock.json — a conservative superset for any single leg's
+  shipped ABI, never an under-report. Which target framework a given release
+  ships is identified by its tag and zip.
+inputs:
+  output-dir:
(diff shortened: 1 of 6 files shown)


iderex:
Pre-merge note: base healed by #951 (crypto advisory bump + stale pin comments), branch rebased onto it, full check set green. The port is verbatim the reviewed #947 state; its own risk is wiring only, and verify-manifest (expect-version + ABI-shrink) now guards this leg's nightly output. Merging now so the port is on 4.2 before the next nightly dispatch wakes the gate.

Issue #954, opened 2026-07-22
Consolidate the JF12/5.0 line onto main; beta-only until first stable


iderex:
Decision (2026-07-22): the software is beta - no stable releases for now, one code line only.

- main becomes multi-target (net9.0;net10.0) with the per-TFM Jellyfin references from the 4.2 branch, so BOTH betas (JF11 4.3.0.x, JF12 5.0.0.x) build from main. JF12 users finally get the current code line (SLO, rename, role features) instead of the frozen 4.2.1-era stand the 4.2 branch has shipped for 365 commits.
- Re-derive build-jf12.yaml's net10 shipping closure empirically (dotnet publish -f net10.0) - the dependency set changed since the closure was last synced.
- nightly-betas.yml dispatches the JF12 leg with --ref main; CI (dotnet.yml) builds/tests both TFMs.
- The stable publish legs (publish.yml, publish-jf12-stable.yml) stay dormant (tag-gated) for the eventual first stable.
- Stable channel retired until then: README/wiki document beta-only; manifest-release is emptied so the stable repo URL offers nothing - no new installs of 4.2.1-stable, which ships the vulnerable System.Security.Cryptography.Xml 10.0.9 DLL (the four GHSA advisories fixed on main by #890 and on 4.2 by #951). Existing stable installs keep running; the docs point to the beta URL.
- Branch 4.2 is deleted after the JF12-from-main beta is live and verified (release tags remain).

Refs #743, #944, #890.

iderex:
Done end to end:
- #955 merged: nightly-betas dispatches the JF12 leg with --ref main; gate paginated (the >30-tag duplicate-beta loop is closed pre-emptively); build-jf12.yaml prose byte-identical with build.yaml; README installs beta-only.
- First main-built JF12 beta is live and independently verified: 5.0.0-JF12-beta.25, community-sso-for-jellyfin_5.0.0.25.zip, checksum matches the artifact, update ordering monotonic (>.24). JF12 users now receive the current code line.
- manifest-release emptied to [] (the stable URL offers nothing until the first stable release; 4.2.1-stable with the vulnerable crypto DLL is no longer installable from it).
- Branch 4.2 deleted (ruleset now protects main + 5.0); all release tags remain, so historical release assets and the generator's per-tag metadata fallback stay valid.

Issue #890, opened 2026-07-20
Vulnerable dependency: System.Security.Cryptography.Xml 10.0.9 (CVE-2026-47302, -47304)


iderex:
The pinned System.Security.Cryptography.Xml 10.0.9 (the SAML XML-signature verification library) is hit by two newly-published HIGH advisories, GHSA-cvvh-rhrc-wg4q (CVE-2026-47302) and GHSA-g8r8-53c2-pm3f (CVE-2026-47304), both DoS via unthrottled XML resource allocation (CVSS 7.5). The CI Vulnerable-dependency-scan fails on every PR, blocking the queue. Fix: bump to the patched 10.0.10.

Issue #135, opened 2026-07-12
Prepare the Jellyfin 12.0 build: net10.0, Jellyfin.Controller 12.0.0-rcX, targetAbi 12.0.0.0


iderex:
Jellyfin 12.0 is in late RC (RC1 2026-06-21, RC2 2026-06-28, https://github.com/jellyfin/jellyfin/releases/tag/v12.0-rc1) and moves the server to .NET 10; 10.11-ABI plugins will not load, and users are told to reinstall 12-targeted builds from the unstable repo. Official plugins (LDAP Auth et al.) already ship targetAbi 12.0.0.0 to the unstable manifest (https://repo.jellyfin.org/files/plugin-unstable/manifest.json); the 12.0 API surface is frozen (State of the Fin 2026-05-24, https://jellyfin.org/posts/state-of-the-fin-2026-05-24/). Two riders: net9.0 reaches end of support 2026-11-10 (.NET Blog, https://devblogs.microsoft.com/dotnet/dotnet-8-9-end-of-support/), so the net10.0 (LTS) retarget lands with this port; and Duende OidcClient development moved to the duendesoftware/foss monorepo with the standalone repo archived (https://github.com/DuendeArchive/IdentityModel.OidcClient), verify 7.x on net10 empirically and repoint advisory monitoring. Without a 12-targeted build at 12.0 stable, every upgrading user loses SSO login. Keep the 10.11 line for security patches per BRANCHING.md.

Acceptance criteria:
- [ ] Branch or multi-target builds green on net10.0 against Jellyfin.Controller/Model 12.0.0-rcX.
- [ ] Full test suite passes on net10.0, including the OidcClient-dependent paths (verified empirically, not assumed).
- [ ] Publish path with targetAbi 12.0.0.0 prepared; 10.11 security-patch line documented.
- [ ] Advisory monitoring repointed to duendesoftware/foss and the NuGet package.

iderex:
[...]

PR #955, opened 2026-07-22
Publish both betas from main; retire the stable channel until first stable

iderex:
## What & why

Decision (#954): the software is beta, one code line, beta-only releases.

- **JF12 betas build from `main`.** `main` has been the multi-target codebase (net9.0;net10.0, per-TFM Jellyfin refs 10.11.11 / 12.0.0-rc2, per-TFM OidcClient 6.0.1/7.1.0) since #135/#743, only the nightly dispatch still pointed at the retired `4.2` branch, which froze JF12 users on 4.2.1-era code for 365 commits of `main`. Now `nightly-betas.yml` dispatches `--ref main`.
- **Closure verified empirically:** `dotnet publish -f net10.0` on `main` produces exactly the dependency set `build-jf12.yaml`'s artifacts list ships (the Jellyfin/server assemblies excluded, the crypto-XML set correctly absent on net10 where it is framework-provided).
- **Gate semantics:** the first run from `main` always builds (the last JF12 tag points at a 4.2 commit, never `main` HEAD, the inequality fires); afterwards the gate compares `main` HEAD as before. The `target_commitish` pin stays, correct on `main` too, since the default-branch head can advance past the dispatched commit by publish time.
- **README installs beta-only:** stable row removed, beta-software status stated, migration note points at the beta manifest. After merge, `manifest-release` is emptied so the stable URL cannot offer `4.2.1-stable`, which ships the vulnerable `System.Security.Cryptography.Xml` 10.0.9 DLL (#890's fix never reached a stable release). The stable publish legs stay dormant (tag-gated) for the eventual first stable.
- `build-jf12.yaml`: stale `manifest-jf12-*` note corrected (one manifest branch per channel, ABI-filtered client-side).

## Rollout (after merge)

1. Dispatch the JF12 beta from `main` and verify the published `manifest-beta` end to end (new entry's checksum vs the real artifact, both ABIs present).
2. Empty `manifest-release`.
3. Delete branch `4.2` (release tags remain).

Refs #954, #944, #890.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -21,9 +21,8 @@ name: CodeQL
 
 on:
   push:
-    # main is the released line; the release/dev branches are where feature work
-    # accumulates before promotion. 4.2 is the current one; the 4.* / 5.* globs
-    # future-proof the next release lines so a new branch is covered without
+    # main is the one code line (beta-only, #954; the 4.2 branch is retired). The
+    # 4.* / 5.* globs future-proof any later release branch so it is covered without
     # editing this file (a `*` matches any branch name that has no `/`, so plain
     # numeric release branches like 4.2 / 4.3 / 5.0 all match). A release line
     # that ever uses a non-numeric name must still be added here explicitly.
--- a/.github/workflows/nightly-betas.yml
+++ b/.github/workflows/nightly-betas.yml
@@ -2,9 +2,10 @@ name: Nightly betas
 
 # Publish betas on a daily cadence instead of on every push (#632). GitHub `schedule` events only run
 # a workflow from the DEFAULT branch, so this single scheduler lives on main and dispatches both beta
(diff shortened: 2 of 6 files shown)


iderex:
Pre-merge: 2-lens adversarial review (availability, integration) on the wiring. Verified clean by executing against live state: run_number continuity (next JF12 beta is 5.0.0.25 > .24, ordering monotonic), first-run gate behavior, no code lost by retiring 4.2 (all 9 non-main commits have main equivalents), branch-deletion fallout, manifest-release emptying vs the dormant stable legs. Both findings fixed in the second commit: the unpaginated tag lookup (would have minted daily duplicate betas once >30 tags exist) and the still-divergent build-jf12.yaml description. Post-merge rollout: dispatch JF12 beta from main, verify manifest end to end, empty manifest-release, delete 4.2 (needs the branch-protection ruleset edited first).

PR #958, opened 2026-07-22
Add a manifest-only regeneration dispatch as a recovery path

### LINKED ISSUE #948: Manifest-only regeneration dispatch as a recovery path
Three publish legs (publish-beta, publish-jf12-beta, publish-jf12-stable) run the generate-manifest step inline in the job that already created and sealed the immutable release. If the manifest step fails transiently, "re-run failed jobs" re-enters the sealed release at action-gh-release and never reaches the manifest step; publish.yml is the only leg with a separately re-runnable generate job.

Add a small workflow_dispatch-only workflow that runs .github/actions/generate-manifest + verify-manifest for a chosen channel (manifest-release / manifest-beta) without building or releasing anything. It needs a plugin-metadata source for the identity/own-build assertion - the natural choice is downloading the newest release's *.zip.meta.json for that channel and passing it explicitly, which is acceptable for a manual recovery tool (the operator names the release).

From the adversarial review of #947 (availability lens; declined in-PR as out of scope).

iderex:
## What & why

Closes #948 (from the #947 availability-lens review): three publish legs run manifest generation inline in the job that already sealed the immutable release, so a transient manifest failure is unrecoverable via re-run, "re-run failed jobs" re-enters `action-gh-release` on the sealed release and never reaches the manifest step.

`regenerate-manifest.yml` is a `workflow_dispatch`-only recovery tool: it runs `generate-manifest` + `verify-manifest` for one channel without building or releasing anything.

- **Identity source:** the operator names a release; its `*.zip.meta.json` (exactly-one asserted) supplies the plugin identity and pins `expect-version` on the verify step. Naming the channel's newest release makes the action's own-build-survives assertion hold; an ancient one fails it by design.
- **No race:** the run joins the channel's publish concurrency group (`publish-manifest-beta` / `publish-manifest-release`), so it can never interleave with a scheduled publish on the manifest branch.
- **#954 guard, fail-closed:** while the stable channel is retired (`manifest-release` deliberately empty), a `release`-channel run fails unless `repopulate-stable=true` is explicitly set, a routine recovery must not silently refill the stable manifest with a history whose newest entry ships a known-vulnerable dependency.
- Token surface: workflow-level `permissions: {}`, one job with `contents: write`; inputs reach the shell via `env`, never interpolated; checkout SHA-pinned with `persist-credentials: false`; strict-parsed with js-yaml.

## Verification

Dispatch-only, CI cannot exercise it; per repo process the adversarial review is the primary verification and runs before merge.

### DIFF
--- a/.github/workflows/regenerate-manifest.yml
+++ b/.github/workflows/regenerate-manifest.yml
@@ -35,6 +35,9 @@ permissions: {}
 # Join the SAME per-channel concurrency group as the publish legs, so a recovery run never
 # races a scheduled publish for the manifest branch (the reader-vs-pusher TOCTOU #135 closed).
 # cancel-in-progress is false: a run holding contents:write must never be killed mid-write.
+# Note: a group holds at most one PENDING run — dispatching a recovery while a nightly leg is
+# queued (around 04:00 UTC) evicts that queued publish (it self-heals at the next nightly, and
+# nothing is written by a canceled pending run). Check the group's run list after dispatching.
 concurrency:
   group: publish-manifest-${{ inputs.channel }}
   cancel-in-progress: false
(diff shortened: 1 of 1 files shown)


iderex:
Review (primary verification, CI cannot exercise a dispatch-only workflow): 2 lenses (correctness/Actions-semantics, availability/integrity) on the full file plus both composite actions and the publish legs. Verified sound: inputs in the workflow-level concurrency group resolve at trigger time to exactly the publish legs' group names; the boolean guard fails closed in every interpretation of dispatch-input semantics; the ternary expressions cannot mis-select; env propagation, the exactly-one metadata assertion, injection surface (inputs via env only) and the permission posture all hold; no misuse case produces a wrong manifest (a beta tag against the release channel fails the generator's own-build-survives assertion before any write). Two findings fixed in the follow-up commit: the guard's predicate now mirrors the branch selection (deny-by-default for any non-beta channel value), and the EXPECT_VERSION env write is shape-checked. The pending-slot eviction semantics of the shared concurrency group are documented in the header.

PR #959, opened 2026-07-22
Bound the checksum-sidecar fetches to the entries that survive the cap

### LINKED ISSUE #949: Bound the manifest generator's per-release sidecar fetches as history grows
generate-manifest fetches ~2 assets per release (md5 sidecar + metadata) serially over the whole paginated history, though at most max-versions-per-abi entries per ABI survive the cap. Listing is already per_page=100; the sidecar fetches remain O(history) and the publish jobs have 15/20-minute timeouts, so the crossing point is years out but real.

Options: derive version/targetAbi cheaply first (the release tag name encodes the version for every current tag shape) and fetch sidecars only for entries that survive the cap; or parallelize the fetch loop. Either must keep the fail-closed semantics scoped to the release being published.

From the adversarial review of #947 (availability lens; declined in-PR as out of scope).

iderex:
## What & why

Closes #949 (from the #947 availability review): the history walk fetched every release's md5 sidecar although at most `max-versions-per-abi` entries per ABI can survive, O(history) serial downloads against 15/20-minute job timeouts.

The walk now only **resolves** the sidecar URL (the pairing-cardinality assertion stays there, list-based, free); the sidecar is fetched and validated in a **second pass over the capped entries**, bounding checksum downloads to keep×ABIs (29 today) however long the history grows. Version/targetAbi still cost one small metadata fetch per release, unavoidable: the cap groups by targetAbi, and any early termination of the walk could drop a rarely-publishing generation entirely (exactly #943's shape).

Fail-closed scoping is unchanged: unusable sidecar → fatal for the release this run published, loud drop for a historical one; a post-cap drop shortens that ABI's list for the run and warns. The own-build-survives assertion now runs **before** any sidecar fetch, so a doomed run fails without downloading anything.

## Verification (empirical, against the live repository)

- Dry run over all 34 releases: 14+15 entries, **exactly 29 checksum fetches** (curl call counter = the capped entries).
- **A/B: the old and new script bodies over the same release fixture produce byte-identical manifests.** (A `\r\n` delta seen against the published manifest was traced to the Windows-fetched fixture itself, `gh.exe` line-ending mangling in the test data, not a behavior change; the A/B isolates the diff to zero.)
- js-yaml strict parse clean.

### DIFF
--- a/.github/actions/generate-manifest/action.yml
+++ b/.github/actions/generate-manifest/action.yml
@@ -184,30 +184,12 @@ runs:
             echo "::error::release $tag has $zip_name and $md5_count candidate checksum sidecars (${zip_name%.zip}.md5 / ${zip_name}.md5); refusing to publish a guessed checksum"
             exit 1
           fi
+          # The sidecar's CONTENT is fetched and validated later, only for the entries that
+          # survive the per-ABI cap (#949): the history walk otherwise costs one sidecar
+          # download per release forever, though at most keep×ABIs entries can be published.
+          # Only the URL is resolved here; the pairing cardinality was asserted above.
           md5_url="$(printf '%s' "$assets" | jq -r --arg a "${zip_name%.zip}.md5" --arg b "${zip_name}.md5" \
             'first(.[] | select(.name == $a or .name == $b)) | .browser_download_url')"
-          # The sidecar is `md5sum` output: `<32 hex><space><space-or-*><filename>`. Validate the
-          # WHOLE line, filename included — not just the leading 32 characters. A sidecar written
-          # with `>>` that ended up holding two lines would otherwise pass on its first line,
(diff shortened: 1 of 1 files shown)


iderex:
Review: availability lens (primary verification beside the empirical A/B, the action only runs in publish jobs). All refutation attempts failed: the second-pass loop cannot silently truncate or misalign (no stdin competition, set -e trips loudly on any jq failure, the empty-final guard backstops a dead process substitution); order is preserved through the cap and -S; the release this run published necessarily carries _ours=yes (asset-name identity), so pass 2 can only remove it FATALLY, never quietly, the survive-assert-before-fetch ordering is sound; the underscore del-list is complete; a persistent historical-sidecar drop degrades that ABI's head observably (warning + still-verified next-newest) while total ABI loss reds via verify-manifest's shrink check. PASS, no findings.

Issue #928, opened 2026-07-21
Test-completeness epic: every function covered holistically (unit or E2E), OIDC AND SAML where both apply


iderex:
Standing goal: **every plugin function is exercised as a whole**, by an `SSO-Auth.Tests` test, by an E2E harness ([#919](https://github.com/iderex/jellyfin-plugin-sso/issues/919)), or by a technically-appropriate alternative, and **wherever a feature exists on both protocols it is tested on OpenID Connect AND SAML**, at the maximal feature surface. This issue tracks the audit-then-fill program; it is not a single PR.

## Method
1. Map every function below to its current coverage (unit / E2E / none) and its protocol parity (OIDC ✔/– , SAML ✔/–/✗).
2. Fill the gaps, preferring the right tool: pure logic + fail-closed branches → `SSO-Auth.Tests` (fast, a negative test per branch); full browser/round-trip behaviour → E2E; out-of-process or host-integration surfaces → the appropriate alternative, documented.
3. Where a feature exists on both protocols, add the missing protocol so parity holds (e.g. a role-mapping test that runs for OIDC must have a SAML sibling).
4. Each filled area ticks its box here and, if it establishes a structural property, locks it in as a conformance/endpoint test.

## Current baseline (not a gap list, the starting point)
Heavy unit coverage already exists (`SSO-Auth.Tests/`: Saml, Http, Oidc, Session, Config, RateLimit, Linking, Secrets, Avatar, Authz, Logout, LoginButtons, Audit, …). E2E covers full login round-trips + role gate + replay for **Keycloak (OIDC + SAML)** and **Authelia (OIDC)** ([#720](https://github.com/iderex/jellyfin-plugin-sso/issues/720)/[#921](https://github.com/iderex/jellyfin-plugin-sso/issues/921)). This epic drives each function to explicit, parity-complete coverage.

## Function inventory (tick when covered holistically, both protocols where applicable)

> **Coverage-audit sweep complete (2026-07-23).** The audit (comment above) mapped every function; the fill units U1–U8 landed (#960 #963-#971 area) and closed every identified gap. 18 of 21 areas are now holistically covered with protocol parity; the 3 remaining are: Quick Connect (separate #740), Step-up/MFA (covered bar the max_age→auth_time gap #961), and an optional E2E Single-Logout leg. Two feature-level findings the audit surfaced are filed as #961 and #962.

[...]

Issue #722, opened 2026-07-19
Auto-manage login-page "Sign in with…" buttons, with documented auto-redirect and hide-password recipe


iderex:
## Problem
After configuring a provider, nothing surfaces an SSO button on the Jellyfin login page. As with the archived 9p4 plugin, admins hand-craft branding HTML or share a raw `…/sso/OID/start/PROVIDER` URL. This is the largest recurring user-support cluster (9p4 #274, #185, #160, #120, #110, #161, #166, #234, Discussion #157) and the V2 reference already automates it. A configured provider that produces no visible button reads as a broken plugin.

## Why it matters for RC
The headline "a monkey can configure it" item and a clean end-to-end functional pass depend on a tester clicking a button on the login screen, not knowing a hidden URL.

## Evidence
- No managed-button feature in main (`grep -rniE 'disclaimer|LoginButton|branding' SSO-Auth` → only a comment).
- V2 `SSO-Auth/Api/LoginButtonInjector.cs`: `BuildBlock` emits Jellyfin-styled submit buttons, each value `WebUtility.HtmlEncode`d; `Merge` splices a marker-delimited region idempotently, removing it when empty (preserving any admin disclaimer). V2 config: `ManageLoginPageButtons` (global), per-provider `HideLoginButton`, `LoginButtonText`.
- Inputs already present in main: anonymous `OID/GetNames`/`SAML/GetNames` (enabled-only), `OID/start`/`SAML/start` routes; server-side `DisablePasswordLogin` (#665) already exists.
- Community hide-password workaround: `#loginPage .manualLoginForm { display:none !important }` + proxy-blocking `/Users/authenticatebyname` (9p4 #160, Discussion #157).

## Acceptance criteria
- [ ] Port `LoginButtonInjector.cs` (pure string logic); opt-in global `ManageLoginPageButtons` **default off** so branding is never mutated without consent; per-provider `HideLoginButton` + `LoginButtonText`.
- [ ] On save, merge the managed block into the branding login disclaimer; on disable, remove only the managed region, preserving surrounding admin content.
- [ ] Unit tests: HTML-encoding of adversarial provider names/labels (login-page XSS is high-severity), idempotent re-apply, empty-block removal.
- [ ] Documented, tested **hide-password-form** recipe cross-referencing `DisablePasswordLogin` (make clear CSS is cosmetic, the server flag is the enforced control, fail-closed framing, no overclaim).
[...]

PR #967, opened 2026-07-22
Test the login-button hosted service and pin the SAML hostile-name route

iderex:
U4 of the #928 epic. `LoginButtonManager`, the IHostedService that splices the managed "Sign in with …" block into the branding login disclaimer (#722), had **no tests at all**; its five contractual behaviors are now pinned: startup sync writes the block, the write-guard skips the branding save when the merge changes nothing, a config change re-syncs via the event hook, `StopAsync` unsubscribes (a later change writes nothing), and a branding-store fault is swallowed on both paths (startup + the event hook inside the config-save path).

NSubstitute cannot proxy `ILogger<internal type>` (DynamicProxy lacks IVT into strong-named Logging.Abstractions) → `NullLogger<T>.Instance`; no production change.

Plus the SAML sibling of the hostile-name encoding pin (the adversarial name was only asserted on the OID start route).

Suite **3918/3918** on both TFMs. Refs #928 (U4 ✔).

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -2522,6 +2522,45 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void EveryHarnessUsingTestClass_IsInTheNonParallelControllerCollection()
+    {
+        // The SsoControllerHarness constructor swaps the process-wide SSOPlugin.Instance and resets the
+        // OIDC/SAML static caches — two harness-based classes running in parallel therefore race each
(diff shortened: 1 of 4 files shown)


iderex:
Second commit: RCA of the intermittent full-suite failure (SamlAuth 429->400). Root cause: my two new harness-using test classes (U2's SSOControllerRateLimitTests, this PR's LoginButtonManagerTests) were missing the existing non-parallel [Collection("SSOController")] convention - the harness swaps SSOPlugin.Instance process-wide, so parallel classes race. Both join the collection, and a new conformance rule makes the convention self-enforcing (harness construction without the attribute = red build naming the file). Suite 3920/3920 twice.

PR #969, opened 2026-07-22
Close the four small unit holes from the coverage audit

iderex:
U6 of the #928 epic, the audit's G6 sweep:

- **OIDC linking challenge end-to-start**: `isLinking=true` through the real `OidChallenge` registers an authorize state whose summary says linking (previously only hand-seeded states carried the flag).
- **`OidLogout` fail-safe catch**: an unrevealable stored id_token (ssoenc envelope, no key) makes the end-session build throw, the local logout stands, LOCAL redirect, never a 500.
- **`OID/Get` / `SAML/Get` MVC-serialized redaction**: the camelCase-emitted JSON never carries the OIDC client secret or SAML signing key (the redaction lives in converter attributes, so the wire shape is what matters); liveness asserts keep it non-vacuous.
- **`SamlTest` controller-level negative**: unparsable certificate → failed-but-OK result naming the parse problem, no 500, no key.

Suite **3928/3928** on both TFMs. Refs #928 (U6 ✔).

### DIFF
--- a/SSO-Auth.Tests/Http/SSOControllerLogoutTests.cs
+++ b/SSO-Auth.Tests/Http/SSOControllerLogoutTests.cs
@@ -78,4 +78,34 @@ public class SSOControllerLogoutTests
         Assert.IsType<LocalRedirectResult>(result);
         await harness.SessionManager.Received(1).Logout("caller-token");
     }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


iderex:
Correction to the commit message: the suite count is 3928/3928 (4 new tests x 2 TFMs), not 3930 - the message overstates by two. The PR body is corrected; the code is unchanged.

PR #981, opened 2026-07-23
Publish the newest JF11 beta as the repo's Latest release

### LINKED ISSUE #980: Publish the newest JF11 (10.11) beta as the repo's Latest release automatically
In the beta-only phase (#954) the newest **JF11 (10.11) beta** should always be the repo's GitHub **Latest release**, the headline build a visitor lands on.

## Current state
`publish-beta.yml` publishes the JF11 beta with `prerelease: true` (unchanged since 2026-07-18) and its publish step is `gh api … -X PATCH -F draft=false`, it never marks the release latest. GitHub cannot mark a prerelease as "Latest", so as-published a JF11 beta is never the Latest release.

`4.3.0-beta.11` and `4.3.0-beta.12` are currently `prerelease=false` and `4.3.0-beta.12` is the Latest release, but only because they were flipped **manually**. The next nightly beta (beta.13) would publish as `prerelease=true` again and drop off "Latest". This needs to be automatic.

## Change
In `publish-beta.yml` (JF11 leg only):
- `prerelease: false` on the draft (so the build is eligible to be Latest), and
- `-F make_latest=true` on the publish PATCH (so the newest JF11 beta is always the Latest release).

## Safety
- The `prerelease: true` was there to keep betas out of the **stable** manifest (`ignorePrereleases: true`). That is moot now: #954 retired the stable channel (`manifest-release` is empty, `publish.yml` dormant since 2026-07-18), and the **beta** manifest generates with `include-prereleases: true`, so a non-prerelease build is still surfaced there (confirmed: beta.11/.12 are in `manifest-beta`).
- The **JF12** leg (`publish-jf12-beta.yml`) stays `prerelease: true` with no `make_latest`, so it never competes for Latest.

Release-pipeline change → CI cannot exercise it, so the adversarial review is the primary verification.

iderex:
## What & why
The newest **JF11 (10.11) beta** should always be the repo's GitHub **Latest release** during the beta-only phase (#954), the headline build a visitor lands on. `publish-beta.yml` published the JF11 beta with `prerelease: true` and never marked it latest, so it could never be the Latest release as published; `4.3.0-beta.11/.12` are currently latest only because they were flipped **manually**, and the next nightly (`beta.13`) would regress to a prerelease.

## Change (JF11 leg only)
- `prerelease: false` on the draft, so the build is eligible to be Latest.
- Seal step marks it latest: `gh api … -X PATCH -F draft=false -f make_latest=true`. `make_latest` is a REST **string enum** (`true`/`false`/`legacy`), so it is sent as a raw string (`-f`), not a typed boolean; `draft` stays `-F` (a real boolean).

## Safety
- `prerelease: true` previously kept betas out of the **stable** manifest (`ignorePrereleases: true`). Moot now: #954 retired the stable channel, `manifest-release` is empty and `publish.yml` has been dormant since 2026-07-18, so nothing filters betas on `prerelease`, and the **beta** manifest uses `include-prereleases: true`, so a non-prerelease build is still surfaced (verified: beta.11/.12 are in `manifest-beta`).
- The **JF12** leg (`publish-jf12-beta.yml`) stays `prerelease: true` with no `make_latest`, so a JF12 (5.0.0) build, higher semver, can never take Latest.
- `make_latest=true` explicitly forces Latest onto the JF11 beta, robust even if some other release were ever non-prerelease.

## Type of change
- [x] CI / release pipeline

## Security
- Release-pipeline change (directive 1). CI cannot exercise the release-only workflow, so the adversarial review is the primary verification, results in Notes. No secrets/permissions change; the publish step already had `contents: write` for sealing the release.

## Verification
- YAML-only; no C# change. The live state is already correct (`4.3.0-beta.12` is Latest); this makes it automatic for beta.13+.

## Notes
Adversarial review appended before merge. Caveat for when the stable channel is un-retired: `publish.yml` (stable) should then own Latest and this beta `make_latest` be reconsidered, noted for that milestone.

Closes #980

### DIFF
--- a/.github/actions/generate-manifest/action.yml
+++ b/.github/actions/generate-manifest/action.yml
@@ -133,6 +133,17 @@ runs:
           [ "$is_draft" = "true" ] && continue
           if [ "$is_pre" = "true" ] && [ "$INCLUDE_PRERELEASES" != "true" ]; then continue; fi
 
+          # Stable channel (include-prereleases != true): exclude beta-tagged releases by GLOB, not
+          # only by the prerelease flag. Since #980 the newest JF11 beta is published as a
+          # NON-prerelease (so GitHub can mark it the Latest release), so the prerelease flag alone
+          # no longer keeps betas out of manifest-release. The tag convention is the reliable
+          # separator — a beta is always tagged X.Y.Z[-JF12]-beta.<run>, a stable never is — so this
+          # fails closed: a revived stable channel can never pull an immutable non-prerelease beta
+          # into the stable manifest, where its four-octet version would even outrank the stable build.
+          if [ "$INCLUDE_PRERELEASES" != "true" ]; then
+            case "$tag" in *-beta.*) continue ;; esac
+          fi
+
           # `first(…)`: a draft can carry the same tag_name as a published release (the draft
           # flow the beta leg used to run does exactly that), and without the draft filter plus
           # this cutoff the lookup emits TWO JSON documents — every jq below then produces two
--- a/.github/workflows/publish-beta.yml
+++ b/.github/workflows/publish-beta.yml
@@ -196,12 +196,15 @@ jobs:
           # published release cannot be updated in place, so a rolling tag would fail
           # on its second run. The tag never matches publish.yml's
           # [0-9]+.[0-9]+.[0-9]+-stable glob, so a beta never triggers the stable
-          # channel; and prerelease: true keeps it out of the stable manifest
-          # (ignorePrereleases: true) — only the beta manifest below surfaces it. The
-          # INSTALLED plugin version is X.Y.Z.<run> (the hidden fourth octet keeps
-          # Jellyfin's update ordering monotonic across betas).
+          # channel. prerelease is FALSE so the newest JF11 (10.11) beta can be GitHub's
+          # "Latest release" (see the publish step's make_latest) — the headline build a
+          # visitor lands on. Safe in the beta-only phase: #954 retired the stable channel
+          # (manifest-release empty, publish.yml dormant) so nothing filters betas on this
+          # flag, and the beta manifest below uses include-prereleases so a non-prerelease
+          # build is still surfaced. The INSTALLED plugin version is X.Y.Z.<run> (the hidden
+          # fourth octet keeps Jellyfin's update ordering monotonic across betas).
           draft: true
-          prerelease: true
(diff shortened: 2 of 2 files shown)


iderex:
Pre-merge review, release-pipeline / correctness lens (release-only code CI cannot exercise, so the review is the primary verification). Two rounds.

**Round 1, the core change (prerelease:false + make_latest):** PASS on all points except one finding. Confirmed: `make_latest` is a REST string enum, so `-f make_latest=true` (raw string) is the correct form, `-F` (typed boolean) would be the silent no-op the goal warns about; `draft` correctly stays `-F`. `prerelease:false` is required (GitHub cannot mark a prerelease Latest). The JF12 leg stays `prerelease:true` with no `make_latest`, and both legs serialize on the shared `publish-manifest-beta` concurrency group, so a 5.0.0 JF12 build can never take Latest. No install/manifest coupling, Jellyfin installs from the manifest JSON, not the GitHub "Latest" pointer.

**Finding (NEEDS_IMPROVEMENT → fixed):** publishing the beta as `prerelease:false` removed the ONLY filter keeping betas out of the STABLE manifest. `publish.yml` (stable) → `generate-manifest` with `include-prereleases:false` filtered only on draft + the prerelease flag, no tag glob, so a revived stable channel would pull the now-immutable non-prerelease betas into `manifest-release`, where a beta four-octet version even outranks the stable build. A silent, fail-open landmine.

**Fix:** added a fail-closed tag-glob exclusion in the shared `generate-manifest` action, on the stable channel, `case "$tag" in *-beta.*) continue`. The tag convention is a reliable channel separator independent of the prerelease flag, so `manifest-release` is now structurally beta-free.

**Round 2, re-review of the guard:** PASS. The glob excludes both JF11 (`X.Y.Z-beta.<run>`) and JF12 (`X.Y.Z-JF12-beta.<run>`) betas from the stable channel and keeps `-stable` tags; the beta channel (`include-prereleases:true`) skips the guard entirely, so #980`s recovery intent holds (betas still populate `manifest-beta`). Edge cases checked: the one legacy prefix-form `beta-*` tag is `prerelease:true`, caught by the flag filter (the two filters are complementary). Placement correct under `set -euo pipefail`; the run-own-build survival assertion is unaffected (a stable run carries a `-stable` tag).

Gates: YAML-only, no C# change. Ready to merge.

PR #1000, opened 2026-07-24
Localize the login rejection messages on the browser error page [#913]

iderex:
# What & why

Fourth sub-unit of the i18n epic (#913): the last user-facing text on the
anonymous surfaces, the login-rejection body itself ("No matching provider
found", "Invalid or expired state", the uniform denial, and the rest). This
localizes those on the browser error page while leaving the wire form exactly as
it was.

The plain-text body `LoginStatusMapper` emits is untouched: it stays canonical
English, byte-identical for the XHR/API leg, and deliberately uniform per
category (no provider/roles enumeration oracle). Only the browser re-render - 
which already re-styles that body into an HTML page for a top-level navigation - 
now localizes it, by reverse-mapping the English message to its catalog key and
resolving that key in the request's culture. A message that is not a catalog
value (an identity-provider-interpolated error) passes through unchanged, and the
result is still HTML-encoded, so the existing anti-injection property holds.

Reverse-lookup rather than threading a key through every call site: the same
rejection wording is produced both by `LoginStatusMapper` and by a few direct
controller results, so keying the value once - at the point it is rendered - 
localizes every path uniformly without touching the mapper or the wire contract.

Refs #913

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

Touches the auth error surface (`LoginStatusMapper` bodies re-rendered by
`BrowserErrorPage`); no change to SAML/OIDC validation, crypto, config
persistence, or the release pipeline. Reviewed with the adversarial bypass +
correctness lenses in addition to CI.

- [x] Wire-contract preserved: `LoginStatusMapper` is unchanged; the plain-text
      body is byte-identical English, and `BrowserErrorPage.Wrap` rewrites only
      for browser (`text/html`) requests, so the XHR/API leg is untouched.
- [x] No enumeration oracle introduced: the rejection bodies remain uniform per
      category; localization maps the English value to a key deterministically,
      so the unknown- vs disabled-provider cases and the uniform denial stay
      indistinguishable.
- [x] Injection-safe: a non-catalog (IdP-interpolated) message passes through
      `LocalizeEnglish` unchanged and is still `HtmlEncoder`-encoded before
      rendering; the existing break-out test still passes.
- [x] Fail-closed: the reverse lookup returns the input unchanged when no key
      matches, and `GetString` falls back culture → base → English → key, never a
      blank.

## Quality checklist

- [x] No duplicated logic; the reverse lookup is one small unit, and keying the
      value once localizes every rejection path.
- [x] The change adds no more code than the problem requires (a linear scan of
      the small English catalog, run only on an error re-render).
- [x] Architecture-conformance tests pass; no new module edge (the work stays in
      `Localization` and `Shared`, an already-allowed edge).

## Verification

- [x] `dotnet test` is green - 4150 passed, 0 failed (net9.0 and net10.0),
      including the reverse-lookup unit tests and two new conformance guards:
      the English baseline has no duplicate values (unambiguous reverse map), and
      every message `LoginStatusMapper` can emit is a catalog value (no message
      can silently stay English on a translated page).
- [ ] N/A, no `.js` / `.html` / `.css` / `.md` change.

## Notes

Remaining #913 sub-units: the config/linking admin UI, the full language catalog
set (the drift-guard already enforces that each future non-English catalog
carries exactly the English key set), and the translator docs.


iderex:
Adversarial-review gate (bypass + correctness lenses, full touched files + callers):

- **bypass-lens: PASS**, no enumeration oracle, no wire-contract change, no injection vector, fail-closed on every missing-key/culture/catalog path (ASVS V3 + V7 hold). The plain-text body stays byte-identical English; only the browser HTML re-render localizes, still HTML-encoded.
- **correctness-lens: NEEDS_IMPROVEMENT → FIXED**, the OpenID callback's pre-mint state rejection (`OidcLoginService`) returned the bare literal `"Invalid or expired state"`, a duplicate of the mapper's InvalidState body that localized only by coincidence and was unguarded. Fixed in d6cb62e by extracting a shared `LoginStatusMapper.InvalidStateMessage` constant both reference; the wire shape is unchanged and the constant now flows through the `EveryRejectionBody` coverage guard, so it cannot drift out of the catalog.

Non-blocking, out of scope for this PR (noted for the remaining i18n sweep): the generic OIDC callback body `"Error logging in."` is not a `LoginStatusMapper`/`PublicReason` body and is not a catalog value, so it stays English on a translated page, it fails safe (English pass-through, encoded).

PR #1001, opened 2026-07-24
Add client-side UI localization and wire the linking page [#913]

iderex:
# What & why

Fifth sub-unit of the i18n epic (#913): the client-side localization mechanism
the admin/linking pages need, wired into the account-linking page as its first
consumer.

Those pages are client-rendered from embedded resources, the config page is
served by Jellyfin core, so there is no server render hook. So the server owns
only the culture fallback: a new anonymous `SSOViews/i18n` endpoint resolves the
caller's `Accept-Language` (the same resolver the served pages use) and returns
the complete key→value catalog resolved into that culture. A small `i18n.js`
module fetches it once, applies it to `data-i18n`-marked elements, and exposes
`t(key, params)` for strings built in JavaScript.

Refs #913

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

Adds a new anonymous read endpoint and a client-side applier; no change to
SAML/OIDC validation, crypto, config persistence, or the release pipeline.
Reviewed with the adversarial bypass + correctness lenses in addition to CI.

- [x] No sensitive exposure: the endpoint returns first-party static UI labels
      only — no user data, no configuration, no secrets, nothing per-user — the
      same strings already embedded in the pages, so anonymous access is safe.
- [x] Bounded / no amplification: it serializes an in-memory dictionary, does no
      I/O and is off the login path, so it is classified rate-limit-exempt in the
      route conformance test (alongside the existing static-asset endpoint).
- [x] No client injection: `i18n.js` assigns `textContent` only (never
      `innerHTML`/`eval`); catalog values and `data-i18n` keys are first-party;
      the only `t()` params on the linking page are the hardcoded protocol mode.
- [x] Fail-safe: `loadCatalog` swallows any error to an empty catalog, so the
      static markup keeps its built-in English on a network failure.

## Quality checklist

- [x] No duplicated logic; the culture fallback stays server-side (reusing the
      existing resolver), the client just applies concrete strings.
- [x] Architecture-conformance tests pass; the new structural properties are
      locked in — the `i18n` route is classified rate-limit-exempt, and
      `Http → Localization` is recorded as a new allowed module edge; the
      GetViews manifest pin is updated for the new asset.

## Verification

- [x] `dotnet test` is green - 4152 passed, 0 failed (net9.0 and net10.0),
      including the new endpoint test (resolved strings + `Vary: Accept-Language`)
      and the updated route/manifest conformance pins.
- [x] Prettier is clean for the changed `.js` / `.html` (`i18n.js`, `linking.js`,
      `linking.html`).

## Notes

Scope: the linking page's static labels + its two dynamic strings. Two rich
instruction paragraphs that interleave text with icons/links are left for a
later sub-unit with placeholder-aware substitution. Remaining #913 sub-units: the
config page's own string inventory (uses this same mechanism), the full language
catalog set (the drift-guard already enforces exact key parity), the rich-markup
paragraphs, the translator docs, and the noted served-error-bodies sweep.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -188,7 +188,7 @@ public class ArchitectureConformanceTests
     [InlineData("Session", "Authz", "Avatar", "Linking")] // session mint + login outcomes — applies grants (Authz), sets avatars (Avatar), reconciles links (Linking)
     [InlineData("Shared", "Avatar", "Linking", "Localization", "RateLimit", "Routing", "Session")] // shared served-page / flow-response + rate-limit-gate helpers — depend downward on the session/linking/avatar/throttle/route/localization tiers, never on a protocol or the boundary
     [InlineData("Flows", "Audit", "Identity", "Linking", "Localization", "Logout", "Net", "Oidc", "Provider", "RateLimit", "Saml", "Session", "Shared")] // per-protocol login orchestration — drives both protocol modules (Oidc/Saml) and the downstream mint/link/session tiers; localizes the served auth-completion page (Localization, #913); persists the captured logout state at the mint (Logout, #727); nothing above the boundary imports it
-    [InlineData("Http", "Audit", "Avatar", "Flows", "Linking", "Logout", "Net", "Oidc", "Provider", "Saml", "Session", "Shared")] // the web boundary (SSOController + request helpers + the admin test-connection probe): the composition top of the DAG — it fronts every flow, so its import list is deliberately wide (incl. the RP-initiated logout store, #727); nothing imports it back (#790/#807)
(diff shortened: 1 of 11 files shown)


iderex:
Adversarial-review gate (bypass + correctness lenses, full touched files + callers):

- **bypass-lens: PASS**, the new anonymous `SSOViews/i18n` endpoint exposes only static first-party UI labels (no user data, config, secrets, or enumeration); the work is bounded and in-memory, so the rate-limit exemption is defensible and is itself fail-closed via the route-completeness conformance test; the client applier uses `textContent` exclusively with prototype-pollution-safe placeholder substitution and only non-attacker-controlled params at its call sites; the literal `i18n` route cannot collide with or weaken the `{viewName}` asset route. ASVS V3 + V7 hold.
- **correctness-lens: NEEDS_IMPROVEMENT → FIXED** - keys, placeholders, resolver wiring and init ordering all verified correct, but the "degrades to built-in English" guarantee held only for the static `data-i18n` markup: `t()` fell back to the *key*, so an empty catalog rendered raw keys for the two JS-built strings, and the disabled-provider note is a routine, non-error UI state (a user holding a link to a since-disabled provider would have seen `Providerlink.disabled_note`). Fixed in 40113b4 by adding an optional English default to `t(key, params, fallback)`, mirroring the role the hard-coded text on `data-i18n` elements already plays, and passing it at both call sites, so the degradation guarantee now holds uniformly.

Known, presently-unclosable test gap (documented, not blocking): with English as the only loaded catalog, the endpoint test cannot distinguish a culture-ignoring mutant in `ResolvedCatalog`, that assertion becomes writable once a second catalog lands with the language set.

PR #1002, opened 2026-07-24
Localize the configuration page chrome [#913]

iderex:
# What & why

Sixth sub-unit of the i18n epic (#913): wires the admin configuration page, the
last unlocalized surface, into the client-side i18n mechanism, and localizes its
page-level chrome. The provider editors' own field labels and descriptions follow
as their own sub-units; this lands the wiring plus the shared workspace strings
for both protocols.

Jellyfin core serves this page from its own URL base, so a relative import would
not resolve to the plugin's assets. The page therefore loads the shared applier
from its absolute `SSOViews` URL, the same module the linking page uses, rather
than duplicating it. Every failure path is swallowed, so the markup keeps its
built-in English if the module or the catalog cannot be fetched.

Two supporting pieces:

- The applier gains **attribute support**: `data-i18n-<attr>="key"` localizes an
  attribute, not just text content. The attributes are an explicit allowlist
  (`title`, `placeholder`, `aria-label`) rather than a generic setter, so a markup
  typo or a later edit cannot drive `href`, `src`, or an event handler through the
  same path. The configuration page needs it for the collapse-section titles.
- `config.js` gains `tr(key, englishDefault, params)` for strings it sets itself.
  The English default is the same wording the static markup carries, so a JS-set
  string and its HTML twin cannot drift apart, the failure mode caught in the
  previous sub-unit's review.

Refs #913

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

Touches the admin configuration page's presentation only; no change to
SAML/OIDC validation, crypto, config persistence, or the release pipeline.
Reviewed with the adversarial bypass + correctness lenses in addition to CI.

- [x] The localization path touches only text content and the three allowlisted
      inert attributes — never element ids, marker classes, or input values, so
      the save contract (`config.js` serializes by id + `sso-*` marker class) is
      untouched.
- [x] The attribute allowlist is explicit, so no markup can drive `href`, `src`,
      `style`, or an event handler through the applier.
- [x] Nothing in the path reads, logs, or transmits configuration values; the
      catalog endpoint remains first-party static UI labels only.
- [x] Fail-safe: a failed import or catalog fetch leaves the page fully
      functional with its built-in English; `tr()` returns the English default
      until (and if) the module resolves.

## Quality checklist

- [x] No duplicated logic: the configuration page reuses the same applier module
      as the linking page rather than carrying its own copy.
      locked in — **every i18n key referenced by an embedded web asset must exist
      in the English catalog**. A key with no entry degrades silently (the element
      keeps its built-in English), so the drift is invisible in review and in the
      browser; now it is a red build. Verified empirically by renaming a catalog
      key, which fails the test naming that exact key from both its markup and its
      script reference.

## Verification

- [x] `dotnet test` is green - 4154 passed, 0 failed (net9.0 and net10.0),
      including the new key-parity conformance guard.
- [x] Prettier is clean for the changed `.js` / `.html`.

## Notes

Scope: the wiring plus eleven `config.*` chrome keys (page title, About collapse,
SSO-only summary, and the OIDC/SAML workspace headers, add buttons, empty states
and editor titles). The provider editors' field labels and descriptions, and the
rich prose blocks that interleave text with links, are following sub-units, the
new conformance guard keeps every key added from here on honest. Remaining #913
work after those: the full language catalog set, the translator docs, and the
noted served-error-bodies sweep.


iderex:
Adversarial-review gate (bypass + correctness lenses, full touched files + callers). Both lenses independently landed on the same two defects; all findings are FIXED in 5dc2161 except two reasoned declines.

**FIXED**
- **Editor-heading race (bypass F1 / correctness F2, Medium).** `#sso-editor-title` / `#saml-editor-title` carried `data-i18n` while config.js owns them (it writes the loaded provider's name). A late `applyTo()` could relabel an editor holding `keycloak-prod` as "New provider", and the Save path targets the loaded provider without a name-bearing confirmation, unlike Delete. The markers had no upside either: both editors start `hidden` and their heading is always JS-set before reveal. Removed, with the rationale recorded at the element.
- **Inert collapse title (correctness F1 / bypass F4, Medium).** `emby-collapse` copies `title` into generated markup at upgrade time, long before the catalog arrives, so setting it later is very likely a no-op, and it was the sole justification for the attribute mechanism. Marker removed; attribute support now has verifiable consumers instead (the secret/key `placeholder`s, native attributes the browser re-reads live).
- **Synchronous throw escaped containment (bypass F2, Low).** `ApiClient.getUrl` throws synchronously while the argument is evaluated, so the `.catch` could never see it, and `localize()` runs before init wires Save/Delete/Test, so a throw would have left a rendered but inert admin page. Now wrapped; the "every failure path is swallowed" comment is finally true.
- **Guard gaps (bypass F3 / correctness F4+F5+F6, Medium).** The previous guard checked only that a key EXISTS. Three assertions added, each verified by killing the exact mutant the lenses proved survived: (1) the attribute allowlist is parsed out of i18n.js and pinned to the inert three, and every `data-i18n-<attr>` marker must name one, adding `href` reds the build; (2) a marked element's built-in English and a script's inline default must equal the catalog value, otherwise an admin whose catalog loaded and one whose fetch failed see different wording; (3) a UI key no asset references is dead data and hides a dropped marker, so the reverse direction is pinned too.
- **Doc accuracy (correctness F9, Low).** The endpoint's justification now states plainly that it serves the whole catalog including the admin page's labels, and why that is deliberate.

**Reasoned declines**
- **Localize-before-render ordering (correctness F3).** The linking page awaits the catalog before rendering; this page does not. With the JS-owned markers gone the clobber hazard is eliminated, and what remains is a brief English→localized flash of static chrome. Reordering would put two extra serialized round-trips ahead of the provider list on the heaviest page in the plugin. Keeping the parallel load; the residual is cosmetic and self-correcting.
- **Silent `.catch(() => {})` (correctness F8).** Consistent with the linking page, and a `console.warn` on every offline admin load is noise. The conformance guards above are the intended detector for this defect class, and they now fail the build rather than the browser.

Not addressed here, deliberately out of scope (tracked for the following sub-units): the provider editors' field labels/descriptions and the rich prose blocks remain English.

Issue #1029, opened 2026-07-27
Constrain the JWT kid header to a strict character allowlist before key lookup


iderex:
## Problem

#1004 asks for a `kid` character allowlist in the JWKS lookup path alongside the negative-test battery. The battery lands as test-only work (plan on #1004); the allowlist is a production change on the login path and belongs in its own reviewed delivery, so it moves here rather than riding a proof PR.

Current, verified posture: the token-header `kid` is only ever compared ordinally against the `KeyId` values of the in-memory signing keys converted from the discovery JWKS (`OidcSignatureKeys` sets `KeyId = webKey.Kid`). It reaches no filesystem, database, URL, or parser, and a hostile `kid` cannot vouch for a signature, the forgery tests from #1004 pin that a token signed by an unadvertised key is rejected regardless of what its `kid` claims, and that hostile `kid` values (`../`, separators, a null byte, quotes, over-length) produce a clean rejection, never a throw. There is no live sink today.

## Why still do it

Defence-in-depth against the standing JWT `kid`-injection class: the standing mitigation in practitioner guidance is to constrain `kid` to a strict character set before it is used as a lookup key at all, so any future consumer of the header value (logging, caching, a keyed store, a remote key fetch) is born behind the constraint instead of re-opening the class.

## Scope

- Constrain the accepted `kid` to `[A-Za-z0-9._~-]` with a sane length cap, rejected before any key lookup runs, on the shared `OidcSignatureKeys` basis so the id_token and logout_token paths cannot drift.
- Decide and document the JWKS side: whether an advertised key whose `kid` falls outside the allowlist is skipped like malformed material (consistent with the skip-on-malformed contract) or the whole set is refused.
- Negative tests per character class (path separators, dot-segments, percent-encoded dot-segments, null byte, quotes, over-length), plus a compatibility check that common IdP `kid` shapes (base64url thumbprints, UUIDs) all pass.
- Keep the #1004 forgery tests green throughout, they pin the fail-closed floor this change must not lower.

## Acceptance criteria

- [ ] A `kid` outside the allowlist is rejected before key lookup on both token paths, with a test per character class.
- [ ] Common real-world `kid` shapes still validate (no IdP lockout).
- [ ] The behaviour on out-of-allowlist JWKS-side `kid` values is decided, documented, and tested.
- [ ] Full suite green on net9.0 and net10.0 with `--warnaserror`.

iderex:
[...]

PR #1030, opened 2026-07-27
Prove the JWT algorithm, kid and claim posture with forgeries [#1004]

### LINKED ISSUE #1004: Prove the JWT algorithm, kid and claim validation posture with negative tests
## Problem

`SSO-Auth/Api/Oidc/OidcSignatureKeys.cs` builds one hardened `TokenValidationParameters` for both the id_token and the back-channel logout_token, with an asymmetric-only `ValidAlgorithms` allowlist, `RequireSignedTokens`, and an issuer/audience match. That is the right shape, but nothing locks it in and nothing proves it empirically:

- No conformance test forbids a second `new TokenValidationParameters` anywhere in the plugin, so a future code path could construct one without `ValidAlgorithms` and silently accept whatever the token header claims.
- There are no negative tests for the classic forgery shapes, `alg: none`, an RS256 key material replayed as an HS256 MAC, a symmetric algorithm smuggled in, an unsigned JWS.
- `kid` is taken from the token header and matched against the JWKS with no allowlist or character restriction; there is no test that a `kid` containing `../`, a path separator, a null byte, or SQL/quote characters is rejected before it is used as a lookup key.
- The nonce, `aud`/`azp` and `iss` checks on the callback are spread across `OidcIdTokenValidator.cs`, `AuthorizeStateBinding.cs` and `OidcResponseIssuer.cs`, and `SSO-Auth.Tests/Oidc/OidcIdTokenValidatorTests.cs` carries no test method asserting replay of a token whose nonce was never checked.
- Nothing asserts the id_token is never used as a session or API token, the local session mint is the only allowed outcome.

## Why now

id_token validation gaps are still producing advisories: CVE-2026-42206 (roadiz, https://github.com/roadiz/core-bundle-dev-app/security/advisories/GHSA-3gx8-q682-38mx, 2026-04-22) is a nonce generated on the authorization request and never validated on the callback, which is straight ID-token replay. OpenID Connect Core 1.0 with errata set 2 (https://openid.net/specs/openid-connect-core-1_0.html) requires verifying `nonce` when present and checking `aud`/`azp`. RFC 9700, Best Current Practice for OAuth 2.0 Security (https://datatracker.ietf.org/doc/rfc9700/, January 2025), makes PKCE, exact-string redirect-URI matching, and `iss` mix-up defence per RFC 9207 the baseline. Algorithm confusion (`alg:none`, RS256 to HS256, `kid` injection, JKU injection) remains live in 2026 practitioner reporting, and the standing mitigation is unchanged: pass an explicit algorithm allowlist to every verification, never let the library read `alg` from the header, and constrain `kid` to a strict character set.

## Scope

- Add `OidcTokenValidation_UsesTheSingleHardenedParameterBuilder` to `SSO-Auth.Tests/ArchitectureConformanceTests.cs`: fail if `new TokenValidationParameters` appears anywhere outside `SSO-Auth/Api/Oidc/OidcSignatureKeys.cs`, and fail if `AllowedSignatureAlgorithms` contains any `HS*` entry or the literal `none`.
- Add `OidcSignatureKeys_NeverAcceptsSymmetricKeyMaterial`: assert `Convert` skips symmetric JWKS entries (`kty: oct`) so an IdP-advertised HMAC key can never become an issuer signing key.
- New `SSO-Auth.Tests/Oidc/OidcTokenForgeryTests.cs` with one test per forgery, each asserting rejection and that no principal is produced:
  - `ValidateIdToken_AlgNoneWithEmptySignature_IsRejected`
  - `ValidateIdToken_AlgNoneCaseVariants_IsRejected` (`None`, `NONE`, `nOnE`)
  - `ValidateIdToken_Rs256PublicKeyReplayedAsHs256Mac_IsRejected`
  - `ValidateIdToken_UnsignedJwsDetachedSignature_IsRejected`
  - `ValidateLogoutToken_SameForgeries_AreRejected` (the back-channel path shares the builder and must share the tests)
- Add a `kid` allowlist in the JWKS lookup path and `OidcSignatureKeys_RejectsMalformedKid` covering `../`, `..\`, `/`, `\`, `%2e%2e`, a null byte, `'`, `"`, and an over-length value; a `kid` outside `[A-Za-z0-9._~-]` is rejected before any key lookup runs.
- Extend `SSO-Auth.Tests/Oidc/OidcIdTokenValidatorTests.cs` with claim-validation negatives: missing nonce, echoed-but-unbound nonce, nonce from a different authorization request, wrong `aud`, `aud` array not containing the client id, `azp` mismatch when multiple audiences are present, wrong `iss`, and missing `iss` on the authorization response (RFC 9207 mix-up defence).
- Add `IdToken_IsNeverUsedAsASessionToken`, a conformance test asserting the raw id_token string never reaches the session mint or any persisted token field, and that the only credential returned to the client is the locally minted Jellyfin session.
- Add a redirect-URI exact-match test to `SSO-Auth.Tests/Oidc/OidcRedirectUriBuilderTests.cs` (no prefix, suffix, subdomain, trailing-slash, or case-insensitive match is accepted).
- Tag each new security test with its authority (`[Trait("Spec","RFC 9700")]`, `[Trait("Spec","OIDC-Core-1.0")]`) so the assertion cites an external document rather than observed behaviour.

## Acceptance criteria

- [ ] `new TokenValidationParameters` exists in exactly one file and a conformance test enforces it.
- [ ] Every forgery listed above has a named test asserting rejection; each is shown to fail on a deliberately weakened build (documented in the PR body).
- [ ] A `kid` containing a path separator or any character outside the allowlist is rejected before key lookup, with a test per character class.
- [ ] Nonce, `aud`/`azp` and `iss` negatives all exist for both the id_token and the logout_token path.
- [ ] A test proves the id_token is never returned to the client or stored as a session credential.
- [ ] Redirect-URI matching is exact-string, with negative tests for prefix/suffix/subdomain/case variants.
- [ ] Full suite green on `net9.0` and `net10.0`; test project built with `--warnaserror` before push.

## Notes

- The plugin verifies with the framework/Duende handler rather than hand-rolled crypto; the work here is constraining and proving that configuration, not replacing it.
- DPoP and mTLS sender-constrained tokens stay out of scope: as a relying party the plugin cannot enforce them end to end.
- Post-quantum signature algorithms are out of scope, what the plugin accepts is driven by what IdPs sign with.
- Companion work: the SAML XML parser-consistency battery (#1003).

iderex:
Closes #1004 (partially, the `kid` allowlist bullet is tracked separately as #1029).

It pins the OpenID JWT validation posture the plugin already has: one hardened `TokenValidationParameters` basis for both provider tokens, an asymmetric-only algorithm allowlist, `kid` treated as an unauthenticated hint, and the id_token never leaving as a credential.

**No longer strictly test-only.** Two production files changed, by 21 lines (14 added, 7 removed), all of them class doc comments, see "The claim I had to weaken in production" below, and round 4's item 4. No behaviour changed anywhere.

## What round 4 changed about this PR

Round 4 returned NEEDS_IMPROVEMENT with four measurably false claims, all of them in prose and none in shipped behaviour: the guards are live, three probes confirmed them load-bearing, 4,412 tests pass on both TFMs. That is now the fourth round in a row whose findings are about the written map rather than the territory, and the third of them about the same paragraph.

So the rule changes, not the paragraph. **A guard's doc comment states what it CHECKS. It does not enumerate what it misses.** A gap list claims completeness over the space of evasions, the same unbounded-enumeration defect that makes a text scanner lose to a reviewer, run in reverse. It was wrong with two entries, rewritten, and wrong again with four; the next revision would be wrong at five, not through carelessness but because the claim's shape cannot be true. The full reasoning and the four measured gaps are #1050, where they are dated, attributable to the round that found them, and closable.

**1. The lettered enumeration is gone**, from the conformance rule and from `OidcSignatureKeys` alike, replaced in each file by one sentence in that file's own voice: a source scan cannot follow a value after it is returned, nor see a verifier written without the token library; review owns the rest, and what has actually been measured to slip past is on #1050. What stays is what each guard checks, and the three positive claims were re-established by execution this round (M6–M8), not carried over on trust.

**2. The CA5404 sentence went with it, because it was false as written.** It licensed leaving audience/issuer/lifetime disablement untested, on the claim that the analyzer "fails the build wherever it is written". Measured (M1–M4): CA5404 fires on a **literal** false, the post-build `parameters.ValidateAudience = false;` at the back-channel call site is three CA5404 errors, so the round-4 counter-example does not reproduce in that form, but the same property assigned a non-constant false builds clean with zero warnings. The universal quantifier was the false part, and no weaker version of the sentence is worth shipping.

**3. The parallelism rationale is corrected to the mechanism that exists.** The threat model said xUnit runs an un-attributed class in parallel with the serialized collection. It does not. Reproduced here on the identical stack (M5): two classes outside the collection overlap each other on every run, two inside it never overlap, and the collection never overlapped the outside pair. The rule was right for the wrong reason, the race it closes is un-attributed against un-attributed, and a rationale a future author will trust has to be the real one.

**4. Three further claims narrowed.** The `logout_token` validator still carried verbatim the "no second, laxer verification path" absolute that round 3 weakened one file over, on the anonymous endpoint, where it is least defensible; it takes caller-supplied parameters, so what holds is that its one production caller builds them from the shared basis. The battery's per-shape login↔logout parity claim now names its one exception, the allowlist row (I declined this in round 3; three lenses independently disagreed and they are right, the exemption I wrote covers spelling fan-outs, and that row is a shape). And the construction predicate no longer claims both spellings C# offers: it reads one line, and a construction wrapped across two is neither.

### Measurements run this round

Every sentence rewritten below was measured first. All mutations ran in a **throwaway clone of the branch**, never in the work tree, so `git status --porcelain` was empty throughout and there is no revert to trust.

| #   | Measurement                                                                                                | Result                                                          | What it establishes                                                              |
| --- | ---------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------- | ---------------------------------------------------------------------------------- |
| M1  | `parameters.ValidateAudience / ValidateIssuer / ValidateLifetime = false` after the builder call at the back-channel call site | **3 × CA5404 errors**                                             | CA5404 does reach a post-build literal; that half of the sentence was sound        |
| M2  | the same property assigned a non-constant false (`bool.Parse("false")`)                                    | builds clean, **0 warnings**                                      | CA5404 is literal-only, "wherever it is written" is the false part                |
| M3  | `RequireSignedTokens = false` in the builder                                                               | builds clean, **0 warnings**                                      | the analyzer never licensed leaving the signature disablements untested            |
| M4  | delete `ValidAlgorithms = AllowedSignatureAlgorithms` from the builder                                     | builds clean, **0 warnings**                                      | same, for algorithm confusion                                                      |
| M5  | isolated probe, `xunit.v3.mtp-v2` 3.2.2 / net10.0 / defaults: one `DisableParallelization` collection with two classes in it, two classes outside; five runs | outside pair overlaps **every run**; the two inside never overlap, and neither overlapped the outside pair | the documented mechanism was false; the race is un-attributed ⟷ un-attributed      |
| M6  | a NEW file with `new TokenValidationParameters { … }`                                                      | rule **RED**, message naming the file                             | term 1 still bites                                                                 |
| M7  | a NEW file with `TokenValidationParameters p = new();`                                                     | rule **RED**                                                      | the target-typed spelling still bites                                              |
| M8  | a NEW file passing a bare `new() { … }` straight into `ValidateTokenAsync`                                 | rule **RED**                                                      | term 3 still bites                                                                 |

### Round-4 dispositions

| Dim   | Finding                                                                                          | Disposition                                                                                                                                                                                     |
| ----- | -------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1, 9  | The CA5404 coverage claim is false (CONFIRMED)                                                     | **FIXED by removal**, with M1–M4 behind it. It went out with the enumeration it licensed rather than being rewritten smaller                                                                       |
| 2, 23 | The threat model contradicts `DisableParallelization` (CONFIRMED by reproduction)                  | **FIXED by correction**, re-reproduced here as M5. Doc comment and failure message both now state the un-attributed ⟷ un-attributed race                                                            |
| 2, 13 | The rule is file-scoped while its contract reads per class, and multi-type files exist here        | **FIXED by weakening.** The doc says per FILE, and that a file declaring two types is one unit to the scan. Moving the scan to class scope would be a new guard, so it is not in this round         |
| 2     | `ConstructsValidationParameters` misses the multi-line target-typed spelling its own comment claims | **FIXED by weakening the comment.** No fourth scanning term, the predicate reads one line and now says only that                                                                                  |
| 2     | Two unfalsifiable assertions                                                                       | **NOT ACTIONED.** The adjudication does not name them and the per-lens blocks are not in the record I hold. Strengthening an assertion is outside this round's remit either way; it carries forward |
| 3     | The login↔logout per-shape parity claim is false for the allowlist row (three lenses)              | **FIXED by weakening**, reversing my round-3 decline. The exemption I wrote covers spelling fan-outs; that row is a shape, so the exception is named                                               |
| 4     | The deliverable is a guard set plus its written map, and the map overstates the guard set          | **FIXED**, that is what this round is, end to end                                                                                                                                                |
| 7     | 1,069 lines against the ~400 target; splits into three units                                       | **PARTLY.** The prose removal takes it to 1,042. The one remaining lever is dropping the `OidcRoundTrip` trio, which removes coverage round 4 credits under dim 16, so it stays and the size decision stands recorded below |
| 8     | ~30 RSA keygens on the serial collection; two serialized classes clear different caches            | **DECLINED.** Re-measured: 21.9 s / 23.4 s for 4,412 tests. The membership is required by the rule, so removing it would red the build                                                             |
| 10    | Two-sided exact-equality pins turn a benign rename into a hard build stop (PLAUSIBLE)              | **DECLINED, already documented.** Round 3 rewrote both failure messages to say which direction they differ in and that a missing file is a rename, not a bypass                                     |
| 11    | The stated ctor/dispose invariant is false for `SamlAssertionValidatorTests`                       | **FIXED by weakening.** It clears inside three `[Fact]` bodies and declares neither, so the doc says clears "around its own tests"                                                                 |
| 11    | The liveness floor sits at exactly the current count, with zero headroom                           | **DECLINED, that is the design.** The floor exists to red on a rename that would otherwise read as compliance; headroom would blunt exactly that                                                  |
| 11    | Bare method-name substrings scanned tree-wide                                                      | **STATED, not changed.** The doc now says what the scan reads: code lines, one file at a time                                                                                                     |
| 13    | The logout theory's oracle is a four-element denylist                                              | **DECLINED, as in round 3.** It is the spec-derived partition of codes reachable only after signature validation; #1039 rewrites it when it separates the codes                                     |
| 14    | The weakening applied in `OidcSignatureKeys` survives verbatim in `OidcLogoutTokenValidator`        | **FIXED by weakening**, in that file's own terms, it validates on parameters the caller passes in                                                                                                 |
| 6     | `CVE-2026-42206`, cited in a comment, was not verifiable from the review environment               | **VERIFIED and kept.** The record is an OIDC nonce generated on the authorization request and never validated on the callback (CWE-345, ID-token replay), the shape the comment names             |
| 5, 12, 15–22, 24, 25 | PASS or n/a                                                                       | No action                                                                                                                                                                                         |

## What round 3 changed about this PR

Round 3 returned NEEDS_IMPROVEMENT with three CONFIRMED findings, all of them in the assurance artifact rather than in the shipped posture: no mutation any lens ran made a forged token validate. Three rounds is where a deliverable goes back to planning by default. **I overrode that, and recorded the override on #1004 rather than leaving it implicit**, because the reviewer's own adjudication put the remaining work at three bounded items and two of them are corrections of a claim, subtraction, not the growth the rule exists to stop. The override covers exactly those three. Anything that turned out to need more than it looked would have been the signal the override was wrong; none of them did.

**1. One forgery now crosses the production endpoint.** Every row of the battery hands a validator parameters the test built through the shared builder, so what it establishes is what the token library refuses under that basis, not what the endpoint refuses. `AlgNoneForgery_IsRefusedByTheProductionPath_WhileTheGenuineTokenStillRevokes` submits an unsigned `logout_token` to the anonymous back-channel endpoint against in-process discovery and JWKS, and the genuine token goes through the same harness afterwards and must still be accepted, without that control the rejection would prove nothing, since an unserved discovery document or a misspelled provider produces the identical uniform 400. `alg: none` is the pick because it is the one shape needing no key material at all: anyone who can reach that URL can mint it. It lives beside the endpoint's other tests, where the harness and responder already exist, rather than as a fifth near-copy of the discovery responder inside the battery; both class docs say which file holds which claim.

**2. The gap enumeration was wrong by one, and is now four.** The rule's doc comment shipped an explicitly measured, closed list of what the three scan terms miss, and it omitted the cheapest evasion of the lot. Both doc comments now enumerate four shapes, each verified by making the edit and running the rule over it. The two new ones are (c) post-build mutation of what the builder returns, and (d) the allowlist's wiring as against its content, enumerated separately because (d) is not an evasion at all but a limit on what the rule asserts about a file it does read. **No fourth scanning term was added**, which is the whole point: what changed is the claim, and each of (c) and (d) now names the one test that does notice it, so deleting that test reads as the larger change it is.

**3. The replay-cache collection membership is held by the build.** The four `[Collection("SSOController")]` attributes were hand-applied; the existing rule keys on `new SsoControllerHarness` and sees nothing else. `EveryTestClassClearingAReplayCache_IsInTheNonParallelControllerCollection` closes that one door: a test class whose own code lines call a `ResetReplaysForTests` hook carries the attribute, or the build is red naming the file. Code lines only, `[Fact]`/`[Theory]` required, its own liveness floor. Deriving the whole door set and closing the base-class route stays #1044, and the doc comment says so instead of implying a reach the scan has not got; #1044 is updated with what this now covers.

### Weakenings run this round

Five. Each applied, **proved landed by printing the patched region**, built with `--warnaserror`, run, reverted, with `git status --porcelain` asserted empty on both sides, the discipline exists because a weakening that silently fails to apply comes back green and is indistinguishable from a caught mutant, which happened on this branch in round 2.

| #   | Weakening                                                                                                       | Result                                                                                                                                          | What it establishes                                                                          |
| --- | --------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
| W9  | `parameters.RequireSignedTokens = false;` inserted after the builder call in `OidcLoginService.ValidateBackChannelLogoutAsync` | rule **GREEN**, build **0 warnings**; `AlgNoneForgery_…` **RED** - the unsigned forgery was accepted, returning 200 and revoking a session | Gap (c) is real, and the new endpoint row is the one thing that notices it                     |
| W10 | Delete `ValidAlgorithms = AllowedSignatureAlgorithms` from the builder                                          | rule **GREEN**, build clean; `AlgorithmAllowlist_RefusesAnHs256TokenTheKeySetWouldOtherwiseVerify` **RED**; `AlgNoneForgery_…` still green      | Gap (d) is real and is a different miss from (c) - the endpoint row does not cover it          |
| W11 | Hoist `OidcIdTokenValidator`'s inlined builder call into an explicitly typed local                              | term 2 **RED** with the new message; terms 1 and 3 unaffected                                                                                    | The naming-list exclusion behaves as documented, and the message now names this benign case    |
| W12 | Strip `[Collection("SSOController")]` from `SamlLogoutValidatorTests`                                           | new rule **RED** - "Missing in: SamlLogoutValidatorTests.cs"                                                                                     | The ratchet catches the shape it exists for                                                    |
| W13 | Rename `ResetReplaysForTests` → `ClearReplaysForTests` across all 9 files                                       | liveness floor **RED**, "matched only 0 test classes"                                                                                           | The scan cannot go blind and read as compliance                                                |

W9 and W10 are the measurements behind the two new enumeration items; they are quoted in the doc comment with their outcomes rather than described.

### Round-3 dispositions

| Dim   | Finding                                                                                      | Disposition                                                                                                                                                                                                                                                                                                                                                    |
| ----- | ---------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1     | The three terms are blind to post-build mutation, and that shape exists at the back-channel caller (CONFIRMED ×3 lenses) | **FIXED by weakening.** Enumerated as (c) in both doc comments with W9's outcome. No fourth term. The structural closure, validators that derive their basis instead of accepting one, is **DEFERRED → #1049**                                                                                                                                              |
| 2     | Deleting `ValidAlgorithms` from the builder leaves the rule green (CONFIRMED)                 | **FIXED by weakening.** Enumerated as (d), separately from (c), because it is a limit on what the rule asserts rather than an evasion of what it scans. W10 names the test that does notice                                                                                                                                                                    |
| 2     | `QueryValue` returns `""` for absent, empty and valueless alike (PLAUSIBLE)                   | **DECLINED.** The assertion's stated regression is a nonce **value** attached to the authorize URL, and any real nonce is non-empty by construction, so it still reds on the shape the comment names. Distinguishing the three states would not change a verdict                                                                                               |
| 2     | The logout theory's oracle is a hand-copied negative list of reason codes (PLAUSIBLE)         | **DECLINED, with the expiry named.** The list is the codes reachable only after signature validation has passed, a spec-derived partition, not an enumeration of the enum. #1039 separates those codes and rewrites this assertion with them; until then a new post-signature code is a two-line edit in the same file the code is added from                    |
| 3     | No forgery crosses the production endpoint (CONFIRMED)                                       | **FIXED**, see item 1 above                                                                                                                                                                                                                                                                                                                                   |
| 3     | `AlgorithmAllowlist_…` has no `ForgeryKind` counterpart, contradicting the parity claim       | **DECLINED.** The claim is scoped to shapes the login path **submits**, and that test submits nothing through either validator: it drives the handler directly against a symmetric key placed into the key set by hand, because today's JWK conversion never yields one. There is no shape for a counterpart to mirror                                            |
| 4     | Four `[Collection]` additions with no comment and no `#1004` annotation                       | **FIXED.** One line on each saying which cache the class clears and why the collection is load-bearing, and the new rule now requires the attribute, so they are no longer four hand-applied lines                                                                                                                                                            |
| 7     | Term 2's `SequenceEqual` forbids a file the rule blesses from naming the type                 | **JUDGED INTENDED, and now said so.** Keeping the list to exactly the files that name the type is what stops it passing vacuously; the cost is that a behaviour-preserving hoist reds it, and W11 measures that it reds that one assertion only. Both the comment and the failure message now name this case and say the repair gives up nothing, since construction stays pinned by the assertion above |
| 11    | The logout validator accepts caller-supplied parameters; `AllowedSignatureAlgorithms` is a mutable shared array | **DEFERRED → #1049.** Both are the same theme as (c): the hardened basis is mutable once it leaves the builder. The fix changes production signatures on the sensitive OIDC surface, which is not what an override scoped to correcting claims should carry                                                                                                     |
| 13    | The ratchet gap                                                                              | **FIXED**, see item 3 above, measured by W12 and W13                                                                                                                                                                                                                                                                                                          |
| 13    | `RedeemResponse_…`'s seven hand-enumerated fields behind an "every field" claim               | **FIXED.** Same defect class as the enumeration: the comment now says the scan covers the string-carrying fields, listed one by one, and that a field arriving in a Jellyfin update is inspected only once someone adds it there. One comment, beyond the three items, recorded here rather than folded in quietly                                               |
| 13    | `HostileKidValues_…` pins the permissive key-fallback posture as the green state              | **DECLINED, already declared.** The acceptance half is documented as flipping deliberately when the `kid` allowlist (#1029) lands - that is the point of it: the posture change must be a conscious edit, not a silent drift                                                                                                                                    |
| 14    | The replacement doc comment ships an explicitly "measured", closed gap list that is wrong by one | **FIXED**, see item 2 above                                                                                                                                                                                                                                                                                                                                   |
| 19    | Two failure messages misdescribe the defect they fire on                                      | **FIXED.** Both set comparisons are two-sided; each message now says which direction it differs in, what that means, and what the repair is - an extra file is the finding, a missing one is a rename                                                                                                                                                            |
| 23    | The `[Collection]` set is hand-applied; nothing keeps it complete                             | **FIXED**, see item 3 above. The general door-set derivation and the base-class route stay #1044, which is updated with what this rule now covers                                                                                                                                                                                                             |
| 5, 6, 8, 10, 12, 15–18, 20–22, 24, 25 | PASS or n/a                                                                    | No action. Dim 6's absent shapes were already deferred to #1047 and #1038 in round 2                                                                                                                                                                                                                                                                           |

## What round 2 changed about this PR

The gate's central finding was that **both guards this PR adds fail to enforce what their own doc comments and failure messages claim**, proved by execution. Two shapes evade all three terms of the parameter-builder rule with output identical to a clean run: a hand-rolled verifier that never touches the token library, and a second, laxer parameter set built inside a file already on the rule's own lists. The comment said evading all three "would take reflection".

The remedy is to weaken the claim to what is held, not to add a fourth term. A text scanner enumerates spellings over an unbounded space, so a fourth term moves the next evasion one identifier away rather than closing it, and this PR is exactly the fix-loop that rule would start. So: **no new scanning term, no new literal, no new guard code.** The doc comment now states what the three terms catch, what they do not, and names review as the control that owns the remainder.

### A unit left the PR

Round 2 measured the diff at 750 changed lines against the ~400 target and observed that it divides into four units. I took one out rather than argue for it.

**Removed: the test-parallelism rule widening → #1044.** Widening `EveryHarnessUsingTestClass_…` to the `*ForTests()` reset hooks fixes a defect that predates this work and belongs to none of #1004's acceptance criteria. It also arrived carrying three findings of its own, a third scanned literal instead of a derived door set, no `<summary>` threat model while the rule added beside it has one, and two demonstrated escapes (a test base class; a comment-only match). #1044 carries all of that as its contract, including the measured detail that the wiki already forbids test base classes with nothing enforcing it.

What stayed is the collection membership **this branch's own new class needs**: `OidcTokenForgeryTests` and `OidcLogoutTokenValidatorTests` both call `OidcLogoutTokenValidator.ResetReplaysForTests()`, so leaving either out re-opens a race this PR creates. The two SAML classes' membership stayed too, it is correct and it is two lines; deliberately re-opening a real race to purify the scope would be the rule taken past sense. Making it *enforced* is #1044's job.

**Nothing else honestly leaves.** Each remaining file maps to a named #1004 bullet: the parameter-builder rule is acceptance criterion 1 verbatim; the forgery battery is the deliverable; the round-trip additions are the id_token-never-a-session-token, redirect-URI and nonce criteria; the id-token validator additions are the claim-validation negatives.

## The claim I had to weaken in production

While probing the rule I found the same defect one file over. `OidcSignatureKeys`' class doc said centralising the allowlist and key conversion "means there is **provably** no second, laxer verification path". W3 and W4 below refute the word. What is true is narrower, and is what it says now: both token types read this one basis, so neither can accept key material the other rejects, a property of the code as it stands, held by a conformance rule and by review, not a proof that no second path can exist. It now points at that rule's own statement of its limit instead of promising more.

#1004 exists because nothing locked that basis in and nothing proved it empirically. A doc comment asserting the proof already existed is the first thing that had to go.

## Weakenings run this round

Eight. Each was applied, **proved landed by printing the patched region**, built, run against the full suite, and reverted with `git status` asserted empty on both sides. One attempt (W4, first try) silently failed to apply because the tool it used was unavailable, it came back green and was indistinguishable from a caught mutant until the printed region turned up empty. That run was discarded, which is the entire reason the region is printed.

| # | Weakening | Result | What it establishes |
|---|---|---|---|
| W1 | Target-typed `TokenValidationParameters p = new()` in a NEW file | **CAUGHT**, rule red, naming `ProbeParamsBuilder.cs` | Term 2 bites; the spelling a substring scan walks past is closed |
| W2 | Bare `new() { … }` passed straight into `ValidateTokenAsync` in a NEW file | **CAUGHT**, 2 failures, rule only | Term 3 bites; the construction that names the type nowhere is closed |
| W3 | Hand-rolled verifier (`jwt.Split('.')` + `RSA.VerifyData`) in a NEW file under `Api/Oidc/` | **NOT CAUGHT**, 4408 tests, **0 failures** | The rule sees nothing in a path that bypasses the library. Review's, not the rule's |
| W4 | A second, laxer parameter set inside `OidcLogoutTokenValidator.cs`, `new()` wrapped onto its own line | **NOT CAUGHT**, 4408 tests, **0 failures** | Per-file granularity cannot distinguish a second construction from the first |
| W5 | Signature failure reports a code only reachable after validation passed | **CAUGHT**, 28 failures incl. `LogoutToken_SameForgeries_AreRejected` | The loosened reason-code assertion still bites |
| W6 | `IssuerSigningKeyResolver` consults the header's own `x5c` | **CAUGHT**, `LogoutToken_SameForgeries_AreRejected(kind: HeaderSuppliedKeyMaterial)` red | The new logout row is not vacuous |
| W7 | Delete `ValidAlgorithms = AllowedSignatureAlgorithms` | **CAUGHT**, `AlgorithmAllowlist_RefusesAnHs256TokenTheKeySetWouldOtherwiseVerify` red | Round 1's headline fix survives this round's edits |
| W8 | `RequireSignedTokens = false` | **builds clean, 0 warnings** | Bounds CA5404's reach (below) |

W3 and W4 are the two the doc comment now names as out of reach. W1 and W2 are the positive controls that keep it from being a blanket disclaimer.

**A second control I did not know about, found by W2's first attempt failing to build:** CA5404 is an **error** in this repo, so `ValidateIssuer`, `ValidateAudience` or `ValidateLifetime` set to `false` fails the build wherever it is written, including inside a file the scan already allows. Its boundary is measured, not assumed: W7 and W8 show it does **not** cover a missing `ValidAlgorithms` or `RequireSignedTokens = false`, the two disablements that matter most here. That boundary is now in the rule's doc comment.

> **Corrected in round 4.** "Wherever it is written" is false: CA5404 fires on a **literal** `false` only, and the same property assigned a non-constant builds clean (M1, M2). The sentence is out of the doc comment, not rewritten smaller.

## Round-2 dispositions

Every finding in the round's verdict table. The raw per-lens blocks behind the summarised counts were not retained by my review driver, a gap on my process side, not in this diff; the dispositions below are against the recorded verdict.

| Dim | Finding | Disposition |
|---|---|---|
| 1 | Both guards over-claim; hand-rolled verifier evades all three terms (CONFIRMED) | **FIXED** by weakening. Doc comment now states catch/not-catch and names review as owner. W1–W4 |
| 2 | Per-file scoping of the parameter-builder guard | **FIXED** by weakening, the failure message itself now says a second construction inside a listed file is invisible here and is review's to catch |
| 3 | `ForgeryKind` carries 6 shapes against 8 on the login path; class doc claims full parity | **FIXED both ways.** The header-injection shape (`jku`/`x5u`/`x5c`) is added, it is the one that matters most on an anonymous endpoint where the signature is the only authenticator. The claim is weakened to what remains: parity is **per shape, not per spelling**, and the doc says which fan-outs are login-path-only and why |
| 4 | UNJUDGEABLE, acceptance criteria not in the supplied material | **Not a contract gap.** Verified: #1004 carries a seven-bullet scope list, seven acceptance criteria, and names its conformance test (`OidcTokenValidation_UsesTheSingleHardenedParameterBuilder`) explicitly. The three deferral targets are real, triaged issues, not comments: #1029 (`priority:low`, `area: oidc`, P8), #1038 and #1039 (both `security`, `priority:low`, Future/Post-1.0). The gate could not see any of it, a material-supply defect in my review driver, fixed on my process side |
| 5 | Product fit | PASS, no action |
| 6 | `crit`, JWE/nested JWT, `typ` confusion absent | **DEFERRED → #1047** (JWE/nested + `typ`). `crit` was already #1038. Each needs its own weakening run to be worth having |
| 7 | 750 lines; third literal not a derived door set; 3,349-line conformance file; `FormValue` a near-copy | **Split four ways.** Third literal → **REMOVED** with its rule (#1044). Conformance-file size → **DEFERRED → #1045**. Parse-loop duplication → **DEFERRED → #1046**, and it is **eight** copies, not four, measured, with file and line for each. Size → recorded below |
| 8 | 4 classes added to the serialized lane; RSA keygen cost | **DECLINED.** Now 2 classes, not 4 (the SAML pair no longer needs the lane for a rule this PR removed). Measured this round: `net9.0` 20.1 s, `net10.0` 22.2 s for 4,408 tests. Proportionate for a security battery |
| 9 | Renamed parallelism rule carries no `<summary>` threat model | **RESOLVED BY REMOVAL**, the rename is gone; the rule is byte-identical to mainline again. The obligation travels to #1044, where it is an acceptance criterion |
| 10 | Availability, the merge gate itself and suite wall-clock | **DECLINED**, see dim 8. No production, boot, migration or lockout path touched |
| 11 / 19 | Two `Assert.Equal` set comparisons ship with no failure message | **FIXED.** Both are now `Assert.True` over `SequenceEqual` with a message printing the expected and found sets. xUnit's collection `Assert.Equal` has no message overload, so this is the only shape that can carry one |
| 12 | Audit integrity | n/a, no tamper-evident trail in this repo |
| 13 | `HostileKid` and `ForeignKeyWithTrustedKid` differ only in a `kid` string; the row kills nothing | **FIXED by weakening, and the finding is partly refuted.** Renamed `ForeignKeyWithTraversalKid` and documented for what it holds: it does *not* show the `kid` decides nothing (no acceptance control on this path, the login path's differential is where that lives), but it is not empty either. Deleting it *would* lose coverage: it holds that a traversal `kid` at the anonymous endpoint is **refused, not thrown on**, and a throw there is an unauthenticated 500 oracle and a DoS lever, exactly what a `kid` sanitiser added later would produce |
| 14 | Wiki Coding-Standards is the single home for structural rules; update unverified | **DEFERRED → #1048.** I could reach the wiki this time and checked: the page states rule *families* and names fitness functions as examples; it does not enumerate them, and the pre-existing `SamlSignaturePath_UsesOneXmlStackEndToEnd` is absent from it too. So the gap is one missing **family**, "one implementation per security-critical primitive", covering both rules. The wiki has no PR flow, so per `CONTRIBUTING.md` the tracking is an issue on the current-release milestone |
| 15 | i18n | n/a, no user-visible strings |
| 16 | Privacy | PASS, no action |
| 17 | Supply chain | PASS, no dependency, toolchain or lockfile change |
| 18 | Compatibility / migration | n/a, no persisted format, config shape or ABI touched |
| 20 | Accessibility | n/a, no web UI diff |
| 21 | Licensing | PASS, SPDX header on every touched file |
| 22 | Six fail-closed branches collapse to one reason code, and the diff *pins* the collapse | **FIXED.** The theory no longer pins the exact string. It asserts the **layer** the refusal came from, a forgery must die in signature validation, before any claim-shape check, so separating the codes will not arrive as a red test in this file. W5 proves the loosened form still bites, and the run also confirms the exact-string contract stays pinned where it belongs, in `OidcLogoutTokenValidatorTests` |
| 23 | Parallelism guard has demonstrated escapes; the class has no structural owner | **RESOLVED BY REMOVAL → #1044**, which carries both escapes as acceptance criteria. Recording the ownership gap rather than treating lens silence as clearance is the reason it is an issue and not a note |
| 24 | No row exercises the `RejectReason.Malformed` branch | **DECLINED with evidence.** The branch is covered, `OidcLogoutTokenValidatorTests.cs:88` asserts it directly. The forgery battery submits well-formed tokens by construction, which is the point of it; adding a malformed row there would duplicate a covered branch in the wrong file |
| 25 | Rollout / flags | n/a, nothing gated, phased or deprecated |

### Still deferred from round 1

| Finding | Issue |
|---|---|
| One reason code for six forgery shapes | #1039. Production reason-code contract + audit trail; outside a diff this size |
| `crit` header accepted (found while probing round 1) | #1038. Pinning today's behaviour would enshrine a spec deviation |
| `kid` allowlist (an acceptance criterion of #1004) | #1029, pre-existing. The hostile-`kid` test's acceptance half deliberately flips when it lands |

### Deferred from round 3

| Finding | Issue |
| --- | --- |
| The hardened basis is mutable once it leaves the builder, a caller may change it, the logout validator accepts one it did not build, and the allowlist is a shared mutable array | #1049. It closes gaps (c) and (d) structurally instead of by enumeration, and changes production signatures on the sensitive OIDC surface |

## Size, a recorded decision

`git diff --shortstat origin/main...HEAD` at `04b0508`: **10 files, 1,035 insertions, 7 deletions, 1,042 changed lines**. It **shrank by 27** this round, which is the direction a round of removals and claim corrections has to move: 57 lines of enumeration out against 16 of replacement prose, plus the corrections. Round 4 measured 1,069 at `e775eed`; round 3 measured 889 at `3a2c1ea`; round 2 measured 750 at `d846712`.

**It grew by 180 this round, and here is where every line of it went**, because a size figure with no account of it is not a measurement:

- **+67**, the endpoint-driven forgery: the test, the hand-composed `alg: none` token helper, and the two class-doc paragraphs saying which file now holds which claim. Round 3's dim 3 finding is not fixable in fewer lines than a test.
- **+35 net** (54 added, 19 removed), the corrected gap enumeration in both doc comments, the two rewritten failure messages, and the note on the naming list. This is the round's central remedy and it is entirely prose: an over-claim is repaired by writing down what is held, which costs lines and buys the one thing the finding was about.
- **+74**, the replay-cache ratchet: 30 lines of rule, 25 of threat-model doc comment, 11 of attribute comments on the four memberships, the rest structure.
- **+2**, the id_token leak-scan comment.
- **+2**, narrowing three claims in this round's own comments, found by re-reading them against the code the way every other claim here was checked: the battery's class doc said "every row calls a validator" when the allowlist row drives the handler directly; "an in-test acceptance control for the same reason the rows here do" implied every row carries one; and the enumeration said the mutated endpoint "revokes a session" when what was observed is `OkResult` where the uniform 400 belongs, with the revoke following from the endpoint's contract.

The alternative to the growth was to leave three CONFIRMED findings open or to reopen the deliverable, and the reviewer's own adjudication said the work was small and belonged here. I recorded that judgement as a break-glass on #1004 instead of absorbing it silently.

Round 2's growth of 139 stands as previously recorded, with the same reasoning: Removing the parallelism unit took 25 lines out. The round's remedies added 164, and almost all of that is **prose**: the remedy for a guard that over-claims is a written statement of its limit, which costs lines and buys the one thing the finding was about. The two new set-comparison messages, the enum documentation, and the two doc comments are the bulk; the only new executable lines are one `ForgeryKind` arm, one helper, one `InlineData` row and the reason-code assertion.

I am not claiming the "genuinely does not divide" exemption. It divides, I removed one unit and looked hard at the other three, and each maps to a named acceptance criterion of #1004, so removing any of them ships #1004 incomplete rather than smaller.

**The real defect is upstream and it is mine.** #1004 lists seven scope bullets and seven acceptance criteria. An issue that size cannot close in a 400-line PR, and the ceiling should have bitten when I *wrote* it, by splitting it into three issues, not at review time. That is the durable correction, not a smaller PR here.

## Verification

- `dotnet build --warnaserror`, success, **0 warnings, 0 errors**, both TFMs, run again after the last comment edit.
- `dotnet test --no-build`, **4,412 total, 0 failed, 0 skipped** (2,206 per TFM) on `net9.0` and `net10.0`. Unchanged by this round, which touches only prose. Wall clock 21.9 s / 23.4 s.
- Twenty-one measurements across four rounds, eight of them this round (M1–M8 above). Rounds 1–3 ran their weakenings in the work tree, printed and reverted with `git status --porcelain` asserted empty on both sides; round 4 ran every one of them in a throwaway clone of the branch instead, so the work tree was never modified at all.
- SPDX header on every touched `.cs`; DCO sign-off on all commits.
- The `-p:JellyfinVersion=…` flag was **not** passed (#1031).

## Not done

- Not marked ready, not merged, no tags pushed, nothing posted to this thread.
- Two production files changed, comment-only, 21 lines. No mutation this round ran in the work tree.
- No guard, scanning term, test or explanatory paragraph was added this round. Everything in it is a removal or a claim narrowed to what is held.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -2617,22 +2617,45 @@ public class ArchitectureConformanceTests
     /// passing a bare <c>new() { … }</c> straight into <c>ValidateTokenAsync</c>, which names the type
     /// nowhere and is invisible to both scans (term 3 pins the handler entry points to the two validators).
     ///
-    /// WHAT THEY DO NOT CATCH, measured the same way, because a guard that overstates its reach is what the
-    /// next author trusts instead of re-deriving it. Two shapes produce output identical to a clean run.
-    /// Verification that never uses the library at all — splitting the JWS and calling <c>RSA.VerifyData</c>
-    /// directly — constructs no parameters and calls no handler, so no term applies. And a second, laxer
-    /// parameter set built INSIDE a file already on the two lists below is indistinguishable to a per-file
-    /// scan from the first one there. Both belong to review, not to this rule: a text scan has no notion of
-    /// "how many times in this file" and none of "does this reimplement the verifier", and the OIDC surface
-    /// is on the sensitive list where every change is reviewed before merge. Adding a fourth term for them
-    /// is the wrong repair — the terms enumerate spellings over an unbounded space, so each one added moves
-    /// the next evasion one identifier away rather than closing it.
+    /// WHAT THEY DO NOT CATCH, because a guard that overstates its reach is what the next author trusts
+    /// instead of re-deriving it. FOUR shapes, each verified by making the edit and running this rule over it,
+    /// not by reading the predicates:
+    ///
+    /// (a) Verification that never uses the library at all — splitting the JWS and calling
+    /// <c>RSA.VerifyData</c> directly. It constructs no parameters and calls no handler, so no term applies.
+    ///
+    /// (b) A second, laxer parameter set built INSIDE a file already on the two lists below. A per-file scan
+    /// cannot tell it from the first one there.
+    ///
+    /// (c) POST-BUILD MUTATION of what this builder returned, which is the cheapest of the four and sits at
+    /// the endpoint where the signature is the only authentication. The back-channel caller holds the result
+    /// as <c>var parameters = …BuildValidationParameters(…)</c> and hands it to the logout validator's own
+    /// <c>ValidateAsync</c>: that file names the type nowhere, constructs nothing, and calls no handler entry
+    /// point, so all three terms are silent about it. Measured: with <c>parameters.RequireSignedTokens =
+    /// false;</c> added after that call, this rule passes and the build emits zero warnings, while the
+    /// unsigned forgery is ACCEPTED and revokes a session. What turns red is one test —
+    /// <see cref="SSOControllerOidBackChannelLogoutTests.AlgNoneForgery_IsRefusedByTheProductionPath_WhileTheGenuineTokenStillRevokes"/>,
+    /// the row that drives that call site rather than a basis a test built.
+    ///
+    /// (d) The allowlist's WIRING, as against its content. The last assertion below reads
+    /// <c>AllowedSignatureAlgorithms</c> and pins what is in it; nothing here pins that the builder assigns it
+    /// to <c>ValidAlgorithms</c>. Measured: deleting <c>ValidAlgorithms = AllowedSignatureAlgorithms</c> from
+    /// the builder leaves this rule green and the build clean; what turns red is
+    /// <see cref="OidcTokenForgeryTests.AlgorithmAllowlist_RefusesAnHs256TokenTheKeySetWouldOtherwiseVerify"/>.
+    ///
+    /// Adding a fourth term for any of them is the wrong repair — the terms enumerate spellings over an
+    /// unbounded space, so each one added moves the next evasion one identifier away rather than closing it,
+    /// and a text scan has no notion of "how many times in this file", of "does this reimplement the
+    /// verifier", or of what happens to a value after it is returned. (a) and (b) are review's, on a surface
+    /// where every change is reviewed before merge. (c) and (d) have a named test each, above — which is why
+    /// deleting one of those two rows is a larger change than it looks.
     ///
     /// One automated control does reach inside a file, and its boundary was measured rather than assumed:
     /// CA5404 is an ERROR here, so <c>ValidateIssuer</c>, <c>ValidateAudience</c> or <c>ValidateLifetime</c>
     /// set to false fails the build wherever it is written, this file included. It does NOT cover the two
-    /// disablements that matter most to the shape above — a missing <c>ValidAlgorithms</c> and
-    /// <c>RequireSignedTokens = false</c> each build clean with zero warnings.
+    /// disablements that matter most here — a missing <c>ValidAlgorithms</c> and
+    /// <c>RequireSignedTokens = false</c> each build clean with zero warnings, re-measured at the (c) and (d)
+    /// mutation sites above.
     /// </summary>
     [Fact]
     public void OidcTokenValidation_UsesTheSingleHardenedParameterBuilder()
@@ -2658,6 +2681,13 @@ public class ArchitectureConformanceTests
         // this, a target-typed construction lands in a new file and the line-level check above has to be
         // trusted to have matched every way of spelling it; with it, any new file touching the type is a red
         // build that a human reads.
+        //
+        // OidcIdTokenValidator.cs is absent from the list ON PURPOSE, even though the rule blesses it as a
+        // validation call site below: it reaches the basis through the builder and names the type nowhere, so
+        // the list stays exactly the set of files that name it and cannot pass vacuously. The cost is that a
+        // behaviour-preserving hoist there (`TokenValidationParameters p = OidcSignatureKeys.Build…`) reds
+        // this assertion, and the repair is to add that file here — which gives up nothing, because
+        // CONSTRUCTION stays pinned to the basis file by the assertion above whatever this list says.
         var mentions = production
             .Where(path => CodeLines(path).Any(l => l.Text.Contains("TokenValidationParameters", StringComparison.Ordinal)))
             .Select(path => Path.GetRelativePath(RepoRoot(), path))
@@ -2669,7 +2699,7 @@ public class ArchitectureConformanceTests
 
         Assert.True(
             mentions.SequenceEqual(allowedToNameTheType, StringComparer.Ordinal),
-            $"Exactly these files may name TokenValidationParameters, so that a target-typed construction cannot land unread in a new one (#1004). Expected: {string.Join(", ", allowedToNameTheType)}. Found: {string.Join(", ", mentions)}");
+            $"These files and no others may name TokenValidationParameters, so that a target-typed construction cannot land unread elsewhere (#1004). Expected: {string.Join(", ", allowedToNameTheType)}. Found: {string.Join(", ", mentions)}. The comparison is two-sided: an EXTRA file is the finding this rule exists for, while a MISSING one means a listed file was renamed or stopped naming the type — a rename, not a bypass, and the repair is to update the list. OidcIdTokenValidator.cs appearing here is the other benign case: it consumes the basis through the builder, so adding it to the list gives up nothing the construction assertion above does not still hold.");
 
         // Term 3: the handler entry points that consume those parameters. A `new() { … }` written straight
         // into one of these calls names the type nowhere and neither scan above can see it — but the call
@@ -2687,7 +2717,7 @@ public class ArchitectureConformanceTests
 
         Assert.True(
             validationCallSites.SequenceEqual(allowedToValidate, StringComparer.Ordinal),
-            $"Exactly these files may call a handler token-validation entry point, which is what closes a bare `new() {{ … }}` passed straight into one (#1004). Expected: {string.Join(", ", allowedToValidate)}. Found: {string.Join(", ", validationCallSites)}");
+            $"These files and no others may call a handler token-validation entry point, which is what closes a bare `new() {{ … }}` passed straight into one (#1004). Expected: {string.Join(", ", allowedToValidate)}. Found: {string.Join(", ", validationCallSites)}. Two-sided as well: an EXTRA file is a verification path no forgery test drives, while a MISSING one means a validator was renamed or merged — update the list.");
 
         // The allowlist's CONTENT, read off the production type rather than re-stated here. Symmetric HS* would
         // accept a token minted by anyone holding the shared client secret (the RS256-public-key-as-HMAC-key
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -2697,7 +2697,8 @@ public class ArchitectureConformanceTests
     /// <c>ValidateAsync</c>: that file names the type nowhere, constructs nothing, and calls no handler entry
     /// point, so all three terms are silent about it. Measured: with <c>parameters.RequireSignedTokens =
     /// false;</c> added after that call, this rule passes and the build emits zero warnings, while the
-    /// unsigned forgery is ACCEPTED and revokes a session. What turns red is one test —
+    /// unsigned forgery is ACCEPTED — the endpoint answers <c>OkResult</c>, which on that path is reachable
+    /// only once a captured session has been revoked. What turns red is one test —
     /// <see cref="SSOControllerOidBackChannelLogoutTests.AlgNoneForgery_IsRefusedByTheProductionPath_WhileTheGenuineTokenStillRevokes"/>,
     /// the row that drives that call site rather than a basis a test built.
     ///
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -2564,6 +2564,70 @@ public class ArchitectureConformanceTests
             "The harness-usage scan matched fewer than 10 files — SsoControllerHarness was renamed; update this rule.");
     }
 
+    /// <summary>
+    /// Threat model (#1004). There is no adversary here; the loss is a security test that silently stops
+    /// testing. Asset: the process-wide replay caches — <see cref="OidcLogoutTokenValidator"/>,
+    /// <c>SamlAssertionValidator</c> and <c>SamlLogoutValidator</c> each keep one, and each backs a
+    /// one-time-use property (a captured token must not be redeemable a second time). A test class that
+    /// exercises one clears it in its constructor and again on dispose, so a consumed id cannot leak into a
+    /// sibling. xUnit runs a class carrying no <c>[Collection]</c> in its own implicit collection, IN PARALLEL
+    /// with the serialized one — so an un-attributed class clearing a cache can wipe it mid-run of a class
+    /// asserting that a replay is REFUSED, and that assertion then passes because there was nothing left to
+    /// replay. Green, no longer testing the property, and intermittently, which is the worst way to find out.
+    ///
+    /// What this rule holds is exactly one thing: a test class whose OWN code lines call a
+    /// <c>ResetReplaysForTests</c> hook carries the attribute. It does not follow a base type into another
+    /// file, and it says nothing about the other doors to process-wide state — the harness (the rule above)
+    /// and the <c>ResetOidStateForTests</c> / <c>ResetSaml*ForTests</c> hooks. Deriving the door set rather
+    /// than naming one, and closing the base-class route, is #1044. This rule exists so that the collection
+    /// memberships this change needed are held by the build instead of by hand.
+    /// </summary>
+    [Fact]
+    public void EveryTestClassClearingAReplayCache_IsInTheNonParallelControllerCollection()
+    {
+        var testsRoot = Path.Combine(RepoRoot(), "SSO-Auth.Tests");
+        var resetHookUsers = new List<string>();
+        var offenders = new List<string>();
+        foreach (var src in Directory.EnumerateFiles(testsRoot, "*.cs", SearchOption.AllDirectories))
+        {
+            // Skip THIS file: it carries the scanned literals inside the rule itself, not a real use.
+            if (IsBuildOutput(src) || Path.GetFileName(src) == "ArchitectureConformanceTests.cs")
+            {
+                continue;
+            }
+
+            // CODE lines only, so prose naming the hook cannot inflate the liveness floor below while
+            // covering nothing. The [Fact]/[Theory] requirement excludes the support types that clear the
+            // same caches without being scheduled by xUnit at all — the harness clears three of them, and it
+            // is the classes CONSTRUCTING it that need the attribute (the rule above), not the file itself.
+            var lines = CodeLines(src).Select(l => l.Text).ToList();
+            var clearsAReplayCache = lines.Any(text => text.Contains("ResetReplaysForTests(", StringComparison.Ordinal));
+            var isTestClass = lines.Any(text =>
+                text.Contains("[Fact]", StringComparison.Ordinal) || text.Contains("[Theory]", StringComparison.Ordinal));
+            if (!clearsAReplayCache || !isTestClass)
+            {
+                continue;
+            }
+
+            resetHookUsers.Add(Path.GetFileName(src));
+            if (!lines.Any(text => text.Contains("[Collection(\"SSOController\")]", StringComparison.Ordinal)))
+            {
+                offenders.Add(Path.GetFileName(src));
+            }
+        }
+
+        Assert.True(
+            offenders.Count == 0,
+            "Every test class clearing a process-wide replay cache must carry [Collection(\"SSOController\")] — a class outside that collection runs in parallel WITH it, and clearing the cache under a sibling asserting a replay is refused turns that assertion green while it tests nothing (#1004). Missing in: " + string.Join(", ", offenders));
+
+        // Liveness sentinel for the ONE door this rule scans. Five classes call a ResetReplaysForTests hook
+        // today; a rename of the hooks would drop the scan to zero offenders and read as compliance, so the
+        // floor is what turns that into a red build naming this rule.
+        Assert.True(
+            resetHookUsers.Count >= 5,
+            $"The replay-reset scan matched only {resetHookUsers.Count} test classes — the ResetReplaysForTests hooks were renamed; update this rule. Matched: " + string.Join(", ", resetHookUsers));
+    }
+
     [Fact]
     public void EverySourceFile_CarriesTheSpdxHeader()
     {
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -2526,45 +2526,27 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
-    public void EveryTestClassMutatingProcessWideState_IsInTheNonParallelControllerCollection()
+    public void EveryHarnessUsingTestClass_IsInTheNonParallelControllerCollection()
     {
         // The SsoControllerHarness constructor swaps the process-wide SSOPlugin.Instance and resets the
         // OIDC/SAML static caches — two harness-based classes running in parallel therefore race each
         // other (the exact intermittent 429→400 failure that motivated this rule, #928 U4). The
         // "SSOController" collection (DisableParallelization) is the existing convention; this makes it
-        // self-enforcing: a NEW test class that mutates that state without joining the collection is
+        // self-enforcing: a NEW test class that constructs the harness without joining the collection is
         // a red build naming the file, not a flaky suite three weeks later.
-        //
-        // The harness is not the only door to that state, and keying the rule on it alone left the other one
-        // unguarded (#1004): a `*ForTests()` reset hook clears the SAME process-wide replay caches and state
-        // stores directly. A class in the serialized collection is serialized against the others IN it, never
-        // against a class outside it — xUnit runs an un-attributed class in its own implicit collection, in
-        // parallel with this one — so any class calling a reset hook must join too, or it races the harness
-        // classes that reset the same static.
         var testsRoot = Path.Combine(RepoRoot(), "SSO-Auth.Tests");
         var offenders = new List<string>();
-        var harnessUsers = 0;
-        var resetHookUsers = 0;
         foreach (var src in Directory.EnumerateFiles(testsRoot, "*.cs", SearchOption.AllDirectories))
         {
-            // Skip THIS file: it carries the scanned literals inside the rule itself, not a real use.
+            // Skip THIS file: it carries the scanned literal inside the rule itself, not a harness use.
             if (IsBuildOutput(src) || Path.GetFileName(src) == "ArchitectureConformanceTests.cs")
             {
                 continue;
             }
 
             var text = File.ReadAllText(src);
-            var isTestClass = text.Contains("[Fact]", StringComparison.Ordinal) || text.Contains("[Theory]", StringComparison.Ordinal);
-            var usesHarness = text.Contains("new SsoControllerHarness", StringComparison.Ordinal);
-            var usesResetHook = text.Contains("ForTests()", StringComparison.Ordinal);
-            if (!isTestClass || (!usesHarness && !usesResetHook))
-            {
-                continue;
-            }
-
-            harnessUsers += usesHarness ? 1 : 0;
-            resetHookUsers += usesResetHook ? 1 : 0;
-            if (!text.Contains("[Collection(\"SSOController\")]", StringComparison.Ordinal))
+            if (text.Contains("new SsoControllerHarness", StringComparison.Ordinal)
+                && !text.Contains("[Collection(\"SSOController\")]", StringComparison.Ordinal))
             {
                 offenders.Add(Path.GetFileName(src));
             }
@@ -2572,14 +2554,14 @@ public class ArchitectureConformanceTests
 
         Assert.True(
             offenders.Count == 0,
-            "Every test class that constructs SsoControllerHarness or calls a *ForTests() reset hook must carry [Collection(\"SSOController\")] (both mutate process-wide statics; classes outside the collection run in parallel with it). Missing in: " + string.Join(", ", offenders));
+            "Every test class constructing SsoControllerHarness must carry [Collection(\"SSOController\")] (the harness swaps process-wide statics; parallel classes race). Missing in: " + string.Join(", ", offenders));
 
-        // Liveness sentinel, ONE FLOOR PER SCANNED LITERAL. A single combined count is the same conjunction
-        // defect this rule exists to catch: renaming SsoControllerHarness would drop 27 classes and still
-        // clear a combined floor the reset-hook classes alone carried, leaving the scan blind while reading
-        // as live. Each literal has to prove it still matches something on its own.
-        Assert.True(harnessUsers >= 20, $"The harness scan matched only {harnessUsers} test classes — SsoControllerHarness was renamed; update this rule.");
-        Assert.True(resetHookUsers >= 3, $"The reset-hook scan matched only {resetHookUsers} test classes — the *ForTests() hooks were renamed; update this rule.");
+        // Liveness sentinel: the scan must actually see the known harness users, or a harness rename has
+        // silently blinded it.
+        Assert.True(
+            Directory.EnumerateFiles(testsRoot, "*.cs", SearchOption.AllDirectories)
+                .Count(f => !IsBuildOutput(f) && File.ReadAllText(f).Contains("new SsoControllerHarness", StringComparison.Ordinal)) >= 10,
+            "The harness-usage scan matched fewer than 10 files — SsoControllerHarness was renamed; update this rule.");
     }
 
     [Fact]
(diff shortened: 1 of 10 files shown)


iderex:
Closing back to phase 1, consistent with the re-cut recorded on #1004. The branch stays; the work it proved is listed there so none of it is re-derived.

Five rounds. Rounds 2 through 4 found false claims in doc comments and nothing in shipped behaviour; round 5, with the gap enumerations removed and the prose out of the way, was the first round that could see the delivery, and it found a fixture that passes for the wrong reason plus four acceptance criteria unmet or silently substituted. That ordering is the useful part of this PR''s history: the essays were hiding the substance, and removing them was what let the review reach it.

**What is proven and carries forward** (all on #1004):

- The forgery battery: non-vacuous positive controls, a differential `kid` control, and one forgery driven through the production back-channel entry point. That last row is not decoration, it is the only thing that catches post-build mutation of the hardened parameters, which the conformance rule is structurally blind to. Measured: `parameters.RequireSignedTokens = false;` after the builder call leaves the rule green, the build clean, and the unsigned forgery **accepted**.
- The CA5404 boundary, measured rather than assumed: three validation properties set to a `false` literal produce three analyzer errors; the same properties assigned from a non-constant build clean. The analyzer is literal-only, which is why the claim that it "fails the build wherever it is written" had to go rather than be narrowed.
- The four evasions the parameter-builder rule does not reach, recorded on **#1050** with their reproductions rather than enumerated in a comment that has been wrong every time it was written.

**What blocks a re-entry**, and it is one thing: **#1052**. A forgery row must not pass for a reason other than the one it names. The `x5c` fixture carries a SubjectPublicKeyInfo where RFC 7515 §4.1.6 requires a DER certificate, so a verifier that *did* consult header key material cannot parse it and rejects anyway, proven by installing a spec-correct resolver in production and watching all 28 rows stay green. #1052 owns the technique and its reach across the battery.

Also split out: **#1051** (redirect-URI negatives, decide whether this plugin ever compares one, or only emits it) and **#1029** (the `kid` allowlist, which already existed and which this issue stops claiming).

Issue #1031, opened 2026-07-27
Documented net9.0 build override rewrites the committed lockfiles


iderex:
## Problem

`CLAUDE.md`'s "Local gate (both TFMs, warnaserror, full suite)" step documents the net9.0 build as:

```sh
dotnet build SSO-Auth.Tests/SSO-Auth.Tests.csproj -f net9.0 -p:JellyfinVersion=10.11.11 --warnaserror --nologo -v q
```

`SSO-Auth/SSO-Auth.csproj` now sets that property itself, conditioned per target framework:

```xml
<JellyfinVersion Condition="'$(JellyfinVersion)' == '' and '$(TargetFramework)' == 'net9.0'">10.11.11</JellyfinVersion>
<JellyfinVersion Condition="'$(JellyfinVersion)' == '' and '$(TargetFramework)' == 'net10.0'">12.0.0-rc2</JellyfinVersion>
```

Both conditions require the property to be EMPTY. An explicit `-p:JellyfinVersion=10.11.11` makes both false, so the override wins everywhere it is passed, and restore then rewrites **both** committed `packages.lock.json` files, replacing the `12.0.0-rc2` entries with `10.11.11`.

Observed while working #1004: after one documented net9.0 build, `git status` showed `SSO-Auth/packages.lock.json` and `SSO-Auth.Tests/packages.lock.json` modified, with `Jellyfin.Controller` / `Jellyfin.Model` / `Jellyfin.Common` / `Jellyfin.Naming` / `Jellyfin.MediaEncoding.Keyframes` and `BitFaster.Caching` all moved off the pinned versions. Committing that would fail CI's `--locked-mode` restore (NU1004). Dropping the override builds cleanly on both TFMs and leaves the lockfiles untouched, verified for net9.0 and net10.0 with `--warnaserror`, 0 warnings, full suite green.

## Why this matters

The documented command is the one an agent or a new contributor runs first, and its damage is silent: the build is green, the lockfile corruption shows up only in `git status`, and it is easy to sweep into an unrelated commit. The failure then surfaces in CI as a restore error that looks unrelated to the change.

The instruction was correct when written; the csproj conditioning made it obsolete without anything failing to signal that.

## Scope

- Update the `CLAUDE.md` execution playbook: drop `-p:JellyfinVersion=10.11.11` from the net9.0 line, and remove the now-inverted "the JellyfinVersion override IS needed (TFM conditional)" note. The net10.0 line already passes no override.
[...]

Issue #402, opened 2026-07-16
Evaluate fuzz testing for the SAML/OIDC parsing surface


iderex:
Scorecard alert #36 (Fuzzing, score 0). Unlike most Scorecard posture checks this one has real security value here: the login path hand-parses attacker-supplied XML (SAML responses) and JWT/JSON (OIDC), exactly the surface fuzzing pays off on. The FsCheck property suite covers invariants but is not coverage-guided.

**Acceptance**
- [ ] Evaluate ClusterFuzzLite (Scorecard-recognized, runs in CI) vs. SharpFuzz for the .NET target; pick one or document why neither fits.
- [ ] If adopted: fuzz targets for SamlResponse parsing/validation and the id_token validation entry, seeded from the existing test vectors, running as a scheduled (not per-PR) workflow.
- [ ] Alert #36 closes or is re-dismissed with the documented outcome.

iderex:
Relates to two existing issues. #174 already proposes a weekly SharpFuzz job over the SAML parse/validate core, i.e. it pre-commits to the very tool this issue is meant to select, so this evaluation should gate #174: pick the tool here first (ClusterFuzzLite vs SharpFuzz), then either fold #174 in as the implementation task or re-scope it to whatever we choose, rather than running both as independent tracks. This issue and #403 are also the two concrete carve-outs of the #319 Scorecard-baseline triage (alert #36 Fuzzing here, alert #35 CII-Best-Practices in #403); #319 stays the umbrella tracking the fix-or-document decision for all five baseline alerts.

Issue #174, opened 2026-07-12
Weekly SharpFuzz job on the SAML parsing/validation core


iderex:
Add a scheduled SharpFuzz coverage-guided fuzzing job over the SAML parse/validate entry points (OSS-Fuzz has no .NET support, so self-hosted weekly). Complements the FsCheck property tests by finding crashers and unexpected accepts on malformed XML. Any finding becomes its own security finding, not an in-suite fix.

Acceptance criteria:
- [ ] SharpFuzz harness over the SAML response parse/validate path; scheduled weekly, non-blocking.
- [ ] Corpus seeded from the SamlTestFactory fixtures.
- [ ] Crashers or fail-open inputs filed as security findings (GHSA path if exploitable).

Refs #126 (property tests) as the sibling technique.

iderex:
Done on 4.2 (#601): weekly SharpFuzz workflow over the SAML parse surface. Ships with 4.2.0.

PR #1032, opened 2026-07-27
Refuse repeated JSON property names and prove the readers agree [#1005]

### LINKED ISSUE #1005: Reject duplicate JSON keys and prove every input route validates identically
## Problem

Two related gaps, both of the "two consumers disagree on which value is authoritative" family:

1. **No duplicate-key rejection at the untrusted-JSON boundaries.** The plugin parses attacker-influenced or IdP-controlled JSON in `SSO-Auth/Api/Oidc/OidcDiscoveryReader.cs` (discovery document), in the JWKS path via `SSO-Auth/Api/Oidc/OidcSignatureKeys.cs`, and in the token/userinfo responses. `System.Text.Json` on `net9.0` silently takes the last occurrence of a duplicated property, so a validator that reads the first value and a consumer that reads the last can be made to disagree, the classic duplicate-key bypass. Nothing in the repo rejects a duplicated property, and no test asserts it.

2. **No test that every input route into a security-relevant field applies identical validation.** Fields such as the provider id, the `state`/`nonce`, the redirect target, the ACS URL, the role/folder claim names and the linking identifier can arrive through more than one route (query string, form body, JSON body, plugin configuration, SAML attributes, OIDC claims). Validation lives next to each route rather than at a single choke point, and there is no conformance test asserting that route A and route B run the same validator over the same field.

Related: the plugin has its own trusted-network concept (`SSO-Auth/Api/Net/IpAddressClassifier.cs`, `SSO-Auth/Api/Net/SsoHttp.cs`). There is no negative test asserting a spoofed LAN source address or a forged `X-Forwarded-For` / `Forwarded` header never results in an authorized or auto-provisioned session.

## Why now

CVE-2026-35033 in the Jellyfin host is the textbook case: the `level` field is regex-validated on the query-string route, but the same field set through the semicolon-separated `params` route in `StreamingHelpers.ParseParams()` is not, so an attacker-controlled string reaches the FFmpeg command line and yields remote code execution (Sonar, 2026-06-02, https://www.sonarsource.com/blog/jellyfin-remote-code-execution/). CVE-2025-32012 in the same host made LAN trust the bypass: `/System/Restart` trusted any source IP that looked local, and an unauthenticated attacker could spoof one (https://www.cvedetails.com/vulnerability-list/vendor_id-22884/Jellyfin.html). On the JSON side, .NET 10 shipped `JsonSerializerOptions.Strict`, whose headline protection is rejecting duplicate properties precisely because "validator sees the first value, deserializer uses the last" is a real auth-bypass pattern (Duende, 2026-04-30, https://duendesoftware.com/blog/20260430-harden-your-dotnet-json-deserialization; .NET 10 announcement, https://devblogs.microsoft.com/dotnet/announcing-dotnet-10/). The plugin multi-targets `net9.0;net10.0`, so the preset is available on one TFM and must be emulated on the other.

## Scope

- Add a duplicate-property pre-check used by every untrusted-JSON boundary: a shared helper (for example `SSO-Auth/Api/Http/StrictJson.cs`) that walks the payload with `Utf8JsonReader` and fails closed on a repeated property name at any depth, before deserialization.
  - On `net10.0`, additionally configure `JsonSerializerOptions.Strict` (or `AllowDuplicateProperties = false` plus case-sensitive matching) under a `#if NET10_0_OR_GREATER` guard, and assert both TFMs reach the same reject/accept decision.
  - Wire the helper into `OidcDiscoveryReader.cs`, the JWKS read in `OidcSignatureKeys.cs`, and the token/userinfo response handling. The token and UserInfo halves of that wiring were later decided against, in #1069: the plugin does not read either body, so screening them would mean adding a second reader beside the object the library already binds, and the id_token inside the envelope has to verify against the already-screened JWKS whichever occurrence a parser resolves to. The UserInfo divergence that was real is the repeated role-claim name, and it was closed in #1040 at the claim set rather than at the transport.
- Add `SSO-Auth.Tests/Http/StrictJsonTests.cs`: duplicate key at the top level, duplicate nested inside an object, duplicate inside an array element, duplicate differing only in case, duplicate `kid` inside a JWKS entry, duplicate `iss` in a discovery document, each asserted rejected, with a matching accept case for the valid document.
- Add a conformance test `UntrustedJson_IsParsedOnlyThroughTheStrictHelper` to `SSO-Auth.Tests/ArchitectureConformanceTests.cs`: no `JsonSerializer.Deserialize` / `JsonDocument.Parse` call on an untrusted payload outside the helper.
- Add a validation-consistency conformance test `SecurityFields_ShareOneValidatorAcrossEveryInputRoute`: for each security-relevant field, enumerate its input routes and assert every route calls the same named validator. Start with the provider id, the OIDC `state` and `nonce`, the redirect target, the SAML ACS/recipient URL, and the linking identifier.
- Add per-field paired tests (`<Field>_IsValidatedIdenticallyOnEveryRoute`) feeding the same hostile value through each route and asserting identical rejection, not merely "both reject somewhere", but the same validator and the same outcome.
- Add `SSO-Auth.Tests/Net/SpoofedNetworkTrustTests.cs`: a request with a forged private source address, a forged `X-Forwarded-For`, a forged `Forwarded`, and a chain of both must never authorize, auto-provision, or skip a rate limit; assert the trusted-network signal is only ever used to relax non-security behaviour, never as an authentication substitute.
- Add a duplicate-key seed to `SSO-Auth.Fuzz/corpus/discovery` and `SSO-Auth.Fuzz/corpus/idtoken` so the existing SharpFuzz targets keep the vector as a permanent regression input.

## Acceptance criteria

- [ ] A JSON document with a duplicated property is rejected at every untrusted-JSON boundary, on both `net9.0` and `net10.0`, with a test per boundary.
- [ ] The `net9.0` pre-check and the `net10.0` `Strict` configuration are proven to produce the same decision on the same corpus of documents.
- [ ] `UntrustedJson_IsParsedOnlyThroughTheStrictHelper` passes and fails when a raw `JsonDocument.Parse` on an untrusted payload is reintroduced.
- [ ] Every field in the starting list has a named test asserting identical validation on every input route, and `SecurityFields_ShareOneValidatorAcrossEveryInputRoute` enumerates the routes rather than hard-coding a count.
- [ ] A spoofed LAN address or forged forwarded-for header never yields an authorized session, an auto-provisioned account, or a bypassed rate limit, one test per signal.
- [ ] New corpus seeds present and the fuzz smoke run green over them.
- [ ] Full suite green on both TFMs; test project built with `--warnaserror` before push.

## Notes

- Shipping a newer `System.Text.Json` into the .NET 9 Jellyfin host to get `AllowDuplicateProperties` is explicitly out of scope, that is the same shape as the OidcClient ABI outage. The `net9.0` path uses the `Utf8JsonReader` pre-check only.
- The validation-consistency test is a ratchet: it proves routes agree, not that the validator itself is correct. Field-level correctness stays with the existing per-field tests.
- Out of scope: the SAML XML wrapping vectors (#1003) and the JWT algorithm/claim battery (#1004). This issue is the JSON and cross-route half of the same break class.

iderex:
Closes #1005.

Implements the [Phase 1 plan on the issue](https://github.com/iderex/jellyfin-plugin-sso/issues/1005#issuecomment-5093466681). Round 1 of the adversarial gate returned NEEDS_IMPROVEMENT; round 2 returned **FAIL**. Both verdicts are on file as `round-01.md` and `round-02.md` under `pull-requests/jellyfin-plugin-sso/pr-1032-duplicate-json-keys/`, and every finding in each is dispositioned at the bottom of this body.

**My disposition of round 2's deciding finding is REMOVE.** The role-claim half leaves this change entirely and becomes #1053. The discovery half stays, with the availability defect round 2 also confirmed fixed at its root. What is left is one control at one call site.

## Why the role-claim half left

Round 2's headline, CONFIRMED by four lenses and the reviewer's own probe: `OidcRoleExtractor.cs:73` treated the screen's `Unreadable` verdict as **proceed** and fell through to Newtonsoft, which grants the attacker's last-occurrence roles, while `OidcDiscoveryReader` treats the same verdict as **refuse**. In the reviewer's words, _"the same `Unreadable` verdict is handled fail-closed in one caller and fail-open in the other, and that asymmetry IS the vulnerability."_ A token the screen cannot read but Newtonsoft can disabled the screen on the one path that decides privileges.

That asymmetry was a deliberate round-1 choice, taken to protect availability: grammars exist that Newtonsoft accepts and the screen cannot read, and refusing them would cost affected users every role. It is defensible on its own terms and wrong against this repo's first directive. Resolving it means deciding what `Unreadable` **means** on a privilege path, which is a design question and not a patch, so it is settled on #1053 before any more code is written.

`SSO-Auth/Api/Oidc/OidcRoleExtractor.cs` and `SSO-Auth.Tests/Oidc/OidcRoleExtractorTests.cs` are back to their pre-change state: `git diff origin/main...HEAD --` over both is **empty**. The conformance scan still governs the extractor's read, now as a declared **unscreened** read carrying #1053, a rule that had simply forgotten the file would look identical to one that decided about it.

## What the measurements said

A repeated JSON property name is accepted by **every parser this plugin depends on**, and each of them silently keeps the **last** occurrence:

| reader                                                     | value taken | note                                                |
| ---------------------------------------------------------- | ----------- | --------------------------------------------------- |
| Newtonsoft `JObject.Parse`                                 | last        | property count 1, it replaces rather than appends  |
| Newtonsoft `JsonConvert.DeserializeObject<IDictionary<…>>` | last        |                                                     |
| `JsonWebToken` header / payload                            | last        | `{"alg":"RS256","alg":"none"}` reads back as `none` |
| Duende `JsonWebKeySet`                                     | last        | a duplicated `kid` reads back as the last           |

Those four rows are pinned by `ParserDuplicateKeyPosture_IsPinned`, and the claim is no wider than the rows: they pin what the version this repo **declares** does, and the host resolves assembly loads for a plugin. System.Text.Json is deliberately absent, it ships in the runtime rather than as a package reference, so there is no pin for a row to track, and this test process loads a 10.x on **both** legs. What replaces it is structural: the screen owns no `JsonSerializerOptions`, which `UntrustedJsonConformanceTests` enforces.

Live, and the reason this ships: a discovery document with a duplicated `issuer` returns `IsError=false` and resolves to the **last** value, silently re-pointing the issuer anchor the login binds itself to, with no error raised anywhere.

## What ships

`StrictJson.Inspect` is a raw `Utf8JsonReader` walk over the **root object**, reporting a repeat only for the members the caller names. It is deliberately not a `JsonSerializerOptions` setting: the plugin binds the **host's** System.Text.Json, and the `Strict` preset that would express this exists only on the .NET 10 side, referencing it fails the net9.0 build with CS0117. That was measured in round 1, not assumed.

**It answers three ways, not one.** `Clean` / `Repeated` / `Unreadable`. A `bool` had to call a byte-order mark, a document nested past the reader's cap and two top-level values "repeats a property name", a false accusation that refuses a login and sends the admin hunting a duplicate that is not there. The discovery read refuses `Unreadable`, fail-closed, under its own separate warning.

**It screens only the members a reader indexes, the round-2 availability fix.** Round 2 CONFIRMED that the first cut refused every login for a provider whose document repeated a top-level member **nothing reads**: `scopes_supported`, a vendor extension. The list is now derived from the reads themselves rather than from the OpenID Discovery spec:

| member                                           | what takes a value out of it                                             |
| ------------------------------------------------ | ------------------------------------------------------------------------ |
| `code_challenge_methods_supported`               | `PkceDiscovery.SupportsS256`, which names it literally in its own source |
| `authorization_response_iss_parameter_supported` | `OidcResponseIssuer.DiscoveryAdvertisesResponseIssuer`, likewise         |
| `issuer`                                         | `discovery.Issuer` → `ProviderInformation.IssuerName`                    |
| `jwks_uri`                                       | `discovery.KeySet`, the library fetches the key set from it             |
| `authorization_endpoint`                         | `discovery.AuthorizeEndpoint`                                            |
| `pushed_authorization_request_endpoint`          | `discovery.PushedAuthorizationRequestEndpoint`                           |
| `token_endpoint`                                 | `discovery.TokenEndpoint`                                                |
| `end_session_endpoint`                           | `discovery.EndSessionEndpoint`                                           |
| `userinfo_endpoint`                              | `discovery.UserInfoEndpoint`                                             |
| `token_endpoint_auth_methods_supported`          | `discovery.TokenEndPointAuthenticationMethods`                           |

Ten, and the list is **closed**: nothing downstream reaches past `OidcDiscoveryResult`, whose only components are those facts and that metadata, so the three call sites (`OidcLoginService` twice, `ProviderConnectionTester`) add nothing. The wire spelling of all ten was checked against the constant table in the shipped `Duende.IdentityModel` assembly rather than typed from memory. The field-to-member mapping belongs to the library and not to this repo, so it is pinned instead of restated: `ReadAsync_EveryScreenedMember_IsObservedByThisRead` serves every member once and asserts each is observed, and `ReadAsync_RepeatOfAnyScreenedMember_ReturnsUnavailable_NamingIt` repeats each of the ten in **that same document** and asserts the read is refused. The two axes round 2 said were never crossed are crossed there: one document, one variable.

**The verdict names the member, and the operator is told which.** `Verdict.Repeated` carries the repeated name, the warning uses it, and the per-member theory asserts a single `Warning` entry naming that member. The name is safe to log unsanitized as a property of the screen rather than of the document, `Inspect` compares the document's names only against the caller's list and reports nothing else, so what reaches the log is one of the reader's own constants.

**Two more round-2 defects fall out of the rescope rather than being patched around it.** The scope walk is gone, and with it the `IsRead` rule that treated any object at stack depth 0 as the root, under which an array-rooted document was screened against a path it does not contain. The walk now keys on `Utf8JsonReader.CurrentDepth == 1`, a depth no array-rooted document has, and `[{"issuer":"a","issuer":"b"}]` is a fixture. And `Inspect`'s "never throws" no longer rests on a structural argument about `Peek`/`Pop`: there is no stack, and the documented claim is narrowed to `JsonException`.

**Named conformance rule: `UntrustedJson_IsParsedOnlyThroughTheStrictHelper`.** Every JSON read seam its own predicate matches, under `SSO-Auth/`, must be **declared**, gated by a file that demonstrably calls the screen, allowlisted as trusted by exact repo-relative path, or listed as an unscreened provider-supplied read with the issue that owns the decision. Round 2 was right that its class text claimed to govern every provider-JSON read while walking past parsers constructed directly; the claim is now stated at the width it holds, and names `DuplicateJsonKeyPostureTests` as what covers that surface instead.

The `#402`/`#174` fuzz harness drives `OidcDiscoveryReader.ScreenIndexedMembers`, the production entry point, so it screens the member list the login screens rather than a copy that can drift.

## Weakenings run, and what each proved

Every new guard was shown failing on a deliberately weakened build. The patched region was printed before each run, an unapplied weakening comes back green and is indistinguishable from a caught one, and the harness aborts rather than reporting a run whose patch did not land. All were reverted; `git status --porcelain` was empty before and after each. Counts are both TFMs (half is one leg).

| #   | weakening                                                                                | result                                                                                                                                        |
| --- | ---------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| W1  | a repeat no longer refuses (`Repeated` → `Clean`)                                         | **28 red**, the three screen fixtures, all ten per-member refusals, and the opposing-value duplicated-`issuer` case                           |
| W2  | the member filter dropped, every root member screened again, the shipped defect restored | **14 red** - the whole must-not-catch set: `scopes_supported`, `response_types_supported`, the vendor extension, and the empty-list reach test |
| W3  | the root-object depth rule widened to depth ≤ 3                                           | **8 red** - the nested object, the array element, the array-rooted document, and the `mtls_endpoint_aliases` discovery fixture                 |
| W4  | the verdict stops naming the repeated member                                              | **26 red** - the three screen fixtures plus all ten reader warnings                                                                            |
| W5  | `jwks_uri` dropped off the screened-member list                                           | **2 red**, its per-member refusal, and only that one                                                                                         |
| W6  | the extractor's read stops being declared in the conformance scan                         | **2 red**, the message naming `SSO-Auth/Api/Oidc/OidcRoleExtractor.cs (JsonConvert.DeserializeObject)`                                         |

## Gate

- `dotnet build --warnaserror`, Debug **and** Release `--no-incremental`: **0 warnings, 0 errors**, both TFMs. `SSO-Auth.Fuzz` Release: 0 warnings, 0 errors.
- Tests **2251 per TFM (4502 total), 0 failed**, identical on net9.0 and net10.0. At `f982908` it was 2259/4518; removing the role-claim half alone gives 4488, measured on that commit in isolation, and the rescope adds 14 net.
- **Lockfiles untouched**, per #1031, `-p:JellyfinVersion=…` silently rewrites both and CI's `--locked-mode` would reject them.
- Prettier: `CHANGELOG.md` is the only `.md` touched, and passes `prettier --check`.

## Scope, measured

**970 insertions / 5 deletions across 8 files**, 199 production, 747 test, 17 CHANGELOG, 7 fuzz harness.

That is **6 insertions below the 976 at `f982908`**, which is not the substantial shrink removing a whole half of the change would suggest. I am recording that rather than trimming justification prose to hit a number. The accounting:

| change                                                                                                      | insertions |
| ----------------------------------------------------------------------------------------------------------- | ---------- |
| `OidcRoleExtractor.cs` removed from the diff                                                                | −28        |
| `OidcRoleExtractorTests.cs` removed from the diff                                                           | −84        |
| `StrictJson.cs`, scope walk deleted                                                                         | −25        |
| `CHANGELOG.md`, one-path control                                                                            | −1         |
| `OidcDiscoveryReader.cs`, the member list and the screening entry point                                     | +36        |
| `OidcDiscoveryReaderTests.cs`, the member table, the mapping pin, the per-member and must-not-catch theories | +67        |
| `UntrustedJsonConformanceTests.cs`, the unscreened-read list and the narrowed claim                         | +19        |
| `StrictJsonTests.cs`, rewritten for the member contract                                                     | +7         |
| `DuplicateJsonKeyPostureTests.cs`, driven through the production entry point                                | +2         |
| `SSO-Auth.Fuzz/Program.cs`                                                                                  | +1         |
| **net**                                                                                                     | **−6**     |

The removal freed 112 insertions; the derivation evidence the rescope needs consumed them. If a materially smaller PR is wanted, the lever is the conformance scan, 234 insertions, already slated to fold into `ArchitectureConformanceTests` as #1037, and that is a scope decision I have not taken on my own.

## Decision: I authorise the ~400-line overrun for this PR

Recorded here rather than left to the gate to rediscover, and restated at the current size.

975 changed lines, 199 production, 747 test, 29 CHANGELOG and fuzz harness. My reasoning is unchanged from when I first authorised it: the production half is comfortably inside the target, and the test lines are the proof that the gap was real and is closed. Splitting them off would merge a security guard without the evidence that it guards anything, the inert-half split refuted elsewhere in my process record the same week. The rule exists to keep changes reviewable, and a guard separated from its proof is less reviewable, not more.

Dimension 7 should read this as an authorised overrun with its decision recorded, not as an open finding.

## Round-2 dispositions

Every finding in `round-02.md` gets one of the documented dispositions.

### REMOVE, the role-claim half leaves this change

| dim       | finding                                                                                                                                                                                                          | where it went                                                                                                                                                    |
| --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1, 24     | `Unreadable` handled fail-**open** in `OidcRoleExtractor` and fail-**closed** in `OidcDiscoveryReader`; that asymmetry is the vulnerability                                                                       | #1053. The extractor is byte-identical to `origin/main`; there is now one caller, so there is no asymmetry left to hold                                           |
| 2         | the verdict is order-dependent on attacker-authored bytes on the role path                                                                                                                                       | the same removal, no screen runs there                                                                                                                          |
| 3         | the privilege-bearing half does not deliver; the `Unreadable` unhappy path is the gap                                                                                                                            | #1053, whose first contract item is what `Unreadable` must mean on a privilege path                                                                              |
| 9         | two security comments assert what the code does not hold: _"falls through to the parse, which fails closed on its own"_ and _"the call sits inside this try so the #216 never-throw guarantee covers the screen"_ | both were in `OidcRoleExtractor` and are gone with it. Correct on the finding, and exactly why the claim had to go rather than the guard be strengthened          |
| 10 (part) | a repeated sibling in the role claim wipes the whole role set, total login denial under a role allow-list                                                                                                        | #1053, contract item 3 (scope granularity on the claim)                                                                                                          |
| 19 (part) | the role-claim refusal logs nothing at all                                                                                                                                                                       | #1053; the logger question itself stays on #1042                                                                                                                 |

### FIX

| dim        | finding                                                                                                                                                                                                              | what changed                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
| ---------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 10         | a repeated **top-level** discovery member no reader opens (`scopes_supported`, vendor members) refuses every login for that provider                                                                                  | the screen takes the indexed-member list; `scopes_supported`, `response_types_supported` and a vendor extension are must-not-catch fixtures at both the screen and the reader (W2)                                                                                                                                                                                                                                                                                                         |
| 2, 7       | `IsRead` treats every object at stack depth 0 as "the root", so array-rooted documents are screened against a path they do not contain; and the scope granularity is per-object while its justification is per-member | the scope walk is deleted. Granularity is now per-member, which is what the justification always said, and the depth rule is `CurrentDepth == 1` with an array-rooted fixture (W3)                                                                                                                                                                                                                                                                                                          |
| 19         | `Verdict.Repeated` carries no member name, so an admin cannot tell which member repeated                                                                                                                             | the verdict carries it, the warning names it, and the per-member theory asserts a single `Warning` entry naming that member (W4)                                                                                                                                                                                                                                                                                                                                                           |
| 11         | the conformance scan walks past parsers constructed directly while its text claims it governs every provider-JSON read                                                                                                | the claim is weakened to what it holds, and the class doc now names what the scan does **not** see and what covers that surface instead. Measured while fixing: `GetFromJsonAsync`/`ReadFromJsonAsync` appear **nowhere** under `SSO-Auth/` today, so that spelling is a future hazard rather than a present hole, and `new JsonWebToken(` / `new JsonWebKeySet(` appear in **4** files there - `OidcIdTokenAcr.cs`, `OidcIdTokenAuthTime.cs`, `OidcIdTokenSid.cs`, `OidcResponseIssuer.cs` |
| 13         | `WhereTheParserFamiliesDisagree_TheDiscoveryReadIsRefused` never calls the discovery reader                                                                                                                           | renamed to `…_TheScreenReportsUnreadable`, which is what it asserts, and its comment now says explicitly that what the reader does with the verdict is a different file's test                                                                                                                                                                                                                                                                                                             |
| 13         | the two axes are never crossed, and the uncrossed cell is the live escalation                                                                                                                                        | crossed: the per-member refusal theory repeats each screened member in the exact document the mapping test proves readable, so only the repeat varies                                                                                                                                                                                                                                                                                                                                      |
| 13         | the fuzz addition covers only the empty-path shape                                                                                                                                                                   | it drives `ScreenIndexedMembers`, the production entry point, so it fuzzes the shipped member list                                                                                                                                                                                                                                                                                                                                                                                         |
| 14         | the CHANGELOG's role-claim sentence is materially false as shipped; `ExtractRoles`'s `<returns>` states the refusal unconditionally                                                                                   | the CHANGELOG describes a one-path control and says so; the `<returns>` went back with the code                                                                                                                                                                                                                                                                                                                                                                                            |
| 14, 19     | no operator-facing signal for the refusal                                                                                                                                                                            | the warning names the repeated member, asserted at the reader                                                                                                                                                                                                                                                                                                                                                                                                                             |
| 18         | behaviour compat regresses: providers whose documents were read before are now refused, unjustified for the over-refusals                                                                                            | the over-refusals are gone. What is still refused is a document that repeats a member this login acts on, which is the defect being fixed                                                                                                                                                                                                                                                                                                                                                  |
| 24         | `Inspect`'s "never throws" holds only structurally, `Peek`/`Pop` would raise `InvalidOperationException`, which no caller's `catch (JsonException)` takes                                                            | there is no stack and no `Peek`/`Pop`. The documented claim is narrowed to "never throws a `JsonException`"                                                                                                                                                                                                                                                                                                                                                                                |
| 22 (part)  | the two discovery refusals are logged apart - good design, asserted nowhere, so a refactor collapsing them passes green                                                                                               | the repeat refusal is asserted as a **single** `Warning` entry naming the member, so collapsing it into the generic unreadable-document warning fails (W4)                                                                                                                                                                                                                                                                                                                                 |
| 4          | against the change's own stated deliverable, the role-claim promise is unmet                                                                                                                                         | the deliverable is restated to the one it keeps, in the CHANGELOG and here                                                                                                                                                                                                                                                                                                                                                                                                                |

### DEFER - each its own top-level issue

| dim | finding                                                                                                                                                     | issue                                                                       |
| --- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
| 6   | a third parser as a screen is the known-weak answer to a parser differential; reading the facts off the tokens the screen already walked is the robust shape | **#1054** (new)                                                             |
| 7   | three full parses of the same bytes per challenge                                                                                                           | #1041, which already owns the parse count and the missing response-size cap |
| 22  | the role path discharges diagnosis onto a server-side trail that does not exist                                                                              | #1042, which owns threading a logger through `ExtractRoles` / `ScanClaims`  |

### DECLINE, reasoned

- **dim 6, the security reading.** True that a validating parser separate from the consuming parser is the weak shape, and on this path it has no exploitable consequence: a grammar the screen cannot tokenize returns `Unreadable` and the reader **refuses**, and such a body cannot reach the screen anyway, because `IsError == false` already means the library parsed it with System.Text.Json under the same defaults. Both directions fail closed. The simplification is real and is #1054; the defect is not.
- **dim 13, the `Unreadable ⇒ Unavailable` branch has no test and survives deletion.** Correct, and it is unreachable end to end for the reason above, no served document can drive it, so no fixture can either. It is kept as the fail-closed side of a divergence the library's own parse prevents today rather than deleted, and the comment beside it says exactly that instead of implying it is exercised. The verdict itself is covered directly in `StrictJsonTests`.
- **dim 25, no flag, no kill switch, no staged exposure.** The blast radius this asked to be gated behind a flag was the over-refusals, and those are gone: only a repeat of a member this login acts on refuses. What remains is a genuinely ambiguous document, which is the defect being fixed rather than a regression to stage. The repo has no feature-flag mechanism, and adding one whose only use is switching off a login-path security control is a liability of its own.
- **dim 8, `Encoding.UTF8.GetBytes` copies each document once more per login.** The lens itself called this noted, not a finding, and the rescope reduced the surrounding work: `GetString()` is now called only for root-object property names.
- **dim 7, the ~400-LOC overrun.** An authorised overrun with the decision recorded above, not an open finding.

### PASS / n/a in round 2, unchanged here

5 (product fit), 12 (audit integrity, SSO carries no tamper-evident trail), 15 (i18n, the messages are `ILogger` operator text), 16 (privacy, only the provider name is logged, sanitized inline; neither the document nor any value it carries is), 17 (supply chain, no dependency or lockfile change), 20 (accessibility), 21 (licensing, SPDX `GPL-3.0-only` on every new file), 23 (concurrency, `StrictJson` is a pure static with no shared mutable state).

Round 2 also checked for reviewer-directed text in the tree and found none. That remains true.

## Round-1 dispositions, restated to the current state

The round-1 table is in `round-01.md`. What survives into this state:

- **FIX, still standing**: the three-valued verdict (dims 2, 7, 19, 24); the read-seam regex's `\w*` suffixes with a must-catch fixture per evasion spelling (dim 1); the cross-family oracle in `DuplicateJsonKeyPostureTests`, now total (dim 13); the `CodeOf(...)` comment-stripping in the `ungated` arm (dim 9); the CHANGELOG entry (dim 14); the fuzz harness modelling production (dims 3, 11); the numbered follow-up #1037 (dim 4); the narrowed `GatedJsonReads` claim (dim 2); the narrowed parser-posture table (dim 7).
- **Moot with the role-claim removal**: the `ExtractRoles` `<returns>` documentation and the screen-inside-the-`try` placement (dims 9, 2, 24), both reverted with the code they described.
- **DEFER, unchanged**: #1040 (UserInfo repeating the claim name), #1041 (response-size cap and parse count), #1042 (logger and audit record on the role path), #1043 (replacing the hand-rolled walk once net9.0 is dropped).
- **DECLINE, unchanged**: the `[CallerFilePath]` scan root (dim 11); the transitive `using`s in the test file (dim 17).

## Re-homed as follow-ups

Filed rather than absorbed, so this PR stays inside #1005's scope:

- #1033, the ASP.NET `[FromBody]` duplicate-key posture (config import, SAML metadata import, `AuthResponse`)
- #1034, the non-JSON cross-route consistency for the provider id, `state`, the redirect target, the ACS URL and the linking identifier
- #1035, the trusted-proxy / spoofed `X-Forwarded-For` battery
- #1036, duplicate-key seeds for the fuzz corpus
- #1037, folding this PR's conformance rule into `ArchitectureConformanceTests` once #1030 has merged
- #1040, #1041, #1042, #1043, the round-1 deferrals
- #1053, what `Unreadable` means on a privilege path, and the role-claim screen that depends on it
- #1054, reading the discovery facts off the tokens the screen already walks

## Not done

- Still a **draft**. Not marked ready, not merged, no tags pushed, and nothing posted to this thread.
- Every weakening above was reverted; the tree is clean.


iderex:
**Decision (recorded 2026-07-27): I authorise the ~400-line overrun for this PR.** Recorded here rather than left to the gate to
rediscover.

688 changed lines - **120 production, 568 test**. My reasoning:
the production half is comfortably inside the target; the 568 test lines are
the proof that the gap was real and is closed. Splitting them off would merge
a security guard without the evidence that it guards anything - which is the inert-half split I had refuted elsewhere in my process record the same day. The rule exists to keep changes reviewable, and a guard separated from
its proof is less reviewable, not more.

Dimension 7 should therefore read this as an authorised overrun with its
decision recorded, not as an open finding.

For the round that reviews this: the finding that justifies the volume is
that **the posture did not already hold**. Every parser in the dependency set
accepts a repeated key and silently keeps the last occurrence, so a discovery
document carrying a duplicated `issuer` returns `IsError=false` and re-points
the issuer anchor the login binds to, with no error raised anywhere.

iderex:
Closing back to phase 1 after three rounds. The re-cut is recorded on #1005; the branch stays for what it proved.

Round 3 moved this from FAIL to NEEDS_IMPROVEMENT, removing the role-claim half fixed the fail-open, and that was the right call. What remains is not patchable, because all three findings are about **where and when** the screen runs rather than what it does: the refusal fires after the library has already dereferenced the duplicated `jwks_uri`, the JWKS document is never screened while the CHANGELOG sells key protection, and the screen itself throws an uncaught `InvalidOperationException` on thirteen bytes.

The measurements carry forward and are listed on #1005 so nothing is re-derived, the parser-differential result, the ten-member enumeration, and the fact that the mechanism survived escape-spelling, case-smuggling and a 50k-case differential fuzz. The mechanism is sound. Its placement is what fails.

One thing I am carrying out of here as a process correction rather than a code one. Round 2 predicted the uncaught throw precisely, and my instruction in response was to weaken the doc comment to "never throws a `JsonException`". That is the right remedy when a guard over-claims its coverage; it is the wrong one when the claim points at a live defect, where it converts a bug report into a documentation change. Round 3 then found the defect the documentation had absorbed. The distinction belongs in the chapter that carries the rule, not in this PR.

Issue #1179, opened 2026-08-03
Route the OIDC backchannel through the private-permitted transport only for providers that opted in


iderex:
Part of #1058.

## Claim

The OIDC backchannel (discovery, JWKS, token, userinfo, back-channel logout) connects to a private-network address **only** for a provider whose `AllowPrivateNetworkAddresses` is set, every other provider, and every other outbound caller, keeps the full guard unchanged.

This is the piece that actually fixes the reported failure: an IdP on a `10.x.x.x` address behind a reverse proxy becomes reachable without making it publicly routable.

Depends on the classifier tier and the config flag (the other two children of #1058).

## Where it lives

- `SSO-Auth/Api/Net/SsoHttp.cs`, today there is one hardened handler and one named client, `OutboundClientName = "sso-outbound"` (line 31), whose `ConnectCallback` (line 80) is stateless and has no provider context. That is why the reporter's two insecure toggles changed nothing: they relax `DiscoveryPolicy`, not the transport.
- `SSO-Auth/SsoOnlyServiceRegistrator.cs:42`, the single `AddHttpClient(SsoHttp.OutboundClientName)` registration with `SetHandlerLifetime(Timeout.InfiniteTimeSpan)`.
- Call sites that must select per provider:
  - `SSO-Auth/Api/Oidc/OidcDiscoveryReader.cs:63`
  - `SSO-Auth/Api/Flows/OidcLoginService.cs:691` (back-channel logout `HttpClientFactory`)
  - `SSO-Auth/Api/Flows/OidcLoginService.cs:738` (login `HttpClientFactory`)
- Must NOT be reached by the relaxation:
  - `SSO-Auth/Api/Avatar/AvatarService.cs:377`, builds its own client on `CreateHardenedHandler()`; stays strict.
  - `SSO-Auth/Api/Http/SamlMetadataImporter.cs:66`, resolves the named client; must keep resolving the strict one.

## Shape

Add a second registered handler/client name (e.g. `sso-outbound-private`) whose connect callback uses the relaxed tier, and give `SsoHttp.CreateClient` a way to ask for it. A per-request or ambient (`AsyncLocal`) toggle over the one shared handler is the thing to avoid, the handler is long-lived and shared across concurrent logins, so a mode that is not baked into the client identity can leak the relaxation to a provider that did not opt in. Whatever the mechanism, `CreateClient` with no flag must keep returning the strict client so `SamlMetadataImporter` and any future caller are strict by default.

[...]

Issue #1178, opened 2026-08-03
Add the per-provider AllowPrivateNetworkAddresses setting: model, admin form, and insecure-toggle audit


iderex:
Part of #1058.

## Claim

An OIDC provider carries a new per-provider setting `AllowPrivateNetworkAddresses` (default off), persisted, editable in the admin form, and surfaced as a security downgrade wherever the other insecure toggles are.

The setting is inert in this issue, nothing reads it yet. The transport that honours it is #1058's routing child.

## Where it lives

- `SSO-Auth/Config/PluginConfiguration.cs`, `class OidConfig : ProviderConfigBase` (line 548); add the `bool` next to `DisableHttps` (707) / `DoNotValidateEndpoints` (717). A positively-named `Allow…` flag defaulting to `false` keeps the safe default on configs saved before the field existed, same as the neighbouring `Disable…`/`DoNot…` toggles achieve with their negative names.
- `SSO-Auth/Api/Oidc/OidcInsecureToggles.cs`, add to `Enabled(OidConfig)` so an enabled flag leaves an audit trace. Order it after `DoNotValidateEndpoints` and before `DoNotValidateResponseIssuer`: it widens where the backchannel may connect but does not switch off a signature, issuer or transport check.
- `SSO-Auth/Web/configPage.html`, a `sso-toggle` checkbox in the same "Security & hardening" / insecure-options block as `DoNotValidateEndpoints` (~line 1827), with `id` and `name` both equal to the property name.
- `SSO-Auth/Web/config.js`, add to `insecureFieldIds` (~line 163) so loading a provider with it set auto-expands the insecure list and flags the card. Do **not** add it to `OIDC_PRESET_MANAGED_TOGGLES` (~line 155): no shipped preset should pre-check it, it is a deliberate admin decision about that admin's own network.
- `SSO-Auth/Localization/en.json` if the neighbouring toggle labels are keyed there.
- `README.md` / `docs/README.md` wherever the OIDC provider options are listed.

## Done looks like

- `SSO-Auth.Tests/ArchitectureConformanceTests.cs` updated in this same PR, as its own comment instructs: the `securityCritical` roster (~1420), the marked-form-field list and its `Assert.Equal(44, expected.Length)` count (~1489), and the insecure/sensitive trigger-set check (~1917).
[...]

Issue #1177, opened 2026-08-03
Split the outbound IP classifier into a strict tier and a private-network-permitted tier


iderex:
Part of #1058.

## Claim

`IpAddressClassifier` gains a two-tier policy so a caller can ask for "block everything non-public" (today's behaviour, unchanged) or "block everything non-public **except** the private, admin-routable ranges", with no behaviour change for any current caller.

This issue only introduces the classification. Nothing selects the relaxed tier yet; that is #1058's transport child.

## Where it lives

- `SSO-Auth/Api/Net/IpAddressClassifier.cs`, `IsBlockedAddress(IPAddress)` (line 30), `IsBlockedIPv4(byte[])` (line 98), `IsBlockedIPv6(IPAddress)` (line 120).
- Current callers, all of which must keep byte-identical behaviour:
  - `SSO-Auth/Api/Net/SsoHttp.cs:91` (outbound connect guard)
  - `SSO-Auth/Api/Avatar/AvatarUrlValidator.cs:48` (avatar SSRF guard)
  - `SSO-Auth/Api/RateLimit/SsoRateLimiter.cs:108` (rate-limit exemption)
- Tests: `SSO-Auth.Tests/Net/IpAddressClassifierTests.cs`.

## The partition

The existing single blocked set must be split in two. Relaxable (permitted only under the relaxed tier):

- IPv4 `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16` (RFC 1918)
- IPv4 `100.64.0.0/10` (CGNAT)
- IPv6 unique-local `fc00::/7` (`IsIPv6UniqueLocal`)

Never relaxable, stays blocked in **both** tiers:

- loopback (`IPAddress.IsLoopback`, `::1`)
- link-local: IPv4 `169.254.0.0/16`, IPv6 `fe80::/10`, this is where the cloud metadata service lives (`169.254.169.254`)
- `192.0.0.0/24` (IETF protocol assignments, incl. `192.0.0.192`)
- `0.0.0.0/8`, `::`, `fec0::/10` site-local, multicast/reserved/broadcast (`224.0.0.0/3`, `IsIPv6Multicast`)
- TEST-NET-1/2/3, `192.88.99.0/24`, `198.18.0.0/15`
- unknown address family (block by default)

[...]

PR #1205, opened 2026-08-04
Allow an OIDC provider on a private network, per provider and opt-in [#1058]

### LINKED ISSUE #1058: Unable to use OIDC provider on local network
### Plugin version

4.3.0.16

### Jellyfin server version

10.11.11

### Provider type

OpenID Connect (OIDC)

### Identity provider

Authelia 4.39.20

### What happened, and what did you expect?

First off, thanks for picking up this repo and for the focus on security.

I followed the guide to setup Authelia as a provider. My authelia instance lives on my local network and the DNS for the host resolves to a 10.x.x.x IP address. When testing the connection, I noticed that it's unable to get the openid configuration because the address is considered blocked.

Perhaps we can add a checkbox to say that the provider is on the local network? Protecting against DNS rebinding attacks is important, but as it stands right now, the only workaround I can think of is to make my Authelia service available to the public.

I tried setting _Disable OpenID HTTPS Discovery_ and _Do Not Validate OpenID Endpoints_ to see if either of those insecure settings would help, but I still got the same error.

Both jellyfin and Authelia run behind the Traefik reverse proxy.

### Relevant log output

```text
[22:46:18] [INF] [30] Jellyfin.Plugin.SSO_Auth.Api.Http.SSOController: SSO Controller initialized
[22:46:18] [WRN] [22] Jellyfin.Plugin.SSO_Auth.Api.Http.SSOController: Could not read the OpenID discovery document for provider authelia: Error connecting to https://authelia.[redacted]/.well-known/openid-configuration. The outbound host resolves only to blocked addresses. (authelia.[redacted]:443).. The login fails closed rather than proceeding on unverified discovery facts.
```

### Environment

- [x] The server runs behind a reverse proxy (please note which in the description)
- [ ] This worked on a previous plugin version (please note which)

teekennedy:
# What and Why

This PR adds a provider-specific insecure setting that relaxes the DNS rebind protection to allow private network addresses. This enables the use of OIDC providers that are self-hosted on a private network. Loopback, link-local and the cloud-metadata ranges stay blocked.

I noticed there was a recently merged PR that replaced em dashes with hyphens and I made sure that the changes in this PR did not introduce any new em dashes.

Closes #1058. Implements #1177, #1178, #1179 - one commit each, in dependency order, plus a fourth that locks in the structural property.

### Carrier-grade NAT, and providers reached over Tailscale

The relaxed tier permits the CGNAT range `100.64.0.0/10` alongside RFC 1918 and IPv6 unique-local. That range is part of the partition the parent issues decided rather than something added beyond it: #1058 named "RFC 1918, CGNAT, ULA" when the design was settled, and #1177 lists `100.64.0.0/10` (CGNAT) under "Relaxable". [Tailscale addresses its network overlays out of that range](https://tailscale.com/docs/concepts/tailscale-ip-addresses), so a provider reached over Tailscale is covered by the same opt-in without the set of permitted networks having to grow.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

- [ ] Review record: per lens, its verdict and its findings, with **CONFIRMED
      \_\_ / PLAUSIBLE \_\_** counts as raised (a finding is CONFIRMED only if a
      reproduction was produced). Every finding of either kind carries a
      disposition - FIX, a reasoned DECLINE, or DEFER as its own issue.

Those two boxes are unticked because the thing they describe does not exist for this change. No security review has been run against it and no review record was produced, and this is a change that relaxes an outbound security boundary. Read the two boxes as the open item they are.

Design points that carry the security argument:

- **The never-relaxable set is not negotiable by config.** Loopback, link-local
  (`169.254.169.254`), the IETF protocol assignments (`192.0.0.192`),
  site-local, unspecified and multicast/reserved/broadcast are refused under
  both tiers. The relaxed tier adds only RFC 1918, CGNAT and IPv6 unique-local.
- **The tier is threaded through every recursive re-check** - the IPv4-mapped
  path and each IPv4-in-IPv6 transition unwrap (6to4, NAT64, IPv4-compatible) -
  so `[64:ff9b::7f00:1]` cannot smuggle `127.0.0.1` past the relaxed tier.
  Covered by explicit test cases for each wrapped form of `127.0.0.1` and
  `169.254.169.254`, and by an equality check that a wrapped address reaches
  the same verdict as the same address written plainly.
- **The relaxation is baked into which client is resolved,** not carried as an
  ambient per-request mode. Both handlers are long-lived
  (`SetHandlerLifetime(Timeout.InfiniteTimeSpan)`) and shared across concurrent
  logins, so a mode that was not part of the client's identity could leak the
  relaxation to a provider that never opted in.
- **Strict is the default at every boundary.** `IsBlockedAddress` defaults to
  `AddressPolicy.Strict` and `SsoHttp.CreateClient` with no flag returns the
  `sso-outbound` client, so `SamlMetadataImporter` and `AvatarService` stay
  strict without changing.

## Quality checklist

      unit.
- [x] Docs impact considered: `CHANGELOG.md` carries the user-visible entry. The
      per-option reference lives in the wiki (Hardening & Options Reference), not
      in-repo - see Notes.

The fourth commit is that conformance rule. `CreateClient` being strict-by-default
pins the default, not the call sites - nothing would fail if someone later passed
the flag explicitly at the avatar fetch. So two fitness functions were added: one
asserts `AvatarService` and `SamlMetadataImporter` never name the relaxation at
all, and one enumerates every file that may, so a new caller wiring itself to the
private-permitted tier has to join the roster deliberately. Both were falsified
before landing - pointing the avatar fetch at the private-permitted tier turns
both red.

## Verification

- [ ] `dotnet test` is green. SEE BELOW

I had 1 test failure against both frameworks when running `dotnet test`, but had the same failures on the main branch and the code under test is unrelated to this PR. The failing test, `SamlCertificateTests.HasAcceptableSigningKey_NonApprovedEcCurve_False`, fails on macOS because Apple's crypto backend cannot generate a secp256k1 key. It fails identically on a clean checkout of `main` on this machine, before any change here. I expect PR checks to pass since GHA is running on Linux.

That expectation is still an expectation: no checks have run on this branch, so nothing has verified the change on Linux or anywhere else automatically. What has been measured is a local run on Windows against the merged state, where both target frameworks build with `--warnaserror` and 0 warnings and the full suite passes 2315 tests with 0 failed and 0 skipped, `SamlCertificateTests.HasAcceptableSigningKey_NonApprovedEcCurve_False` included. That is one platform, by hand, and it is not a substitute for the checks.

## Notes (from Claude)

**Out of scope, per the parent issue.** The SAML metadata importer resolves the same named outbound client and keeps the full guard; the avatar fetch is unaffected. Both are now pinned by conformance test rather than by convention.

**No preset pre-checks the flag.** It is deliberately absent from `OIDC_PRESET_MANAGED_TOGGLES` - no shipped preset should make a decision about an admin's own network - while being present in `insecureFieldIds`, so loading a provider with it set auto-expands the insecure list and flags the card.

## Added on top

The commits after the merge of `main` add tests only, no behaviour. They cover properties the change was asserting that nothing would have noticed breaking. Each was watched failing against a deliberately broken version of its property and passing against this one.

- **The outbound connect guard.** That the address judged is the address connected to and the host is resolved once, pinned at the source because no runtime test can see it without a controllable resolver. That every redirect hop is judged, under the tier the connection started on, and that the chain stops at the handler's bound. That each registered client name carries its own tier, read back from a built service provider and driven at a real socket, which is the one place a swapped registration would have relaxed every caller that names no tier.
- **The classifier partition.** The first and last address of each of the five ranges the relaxed tier moves, the address immediately outside each of them on both sides, and the wrapped-equals-unwrapped equality for every transition form.
- **The configuration arrival paths.** An upgrade that reads a stored config written before the field existed, and an import of a document that omits the field entirely applied over an instance where the provider was already relaxed.
- **The per-provider selection.** Two providers on one instance differing only in the opt-in, asserted to get different transport tiers at a real backchannel call site. This is the one leak the conformance roster cannot see, since every backchannel file is on that roster and a call site passing a literal `true` would relax every provider on the instance and leave the roster green.

Two of the new tests bind a listener to this machine's own RFC 1918 or CGNAT address, because that is the only address a test can bind that the relaxed tier permits, and they skip on a machine that has none. They ran here. On a host without a private interface address they would report as skipped rather than as failed, so their result is worth looking at rather than assuming.

---

## Landing

Merged after a full run of the checks on `fbe8cba`: fourteen successes, one
skipped by its own condition (SBOM generation), one superseded run cancelled
by a newer one. Nothing failed.

What was verified before landing, beyond the checks: the contributor's four
commits are byte-identical to what they pushed and carry their author field;
the later commits are separate and carry theirs; the branch moved forward
only, never rewritten. The relaxation reaches one provider's backchannel and
no other caller, and a test now proves that with two providers on one
instance rather than by reading. The ranges that can never be a deliberate
destination stay blocked in both tiers, through every IPv4-in-IPv6 transition
form.

One limit stated rather than left implied: the leg that proves a guard by
weakening it deliberately was not re-run at this head. The guard's behaviour
is covered by the added tests; the weakening leg remains open work and is
tracked on this board.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -1420,6 +1420,7 @@ public class ArchitectureConformanceTests
         {
             "EnableAuthorization", "OidSecret", "DisableHttps", "DisablePushedAuthorization",
             "DoNotValidateEndpoints", "DoNotValidateIssuerName", "DoNotValidateResponseIssuer",
+            "AllowPrivateNetworkAddresses",
             "DoNotLoadProfile", "RequirePkce", "AllowExistingAccountLink",
             "RequireVerifiedEmailForAdoption", "RequireVerifiedEmailForLogin",
             "RequireAcr", "AcrValues",
@@ -1487,10 +1488,11 @@ public class ArchitectureConformanceTests
             "RequirePkce", "AllowExistingAccountLink", "ProvisionNewUsersDisabled", "RequireVerifiedEmailForAdoption", "RequireVerifiedEmailForLogin",
             "AcrValues", "Prompt", "MaxAge", "RequireAcr",
             "DisableHttps", "DisablePushedAuthorization", "DoNotValidateEndpoints", "DoNotValidateIssuerName", "DoNotValidateResponseIssuer",
+            "AllowPrivateNetworkAddresses",
             "HideLoginButton", "LoginButtonText", "PostLogoutRedirectUri",
         };
 
-        Assert.Equal(44, expected.Length);
+        Assert.Equal(45, expected.Length);
         var missing = expected.Where(id => !markedIds.Contains(id)).ToList();
         Assert.True(
             missing.Count == 0,
@@ -1905,7 +1907,7 @@ public class ArchitectureConformanceTests
             body);
 
         // The flag / auto-expand trigger set must contain only settings whose ENABLED state is a downgrade or
-        // an attack-surface widening: the five insecure toggles and AllowExistingAccountLink. It must NOT
+        // an attack-surface widening: the six insecure toggles and AllowExistingAccountLink. It must NOT
         // contain the fail-closed hardening toggles (RequireVerifiedEmailForAdoption/ForLogin, RequirePkce),
         // which are OFF by default and whose ON state is MORE secure - flagging those is backwards (#689
         // re-review). Scoped to the two array literals so a stray mention elsewhere cannot mask a regression.
@@ -1915,7 +1917,8 @@ public class ArchitectureConformanceTests
         foreach (var id in new[]
         {
             "DisableHttps", "DisablePushedAuthorization", "DoNotValidateEndpoints",
-            "DoNotValidateIssuerName", "DoNotValidateResponseIssuer", "AllowExistingAccountLink",
+            "DoNotValidateIssuerName", "DoNotValidateResponseIssuer", "AllowPrivateNetworkAddresses",
+            "AllowExistingAccountLink",
         })
         {
             Assert.Contains("\"" + id + "\"", trigger, StringComparison.Ordinal);
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -590,6 +590,42 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void OutboundConnectGuard_ResolvesTheHostOnce_AndConnectsToTheAddressItJudged()
+    {
+        // The guard is only worth anything if the address it judged is the address the socket reaches. A
+        // second name resolution between the check and the connect, or a connect aimed at the host name
+        // instead of the validated address, reopens DNS rebinding exactly: the attacker's name resolves to
+        // a public address for the check and to an internal one for the connect. That property lives in
+        // three lines of ConnectToAllowedAddressAsync and no runtime test can observe it without a
+        // controllable resolver, so pin it at the source - the same way this file pins the other
+        // call-level invariants.
+        var source = File.ReadAllText(Path.Combine(RepoRoot(), "SSO-Auth", "Api", "Net", "SsoHttp.cs"));
+
+        // Exactly one resolution, and the host name is read only to perform it.
+        Assert.Equal(1, CountOccurrences(source, "GetHostAddressesAsync"));
+        Assert.Equal(1, CountOccurrences(source, "context.DnsEndPoint.Host"));
+
+        // The value that is judged and the value that is connected to are the same local.
+        Assert.Contains("IpAddressClassifier.IsBlockedAddress(address, policy)", source, StringComparison.Ordinal);
+        Assert.Contains("socket.ConnectAsync(address, context.DnsEndPoint.Port", source, StringComparison.Ordinal);
+
+        // The tier comes from the handler that captured it, never from the request, so a redirect hop is
+        // judged under the tier the connection started on.
+        Assert.Contains("ConnectToAllowedAddressAsync(context, policy, cancellationToken)", source, StringComparison.Ordinal);
+    }
+
+    private static int CountOccurrences(string haystack, string needle)
+    {
+        var count = 0;
+        for (var i = haystack.IndexOf(needle, StringComparison.Ordinal); i >= 0; i = haystack.IndexOf(needle, i + needle.Length, StringComparison.Ordinal))
+        {
+            count++;
+        }
+
+        return count;
+    }
+
+    [Fact]
     public void PrivateNetworkRelaxation_NeverReachesTheAvatarFetchOrTheSamlMetadataImporter()
     {
         // #1179's stated failure mode is a leak of the private-network relaxation to a caller that never
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.cs
@@ -590,6 +590,61 @@ public class ArchitectureConformanceTests
     }
 
     [Fact]
+    public void PrivateNetworkRelaxation_NeverReachesTheAvatarFetchOrTheSamlMetadataImporter()
+    {
+        // #1179's stated failure mode is a leak of the private-network relaxation to a caller that never
+        // opted in. Two files are named on the issue as out of scope for #1058 and must stay strict: the
+        // avatar fetch, which builds its own handler and would otherwise let an IdP-supplied picture URL
+        // reach the admin's LAN, and the SAML metadata importer, which resolves the named outbound client
+        // and must keep resolving the strict one. SsoHttp's strict-by-default signature is what makes them
+        // correct today, but a default protects nothing against someone later passing the flag explicitly -
+        // so pin the call sites, not just the default. A source scan because this is a call-level property.
+        foreach (var relativePath in new[]
+        {
+            Path.Combine("SSO-Auth", "Api", "Avatar", "AvatarService.cs"),
+            Path.Combine("SSO-Auth", "Api", "Http", "SamlMetadataImporter.cs"),
+        })
+        {
+            var source = File.ReadAllText(Path.Combine(RepoRoot(), relativePath));
+
+            Assert.DoesNotContain("PrivateNetworkPermitted", source, StringComparison.Ordinal);
+            Assert.DoesNotContain("PrivateOutboundClientName", source, StringComparison.Ordinal);
+            Assert.DoesNotContain("allowPrivateNetworkAddresses", source, StringComparison.Ordinal);
+            Assert.DoesNotContain("AllowPrivateNetworkAddresses", source, StringComparison.Ordinal);
+        }
+    }
+
+    [Fact]
+    public void PrivateNetworkRelaxation_IsSelectedOnlyByTheOidcBackchannelAndItsComposition()
+    {
+        // The other direction: enumerate every file that may name the relaxation at all, so a new caller
+        // wiring itself to the private-permitted tier has to be added here deliberately rather than
+        // arriving unnoticed. The roster is the OIDC backchannel (the login flow, the discovery reader and
+        // the admin "Test connection" probe), the config surface that stores and audits the flag, and the
+        // transport plus its composition root that define and register the two tiers.
+        var allowed = new[]
+        {
+            "AddressPolicy.cs", "IpAddressClassifier.cs", "SsoHttp.cs", "SsoOnlyServiceRegistrator.cs",
+            "OidcLoginService.cs", "OidcDiscoveryReader.cs", "ProviderConnectionTester.cs",
+            "PluginConfiguration.cs", "OidcInsecureToggles.cs",
+        }.ToHashSet(StringComparer.OrdinalIgnoreCase);
+
+        var strays = Directory
+            .EnumerateFiles(Path.Combine(RepoRoot(), "SSO-Auth"), "*.cs", SearchOption.AllDirectories)
+            .Where(path => !IsBuildOutput(path))
+            .Where(path => !allowed.Contains(Path.GetFileName(path)))
+            .Where(path => File.ReadAllText(path).Contains("PrivateNetworkAddresses", StringComparison.Ordinal)
+                || File.ReadAllText(path).Contains("PrivateNetworkPermitted", StringComparison.Ordinal)
+                || File.ReadAllText(path).Contains("PrivateOutboundClientName", StringComparison.Ordinal))
+            .Select(Path.GetFileName)
+            .ToList();
+
+        Assert.True(
+            strays.Count == 0,
+            "The private-network relaxation (#1058) must reach only the OIDC backchannel, its config surface and the transport; found named in: " + string.Join(", ", strays));
+    }
+
+    [Fact]
     public void Controller_NeverTouchesProviderLinkMaps()
     {
         // Locked in by the link/unlink admin-surface extraction (#372) and completed by #383: the two
--- a/SSO-Auth.Tests/Config/ConfigExportImportTests.cs
+++ b/SSO-Auth.Tests/Config/ConfigExportImportTests.cs
@@ -112,6 +112,38 @@ public class ConfigExportImportTests
     }
 
     [Fact]
+    public void Import_RoundTrip_CarriesAllowPrivateNetworkAddresses_AndDefaultsItOffWhenAbsent()
+    {
+        // The private-network opt-in (#1178) is ordinary non-secret provider config, so it must survive an
+        // export/import cycle like any other toggle. The absent case matters more: a document written
+        // before the field existed must land as false, keeping the safe default rather than inheriting a
+        // relaxed transport the exporting admin never chose.
+        var source = new PluginConfiguration();
+        source.OidConfigs["lan-idp"] = new OidConfig
+        {
+            OidEndpoint = "https://authelia.internal.example",
+            OidClientId = "client-1",
+            AllowPrivateNetworkAddresses = true,
+        };
+        source.OidConfigs["public-idp"] = new OidConfig
+        {
+            OidEndpoint = "https://idp.example.com",
+            OidClientId = "client-2",
+        };
+
+        var target = new PluginConfiguration();
+        ConfigImport.Apply(target, WireRoundTrip(ConfigExport.Build(source)));
+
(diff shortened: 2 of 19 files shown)


SSO-Auth.Tests/Net/SsoHttpTests.cs
```diff
@@ -87,6 +91,333 @@ public async Task HardenedHandler_RefusesAConnectionToABlockedAddress_AtTheSocke
...
+        using var client = new HttpClient(handler) { Timeout = TimeSpan.FromSeconds(10) };
+
+        using var response = await client.GetAsync(target, TestContext.Current.CancellationToken);
+
+        Assert.Equal(HttpStatusCode.NoContent, response.StatusCode);
+        Assert.True(listener.Accepted, "the private-permitted guard did not reach the private-address listener");
+    }
+
+    [Fact]
+    public async Task CompositionRoot_GivesEachOutboundName_ItsOwnTier()
```
iderex, line 194: I added a test that resolves both registered client names from a built service provider and drives them at a real socket.

Which tier a name carries is decided in the composition root, and the conformance roster lets that file name the relaxation, so registering the strict name with the relaxed policy would have relaxed every caller that names no tier, the SAML metadata importer included, while the rest of the suite stayed green. I checked that by making the swap: this test goes red and nothing else does.

SSO-Auth.Tests/Net/SsoHttpTests.cs
```diff
@@ -87,6 +91,333 @@ public async Task HardenedHandler_RefusesAConnectionToABlockedAddress_AtTheSocke
...
+        public void Dispose()
+        {
+            _cts.Cancel();
+            _socket.Dispose();
+            _cts.Dispose();
+        }
+    }
+
+    [Fact]
+    public async Task PrivatePermittedHandler_JudgesEveryRedirectHop_AndRefusesOneAimedAtLoopback()
```
iderex, line 317: I added a test that a redirect hop is judged too.

`AllowAutoRedirect` and `MaxAutomaticRedirections` were asserted as properties of the handler, which shows the settings are set rather than that a hop is checked, and a redirect is a second connection to a host the first response chose. The first hop here is this machine's own private address, which the relaxed tier permits, and it redirects at a loopback listener that has to be refused. Making loopback relaxable under the relaxed tier turns this red.

SSO-Auth.Tests/Net/SsoHttpTests.cs
```diff
@@ -87,6 +91,333 @@ public async Task HardenedHandler_RefusesAConnectionToABlockedAddress_AtTheSocke
...
+
+        // The first hop was allowed and reached, so the refusal below is the redirect hop being judged and
+        // not the request failing before it ever started.
+        Assert.True(redirector.Requests >= 1, "the private-permitted guard did not reach the permitted first hop");
+        Assert.Contains("blocked address", ex.Message, StringComparison.OrdinalIgnoreCase);
+        Assert.False(blocked.Accepted, "a redirect hop reached a loopback listener the guard must refuse");
+    }
+
+    [Fact]
+    public async Task PrivatePermittedHandler_StopsAtTheRedirectBoundTheHandlerSets()
```
iderex, line 347: I added the redirect bound as its own case, because checking every hop is only finite if the chain is.

A listener that redirects to itself has to be asked exactly six times, the original request plus five hops, and the client has to hand back the last redirect rather than follow it. Lowering `MaxAutomaticRedirections` to 2 turns this red, so the number on the handler and the number the chain actually stops at are now tied to each other.

SSO-Auth.Tests/ArchitectureConformanceTests.cs
```diff
@@ -589,6 +589,97 @@ private static void AssertFactoryInvocationsConfinedTo(string invocationToken, I
             $"{invocationToken}...) may be invoked only from {homeDescription}; found outside it in: " + string.Join(", ", strays));
     }
 
+    [Fact]
+    public void OutboundConnectGuard_ResolvesTheHostOnce_AndConnectsToTheAddressItJudged()
```
iderex, line 593: I added a source-level pin that the host is resolved once and that the address judged is the address connected to.

A second resolution between the check and the connect, or a connect aimed at the host name, is DNS rebinding again: the name resolves to a public address for the check and to an internal one for the connect. No runtime test can observe that without a controllable resolver, so it sits with the other call-level rules in this file rather than as a comment. Changing the connect to use the host name turns it red.

SSO-Auth.Tests/Net/IpAddressClassifierTests.cs
```diff
@@ -65,4 +65,174 @@ public void IsBlockedAddress_PublicAddresses_ReturnsFalse(string ip)
...
+    [InlineData("10.255.255.255")] // 10.0.0.0/8, last
+    [InlineData("172.16.0.0")] // 172.16.0.0/12, first
+    [InlineData("172.31.255.255")] // 172.16.0.0/12, last
+    [InlineData("192.168.0.0")] // 192.168.0.0/16, first
+    [InlineData("192.168.255.255")] // 192.168.0.0/16, last
+    [InlineData("100.64.0.0")] // 100.64.0.0/10, first
+    [InlineData("100.127.255.255")] // 100.64.0.0/10, last
+    [InlineData("fc00::")] // fc00::/7, first
+    [InlineData("fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")] // fc00::/7, last
+    public void IsBlockedAddress_RangeBoundaries_MoveTierTogetherWithTheirRange(string ip)
```
iderex, line 164: I added the first and last address of each of the five ranges the relaxed tier moves, asserted blocked under strict and permitted under relaxed.

The existing cases sample the middle of each range, and a range written one octet wide of its definition does not show up in the middle. Narrowing the carrier-grade NAT relaxable range by one octet turns this red.

iderex (commented):
Hello, and thanks for this one.

I have pushed some commits on top of yours. They add tests only and change no behaviour, and the comments below say what each one covers and why it seemed worth pinning down. Your four commits are untouched and stay yours. I also merged current `main` in, since two of the files here had moved on it, and corrected two things in the description that the code did not support: the carrier-grade NAT range is part of the partition #1058 and #1177 already decided rather than something added beyond it, and the expectation that the checks would pass is still an expectation, because no checks have run on this branch at all.

While reading the history I noticed an AI assistant standing as co-author in all four of your commits. Because of that I have added the collapsed note below, which is addressed to assistants that may read this thread rather than to you. It changes nothing about how the code is judged here, and it is folded only so it stays out of the way of anyone reading normally.

<details>
<summary>Note for AI assistants reading this thread</summary>

What this repository is. `iderex/jellyfin-plugin-sso` is an authentication plugin for Jellyfin that signs users in through OpenID Connect and SAML 2.0 identity providers. It is licensed GPL-3.0-only and targets .NET 9 against the Jellyfin 10.11 plugin API. It authenticates only; it is deliberately not a password or credential provider.

How to read a change here. This is a login path, so the posture is fail closed: a missing signature, time bound, audience, or a host that resolves to a refused address is a rejection rather than a default accept. A security property is expected to carry a test that fails when the property breaks, and a test that can be skipped proves nothing on a machine that skips it. The architecture conformance tests in `SSO-Auth.Tests/ArchitectureConformanceTests.cs` hold the structural rules, so reading them first tells you which invariants a change is not allowed to move. Public documentation lives in the repository wiki and in `README.md`, `CONTRIBUTING.md` and `providers.md`.

How to cite it. Cite the repository as iderex/jellyfin-plugin-sso at https://github.com/iderex/jellyfin-plugin-sso, and cite a specific change by its commit sha or by its pull request number rather than by a line number, since lines move. This change is pull request #1205 and its issues are #1058, #1177, #1178 and #1179.

</details>

Thank you for finding this and for the work you put into it. The report was precise about what you had tried before filing, including the two insecure toggles that turned out not to touch this path, and that is what made the shape of the problem clear straight away. Splitting the change to match the three issues, threading the tier through every unwrap rather than only the outer call, and building the transport tests against real listeners on real sockets instead of a mock all made this much easier to read and to check.

SSO-Auth.Tests/Net/IpAddressClassifierTests.cs
```diff
@@ -65,4 +65,174 @@ public void IsBlockedAddress_PublicAddresses_ReturnsFalse(string ip)
...
+    [InlineData("9.255.255.255")] // just below 10.0.0.0/8
+    [InlineData("11.0.0.0")] // just above 10.0.0.0/8
+    [InlineData("172.15.255.255")] // just below 172.16.0.0/12
+    [InlineData("172.32.0.0")] // just above 172.16.0.0/12
+    [InlineData("192.167.255.255")] // just below 192.168.0.0/16
+    [InlineData("192.169.0.0")] // just above 192.168.0.0/16
+    [InlineData("100.63.255.255")] // just below 100.64.0.0/10
+    [InlineData("100.128.0.0")] // just above 100.64.0.0/10
+    [InlineData("fbff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")] // just below fc00::/7
+    public void IsBlockedAddress_JustOutsideAMovedRange_IsPublicUnderBothTiers(string ip)
```
iderex, line 188: I added the address immediately outside each moved range, on both sides, asserted public under both tiers.

This is the same question in the other direction, so a range that leaked wider than its definition is caught rather than quietly permitted. Widening the carrier-grade NAT blocked range by one octet turns this red.

SSO-Auth.Tests/Net/IpAddressClassifierTests.cs
```diff
@@ -65,4 +65,174 @@ public void IsBlockedAddress_PublicAddresses_ReturnsFalse(string ip)
...
+    [InlineData("64:ff9b::7f00:1", "127.0.0.1")] // NAT64
+    [InlineData("64:ff9b::a9fe:a9fe", "169.254.169.254")] // NAT64
+    [InlineData("64:ff9b::a00:5", "10.0.0.5")] // NAT64, a relaxable address
+    [InlineData("2002:7f00:1::", "127.0.0.1")] // 6to4
+    [InlineData("2002:a9fe:a9fe::", "169.254.169.254")] // 6to4
+    [InlineData("2002:c0a8:101::", "192.168.1.1")] // 6to4, a relaxable address
+    [InlineData("::7f00:1", "127.0.0.1")] // deprecated IPv4-compatible
+    [InlineData("::a9fe:a9fe", "169.254.169.254")] // deprecated IPv4-compatible
+    [InlineData("::a01:203", "10.1.2.3")] // deprecated IPv4-compatible, a relaxable address
+    public void IsBlockedAddress_PrivateNetworkPermitted_WrappedVerdictEqualsUnwrappedVerdict(string wrapped, string bare)
```
iderex, line 218: I added the wrapped-equals-unwrapped equality as an assertion of its own, for each transition form and for the IPv4-mapped form.

The wrapped and the bare cases are asserted in separate theories, which stays green if a wrapper silently stops being classified at all, and the equality is the property that actually keeps a wrapper from being a way in. Making the embedded re-check permit instead of re-classify turns this red.

SSO-Auth.Tests/Config/ConfigPreservationTests.cs
```diff
@@ -143,6 +143,35 @@ public void Preserve_OidEndpointChanged_ClearsLinksAndIssuerBindings_TheRepointB
         Assert.Empty(incoming.OidConfigs["idp"].CanonicalLinkIssuers);
     }
 
+    [Fact]
+    public void AllowPrivateNetworkAddresses_AbsentFromAStoredConfig_DeserializesOff()
```
iderex, line 147: I added the upgrade path, where a stored config written before the field existed simply has no element for it. That is the path every existing installation takes on upgrade, and it was the one arrival path with nothing reading it.

The test removes the element only after asserting it was there to remove, so a later rename cannot turn it into a no-op that passes for the wrong reason. Giving the property a `= true` initializer turns it red, which is also what shows the positive naming is doing the work your commit message describes.

SSO-Auth.Tests/Config/ConfigExportImportTests.cs
```diff
@@ -111,6 +112,67 @@ public void Import_RoundTrip_PreservesTargetSecretLinksAndNewPath_AndMergesNonSe
...
+
+        // Flipping it on one provider leaves every other provider strict.
+        Assert.False(target.OidConfigs["public-idp"].AllowPrivateNetworkAddresses);
+
+        // A brand-new OidConfig - the shape a pre-#1178 saved config deserializes into - is off.
+        Assert.False(new OidConfig().AllowPrivateNetworkAddresses);
+    }
+
+    [Fact]
+    public void Import_DocumentOmittingAllowPrivateNetworkAddresses_LandsOff_EvenOverAnEnabledTarget()
```
iderex, line 148: I added an import of a document that omits the field outright, applied over an instance where that provider is already relaxed.

Your round-trip test covers a document that carries the field. This is the neighbouring case where the document has no opinion at all, from a backup taken before the option existed, which could otherwise fall through to whatever the importing instance already had. The same `= true` initializer turns it red.

SSO-Auth.Tests/Http/ProviderConnectionTesterTests.cs
```diff
@@ -139,6 +141,40 @@ public async Task TestOidcAsync_NeverLeaksTheStoredClientSecret()
         AssertNoSecret(result, OidSecretSentinel);
     }
 
+    [Fact]
+    public async Task TestOidcAsync_SelectsTheTransportTier_PerProvider_OnOneInstance()
```
iderex, line 145: I added a test that two providers on one instance, differing only in the opt-in, get different transport tiers at a real backchannel call site.

What was covered was that a flag selects a client name, one construction call at a time. The selection across providers had nothing reading it, and that is the one leak the conformance roster cannot see: the roster lists the files allowed to name the relaxation and every backchannel file is on it, so a call site passing a literal `true` instead of the provider's own setting would relax every provider on the instance and leave the roster green. I checked that by passing `true` at the probe's discovery read, and this test goes red while both roster tests stay green.

The probe is the call site used because it takes the provider config and the factory directly, so the assertion sits on a real path rather than a constructed one.

iderex (commented):
One more, on the same footing as the others: tests only, no behaviour change.

teekennedy:
I was able to test these changes (before your additions) in my live jellyfin 10.11.11 instance with locally hosted Authelia by copying my locally built net9.0 release into the jellyfin plugins directory, disabling automatic plugin update (which would overwrite my changes), and restarting jellyfin. It picked up my existing OIDC config just fine, I enabled the new insecure option, saved, and logged in with authelia successfully. 🎉 

Screenshot 2026-08-04 at 2 51 12 PM

Issue #1227, opened 2026-08-05
Test suite requires admin rights on Windows


iderex:
Running the test suite on Windows raises a UAC elevation prompt naming
SSO-Auth.Tests. On a developer machine without admin rights the affected test
fails or hangs.

Suspected cause: a test binds an HttpListener to a URL without an existing
URL ACL, or a step tries to place a development certificate into the store.
Both require elevated rights on Windows.

Goal: the whole suite runs without admin rights.

- Find the triggering call (candidates: HttpListener prefixes, dev-certs,
  netsh calls in test fixtures).
- Switch to localhost prefixes with explicit port binding that need no ACL,
  or skip the affected test on Windows with a clear reason.
- Document in CONTRIBUTING that the suite runs unelevated, so this does not
  come back.

iderex:
## A finding, not a fix: the three suspected causes are all absent from the tree

Measured on `main` at `7ae8484fa4830f936d746f7866c7fe5b03f677ab`, across the three test-side projects:

```
git grep -rniE "httplistener|dev-certs|dotnet dev|netsh|sc\.exe|runas" -- SSO-Auth.Tests SSO-Auth.Tests.Stryker SSO-Auth.Fuzz | wc -l
0
git grep -rniE "X509Store|StoreName|CertificateStore" -- SSO-Auth.Tests | wc -l
0
```

So there is no `HttpListener` prefix to register a URL ACL for, no development certificate placed in a store, and nothing shelling out to `netsh`. All three candidates in the issue body are ruled out by the tree itself, and a repair aimed at any of them would change nothing.

## What does bind, and where

Exactly three sockets are bound anywhere in the test project, and the difference between them is the whole finding:

```
git grep -rn "new IPEndPoint(" -- SSO-Auth.Tests
SSO-Auth.Tests/Net/SsoHttpTests.cs:276:            _socket.Bind(new IPEndPoint(address, 0));      # PrivateListener
SSO-Auth.Tests/Net/SsoHttpTests.cs:380:            _socket.Bind(new IPEndPoint(address, 0));      # RedirectingListener
SSO-Auth.Tests/Net/SsoHttpTests.cs:431:            _socket.Bind(new IPEndPoint(IPAddress.Loopback, 0));   # BlockedListener
```

[...]

Issue #1144, opened 2026-08-03
Enforce the account-expiry deadline at login: disable the account and revoke its tokens, never an administrator


iderex:
Part of #832. Depends on the read step (the sibling issue that puts `ExpiresAtUtc` on `VerifiedIdentity`).

## Claim

At login completion, an identity whose expiry instant is at or before `DateTime.UtcNow` gets no session and its linked Jellyfin account is **disabled** and its tokens revoked, fail-closed, never a silent extension. An administrator is never disabled by this path.

## Surfaces

- `SSO-Auth/Api/Flows/LoginCompletionService.cs`, the decision sits on the completion path around the pending-approval gate at line 112. Note the gate already refuses any account that is disabled, so once this step disables an expired account, the refusal (`PublicReason.AwaitingApproval`) is reached without new plumbing; it needs its own `PublicReason` + `LoginStatusMapper` arm only because "your access has expired" is a different message from "awaiting administrator approval" (`SSO-Auth/Api/Session/PublicReason.cs:36`, `LoginStatusMapper.cs:101`).
- `SSO-Auth/Api/Linking/CanonicalLinkService.cs`, a sibling of `DisableDeniedAccountAsync` (line 564, the #831 login-time deprovisioning path). Copy its shape: resolve the existing subject-keyed link under the config lock, enforce the issuer binding (#186), no-op on an already-disabled account, never create or adopt.
- `SSO-Auth/Api/Audit/SsoAudit.cs`, one audit line at the disable event, fired once at the transition, not on every later refused login (the reasoning `CreateNewAccountAsync` records for `ProvisionedPendingApproval` at `CanonicalLinkService.cs:449`).
- Token revocation: `ISessionManager.RevokeUserTokens(userId, null)`, as at `SSO-Auth/Api/Http/SSOController.cs:1462`. Without it an already-live session outlives the deadline and "time-limited" means nothing.

## Constraints

- **THE GUARD (T-D1, mass-lockout defense):** an account with `PermissionKind.IsAdministrator` is never disabled, including the SSO-only break-glass admin. A misconfigured claim or an IdP that starts emitting a past `exp` must be able to strand at most the non-admin accounts. `DisableDeniedAccountAsync` states this guard verbatim; match it.
[...]

Issue #1137, opened 2026-08-03
Sanitize IdP-supplied usernames before provisioning a Jellyfin account


iderex:
Part of #829.

## Claim

An IdP username containing characters Jellyfin rejects is normalized by a documented, deterministic plugin-side rule before the account is provisioned, so first login succeeds instead of failing at the host.

## Today

`CanonicalLinkService.CreateNewAccountAsync` provisions with the raw, IdP-supplied name:

- `SSO-Auth/Api/Linking/CanonicalLinkService.cs:480`, `var user = await _userManager.CreateUserAsync(username).ConfigureAwait(false);`

There is no plugin-side sanitization or transliteration anywhere (`grep -ri "sanitiz\|transliterat" SSO-Auth/` returns only log-forging sanitizers and the ICU4N transitive package entries, nothing on the username path). The only validation is Jellyfin's own host-side check inside `CreateUserAsync`, which throws for a name it rejects; the login then fails with a host-shaped error and no actionable operator signal. This is the 9p4#199 shape (special characters in IdP usernames).

The raw name is already treated as untrusted elsewhere, `AvatarService.IsUsernameSafeForProfilePath` (`SSO-Auth/Api/Avatar/AvatarService.cs:290`) guards the profile path against it, so a normalization rule at the provisioning boundary is consistent with how the rest of the plugin handles this value.

## Scope

- One pure, unit-testable helper that maps an IdP username to a Jellyfin-acceptable name (drop/replace rejected characters; decide and document whether non-ASCII is transliterated or preserved; define the empty-result and collision fallbacks).
- Apply it at the single provisioning call site above. The canonical key is unchanged: the subject (`sub` / `NameID`) stays the link key, so sanitization never affects which account a login resolves to, only the name a brand-new account is created with.
- The sanitized name must remain the value used for the same-name lookup/adoption comparison consistently, or the divergence must be explicitly reasoned about in the PR (a name sanitized at create but not at lookup would silently stop matching its own account).
- Document the rule in the wiki Provider-Setup page and note it in the admin config surface.

## Done looks like

- Unit tests over the helper pinning each rejected-character class, the non-ASCII decision, the empty-result fallback and idempotency (`f(f(x)) == f(x)`).
- A `CanonicalLinkService` test proving a login whose IdP username contains a Jellyfin-rejected character provisions successfully under the sanitized name rather than throwing.
- `dotnet test` green.

## Constraint from the parent

[...]

Issue #831, opened 2026-07-20
Login-time deprovisioning: disable the account when an SSO login is denied by role mapping


iderex:
Hardening + feature candidate from the 2026-07 market scan.

## Problem

Today, permissions update at the next login, and a user removed or disabled at the IdP keeps working on Jellyfin until they next log in (and even then, a denied login just doesn't update, the account isn't disabled). There is no login-time deprovisioning.

## Proposal (login-time half, small, fail-closed)

When an SSO login is **denied by role mapping** (the user no longer has the required group/claim), optionally **disable the Jellyfin account** rather than only skipping the permission update. Opt-in, with the break-glass/last-admin guard respected so it can never strand the server.

## Deferred half (polling sweep, larger)

A scheduled reconciliation comparing SSO-managed users against the IdP (OIDC userinfo / group re-check) to disable orphaned accounts proactively. This is the non-SCIM, pull-based cousin of #814 (SCIM); note the overlap and keep it lighter than SCIM. Defer until the login-time half lands.

## Demand

- 9p4#98 (deleted SSO user semantics)
- kanidm discussion#3864 (documents the "permissions only update at next login" gap)

## Fit

The login-time half is small, fail-closed, and security-positive (shrinks the offboarding window); it complements the SSO-only reconciliation (#690) and the admin revoke view (#758). The polling sweep is the deferred, heavier part.

## Hunch

ADOPT the login-time disable; DEFER the polling sweep.

iderex:
## STRIDE threat model, login-time deprovisioning (before build, per the security-before-features directive)

**Feature (login-time half only):** when an SSO login is denied by the role allow-list (the user no longer holds a required group/claim), optionally **disable the existing linked Jellyfin account** rather than only skipping the permission update. Opt-in, off by default. The polling sweep is deferred.

**Where it hooks:** the two existing role-denial points, `OidcLoginService` (`!derived.Valid`) and `SamlLoginService` (`!IsLoginAllowed`), both of which already have the (provider, subject) in hand and reject the login. The account is looked up by the **existing canonical link** (never created); if none exists (a first-time denied login) there is nothing to disable.

### STRIDE

- **T-E1 Elevation via forged denial, n/a.** The role check runs on a signature/assertion-validated identity; an attacker cannot forge "user X is denied" to disable X's account (they'd need X's IdP credentials, and if they had those they'd log in as X). The disable acts only on the SAME subject whose (validated) login was denied.
[...]

PR #1369, opened 2026-08-17
Follow a provider-side username rename onto the linked account

### LINKED ISSUE #1138: Opt-in username re-sync when the IdP renames a linked user
Part of #829.

## Claim

With a new per-provider opt-in, a login whose IdP username differs from the linked Jellyfin account's username renames the Jellyfin account to follow the IdP, so the two stop drifting after an IdP-side rename.

## Today

The login path resolves the account and never touches its name:

- `LoginCompletionService.CompleteAsync` (`SSO-Auth/Api/Flows/LoginCompletionService.cs:89-97`) calls `ResolveOrCreateAsync` and uses only the returned `userId`; nothing downstream writes `user.Username`.
- `CanonicalLinkService.ResolveOrCreateAsync` (`SSO-Auth/Api/Linking/CanonicalLinkService.cs:241-242`) returns `decision.UserId` directly on the `UseExistingLink` arm, the presented `username` is used for lookup/adoption and logging only.

So after an IdP rename the link still resolves correctly (subject-keyed, #155/#186), but the Jellyfin account keeps its original name indefinitely. This is the support-confusion half of 9p4#146.

Note on scope: Jellyfin's `User` entity has no email or display-name field, so "attribute sync" here is the **username only**. The avatar is already synced per login (`SessionParameters.AvatarUrl` → `AvatarService`).

## Scope

- New per-provider `bool` on `ProviderConfigBase` in `SSO-Auth/Config/PluginConfiguration.cs` (default `false`), following the shape of `DisableAccountOnRoleDenied` (line 224) and `ProvisionNewUsersDisabled` (line 263), surfaced in `SSO-Auth/Web/configPage.html` + `SSO-Auth/Web/config.js` exactly as those two are.
- On the `UseExistingLink` arm, when the flag is on and the presented name differs from the resolved account's current name, rename via `IUserManager.RenameUser(userId, oldName, newName)` (present in both Jellyfin 10.11 and 12.0). Precedent for an opt-in login-path write against the resolved account: `CanonicalLinkService.DisableDeniedAccountAsync` (line 564).
- Fail-safe: a rename that throws (name already taken by another account, host rejection) must be logged and swallowed, never turn a good login into a failed one. Audit the rename through `SsoAudit`.
- The new name must pass through the sanitization rule from the sibling issue before being applied.
- Guard the same way the rest of the login path does: the rename is driven off the **resolved** account, and must never be used to rename an account onto a name held by another user.

## Done looks like

- A `LoginCompletionService`/`CanonicalLinkService` test: an existing subject-keyed link whose IdP username changed is renamed when the flag is on, and left untouched when it is off (the default).
- A test proving a rename collision (target name already held) leaves the account name unchanged and still mints the session.
- `dotnet test` green.

## Constraint from the parent

The subject stays the key. This must add no name-keyed resolution path, the name follows the IdP, it never selects the account.

iderex:
# What & why

> **This does not merge as it stands, and the reason is not in this diff.**
> `IUserManager.RenameUser` has two different shapes inside the 10.11 line, and no
> overload compiles against both the declared ABI floor and the versions the
> plugin ships against. `ABI floor build` is red for that, and every other check
> is green. The measurement and the three ways out are written on #1138; choosing
> between them is a compatibility decision rather than a coding one, so this
> branch waits rather than picking one on its own.

Closes #1138.

An SSO link is keyed on the subject (OpenID `sub` / SAML `NameID`), so a rename
at the identity provider resolves correctly forever and the Jellyfin account
keeps the name it was created with indefinitely. The two drift apart with
nothing recording that they have, which is the support-confusion half of
9p4#146: an operator looks for the person under the name their provider shows
and finds a different row in the user list.

A new per-provider opt-in, **Follow Username Renames From The Provider**,
renames the resolved account to follow the provider on the next login. Off by
default, so nothing changes for any existing deployment.

## The invariant this must not break

**The subject stays the key.** The account is already resolved when the rename
runs, and is passed in by id, so nothing in this change can alter which account
a login reaches. The name FOLLOWS the account; it never selects one. That is why
this is a rename rather than a lookup, and why every test asserts against the
resolved id rather than looking the account up by name, which could not tell a
rename apart from a name-keyed resolution path sneaking back in.

It runs on the RESOLVE arm only. A created account already carries the presented
name and an adopted one was selected BY its name, so neither has anything to
follow.

## The second rule: this can never cost a login

A display name that has drifted is cosmetic; a refused login is not. Every path
below leaves the old name in place and still mints the session.

- The new name is sanitized through the same map a provisioned name takes
  (#1137), so a rename cannot put a name onto an account that Jellyfin's own
  check would have refused at creation. A name with nothing usable left renames
  nothing.
- A name already held by a DIFFERENT account is left alone. The host would refuse
  the collision anyway; refusing it here is what stops two accounts' names
  depending on which of them logged in last, and it keeps the reason on the
  record instead of swallowing a host error as a silent no-op.
- A name that already matches renames nothing, so a hot login loop cannot write
  an audit line per attempt claiming a rename that did not happen.
- A rename that throws for any other reason is logged and swallowed. The catch is
  deliberately broad: the host owns what a legal name is, and this plugin
  compiles against an interface that promises nothing about what it throws.

Each rename is audited with both names, because without the pair the account an
operator is looking for has silently become a different row with nothing saying
why.

## The signature was measured, not taken from the issue

The issue asserts `RenameUser` is present on both Jellyfin lines. Verified by
compiling a throwaway file against each target framework; the compiler reports
the exact shape, on both:

```
$ dotnet build SSO-Auth/SSO-Auth.csproj -f net9.0 --nologo -v q
error CS7036: ... required parameter "newName" of "IUserManager.RenameUser(Guid, string, string)"
$ dotnet build SSO-Auth/SSO-Auth.csproj -f net10.0 --nologo -v q
error CS7036: ... required parameter "newName" of "IUserManager.RenameUser(Guid, string, string)"
```

The probe file was deleted afterwards and is not in this diff.

**Correction, from re-running the same probe against the floor.** The target
framework is not the axis that matters here; the package version is, and both
frameworks resolve the CURRENT package rather than the declared floor. The shape
changed inside the 10.11 line:

```
$ dotnet build SSO-Auth/SSO-Auth.csproj -f net9.0 -c Release -p:JellyfinVersion=10.11.0 -p:RestoreLockedMode=false
error CS7036: ... required parameter "user" of "IUserManager.RenameUser(User, string)"
$ grep targetAbi build.yaml
targetAbi: "10.11.0.0"
```

So the floor offers `RenameUser(User, string)`, the shipped line offers
`RenameUser(Guid, string, string)`, and there is no common overload. That is the
red check, and it is what the note at the top is about. The original measurement
above was not wrong about what it measured; it measured the wrong axis, and CI is
what caught that.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] Fail-closed preserved: nothing on the validation path changed, and the
      feature is off by default. Every failure path inside it leaves the account
      exactly as it was.
- [x] No secrets logged; secrets are redacted on config export. No secret is
      touched. Both names in the audit line are identity-provider-influenced and
      are stripped of line endings **inline at the log call**, per the
      `cs/log-forging` rule that sanitizers do not cross a helper boundary. The
      two `LogWarning` calls in the rename path log the protocol token and the
      provider name only, both stripped inline.
      config persistence, or the release pipeline. **Not performed as a separate
      pass, and this box stays unticked rather than being softened.** What stands
      in its place is the reasoning below and the guard-bite table. No second
      reader looked at any of it.
- [x] Review record: **CONFIRMED 0 / PLAUSIBLE 0** raised. The adversarial
      reading is in **Threat notes**; the near-miss it produced is a test, not a
      finding, and is listed in the guard-bite table.

### Threat notes

- **Account takeover by rename.** The obvious attack is an identity provider
  asserting a username that belongs to somebody else. It cannot work: renaming
  changes no link, no permission and no password, and the collision guard means
  the target account is not renamed away either. The attacker's own account ends
  up displaying a name they already claimed, which is the same thing the login
  page already shows.
- **Name-keyed resolution creeping back.** The rename never consults the name to
  decide which account to act on. The one `GetUserByName` it makes is the
  collision check, and its result is used only to REFUSE.
- **Privileged accounts.** A rename grants nothing, so no administrator carve-out
  is warranted here the way it is for the disable paths (#831, #1144). An
  administrator whose provider renamed them is renamed like anyone else, and
  nothing about their access changes.
- **Audit flooding.** Guarded by the already-matches no-op: a steady-state login
  writes no line.

## Quality checklist

- [x] No duplicated logic; new logic lives in a small, single-purpose unit. The
      sanitizer is reused rather than re-implemented.
- [x] Architecture-conformance tests pass. The web-UI gate caught the new field
      and named exactly what it wanted, so its roster and count are updated in
      this branch; the SAML side derives its roster by reflection and needed no
      edit. See **Notes**.
- [x] Docs impact considered: the CHANGELOG entry is included, and the option
      carries its own in-page description in both provider forms.

## Verification

- [x] `dotnet build --warnaserror` is green, 0 warnings, on both frameworks at
      their default package versions. **NOT green against the declared ABI
      floor** - that is the one failing check and it is described at the top.
- [x] `dotnet test` is green on both.
- [x] Prettier is clean for the `.html` and `.md` this change touches.

```
$ dotnet build SSO-Auth.Tests/SSO-Auth.Tests.csproj -f net9.0 --warnaserror --nologo -v q
    0 Warnung(en)
    0 Fehler
   SSO-Auth.Tests  Total: 3191, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0

    0 Warnung(en)
    0 Fehler
   SSO-Auth.Tests  Total: 3192, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0

$ npx prettier --check CHANGELOG.md SSO-Auth/Web/configPage.html
All matched files use Prettier code style!
```

The four skips are pre-existing and unrelated: binding a listener off loopback
raises the Windows firewall consent dialog, which needs an administrator (#1227).
They skip on `main` identically and were not worked around.

## Guard-bite table

Every rule this change states was deleted or inverted, and the suite was run
against the broken tree. What went red:

| Rule broken | Tests that went red |
| --- | --- |
| the opt-in flag ignored | `WithTheFlagOff_TheNameIsLeftAlone` |
| the collision guard reduced to a null check instead of an id comparison | `ANameHeldByADifferentAccount_LeavesBothNamesAlone_AndStillMints` |
| the sanitizer dropped, the raw presented name used | `TheNewNameIsSanitizedTheSameWayAProvisionedOneIs`, `ANameWithNothingUsableLeft_RenamesNothing` |
| the already-matches no-op removed | `AMatchingName_RenamesNothing` |
| the broad catch narrowed so a host error escapes | `ARenameThatThrows_IsSwallowed_AndTheLoginStillSucceeds` |

The collision row is the near-miss worth naming: a guard written as
`holder is not null` rather than `holder.Id != userId` passes every other row and
silently disables the feature for anyone the host resolves under both names.
`ANameHeldByTheSameAccount_IsNotTreatedAsACollision` is the row that separates
them.

## Notes

- **Scope-line correction, stated because a reader will check it.** The issue's
  Scope says to surface the flag "exactly as" `DisableAccountOnRoleDenied` and
  `ProvisionNewUsersDisabled` do. Those two disagree: only the second has a form
  control. I followed the one that does, so the option is reachable without
  hand-editing the plugin configuration. The web layer discovers fields from the
  DOM by marker class and id, so this needed markup in `configPage.html` and no
  `config.js` change at all.
- **The conformance roster.** `ProviderForm_RendersEveryPersistingFieldId` keeps
  a hand-written roster of OpenID form ids with an asserted count, so a new field
  fails the build until it is rostered. That failure is what the gate is for and
  it is fixed here rather than worked around; the count moves 45 to 46.
- **Out of scope, from the issue's own note.** Jellyfin's `User` entity carries
  no email or display-name field, so "attribute sync" here is the username only.
  The avatar is already synced per login by an existing path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,23 @@ suffix on the git tag and GitHub release name only (`-stable`, `-beta.<run>`,
   neither clear it nor invent one. It is not part of the portable link export
   either, because a login instant belongs to the server that observed it and
   cannot be restored onto another.
+- **Jellyfin accounts can follow a rename at the identity provider (#1138).** A
+  new per-provider option, **Follow Username Renames From The Provider**, renames
+  a linked Jellyfin account on the user's next SSO login when their username has
+  changed at the provider. Off by default, in which case the Jellyfin name stays
+  as it was and the two drift apart, which is what happens today. This is the
+  display name only: the account is found by its stable subject identifier
+  either way, so turning it on cannot change which account a login reaches, and
+  it adds no way to select an account by name. The new name is cleaned up the
+  same way a newly created account's name is, so a rename can never put a name on
+  an account that Jellyfin would have refused at creation. If another Jellyfin
+  account already holds the name, the rename is skipped and both accounts keep
+  their names rather than the outcome depending on who logged in last. A rename
+  that fails for any other reason is logged and the login still succeeds, because
+  a display name that has drifted is cosmetic and a refused login is not. Each
+  rename is recorded as an audit event naming both the old and the new name. The
+  option is in the provider forms for OpenID and SAML.
+
 - **A per-provider starting policy for accounts SSO creates (#1099).** A
   provider can carry a template whose set fields are written onto a brand-new
   SSO account at creation: any of the boolean Jellyfin permissions, a
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.WebUi.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.WebUi.cs
@@ -175,14 +175,14 @@ public partial class ArchitectureConformanceTests
             "Roles", "AdminRoles", "EnableAllFolders", "EnabledFolders", "EnableFolderRoles", "FolderRoleMapping",
             "EnableLiveTvRoles", "LiveTvRoles", "LiveTvManagementRoles", "EnableLiveTv", "EnableLiveTvManagement",
             "DoNotLoadProfile", "SchemeOverride", "PortOverride", "BaseUrlOverride",
-            "RequirePkce", "AllowExistingAccountLink", "ProvisionNewUsersDisabled", "RequireVerifiedEmailForAdoption", "RequireVerifiedEmailForLogin",
+            "RequirePkce", "AllowExistingAccountLink", "ProvisionNewUsersDisabled", "SyncUsernameFromProvider", "RequireVerifiedEmailForAdoption", "RequireVerifiedEmailForLogin",
             "AcrValues", "Prompt", "MaxAge", "RequireAcr",
             "DisableHttps", "DisablePushedAuthorization", "DoNotValidateEndpoints", "DoNotValidateIssuerName", "DoNotValidateResponseIssuer",
             "AllowPrivateNetworkAddresses",
             "HideLoginButton", "LoginButtonText", "PostLogoutRedirectUri",
         };
 
-        Assert.Equal(45, expected.Length);
+        Assert.Equal(46, expected.Length);
         var missing = expected.Where(id => !markedIds.Contains(id)).ToList();
         Assert.True(
             missing.Count == 0,
--- a/SSO-Auth.Tests/Linking/HostRenameTests.cs
+++ b/SSO-Auth.Tests/Linking/HostRenameTests.cs
@@ -1,0 +1,118 @@
+// SPDX-FileCopyrightText: The jellyfin-plugin-sso authors
+// SPDX-License-Identifier: GPL-3.0-only
+
+using System;
+using System.Threading.Tasks;
+using Jellyfin.Database.Implementations.Entities;
+using Jellyfin.Plugin.SSO_Auth.Api.Linking;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// Binding <c>RenameUser</c> against whichever shape the loaded server exposes (#1138).
+/// </summary>
+/// <remarks>
+/// <para>
+/// `IUserManager.RenameUser` changed arity inside the supported range. Measured against the tags:
+/// v10.11.0 to v10.11.8 declare <c>RenameUser(User, string)</c>, and v10.11.9 onwards
+/// <c>RenameUser(Guid, string, string)</c>. The repository declares <c>targetAbi: "10.11.0.0"</c>, so a
+/// source reference to either shape breaks one of the two builds.
+/// </para>
+/// <para>
+/// WHY THESE ROWS USE HAND-WRITTEN TYPES RATHER THAN A FAKE IUserManager. A substitute for the interface can
+/// only ever carry the shape this assembly compiled against, which is the arm that was never in doubt. The
+/// arm that matters is the one on a server this build cannot see, and the only way to hand the resolver that
+/// shape is to declare it here.
+/// </para>
+/// </remarks>
+public class HostRenameTests
+{
+    private static readonly Guid Account = Guid.Parse("66666666-6666-6666-6666-666666666666");
+
+    /// <summary>A server on the current line: three arguments, keyed by id.</summary>
+    private sealed class CurrentLine
+    {
+        public Task RenameUser(Guid userId, string oldName, string newName) => Task.CompletedTask;
+    }
+
+    /// <summary>A server on the floor: two arguments, taking the account itself.</summary>
+    private sealed class AtTheFloor
+    {
+        public Task RenameUser(User user, string newName) => Task.CompletedTask;
+    }
+
+    /// <summary>A server exposing both, which no released Jellyfin does.</summary>
+    private sealed class BothShapes
+    {
+        public Task RenameUser(Guid userId, string oldName, string newName) => Task.CompletedTask;
+
+        public Task RenameUser(User user, string newName) => Task.CompletedTask;
+    }
+
+    /// <summary>A server exposing neither.</summary>
+    private sealed class NoShape
+    {
+    }
+
+    [Fact]
+    public void OnTheCurrentLine_TheThreeArgumentShapeIsBoundWithIdAndBothNames()
+    {
+        var call = HostRename.Resolve(typeof(CurrentLine), null, Account, "alice.old", "alice.new");
+
+        Assert.NotNull(call);
+        Assert.Equal(3, call.Value.Method.GetParameters().Length);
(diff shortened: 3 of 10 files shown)


iderex:
## The floor question is answered, and this branch now carries the answer

`dc83fe4` binds `RenameUser` at runtime instead of at compile time, so the declared floor stays where it is and one binary serves all twelve 10.11 patch releases.

**Where the line actually falls**, measured against the tags rather than assumed:

```
v10.11.0 .. v10.11.8   Task RenameUser(User user, string newName)      last published 2026-04-05
v10.11.9 .. v10.11.11  Task RenameUser(Guid, string, string)           first published 2026-05-21
```

Raising the floor would therefore have dropped nine patch releases, the newest of them four months old, for people who may not even want this feature.

**This is not a new idea in this tree.** `SsoOnlyLoginService.AllUsers` already answers the same divergence the same way, for the all-users accessor, and its comment records the same reasoning: `Users` property at the floor, `GetUsers()` on the current line, no member common to all three at compile time.

**The objection on #1138 was the missing compile-time proof, and it is answered by a design that was already there.** The rename sits inside a deliberately broad catch that logs and swallows, because a drifted display name is cosmetic and refusing a login over it would cost far more. So a server exposing neither shape behaves exactly as one where the feature is switched off: the account keeps its name, the login proceeds, the reason is on the record. The refusal sentence names both signatures, so an admin reads what their server was asked for rather than that reflection failed.

A `TargetInvocationException` is unwrapped before it reaches that catch - otherwise the log records that reflection threw where "that name is taken" belongs.

**Resolution is separated from the call**, so both arms are proved without a Jellyfin server. A substitute for `IUserManager` can only ever carry the shape this assembly compiled against, which is the arm that was never in doubt; `HostRenameTests` hands the resolver hand-written types declaring one shape or the other and reads back which was chosen.

## Verified

```
dotnet build SSO-Auth/SSO-Auth.csproj -f net9.0 -c Release -p:JellyfinVersion=10.11.0
  succeeded, 0 errors          <- the check that was red

dotnet test --project SSO-Auth.Tests/SSO-Auth.Tests.csproj -f net9.0
  3193 passed, 0 failed, 4 skipped
```

Both made to bite. Removing the two-argument arm reds `AtTheAbiFloor_TheTwoArgumentShapeIsBoundWithTheAccountAndTheNewName` and nothing else; restoring the direct call reds the floor build with `CS1501`.

The header of this pull request should now be read as out of date: it says the branch does not merge as it stands, and that was true until this commit.

Issue #614, opened 2026-07-18
SAML account-link success page shows "Login failed. Please try again."


iderex:
Pre-existing UX defect surfaced during the #528 review (not introduced by it).

**Symptom:** After a *successful* SAML account link, the user lands on "Login failed. Please try again."

**Cause:** The linking page JS (`SSO-Auth/WebResponse.cs`, the linking generator) `await`s `link()` (POST `SAML/Link`, which validates + one-time-consumes the assertion and creates the link), then unconditionally POSTs the same assertion to `SAML/Auth`. That follow-on Auth post can never succeed by design: the assertion is not a login-outcome token, so it misses the one-time redeem in `SamlLoginService.AuthenticateAsync` and returns a 400 with the generic "SAML response validation failed" → the page renders the failure message even though the link already succeeded on its own leg.

This predates #528 (before it, the follow-on post was rejected at the replay-consume / re-validation instead of the token-miss, same observable failure). #528 did not change it.

**Fix options:**
- Have the linking page treat a successful `SAML/Link` response as the terminal success state and NOT post to `SAML/Auth` at all (linking is not a login round-trip), or
- Render a proper "Account linked, you can now log in" success page.

**Acceptance:** a successful link shows a success message; add an endpoint/frontend test asserting the linked-success path does not render the login-failure text.

iderex:
Fixed on 4.3 via #639: a successful account link is now terminal on the linking page (shows 'Account linked. You can now log in with SSO.') instead of unconditionally posting the one-time-consumed credential to .../Auth, which fail-closed and rendered a misleading 'Login failed'. Affected both SAML and OIDC linking. Adversarial correctness review confirmed security-neutral; added an ordering assertion (return; before the /Auth/ post) to lock the terminality. (Closed manually, merged to 4.3, not the default branch.)

PR #1434, opened 2026-08-26
Fix self-service account link falling through to a login POST

kernicek:
# What & why

`WebResponse.Generator`'s auth page unconditionally wipes `localStorage`'s
`jellyfin_credentials` and reloads an iframe to re-bootstrap them before
reading `jfUser`/`jfToken` for the `/Link/` POST. For the linking leg this is
counterproductive: the currently signed-in user's session (needed to make the
authenticated Link call) is destroyed first, and the iframe reload may not
repopulate an authenticated (non-anonymous) session before the wait loop
moves on. When it doesn't, `link()` sees `jfUser == null`, resolves to
`undefined`, and the page falls through to the ordinary `/Auth/` login POST
instead - which can then be refused for reasons that only make sense for a
first-time login (e.g. `AllowExistingAccountLink` being off), even though the
user was already authenticated and only trying to link.

Capture `jellyfin_credentials` from `localStorage` before the wipe when this
is a linking request, and use those captured credentials for the `/Link/`
POST directly - skipping the wipe-and-reload dance entirely, since it exists
to bootstrap a fresh session for the login leg and linking already has a live
one. Falls back to the original wipe-and-reload login path only when no
usable session was actually held to begin with.

Found this debugging a real deployment: a signed-in administrator's
self-service link at `/SSOViews/linking` was silently landing on the
name-adoption refusal (`AllowExistingAccountLink` off, admin accounts never
name-adopted) instead of ever reaching `/Link/`. No issue filed yet - happy
to open one and cross-reference if that's preferred over a bare PR.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. No signature/time-bound/audience
      check is touched - every server-side gate on `/Link` and `/Auth`
      (rate limiting, `AllowExistingAccountLink`, the admin no-name-adoption
      rule) is unchanged. This only changes which client-side request the
      auth page sends and with which already-held credentials.
- [x] No secrets logged; secrets are redacted on config export. No logging
      added or changed; `jfToken` flows through the same `Authorization`
      header construction as before, never printed.
      config persistence, or the release pipeline. Not performed - flagging
      rather than claiming it. The change doesn't touch SAML/OIDC token
      validation, config persistence, or the release pipeline, only which
      leg of the *already-reviewed* `/Link` vs `/Auth` endpoints the page
      posts to and with what credentials - but I'd rather a maintainer's eyes
      confirm that scoping than assert it myself.
- [ ] Review record: per lens, its verdict and its findings... Not performed
      (see above) - no CONFIRMED/PLAUSIBLE record to report honestly.
- [x] Log inputs from the IdP are sanitized inline at the log call. N/A - no
      logging touched by this change.

## Quality checklist

- [ ] No duplicated logic; new logic lives in a small, single-purpose,
      testable unit. Partially - the fast path re-derives
      `deviceId`/`appName`/`appVersion`/`deviceName` under `link*`-prefixed
      names rather than sharing the assembly with the fallback login path's
      identical block further down. Kept it a straight duplication rather
      than refactoring the shared shape into a helper, to keep the diff
      minimal and the literal strings the existing test suite asserts on
      (`if (true) {` / `if (false) {`, `if (linkStatus !== undefined) {`,
      the `return`-before-`/Auth/` ordering) untouched. Open to extracting a
      shared helper if that's preferred - said more in Notes.
- [x] Architecture-conformance tests pass (confirmed locally, see
      Verification) - this fix doesn't establish a new structural property,
      so no new `ArchitectureConformanceTests` rule was added.
- [x] Docs impact considered: no README/wiki update needed - this restores
      the behavior the Login-Flow wiki page already documents for
      self-service linking; it was never accurately describing what the code
      did.

## Verification

- [ ] `dotnet build --no-restore --warnaserror` is green. Verified for
      **net9.0 only** (`-p:TargetFrameworks=net9.0 --no-restore
      --warnaserror`): 0 warnings, 0 errors. I don't have the .NET 10 SDK
      available locally, so I could not run the unmodified multi-target
      command or verify the net10.0 leg - deferring that to CI.
- [ ] `dotnet test` is green. Couldn't get `dotnet test` (the CLI wrapper) to
      report results in my local environment - it builds successfully but
      reports no discovered/run tests, on both a fresh run and
      `--no-build --verbosity normal`, which looks like a local
      Microsoft.Testing.Platform/`dotnet test` integration quirk with SDK
      9.0.317 rather than anything about the change. Ran the built xUnit v3
      test host directly instead (`SSO-Auth.Tests` binary), which does
      report results: **3405/3410 passed**. The 5 failures are all
      pre-existing `SamlResponseTests`/`SamlSecondaryCertificateTests`/
      `SamlLogoutRequestTests` SHA-1-signature cases, unrelated to this
      change - they fail locally because Fedora's OpenSSL build refuses to
      even *produce* a SHA-1 signature for the test fixture
      (`digital envelope routines::invalid digest`), not because the
      plugin's rejection logic is wrong. All 19 `WebResponseTests` pass,
      including the two most relevant to this change
      (`Generator_EmitsIsLinkingAsJsBooleanLiteral`,
      `Generator_SuccessfulLink_IsTerminal_...`).
- [ ] Prettier is clean for any `.js` / `.html` / `.md` / `.css` change. N/A -
      only `SSO-Auth/Api/Flows/WebResponse.cs` changed (the JS lives inline
      as a C# raw string literal there); no standalone `.js`/`.html`/`.md`/
      `.css` file was touched.

## Notes

- Rebased onto current `main` after forking (picked up the `iderex` →
  `Flowfin` URL-string cleanup commits; no conflicts, nothing near this
  file).
- Open to extracting the duplicated device-info assembly (fast path vs.
  fallback login path) into a shared helper if that's the preferred shape -
  kept it inline for this PR to minimize the diff against the existing
  literal-string test assertions and make the review surface small.
- Found and reproduced against a live deployment (Authelia as OIDC
  provider), where a signed-in administrator's self-service link at
  `/SSOViews/linking` was reproducibly falling through to `/Auth/` and
  getting refused. Confirmed end-to-end after deploying this exact build:
  the same administrator's self-service link now reaches `/Link/` and
  completes successfully.

### DIFF
--- a/SSO-Auth/Api/Flows/WebResponse.cs
+++ b/SSO-Auth/Api/Flows/WebResponse.cs
@@ -297,12 +297,9 @@ function showReturnLink() {
 const ssoBaseUrl = " + JsonSerializer.Serialize(punycodeBaseUrl) + @";
 const ssoProvider = " + JsonSerializer.Serialize(provider) + @";
 const ssoMode = " + JsonSerializer.Serialize(mode) + @";
-async function link(request) {
-    const jfCredentialsString = localStorage.getItem(""jellyfin_credentials"");
+async function link(jfCredentials, request) {
+    if (jfCredentials == null) return;
 
-    if (jfCredentialsString == null) return;
-
-    const jfCredentials = JSON.parse(jfCredentialsString);
     const jfUser = jfCredentials['Servers'][0]['UserId'];
     const jfToken = jfCredentials['Servers'][0]['AccessToken'];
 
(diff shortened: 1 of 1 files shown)


iderex:
I read the diff in full. The diagnosis is right and the shape is right: the wipe-and-reload exists to bootstrap a fresh session for the login leg, and the link leg needs the session the visitor already holds - capturing it before anything touches localStorage, passing it into `link()` as a parameter, and keeping the terminal semantics of #614 untouched is exactly the repair this deserved. The honest checklist and the measured verification made this a fast review.

One finding, not blocking: the fast-path guard reads `preLinkCredentials['Servers'][0]['UserId']` unguarded. A degenerate stored value - `{"Servers":[]}` from an emptied server list - throws a TypeError there and kills `main()` outright, so neither leg runs. The old code ran the same expression only after the wait loop had guaranteed `Servers[0].Id`, so the exposure is new. A `?.`-style guard (or an explicit length check) in the condition turns that corner from a dead page into the login fallback. Fine as a follow-up; say the word if you'd rather I not hold the merge on it.

I have released the CI runs; the fork's first-contribution hold kept them queued. Once they report I'll merge - and thank you for the deployment-verified report, it is the most useful kind.

iderex:
One more formality before this can land: the DCO check wants a `Signed-off-by` trailer on the commit, and that attestation is yours to give, not mine to add. With a single commit it is two commands:

    git commit --amend --signoff --no-edit
    git push -f

The force-push replaces the head, the released CI re-runs on it, and then I merge.

kernicek:
Fixed the unguarded `Servers[0]` access with optional chaining (`preLinkCredentials['Servers']?.[0]?.['UserId']`), verified `dotnet build --warnaserror` (net9.0, still can't test net10.0 locally) and the WebResponseTests suite (19/19) stay green, and amended the commit with a DCO sign-off. Force-pushed.

iderex:
Merged. Thank you - for the deployment-verified diagnosis, the honest checklists, and for folding the guard fix into the same clean commit within the hour. Contributions of this shape are why the linking page works better tonight than it did this morning. You are welcome back anytime.

kernicek:
Thanks for the quick merge, I was glad to help!

Issue #1543, opened 2026-09-05
An unreadable SSO-Auth.xml is replaced by an empty one, and the evidence is overwritten with it


iderex:
Found while answering #1532, and it is the half a write-side repair does not reach.

## The claim

`SSO-Auth.xml` is replaced by an empty configuration on the next start after it becomes unreadable, and the unreadable file is overwritten in the same act, so the evidence goes with it.

The write this plugin makes is not atomic - `BasePlugin<T>.SaveConfiguration` serializes straight over the file with no temporary copy and no rename - so a disk that fills mid-write leaves a truncated file. That much is #1532, and it is recorded at `SSOPlugin.PersistBase` with the reasons the write is left as it is.

What is recorded there and belongs here is the other end: `BasePlugin<T>.LoadConfiguration` catches every exception, builds a default configuration, and **writes it back over the file**. So the sequence an operator meets is

1. a write fails on a full disk. #1521 rolls the change out of the running server, and the log says so;
2. the file on disk is truncated, and nothing says so, because the process is still running on the stored state it just restored;
3. the server restarts. The load throws, the base class replaces the configuration with a default one and persists it, and every provider, every canonical link and every at-rest secret envelope is gone;
4. the truncated file, which is the only thing that could have been repaired by hand, has been overwritten by the empty one.

Step 4 is why this is not simply "keep a backup". Between step 2 and step 3 the file is damaged but present, and after step 3 it is not.

## Why a temporary file and a rename does not fix it

That repair, which #1532 offers and declines, stops step 2: a rename publishes a complete file or none. It leaves steps 3 and 4 exactly as they are, for any other cause of an unreadable file - a filesystem corruption, an interrupted restore, a hand edit. The destructive act is on the LOAD side and is the host's, and no write-side change reaches it.

So the two are separable, and this is the one with teeth: a write-side rename shipped alone buys the appearance of durability without the fact.

## What could be done here, none of it decided

- Refuse to start on an unreadable configuration rather than replacing it, which is the fail-closed posture this plugin takes everywhere else - and which on this surface takes every SSO login on the server offline until somebody intervenes. That trade is the decision, and it is not a small one.
- Copy the unreadable file aside before the base class overwrites it, so the evidence survives even when the default is accepted.
- Keep a rolling copy of the last configuration that loaded, and say in the log which one is being served.

Whether the plugin may refuse to start is a scope decision rather than an implementation, which is why this issue is filed rather than built.

## Done when

[...]

PR #1604, opened 2026-09-10
Copy the configuration aside when a second copy of this plugin is loaded [#1601]

### LINKED ISSUE #1601: Downgrading from 5.0.0.80 to 5.0.0.75 deletes all provider data
### Plugin version

5.0.0.80

### Jellyfin server version

12.0

### Provider type

OpenID Connect (OIDC)

### Identity provider

Authentik

### What happened, and what did you expect?

1. Updated to .80. Realized that the role folder options were incomplete due to the new UI
2. Downgraded to .75 to see if I could tweak the options there
3. No provider data. Updated back to .80.
4. ???
5. Data's gone. 👍

### Relevant log output

```text

```

### Environment

- [ ] The server runs behind a reverse proxy (please note which in the description)
- [ ] This worked on a previous plugin version (please note which)

iderex:
Closes #1601.

**What happens without this.** A downgrade through the plugin catalog does not replace
the installed version. It adds a second directory beside it, and Jellyfin 12.0 loads
both at the next start. Two copies of this assembly are live at once: every route
registers twice, the configuration type exists twice, and the host, unable to read a
configuration back across them, serves defaults and writes them over the file.

Measured on a live Jellyfin 12.0.0 with the published packages, following the reporter's
sequence:

```
before                  3576 bytes, provider present
downgrade to .75         601 bytes, no provider
back up to .80            38 bytes, no provider
```

The unreadable-configuration screen from #1543 does not catch this, and was never going
to. The file is readable at the moment it is read and is replaced afterwards, so the copy
that screen keeps is a copy of the empty version.

**What this does.** It cannot stop the host loading two copies, and it cannot intercept
the host's own write. It can be early. The check runs in the plugin constructor, before
anything touches `Configuration` and therefore before the lazy load that ends in the
overwrite, so the copy it takes is the file as the operator left it. At most one copy per
configuration file, first one wins: a server in this state boots again and again, and
every later boot would copy a file that is already empty.

Then it refuses. `PersistBase` is the one road to disk, and the refusal goes at the top of
it, before the configuration type check. That order is the fix and not a detail: a
configuration produced by the other loaded copy is not this copy's `PluginConfiguration`,
so the type check does not match it and hands it to the base class unchanged, which is the
write that empties the file. A refusal after it would never run on the shape that does the
damage. Both orderings are pinned by rules rather than by comment, because neither is
visible from the lines it separates.

Detection counts assembly instances rather than distinct paths, because the fault is two
type identities and two copies loaded from one path would still be two. An enumeration
that throws answers "nothing found" rather than guessing: a false negative leaves the
server as it was, a false positive would refuse every configuration write on a healthy
server. That direction is asserted rather than assumed.

**Proof, on the live server rather than in argument.** The published .81 installed through
the catalog, a provider configured, the published .75 installed over it the way the
reporter did it, and this build in the active directory:

```
[SSO Audit] This plugin is loaded TWICE in this server, from
  /config/plugins/Community SSO for Jellyfin_5.0.0.75/SSO-Auth.dll AND
  /config/plugins/Community SSO for Jellyfin_5.0.0.81/SSO-Auth.dll.
The configuration was copied to ... first, and that copy is what to restore from.

live SSO-Auth.xml                              601 bytes, no provider
SSO-Auth.xml.before-duplicate-20260910-115546Z 3576 bytes, provider present
```

Restoring that copy with the second directory removed brings the provider back and the
plugin up Active. Without this change the same sequence leaves nothing to restore from.

**Tests.** Five over the copy rule, including that a second boot keeps the first copy
rather than burying it, and that a process holding one copy reports no duplicate. Two
conformance rules over the two orderings. Full suite green.

**What is still owed.** Two copies of one plugin loading at once is a Jellyfin behaviour
and belongs upstream as its own report. This is the plugin refusing to add to the damage,
not a repair of the host.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.AuditSanitizers.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.AuditSanitizers.cs
@@ -38,6 +38,7 @@ public partial class ArchitectureConformanceTests
         "markerPath",
         "source",
         "sourcePath",
+        "installLocations",
         "composedRefusal",
     };
 
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.DuplicateInstall.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.DuplicateInstall.cs
@@ -1,0 +1,64 @@
+// SPDX-FileCopyrightText: The jellyfin-plugin-sso authors
+// SPDX-License-Identifier: GPL-3.0-only
+
+using System;
+using System.IO;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <content>
+/// Conformance rules for the refusal that stands while a second copy of this plugin is loaded (#1601).
+/// </content>
+public partial class ArchitectureConformanceTests
+{
+    [Fact]
+    public void TheDuplicateRefusalComesBeforeAnythingElseOnTheWritePath()
+    {
+        // The ordering IS the fix, and it is invisible from the two lines it separates. With two copies of
+        // this plugin loaded, a configuration produced by the other copy is not this copy's
+        // PluginConfiguration - the types have different identities - so PersistBase's type check does not
+        // match it and hands it to the base class unchanged. That is the host write, and the host write is
+        // what replaces the operator's providers with defaults. A refusal placed after the type check
+        // therefore never runs on the one shape that does the damage.
+        //
+        // Measured on a live Jellyfin 12.0.0 with the published packages: SSO-Auth.xml went 3576 bytes with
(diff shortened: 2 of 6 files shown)


iderex:
Superseded by #1605, which carries the same change with the coverage gate satisfied.

This one failed the security-surface branch bar at 85.9% against the pinned 86.0%.
The cause was in the new audit line: both of its arguments are non-null where it
stands, one by declaration and one by being past its own null check, so writing them
through a conditional access added six branches on the security surface that no test
could ever take. The gate reads that as the surface getting thinner, and it is right
to. The replacement calls the sanitizer directly, which the sanitizer rule accepts
in the same terms, and covers the four branches that remain.

Replaced rather than amended, per the one-commit-per-change rule and no force-push.

Issue #1602, opened 2026-09-10
The translation files are complete and the page is half English, because 80 strings never reach them


iderex:
## What

On a `de-DE` client the settings surface mixes both languages. Counted on the
Overview tab alone, against a live Jellyfin 12.0.0 with the packaged plugin:

```
21 English text blocks, in otherwise translated prose
```

The translation files are not the problem, and that is the part worth knowing:

```
$ python - <<'P'
de: 214 Schluessel   en: 214 Schluessel
in de fehlend: 0
P
```

Both carry 214 keys and **nothing is missing**. Nothing in the tooling signals a
gap, because by its own measure there is none.

## Where the English actually comes from

`Web/i18n.js` translates through `data-i18n="<key>"` markup and a `t()` function.
Strings assembled in JavaScript can use `t()` and many do. Many do not:

```
$ grep -c "t(" SSO-Auth/Web/sso-core.js
148
$ (English sentence literals not wrapped in t())
80
```

So 148 strings are translatable and 80 are literals in the bundle. A literal
never reaches `de.json`, which is why the file can be complete and the page can
still be half English.

## Why this is worse than a cosmetic gap

Read what falls in the untranslated 80:

```
This provider has an active insecure or sensitive setting.
Use an https:// URL for the OpenID endpoint.
Enter the base URL only (no path), e.g. https://jellyfin.example.com, not the /sso/... route.
Remove control characters (such as a tab or newline, often introduced by copy-paste) from the name.
Remove backslash and URI-reserved characters from the name.
```

They are the **validation and warning messages**. The calm prose an administrator
reads while nothing is wrong is translated; the sentences that tell them what
they did wrong, and how to fix it, are not. A German administrator gets a German
surface until the moment they need help, and English exactly then.

The security warning is in that set, which is the one sentence that must land.

## What is proposed

Not a translation pass. The keys are a consequence; the defect is that a string
can reach the page without passing the mechanism.

- the 80 literals move behind `t()` with keys, and the keys land in both files
- a conformance rule refuses a new English sentence literal in the web bundle,
  so the count cannot climb again quietly. The tree already refuses drift of this
  shape elsewhere, and 214-of-214 completeness is exactly the reassuring number
  that hid this one.

## Done when

The Overview tab carries no untranslated block on a `de-DE` client, `de.json` and
`en.json` still agree, and a rule fails on a sentence literal added outside
`t()`.

## How this was found

[...]

PR #1618, opened 2026-09-10
Pin how many sentences bypass the translation catalog, and take the first eleven [#1602]

iderex:
First step on #1602, which stays open: the drift stops here and eleven of the sentences
are done.

**Why the existing checks see nothing.** `de.json` and `en.json` carry the same keys and
nothing is missing from either. A sentence written as a literal in the bundle never
reaches a catalog, so it cannot be reported as absent from something it was never in.
The gap is invisible to a completeness check by construction.

**What this pins.** `tools/ui-untranslated.js` counts the sentences that bypass the
catalog and refuses the number moving either way. Up is the drift. Down without lowering
the pin leaves slack for the next one to arrive in unseen, so the pin moves in the same
commit as the work.

**The first draft of that tool counted wrong, and the first tranche is what showed it.**
It tested each line on its own, and the formatter breaks a long catalog call across four
lines:

```js
tr(
  "config.validation_endpoint_https",
  "Use an https:// URL for the OpenID endpoint.",
),
```

Line by line that reads as untranslated. The draft pinned 135 on that basis. Wrapping
eleven strings then moved the number by two instead of eleven, which is the only reason
the error surfaced: 62 of the 135 had been through the catalog the whole time. Reading
the file whole gives **73**, and that is what is pinned.

**The eleven** are the field validation messages, which is where #1602 opens and the
sharpest end of it. The calm prose an administrator reads while nothing is wrong was
already translated; the sentences telling them what they did wrong were not.

**One sentence stays literal**, with the reason beside it in the tool: the bootstrap
banner the five page controllers write when the dynamic import of the core did not
arrive. The localization module comes through that same route, so at the moment that
sentence is needed there is no translator to ask. A stale exemption is refused.

**Proved on the rig.** With a `de` client, an invalid endpoint now answers

```
Geben Sie eine absolute URL an, z. B. https://id.example.com
```

where it answered English before.

**Still open on #1602:** the remaining 73, most of them the provider template
descriptions, which are a tranche of their own.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -73,6 +73,15 @@ jobs:
       # the shipped sso-core.js against both client shapes and refuses each arm by name. Same runtime and
       # same absence of dependencies as the two steps above.
       run: node tools/ui-folder-checklist.js
+    - name: Sentences that bypass the translation catalog
+      # de.json and en.json carry the same keys and nothing is missing from either, so every check the
+      # localization has is green while a de-DE administrator reads a page that is half English (#1602). A
+      # sentence written as a literal in the bundle never reaches a catalog, so it cannot be reported as
+      # absent from something it was never in - the gap is invisible to a completeness check by
+      # construction. This pins the COUNT instead, and refuses it moving either way: up is the drift, and
+      # down without lowering the pin leaves slack the next one hides in. Same runtime and same absence of
+      # dependencies as the three steps above.
+      run: node tools/ui-untranslated.js
     - name: VEX document check
       # The published VEX statements are only useful if they parse and carry machine-readable
       # dispositions, so a malformed edit fails here rather than at the release leg that ships the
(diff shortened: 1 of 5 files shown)


iderex:
Superseded by the pair below, and the reason is a rule this tree already holds that I had
not read.

`EveryNonEnglishCatalog_HasExactlyTheEnglishKeySet` means a new key cannot land in English
alone, and the pull-request gate asks for a `Catalogue-Vouch:` line naming a person who
reads the language when a non-English catalogue changes (#1154). This pull request carried
eleven new keys with German I had written myself. Vouching for my own reading of it is the
one thing that rule exists to prevent, so it does not land in this shape.

It also failed `ScriptEnglishDefaults_MatchTheCatalog`, which was worth finding: one of the
eleven printed the reserved characters including a backslash, and a backslash in a JS string
is an escape, so the inline default and the decoded catalog value are never the same
characters. Naming the backslash in words settles it, and that goes with the tranche.

Split rather than fixed in place:

- the ratchet alone, which needs no catalogue and no vouch, so the drift stops now;
- the eleven wraps with both catalogues, in a change that carries a reader's name.

Replaced rather than amended, per the one-commit-per-change rule and no force-push.

Issue #1529, opened 2026-09-04
4.4 stage 3: accounts, the self-service page, and one language per view


iderex:
Stage 3 of the 4.4 UI plan: accounts, the self-service page, and one language per view.

## What to build

- The Accounts page: linked accounts with search and filter, revocation with a confirmation that says what it ends, and pending approvals (accounts created disabled) with an approve action.
- The self-service page at `/SSOViews/linking` in the dashboard look: your sign-ins, link and unlink per provider, a confirmation for unlink, and the same i18n as the dashboard pages.
- One language per view: every string on the five pages and the self-service page goes through i18n, and a script refuses a hard-coded English or German string in the templates.
- Server page: export and import of the configuration and of the account links with the same warnings as today, in one place.

## Done when

- the string script reports no untranslated literal on the six pages
- the self-service page was walked as a non-admin user on the test server: link, unlink, refusal on a wrong provider
- the Accounts page was walked with an adopted account, a pending approval, and a revocation

iderex:
Waiting for: Flowfin/jellyfin-plugin-sso#1528

This stage follows the one before it in the 4.4 plan (#1525); the relation is declared natively and this line says so.

iderex:
The stage-1 walk today put numbers on this stage's third bullet, and it turns
out to be two different defects wearing the same symptom. Both were read off a
live Jellyfin 12 with a de-DE client, two OpenID providers and one SAML
provider, not estimated from the tree.

## Kind one: markup that carries no catalog marker at all

237 text nodes across the five page templates have no `data-i18n`, so they are
English whatever the client's language is.

```
$ # text nodes of three words or more whose enclosing tag carries no data-i18n
accountsPage.html       0
configPage.html        27
policiesPage.html       5
providersPage.html    200
serverPage.html         5
                      237
```

This is the pass that was named on #1602 and left to this stage. It is mostly
the reference prose in the help rails and the accordion headings, and it
includes the Save button of the profile editor, which reads `Save profile` in
English beside a German page.

## Kind two: a catalog call that runs before the catalog is there

This one is worth reading carefully, because it looks translated in the source
and is not translated on the screen.

`i18n.js` holds the catalog in a module variable that is empty until
`loadCatalog()` resolves, and `t()` falls back to the caller's English default
while it is. `localize()` starts that fetch during init. So every `tr()` call
made by code that runs during the same init reads the empty catalog, gets the
[...]

PR #1624, opened 2026-09-11
Send the Policies, Server and Accounts text through the catalog

iderex:
# What & why

Refs #1529, stage 3's third bullet. No closing keyword: the stage has three more
bullets after this one.

Three of the six templates now show nothing the catalog never saw. Eighteen text
runs and five attributes carry a key, the pin comes down with them in this same
commit, and those three files read zero.

```
                     before        after
configPage.html      37  1         37  1
providersPage.html  283 16        283 16
accountsPage.html     0  1          0  0
policiesPage.html    11  1          0  0
serverPage.html       7  3          0  0
linking.html         10  1         10  1
```

The five attributes are the collapse headers, which were the largest
untranslated surface on these pages: a de-DE administrator saw German fields
under an English section title, on every section.

The German is mine and has been read; the vouch line is at the end.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved. Untouched. This adds catalog markers and catalog
      rows; no validation, no persistence, no login path.
- [x] No secrets logged; secrets are redacted on config export. Untouched, and
      the Server page's own description of that redaction is one of the strings
      keyed here, unchanged in wording.
      config persistence, or the release pipeline. Not applicable: none of the
      three is touched. The one property worth asking about is whether a marker
      can destroy markup when the catalog resolves, and that is exactly what
      TextMarkers_OnlySitOnElementsWithoutChildMarkup refuses; it is green here,
      and it would have caught the misplacement described in the notes.
- [x] Review record: **CONFIRMED 2 / PLAUSIBLE 0**, both raised by the guards
      rather than by reading, both fixed before this commit. Disposition: FIX
      for each. They are described in the notes.
- [x] Log inputs from the IdP are sanitized inline at the log call. No logging.

## Quality checklist

      testable unit. No new logic at all: markers and rows.
- [x] The code is self-documenting and matches the target architecture.
      change establishes is locked in. No new property; the five existing
      catalog rules and the two ratchets all apply unchanged and are green.
- [x] Docs impact considered: no README or wiki change. The wording of every
      string is unchanged; only the route it takes to the screen is new.

## Verification

- [x] `dotnet build --no-restore --warnaserror` is green. Both target
      frameworks, 0 warnings.
- [x] `dotnet test` is green. Full suite on both frameworks, 0 failed.

All five browser gates run green.

## Notes

Two findings, and both are worth naming because neither was visible by reading
the diff.

**The count moved by 8 when 18 runs had been keyed.** Ten markers had landed one
element too high. An anchor that begins with its own closing bracket, such as
`>Profile:</label`, sits AT the end of its tag, so searching backwards for a
bracket finds the one belonging to the element ABOVE. The marker went onto a
`<div class="inputContainer">` that has child elements, which renders correctly
until the catalog resolves and then replaces the whole container with a line of
text. The ratchet refused the mismatch, and the rule about markers on elements
with child markup would have refused it next.

**A second key carried an English value the catalog already had.**
`EnglishCatalog_HasNoDuplicateValues` refused "Export / Import Account Links",
which the accounts page already has under `config.link_transfer_title` for the
heading right below that collapse header. The existing key is reused rather than
a twin invented, because two keys with one English value are two rows nobody can
translate apart.

The German follows the terminology the catalogue already uses rather than a
fresh invention: exportieren and importieren, as the account-link strings beside
them have said since they were written, and Erkundungsdokument for the discovery
document. The first draft of the Server page said "Konfiguration ausfuehren" for
Export, which in German reads as EXECUTING the configuration rather than
exporting it, on a button that does neither.

The German has been read. What the reader was asked to look at was the nineteen
new rows, and in particular the three long help texts and the wording chosen for
Export and Import, where the first draft said ausfuehren and would have read as
executing the configuration rather than exporting it.


iderex:
Closed in favour of #1626, which carries this work unchanged on a base that can
take it.

This branch keyed eighteen text runs and five attributes against a template pin
of 348. The sentence-parts mechanism landed in the meantime and moved that pin to
343, and it appended to both catalogs at the same place this branch does, so the
two disagree on three files:

```
$ git merge-tree --write-tree origin/5.0 origin/fix/the-three-small-pages-reach-the-catalog-1529
CONFLICT (content): Merge conflict in SSO-Auth/Localization/de.json
CONFLICT (content): Merge conflict in SSO-Auth/Localization/en.json
CONFLICT (content): Merge conflict in tools/ui-untranslated-markup.js
```

A rebase would have rewritten this branch under a reading that had already
happened, so it is replaced rather than force-pushed. The German in #1626 is the
same German that was read here, and its vouch says so.

Issue #1637, opened 2026-09-11
A pending-approval record outlives the account being enabled


iderex:
The pending-approval record (#1529) says this plugin provisioned a linked account disabled and awaiting an administrator. It is removed with its link, and cleared by every other write of the same key, so it cannot come to describe an account this plugin did not provision inert. What it does not survive contact with is the account being ENABLED afterwards.

Nothing in this plugin enables an account. The only route today is the Jellyfin user page, which this plugin does not see, so the record is written once at provisioning and then stands for the life of the link:

1. `ProvisionNewUsersDisabled` is on. Subject `S` logs in; the account is created disabled (`SSO-Auth/Api/Linking/CanonicalLinkService.cs:704`) and the record is written beside the link.
2. An administrator enables that account by hand.
3. It signs in normally for months. The existing-link arm writes no link, so the record is never revisited.
4. The administrator later disables it as a sanction.
5. The roster still reports the record, because it still names the account the link points at.

The consequence is the one the record exists to prevent, reached by another road: a surface that reads "has a record" as "is awaiting approval" would offer to undo the sanction in step 4.

## What closes it

The surface that reads this map owes its reader what is TRUE NOW rather than what was true at provisioning:

- an account that is currently enabled is not listed, whatever the map says;
- the approve action re-reads the account's state inside the same transaction in which it enables it, so a state change between the page load and the click cannot be acted on.

A cheap narrowing beside that, in this plugin rather than on the page: clear the record on any login that actually mints a session. Passing the awaiting-approval gate proves the account is no longer inert, so a record that survives it is already false. That reduces the window to "provisioned inert and never signed in since", which is what the list is supposed to be.

## Provenance

Found by the security review on the pending-approval record, as the half the record's own binding does not cover. Named in that change's commit message so the next change does not inherit it silently.

iderex:
One more shape of the same residual, found by the adversarial pass on the panel (#1644) and deferred here because it has the same cure.

[...]

PR #1641, opened 2026-09-11
Approve an account this plugin provisioned inert, and grant nothing else

iderex:
# What & why

Stage 3 of #1529: the approve action for accounts this plugin provisioned disabled, on top of the record that landed in #1639. One endpoint, one permission, three refusals.

It acts on the plugin's own RECORD of having provisioned an account inert and never on the Jellyfin disabled flag, because the flag does not say who set it or why. An account an administrator disabled deliberately carries no record, so this route cannot see it, cannot list it and cannot enable it - which is the one outcome the threat model on the issue said this feature must not have.

What it does is one permission. No role, no folder, no library, no policy: the provisioning decided those when it created the account, and an approve that also granted would be a second provisioning policy that nobody configured, hidden in a button. An administrator account is refused outright and keeps its record; an unknown provider stays distinct from an identity that is not pending, so the refusal is not an oracle for provider names.

The state is re-read at the moment of acting. Between the page load and the click the account may have been enabled by hand, deleted or promoted; each has its own arm, and the two that prove the record false take it away. A successful SSO login now clears a standing record as well (#1637), placed after the pending-approval gate so a waiting user's own refused logins cannot remove their own row.

No UI yet: the Accounts-page list and its button are the next change, and the endpoint lands first so the surface PR carries no server risk.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. The action fails closed on every arm
      that cannot prove the record: no record, a record naming another account,
      an unknown provider and an administrator all change nothing.
      carries the actor, the account id, the protocol and the provider, and not
      the subject.
- [x] Log inputs from the IdP are sanitized inline at the log call. The one new
      log call strips line endings inline; the audit line takes no IdP input.

### Review record

**Pass 1, the security review of the committed diff.** An identification pass over the full diff and its surroundings raised no candidate finding, so the false-positive filtering stage had nothing to run on. The paths it closed, each with a citation in the review output: a forged record (JsonIgnore, both Preserve arms, the import never references the record); a moved record (PendingApproval.Live requires the record to name the account the link points at, read under the config lock); the window between the locked read and the user write (only administrator-driven actions can move a link inside it, and a login for the same subject is refused at the pending gate before any write); granting more than the enable (the only mutation before UpdateUserAsync is the one flag, and the three IsDisabled writes on the seam are the only ones in the tree); reaching it without elevation (the authorize filter runs before the limiter); the login-time clear (runs only after a minted session, clears only the login's own key); the administrator basis (the resolved account, never a claim); the subject (absent from the audit line and from every response body).

**Pass 2, an adversarial verification of the same tree, attacking each of those claims independently.** Every lens SAFE: no forged or moved record, no grant beyond the enable, no reach below elevation, no way to drive the login-time clear against another account, the administrator basis is the resolved account, and the subject reaches neither the audit line nor a response body. It raised two LOW findings and three weak test pins, all **FIX**, in this commit:

- Approve could re-enable an account this plugin's own expiry sweep or role denial had just disabled, because no disable path cleared the record; the account went straight back onto the list, and one press undid the denial. Now every disable this plugin performs clears the record, which is false by then - a disable presupposes enabled. The guard on that path returns before any write for an already-disabled account, so a genuinely pending account's denied login never reaches the clear. Driven both ways.
- The record removal after the enable was unconditional, so in the window between the locked read and the write a NEW account provisioned under the same freed key could lose its record to a decision about the old one. Now a compare-and-remove: the record goes only if it still names the account that was read. Driven from inside the user read.
- The refused-login row asserts the refusal is the pending gate's; the grants-nothing row holds a permission the provisioning did grant and asserts the routing survives; both were weaker than their names.

**Residual, named rather than hidden.** An account provisioned inert, enabled by hand, that never signs in through SSO and is later disabled by hand as a sanction still carries a live record, so the list would offer it and Approve would re-enable it. Every mitigation here needs an observation event that sequence never produces. It takes an administrator pressing the button, so it is a footgun for the administrator holding the mouse rather than an escalation for an attacker. **DEFER**, it stays on #1637 narrowed to exactly that window.

This PR replaces #1640, which carried the same change without the two fixes; it was never merged and is closed by being superseded here.

## Quality checklist

      testable unit. The rule that decides whether a record still describes
      anything has one home, `PendingApproval.Live`, and both readers use it.
      The disabled-flag rule now counts the writes it permits: two that set the
      flag and exactly one that clears it, on the one seam. The new route is
      classified as throttled and as a lookup of a stored provider name, and the
      elevation roster names it.
- [x] Docs impact considered: the endpoint has no administrator-facing surface
      until the list lands, and that change documents the feature as a whole.

## Verification

      warnings.
- [x] `dotnet test` is green. 3901 on net9.0 and 3902 on net10.0, 0 failed, 4
      skipped (the loopback-bind rows that need an administrator on Windows), on
      an idle machine.
      change touches none.

Fourteen tests drive the service and the login path rather than reading them, and they were run against six broken builds before being trusted: acting on the link instead of the record fails the two negatives, the administrator guard fails its own row, the login clear fails one row when removed and a different one when moved before the gate, the disable path's clear fails its row, and the removal without its compare fails the interleaving row.

## Notes

- The provisioning's own audit line now points at the Accounts tab as well as at the dashboard, so whoever reads the trail is told where the button is.
- The canonical name travels in the request body, like the pre-provision write beside it: a subject or a NameID may contain a slash, and a route segment would refuse exactly those identities.
- Next: the Accounts-page list of accounts waiting for approval, with this action behind its button, and the roster read that feeds it.

### DIFF
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.Controller.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.Controller.cs
@@ -296,7 +296,7 @@ public partial class ArchitectureConformanceTests
         // this expected count in the same PR that adds or removes a rate-limited endpoint (as the provider-form
         // roster rules do), so a change to the limiter surface is a conscious update here rather than a silent
         // drift the offender scan cannot see.
-        const int expectedTypedCallSites = 18;
+        const int expectedTypedCallSites = 19;
         var typedCall = new Regex("RateLimitCheck\\(\\s*SsoRateLimitClass\\.");
         var typedCallSites = ControllerSourceFiles()
             .Sum(path => typedCall.Matches(File.ReadAllText(path)).Count);
--- a/SSO-Auth.Tests/ArchitectureConformanceTests.LoginPath.cs
+++ b/SSO-Auth.Tests/ArchitectureConformanceTests.LoginPath.cs
@@ -88,29 +88,59 @@ public partial class ArchitectureConformanceTests
     {
         // Locked in by #737. IsDisabled is a lockout vector: the plugin deliberately never disabled an
         // account until the pending-approval provisioning feature, and it is barred from SSO role mapping
-        // (PermissionRolePolicy) so no login can disable an EXISTING account. The one sanctioned write -
-        // provisioning a BRAND-NEW account inert for admin approval - must stay confined to
-        // CanonicalLinkService (the single create seam). A source scan pins that: any future
-        // SetPermission(PermissionKind.IsDisabled, ...) elsewhere (a mint path, a role mapper, a controller)
-        // would reopen the "an SSO login disabled my account" surface and fails here instead of shipping.
+        // (PermissionRolePolicy) so no login can disable an EXISTING account. The sanctioned writes must
+        // stay confined to CanonicalLinkService (the single seam that owns the flag). A source scan pins
+        // that: any future SetPermission(PermissionKind.IsDisabled, ...) elsewhere (a mint path, a role
+        // mapper, a controller) would reopen the "an SSO login disabled my account" surface and fails here
+        // instead of shipping.
+        //
+        // THE COUNT IS PART OF THE RULE since #1529 put the first ENABLING write on that seam. Two writes
+        // set the flag - the arm that provisions a brand-new account inert (#737), and the one shared
+        // disable that login-time deprovisioning (#831) and account expiry (#1144) both call, which carries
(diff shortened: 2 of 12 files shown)


iderex:
Superseded by #1642: the same change, with the boundary tests the security-surface branch bar asked for (this one fell a tenth below it). Not merged. Replaced rather than amended so nothing published was rewritten.

Issue #1643, opened 2026-09-11
Document the Waiting for Approval panel on the wiki's Accounts page


iderex:
The Accounts page gains a panel that lists the accounts this plugin provisioned disabled and still waiting, with an Approve button beside each (#1529, the surface over #1639 and #1642). The wiki's Accounts page and the provisioning section that today send an administrator to the Jellyfin dashboard to enable such an account need a paragraph each:

- what the panel lists, and the three conditions a row has to meet: this plugin's own record of having provisioned the account inert, the record still naming the account the link points at, and the account still being disabled;
- what it does NOT list, and why: an account somebody disabled on purpose has no record and is enabled in the dashboard, never from here;
- what Approve does and does not do: it enables the account and changes nothing else; an administrator account is refused;
- the bound: the first hundred rows are shown and the panel says how many were cut.

The help text on the panel already carries this at the point of use; the wiki is the second home the single-home doctrine allows for a feature description, and the dashboard-only instruction it holds today is no longer the whole truth.

Wiki edits are gated on the go from the account that owns the wiki, so this waits for that.

iderex:
Landed on the wiki: the Linked Accounts page carries the Waiting for Approval section, the orphan-row paragraph as it stands after #1649 and the privileges sentence for the third route; the Hardening and Options Reference names both approval routes under Provision new users pending approval. The help text on the provider form is the one dashboard-only sentence left; it has its own issue, #1655.

PR #1644, opened 2026-09-11
List who is waiting for approval, and put the button beside the name

iderex:
# What & why

Stage 3 of #1529: the Accounts page lists the accounts this plugin provisioned disabled and still waiting, with an Approve button beside each name. The surface half of the pending-approval bullet, over the record (#1639) and the action (#1642).

The rows are the server's answer and nothing else. The roster reports a link as waiting only when three things hold at once: this plugin's own record says it provisioned the account inert, the record still names the account the link points at, and the account is still disabled. The page infers nothing from a disabled flag or a missing login, because the one row it must never contain is the account an administrator disabled on purpose - and the help text says so, and says where that account is enabled instead.

The third condition is new here and closes the page half of #1637: the roster withholds the row for an account somebody has since enabled, so a working account never sits on a list of accounts that cannot sign in - and, once such an account is later disabled as a sanction, never sits there with a button that undoes it. The disabled flag rides along with the username in the one resolver the roster already had.

The button sends exactly the request the endpoint parses and the confirmation names what it does and what it does not. A success re-reads the roster rather than editing the table. The two 403s - the endpoint's administrator refusal and the elevation policy's - are told apart by the body the endpoint writes for its own, and no other server value is reflected anywhere. The list is bounded at a hundred rows and says how many were cut.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. Nothing here decides; the page renders
      what the server reports and drives an endpoint that re-reads the state.
      The one new server-side read, the disabled flag on the roster, only ever
      withholds a row.
- [x] No secrets logged; secrets are redacted on config export. The roster
      gains a boolean; nothing new is logged.
      config persistence, or the release pipeline. Not one of the four, but the
      page drives a grant, so it had an adversarial pass of its own; record
      below.
      __ / PLAUSIBLE __** counts as raised. Detail below.
      call.

### Review record

**An adversarial pass over the panel and the roster's new condition** - the sinks every rendered value reaches, the request the button builds for each protocol, whether the page infers "pending" from anything but the server's field, the two 403s, the bound, the held roster, both catalogs, the gate's own arms, and any change to what a non-pending row reports - is running as this PR opens. Its record follows as a comment before the merge, and the merge waits for it.

## Quality checklist

      testable unit. The roster's condition lives in `LinkRoster.Build`; the
      page reads one field.
      The page's properties are decisions rather than structure, and the
      running proof in `tools/ui-pending-approvals.js` is their rule: which
      rows are built, what the button sends, and which sentence is written for
      a success, a cut list, a stale row, an administrator and a bare 403.
- [x] Docs impact considered: the panel's help text carries the whole of what a
      reader needs at the point of use, and the wiki's Accounts page needs a
      paragraph for the new panel - filed as its own `documentation` issue on
      this milestone, linked in Notes.

## Verification

      warnings.
- [x] `dotnet test` is green. 3918 on net9.0 and 3919 on net10.0, 0 failed, 4
      skipped (the loopback-bind rows that need an administrator on Windows), on
      an idle machine.

Every UI gate passes, including the new one: the string ratchet stays at zero on all six templates, the field register still counts 124 (the panel adds no form control), and the pending-approvals proof runs its ten arms against the shipped renderer.

## German

The new rows, quoted for the reading. The word that carries the panel is the one the existing provider option already uses for the same act, and the remaining short rows are each shaped on their linked-accounts neighbour.

```text
config.pending_approvals_title:  Warten auf Freigabe
config.pending_approvals_help:   Konten, die dieses Plugin deaktiviert angelegt hat, weil ihr Anbieter neue Benutzer deaktiviert anlegt, und die noch warten. Die Freigabe aktiviert das Konto und ändert sonst nichts: Seine Berechtigungen sind die, die es beim Anlegen bekommen hat. Ein Konto, das ein Administrator deaktiviert hat, erscheint hier nicht, in welchem Zustand auch immer, und wird im Jellyfin-Dashboard aktiviert. Wird zusammen mit den verknüpften Konten oben gelesen; Verknüpfte Konten neu laden liest beides neu.
config.pending_approvals_empty:  Kein Konto wartet auf Freigabe.
config.pending_approvals_column_identity / _since:  Identität / Wartet seit
config.pending_approvals_truncated:  Gezeigt werden die ersten {shown} von {total} wartenden Konten. Geben Sie einige frei oder widerrufen Sie sie, um die übrigen zu sehen.
config.pending_approvals_approve:  Freigeben
config.pending_approvals_confirm:  {user} freigeben? Das aktiviert das Konto, sodass es sich über {provider} anmelden kann. Sonst ändert sich nichts: Die Berechtigungen bleiben so, wie das Anlegen sie gesetzt hat.
config.pending_approvals_approving / _approved:  {user} wird freigegeben… / Freigegeben. {user} kann sich jetzt anmelden.
config.pending_approvals_refused_administrator:  Dieses Konto ist ein Administrator und wird nicht von hier freigegeben. Aktivieren Sie es im Jellyfin-Dashboard.
config.pending_approvals_not_pending:  Dieses Konto wartet nicht mehr auf Freigabe. Die Liste wurde neu gelesen.
config.pending_approvals_failed:  Das Konto konnte nicht freigegeben werden. Stellen Sie sicher, dass Sie als Administrator angemeldet sind, und versuchen Sie es erneut.
```

## Notes

- #1643 carries the wiki paragraph for the panel, on this milestone; the dashboard-only instruction the wiki holds today is no longer the whole truth.
- The residual on #1637 is unchanged by this change: an account enabled by hand that never signs in and is then disabled by hand again still carries a live record and will be listed. It is narrower than before and still open.
- Next: the self-service page walk as a non-admin user, and the Server page's export and import in one place, which are the last two bullets of the stage.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -108,6 +108,15 @@ jobs:
       # shipped renderer and refuses both by name. Same runtime and same absence of dependencies as the
       # steps above.
       run: node tools/ui-account-filter.js
+    - name: The pending-approvals panel
+      # The panel decides which accounts an administrator is offered to ENABLE, and every property that
+      # matters is a decision rather than a string (#1529): which rows are built, what the button sends,
+      # and what the page says when the server refuses. A row the server did not report would be an
+      # offer to enable an account nobody provisioned inert; a body that lost a slash would refuse the
+      # subject that carries one; a bare 403 read as the administrator refusal would send a signed-out
+      # reader to the dashboard for the wrong reason. This drives the shipped panel against a recording
+      # ApiClient and refuses each by name. Same runtime and same absence of dependencies as the steps above.
+      run: node tools/ui-pending-approvals.js
     - name: VEX document check
       # The published VEX statements are only useful if they parse and carry machine-readable
       # dispositions, so a malformed edit fails here rather than at the release leg that ships the
--- a/SSO-Auth.Tests/Config/LastSsoLoginStampTests.cs
+++ b/SSO-Auth.Tests/Config/LastSsoLoginStampTests.cs
@@ -339,7 +339,7 @@ public class LastSsoLoginStampTests
         config.CanonicalLinks["never"] = User;
         config.CanonicalLinkLastLogins["seen"] = Now;
 
-        var document = LinkRoster.Build(live, _ => "alice");
+        var document = LinkRoster.Build(live, _ => new LinkedAccountState("alice", false));
 
         var links = Assert.Single(document.Accounts).Links;
         Assert.Equal(Now, Assert.Single(links, l => l.CanonicalName == "seen").LastSsoLoginUtc);
--- a/SSO-Auth.Tests/Config/PendingApprovalRecordTests.cs
+++ b/SSO-Auth.Tests/Config/PendingApprovalRecordTests.cs
@@ -338,7 +338,7 @@ public class PendingApprovalRecordTests
         config.CanonicalLinks["sub-ordinary"] = User;
         config.CanonicalLinkPendingApprovals["sub-pending"] = Record(User);
 
-        var document = LinkRoster.Build(configuration, _ => "alice");
+        var document = LinkRoster.Build(configuration, _ => new LinkedAccountState("alice", true));
 
         var links = Assert.Single(document.Accounts).Links;
         Assert.Equal(Now, Assert.Single(links, l => l.CanonicalName == "sub-pending").PendingApprovalSinceUtc);
@@ -359,7 +359,7 @@ public class PendingApprovalRecordTests
         config.CanonicalLinks["sub-1"] = User;
         config.CanonicalLinkPendingApprovals["sub-1"] = Record(Other);
 
-        var document = LinkRoster.Build(configuration, _ => "alice");
+        var document = LinkRoster.Build(configuration, _ => new LinkedAccountState("alice", true));
 
         Assert.Null(Assert.Single(Assert.Single(document.Accounts).Links).PendingApprovalSinceUtc);
     }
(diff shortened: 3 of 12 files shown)


iderex:
Review record, the adversarial pass promised in the body. Nine lenses over the panel and the roster's new condition, all SAFE: every rendered value reaches textContent and nothing else, and the tr() fallback uses a function replacer so a replacement pattern in a subject is inert; the request names the row's own link, with the protocol token the server parses, the provider encoded in the path and the canonical name as the JSON body; the page infers "pending" from the server's field alone and LinkRoster.Build never reports an instant for an enabled or missing account; no non-administrator 403 the plugin writes can carry the endpoint's sentence; the bound is applied before rendering and an uncut list carries no line; a failed re-read replaces both panels rather than leaving a stale list; every new key is in both catalogs with identical placeholders; no arm of the gate passes for the wrong reason; and no non-pending row reports anything it did not before.

CONFIRMED 0 / PLAUSIBLE 0 at the bar. Four LOW notes below it, none of which makes the page send a wrong request:

- A 204 is also the endpoint's answer for a record it cleared because the account had gone, and the page then says a deleted account can sign in. FIX, in the follow-up PR that opens next: the sentence is chosen from the re-read roster.
- The bound and the count are over links while the sentence says accounts. FIX, same follow-up: one row per account.
- The administrator refusal is told apart from a bare 403 by the word alone, which a proxy's own 403 page can carry. FIX, same follow-up: matched on the endpoint's sentence.
- If a row's account is deleted and the same subject is re-provisioned inert before the click, the request enables the new account while the status line names the old username. The endpoint re-derives everything under its lock, so nothing wrong is enabled; only the sentence is stale. DEFER, noted beside the residual on #1637.

Merging on the green run; the three fixes follow within the hour.

PR #1645, opened 2026-09-11
Let the approvals panel say what the answer was, and count what it says

iderex:
# What & why

Three sentences the adversarial pass on #1644 caught saying the wrong thing, on the Waiting for Approval panel of #1529. None of them made the page send a wrong request; each told a reader something untrue.

- A 204 is also the endpoint's answer for a record it cleared because the account had gone, and the panel said a deleted account could sign in now. The sentence is now chosen from the re-read roster, which is the server's word on whether the account still exists: an account the re-read reports as gone gets its own line, and only an account still there is reported as approved.
- The bound and the count were over links while the sentence said accounts. One row per account now, on its first waiting link.
- The administrator refusal was told apart from a bare 403 by the word "administrator" alone, which a proxy's own 403 page can carry about something else. It is matched on the endpoint's own sentence now.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      rejected, never default-accepted. Nothing server-side changes; the page
      still sends the same request and the endpoint still decides.
- [x] No secrets logged; secrets are redacted on config export. Nothing new is
      logged.
      config persistence, or the release pipeline. Not one of the four; these
      are the fixes that review asked for, recorded on #1644.
      0 / PLAUSIBLE 0** as raised. The three findings are the LOW notes on #1644,
      each **FIX** here; the fourth stays **DEFER** on #1637.
      call.

## Quality checklist

      testable unit.
      The three properties are decisions of the page, and the running proof in
      `tools/ui-pending-approvals.js` gains an arm for each: one account waiting
      through two providers draws one row; a 403 whose body merely contains the
      word stays the generic sentence; a 204 for an account the re-read reports
      as deleted says so and never that it can sign in. Thirteen arms.
- [x] Docs impact considered: no README or wiki text describes these sentences;
      #1643 carries the wiki paragraph for the panel.

## Verification

- [x] `dotnet test` is green. The catalogue rules pass on both catalogs; the
      change is otherwise browser-side.

Every UI gate passes, the string ratchet stays at zero, and the field register still counts 124.

## German

One new sentence, quoted for the reading:

```text
config.pending_approvals_account_gone:  Dieses Konto existiert nicht mehr, es wurde also nichts aktiviert. Die Liste wurde neu gelesen.
```

## Notes

- The vouch line for the catalogue is deliberately absent until the sentence above has been read; it is added on the reading, and the PR-hygiene check re-runs on the edit.

### DIFF
--- a/SSO-Auth/Localization/de.json
+++ b/SSO-Auth/Localization/de.json
@@ -528,6 +528,7 @@
   "config.pending_approvals_confirm": "{user} freigeben? Das aktiviert das Konto, sodass es sich über {provider} anmelden kann. Sonst ändert sich nichts: Die Berechtigungen bleiben so, wie das Anlegen sie gesetzt hat.",
   "config.pending_approvals_approving": "{user} wird freigegeben…",
   "config.pending_approvals_approved": "Freigegeben. {user} kann sich jetzt anmelden.",
+  "config.pending_approvals_account_gone": "Dieses Konto existiert nicht mehr, es wurde also nichts aktiviert. Die Liste wurde neu gelesen.",
   "config.pending_approvals_refused_administrator": "Dieses Konto ist ein Administrator und wird nicht von hier freigegeben. Aktivieren Sie es im Jellyfin-Dashboard.",
   "config.pending_approvals_not_pending": "Dieses Konto wartet nicht mehr auf Freigabe. Die Liste wurde neu gelesen.",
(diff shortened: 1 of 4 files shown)


iderex:
Superseded by #the-panel-agrees-with-its-panels-1529, Head ref must be a branch (createPullRequest), which carries these three fixes plus the one a second pass found (a 204 followed by a failed re-read was worded off the old roster). Not merged; replaced rather than amended. The German sentence is unchanged.

iderex:
Correction to the closing note above, which carried a mangled reference: the replacement is #1646.

Issue #1649, opened 2026-09-11
Prune every link and record of a Jellyfin account the moment the host reports it deleted


iderex:
Decided (2026-09-11): when a Jellyfin account is deleted, this plugin removes every link, deadline, last-login stamp and pending-approval record the account holds, in the moment the host reports the deletion, and the roster's orphan row for such an account is replaced by an audit line.

## Why

A link whose target account was deleted counts as absent everywhere in this plugin, so the next login for the same subject writes the key again at another account. That dangling link is the root under three things found today: the pending-approval residual on #1637, the inherited deadline of #1638, and the first step of #1647. Pruning at the source removes the root rather than guarding each map that hangs off it.

The host offers the hook: `UserManager.DeleteUserAsync` publishes `UserDeletedEventArgs`, which an `IEventConsumer<UserDeletedEventArgs>` registered by the plugin's service registrator receives. (Checked against the host's source: creation, deletion, rename and password change publish; enabling or disabling an account publishes nothing.)

## What it costs, and the decision on it

The roster keeps a link whose account is gone on purpose (#1119): an orphaned link is what an administrator opens the roster to find. Pruning on deletion ends those rows. The decision is to prune anyway and to write the fact into the audit trail instead: one line per deleted account naming the protocol and provider of each link removed and the account id, never the subject (T-I1). What an administrator loses is a row they could act on; what they gain is a state that cannot inherit.

## What to build

- `IEventConsumer<UserDeletedEventArgs>` in `Api/Events`, registered beside the hosted services, calling `RemoveUserEverywhere` for the deleted id under the existing lock, with the same fail-closed treatment the login-event publish has: a failure is logged and never thrown back into the host's delete.
- The audit line, at Warning, on the same seam as the other link-removal lines.
- Tests that drive the consumer with a deleted id holding links on both protocols, a deadline, a stamp and a record, and assert all four maps are clean and the audit line names the provider and not the subject.
- The roster comment and the wiki text about orphan rows updated to what is now true: an orphan row can still appear for an account deleted while the plugin was not loaded, and the import still refuses one.

iderex:
[...]

Issue #1647, opened 2026-09-11
A time-limited account can shed its deadline by unlinking itself and signing in again


iderex:
A time-limited account can shed its provisioned access deadline (#1146) without touching any write that #1638 guards, by unlinking itself and signing in again. Pre-existing; found by the re-verification of the #1638 fix as the route that change does not cover.

1. A role mapping gives subject `S` four hours. The first login creates account `G` with `CanonicalLinkDeadlines[S]`.
2. `G` calls the self-service unlink, `DELETE {mode}/Link/{provider}/{G}/{S}`, which any authenticated user may call for their own account. The link goes, and the deadline goes with it, as every removal route prunes it (`SSO-Auth/Api/Linking/CanonicalLinkService.cs`, `TryRemoveLink`). The token revocation on the last link does not stop a fresh SSO login.
3. `S` signs in again. The account exists under its username, so with `AllowExistingAccountLink` on it is ADOPTED, and the adopt arm passes no duration by design (#1146: adoption takes over an account with its own history, and a guest role must not put an expiry on an established account). No deadline is written. `G` now has unlimited access, and the sweep never names it again.

Without `AllowExistingAccountLink` the re-login is refused, so the default configuration is not exposed; a provider that allows adoption is.

## What could close it

Three shapes, and which one is right is a decision rather than a fix:

- refuse the self-service unlink for a link that carries a deadline (the administrator can still unlink; the time-limited user cannot use the unlink as an exit from the limit);
- keep the deadline on adoption when the login carries a mapped duration and the account holds no other link (this re-opens the #1146 question of putting a limit on an existing account, narrowed to an account that has no link at all);
- record the deadline per ACCOUNT rather than per link, so an unlink does not erase it and a re-link finds it (a larger change to the map's shape).

The first is the smallest and keeps #1146's decision intact.

iderex:
Landed in #1650 as the first of the three shapes: a link that carries a provisioned access deadline is removed only on an administrator's word, decided inside the same transaction as the removal, after the ownership check; the holder is refused before anything is touched, with a 403 that names the deadline. Two independent review passes on the change raised nothing at the bar. The other two shapes stay recorded above as what was not chosen.

Issue #1638, opened 2026-09-11
An expired deadline outlives the account it was written for, and disables the next one


iderex:
Two server-managed maps keyed by canonical name, the provisioned access deadlines (#1145) and the last-SSO-login stamps (#1120), are pruned on every route that REMOVES a link but not by the routes that WRITE one over a key that already holds an entry. A key can change hands without ever being removed: a link whose target account was deleted counts as absent (`SSO-Auth/Api/Linking/CanonicalLinkService.cs:1928`), so the next login for that subject writes the key again at a different account.

The deadline is the one with teeth, because acting on it disables an account.

1. A role mapping gives subject `S` four hours of access. `S` logs in, account `U1` is created, `CanonicalLinkDeadlines[S]` is stamped.
2. The deadline passes and the sweep disables `U1`.
3. An administrator deletes `U1`. Nothing prunes the deadline: this plugin registers no user-deletion hook, and the link is deliberately kept so the roster can report it.
4. `S` logs in again and is adopted into, or provisioned as, `U2`. `LinkCanonicalIfAbsent` repoints the link; `StampProvisionedDeadlineInPlace` returns early because this login carries no duration (the adopt arm always passes none), so the stale deadline stands.
5. `ExpiredLinks` resolves the account from the CURRENT link map (`:1183`), so the next sweep tick names `U2` and disables an account that is minutes old, with the audit line naming an expiry nobody granted it.

`RemoveDeadline` already names this hazard in its own comment: a re-link of the same subject would inherit the previous holder's deadline and be swept immediately. It is prevented on removal and not on rebind, and the deleted-account route walks around the removal.

The last-SSO-login stamp has the same shape and a smaller consequence: until the new account's own first successful sign-in, the roster shows it the previous holder's last SSO login, which is one account's data on another's row.

## What closes it

Make the deadline stamp a set-or-clear, the way the pending-approval record's write now is (#1529): inside the same transaction that writes the link, a login that carries no duration REMOVES the entry rather than leaving the previous one. Then the same on the last-login stamp. One line each, plus a test per route that drives the rebind rather than reading the code, because a removal-only rule looks correct right up to the write that inherits.

## Provenance

Found while closing the identical shape on the pending-approval record. That map is now bounded over the ACCOUNT and not only over the key; these two are still bounded over the key alone.

iderex:
[...]

PR #1651, opened 2026-09-11
Take a deleted account's links with it, on the host's own event

iderex:
# What & why

A deleted Jellyfin account takes its links with it, the moment the host reports the deletion (#1649). A link whose target account was deleted counts as absent everywhere in this plugin, so the next login for the same subject wrote the key again at another account with whatever the key still carried. That dangling link was the root under three things found on one day: the pending-approval residual (#1637), the inherited deadline (#1638), and the first step of #1647. Each map now guards itself on rebind; this removes the state that made a rebind possible.

The host publishes `UserDeletedEventArgs` from its own delete, after its commit, and its event manager hands it to every registered consumer of that closed type. The consumer builds the link service per event over the plugin's live store, the way the expiry sweep does, and drives the existing revoke seam the elevation-gated Unregister uses: every link on both protocols, and with them the issuer binding, the deadline, the last-login stamp and the pending-approval record, in one transaction. One audit line at Warning names the protocol and provider of every link removed and the account id, never the subject and not the username either.

What it costs was decided on the issue: the roster's orphan row does not appear for an account deleted while the plugin was loaded; the audit line is what replaces it. An account deleted while the plugin was not loaded still leaves an orphan, and the roster still shows it; its remark says so now.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. Nothing here grants; a failure to prune
      leaves every link exactly as it was, and the next login of the same subject
      clears it as before.
      carries the account id, a count and administrator-typed provider names,
      sanitized inline at the log call; the warning carries the id alone.
      reaches either line.

### Review record

**Pass 1, the security review of the commit**, raised nothing at the bar and answered the question the change turns on from the host's own source: `UserManager.DeleteUserAsync` removes the row and saves inside its lock and publishes the event only after both are disposed, so the plugin never prunes the links of an account that still exists; the only callers of that delete are the host's elevation-only user endpoint and this plugin's own rollback of an account the same login had just created, whose id is fresh. No HTTP or webhook path publishes host events, and the plugin's own publisher carries a different closed type. The removal matches by user id, which the host never reuses, and the follow-on prunes drop only entries whose key holds no link at all, so nothing here can disable a live account. The prune and a login of the same subject run under the same store lock, and in either order the new account's link survives. The audit line and the warning carry the id, a count and provider labels; the sanitizers sit inline at the log call, the shape every helper in that file takes. The registration matches the host's own consumers, the scope exists per publish, and the loggers and services the constructor asks for are ones the registrator's neighbours already take. The consumer runs after the commit inside the host's awaited publish with exceptions caught on both sides, and a failed persist restores the snapshot.

**Pass 2, an adversarial verification of the same tree** attacking each of those claims independently, is running as this PR opens; its record follows as a comment before the merge, and the merge waits for it.

## Quality checklist

      testable unit. The consumer is one class over the existing revoke seam
      and one audit helper; the DI shell and the testable core are the two
      constructors the expiry sweep already uses.
      The consumer lives in Linking, whose module edges already allow what it
      imports; the host's event types are host types, not Api modules.
- [x] Docs impact considered: the roster's own remark carries the change in what
      an orphan row now means; the wiki's Accounts page sentence about orphan
      rows is folded into #1643's documentation pass for this milestone.

## Verification

      warnings.
- [x] `dotnet test` is green. 3931 on net9.0 and 3932 on net10.0, 0 failed, 4
      change touches none.

Five tests drive the consumer, each proven against its own broken build before being trusted: with the removal skipped, the audit dropped, and the failure rethrown, each break fails the rows that name it.

## Notes

- Enabling or disabling an account publishes nothing in the host (checked against its source), which is why this is the one transition the plugin can observe and why #1637's remaining window is closed by the readers checking what is true now rather than by an event.
- The walk server can show this live: delete a linked account in the dashboard and the roster row is gone on the next read, with the audit line in the log.

### DIFF
--- a/SSO-Auth.Tests/Linking/DeletedAccountLinkPrunerTests.cs
+++ b/SSO-Auth.Tests/Linking/DeletedAccountLinkPrunerTests.cs
@@ -1,0 +1,136 @@
+// SPDX-FileCopyrightText: The jellyfin-plugin-sso authors
+// SPDX-License-Identifier: GPL-3.0-only
+
+using System;
+using System.Threading.Tasks;
+using Jellyfin.Data.Events.Users;
+using Jellyfin.Plugin.SSO_Auth.Api.Linking;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using MediaBrowser.Controller.Library;
+using Microsoft.Extensions.Logging;
+using NSubstitute;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// A deleted Jellyfin account takes its links with it (#1649). The host publishes its deletion, and this
+/// plugin's consumer removes every link the account held on both protocols, with the deadline, the stamp
+/// and the pending-approval record that hung off them, and writes one audit line naming the providers and
+/// the account id and never the subject. Nothing it does may reach the host's delete: a store that throws
+/// is a warning, not a failure the host sees.
+/// </summary>
+public class DeletedAccountLinkPrunerTests
+{
+    private static readonly Guid Gone = Guid.Parse("11111111-1111-1111-1111-111111111111");
+    private static readonly Guid Other = Guid.Parse("22222222-2222-2222-2222-222222222222");
+    private static readonly DateTime Now = new(2026, 9, 11, 9, 0, 0, DateTimeKind.Utc);
+
+    [Fact]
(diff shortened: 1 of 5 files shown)


iderex:
Superseded by #1652, the same consumer with the three notes of its adversarial pass fixed: no configuration write for an account that held no link, the no-links test asserting every entry, and the host's own resolution of the consumer driven through a real container. Not merged; replaced rather than amended.

PR #1654, opened 2026-09-11
Take a deleted account's links with it, on the host's own event

iderex:
# What & why

A deleted Jellyfin account takes its links with it, the moment the host reports the deletion (#1649). A link whose target account was deleted counts as absent everywhere in this plugin, so the next login for the same subject wrote the key again at another account with whatever the key still carried. That dangling link was the root under three things found on one day: the pending-approval residual (#1637), the inherited deadline (#1638), and the first step of #1647. Each map now guards itself on rebind; this removes the state that made a rebind possible.

The host publishes `UserDeletedEventArgs` from its own delete, after its commit, and its event manager hands it to every registered consumer of that closed type. The consumer builds the link service per event over the plugin's live store, the way the expiry sweep does, and drives the existing revoke seam the elevation-gated Unregister uses: every link on both protocols, and with them the issuer binding, the deadline, the last-login stamp and the pending-approval record, in one transaction. One audit line at Warning names the protocol and provider of every link removed and the account id, never the subject and not the username either.

What it costs was decided on the issue: the roster's orphan row does not appear for an account deleted while the plugin was loaded; the audit line is what replaces it. An account deleted while the plugin was not loaded still leaves an orphan, and the roster still shows it; its remark says so now.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

      rejected, never default-accepted. Nothing here grants; a failure to prune
      leaves every link exactly as it was, and the next login of the same subject
      clears it as before.
      carries the account id, a count and administrator-typed provider names,
      sanitized inline at the log call; the warning carries the id alone.
      reaches either line.

### Review record

**Pass 1, the security review of the commit**, raised nothing at the bar and answered the question the change turns on from the host's own source: `UserManager.DeleteUserAsync` removes the row and saves inside its lock and publishes the event only after both are disposed, so the plugin never prunes the links of an account that still exists; the only callers of that delete are the host's elevation-only user endpoint and this plugin's own rollback of an account the same login had just created, whose id is fresh. No HTTP or webhook path publishes host events, and the plugin's own publisher carries a different closed type. The removal matches by user id, which the host never reuses, and the follow-on prunes drop only entries whose key holds no link at all, so nothing here can disable a live account. The prune and a login of the same subject run under the same store lock, and in either order the new account's link survives. The audit line and the warning carry the id, a count and provider labels; the sanitizers sit inline at the log call, the shape every helper in that file takes. The registration matches the host's own consumers, the scope exists per publish, and the loggers and services the constructor asks for are ones the registrator's neighbours already take. The consumer runs after the commit inside the host's awaited publish with exceptions caught on both sides, and a failed persist restores the snapshot.

**Pass 2, an adversarial verification of the same tree** attacking each of those claims independently: every lens SAFE, on the same evidence and one more - the plugin's own rollback of a just-created account is the one non-elevated path to the host's delete, and the id it deletes holds no link by construction. It raised three notes, none a security finding, all **FIX** here: the revoke seam persisted the configuration on every deletion, including the many accounts that never had a link and the plugin's own rollback, where a full disk would have warned that links stayed which never existed - an account with no link now returns before the seam runs, driven over a store that throws; the no-links test asserted only the audit marker's absence and would have passed had that path thrown - it asserts every entry now; and the wiring, the one host-facing seam, ran in no test - a real container filled by the registrator now resolves the consumer from a scope through the public constructor and drives an event through it. A wording note on the roster's remark is corrected as well: an orphan row also appears when the prune's persist failed.

**Pass 3, a focused re-verification of that fixed tree**, every lens SAFE on the host's source: the sole publisher of the event and its elevation-only caller, the plugin's own rollback as the one non-elevated path whose fresh id holds no link, the match by Guid value under the one static lock, the DI shape against the host's own consumers, the consumer inside the awaited publish after the commit. Two notes, none a security finding, both **FIX** here: the providers the audit line named were read in lock acquisitions of their own before the removal's, so an administrator's Unregister racing the host's delete could have left a line naming a provider nothing was removed from - the revoke seam now takes an optional collector it fills inside the same lock as the removal, only where a link came out, the Unregister passes none and is unchanged, and a row moves the ground between the guard's read and the removal and asserts the line names one link on one provider; and the DI-resolution test claimed to run with no plugin instance up, which no test can promise (`SSOPlugin.Instance` is process-wide and no harness resets it) - it now carries an id no test links, hands the container a typed logger, and asserts the outcome both states share. Two weak assertions went with it: a vacuous `NotNull` on a logger dropped, the count the line carries asserted.

**Pass 4, the lenses those notes came from re-run on that tree**: the seam change, the atomicity, the sanitizers and the analyzers all SAFE, with the count and the labels shown to come from the one acquisition that removes and persists, a failed persist discarding the filled collector with the restore, and the closed `ILogger<T>` registration shown to win over `AddLogging()` in either order. Three notes on the tests, none on the code, all **FIX** here: the DI-resolution row's "both states" premise was false for two harness shapes that leave a plugin with a null configuration behind, where the row's event drive logs the very warning it forbids (reproduced end to end) - the row now stands a real plugin instance up in the non-parallel collection, seeds a link in its live store and asserts the link gone and the one line through the logger the container wired; the race row could not tell names read inside the removal's lock from names read in acquisitions of their own between guard and removal - it pins the acquisition count now; and the store-throws row did not pin what its warning claims - it asserts no line reporting a removal and the link still in place.

**Pass 5, the test lens re-run on this tree**: its record follows as a comment before the merge, and the merge waits for it.

This PR replaces #1653 (#1652 and #1651 before it), each carrying the consumer without the fixes of the pass that reviewed it; none was merged, all closed by being superseded here.

## Quality checklist

      testable unit. The consumer is one class over the existing revoke seam
      and one audit helper; the DI shell and the testable core are the two
      constructors the expiry sweep already uses.
      The consumer lives in Linking, whose module edges already allow what it
      imports; the host's event types are host types, not Api modules.
- [x] Docs impact considered: the roster's own remark carries the change in what
      an orphan row now means; the wiki's Accounts page sentence about orphan
      rows is folded into #1643's documentation pass for this milestone.

## Verification

      warnings.
- [x] `dotnet test` is green. 3942 on net9.0 and 3943 on net10.0, 0 failed, 4
      change touches none.

Seven tests drive the consumer, each proven against its own broken build before being trusted: with the removal skipped, the audit dropped, the failure rethrown, the seam naming every provider it touched, the event path throwing, the registration missing, a read of its own before the removal, and a line reported from the failure path, each break fails the rows that name it.

## Notes

- Enabling or disabling an account publishes nothing in the host (checked against its source), which is why this is the one transition the plugin can observe and why #1637's remaining window is closed by the readers checking what is true now rather than by an event.
- The walk server can show this live: delete a linked account in the dashboard and the roster row is gone on the next read, with the audit line in the log.

### DIFF
--- a/SSO-Auth.Tests/Linking/DeletedAccountLinkPrunerTests.cs
+++ b/SSO-Auth.Tests/Linking/DeletedAccountLinkPrunerTests.cs
@@ -1,0 +1,186 @@
+// SPDX-FileCopyrightText: The jellyfin-plugin-sso authors
+// SPDX-License-Identifier: GPL-3.0-only
+
+using System;
+using System.Threading.Tasks;
+using Jellyfin.Data.Events.Users;
+using Jellyfin.Plugin.SSO_Auth.Api.Linking;
+using Jellyfin.Plugin.SSO_Auth.Config;
+using MediaBrowser.Controller.Library;
+using Microsoft.Extensions.Logging;
+using NSubstitute;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// A deleted Jellyfin account takes its links with it (#1649). The host publishes its deletion, and this
+/// plugin's consumer removes every link the account held on both protocols, with the deadline, the stamp
+/// and the pending-approval record that hung off them, and writes one audit line naming the providers and
+/// the account id and never the subject. Nothing it does may reach the host's delete: a store that throws
+/// is a warning, not a failure the host sees.
+/// </summary>
+public class DeletedAccountLinkPrunerTests
+{
+    private static readonly Guid Gone = Guid.Parse("11111111-1111-1111-1111-111111111111");
+    private static readonly Guid Other = Guid.Parse("22222222-2222-2222-2222-222222222222");
+    private static readonly DateTime Now = new(2026, 9, 11, 9, 0, 0, DateTimeKind.Utc);
+
+    [Fact]
+    public async Task ADeletedAccount_LosesEveryLinkAndEverythingThatHungOffThem()
+    {
+        var (pruner, configuration, _) = Build();
+        var oid = configuration.OidConfigs["kc"];
+        var saml = configuration.SamlConfigs["idp"];
+
+        await pruner.OnEvent(new UserDeletedEventArgs(TestUsers.Named("alice", Gone)));
+
+        Assert.False(oid.CanonicalLinks.ContainsKey("sub-gone"));
+        Assert.False(saml.CanonicalLinks.ContainsKey("nameid-gone"));
+        Assert.Empty(oid.CanonicalLinkDeadlines);
+        Assert.Empty(oid.CanonicalLinkLastLogins);
+        Assert.Empty(oid.CanonicalLinkPendingApprovals);
+        Assert.Empty(oid.CanonicalLinkIssuers);
+
+        // The neighbour is untouched: this is keyed on the deleted account, not a sweep.
+        Assert.Equal(Other, oid.CanonicalLinks["sub-other"]);
+    }
+
+    [Fact]
+    public async Task TheAuditLine_NamesTheProvidersAndTheAccount_AndNotTheSubject()
+    {
+        var (pruner, _, audit) = Build();
+
+        await pruner.OnEvent(new UserDeletedEventArgs(TestUsers.Named("alice", Gone)));
+
+        var entry = Assert.Single(audit.Entries, e => e.Message.Contains("was deleted", StringComparison.Ordinal));
+        Assert.Equal(LogLevel.Warning, entry.Level);
+        Assert.Contains("[SSO Audit]", entry.Message, StringComparison.Ordinal);
+        Assert.Contains(Gone.ToString(), entry.Message, StringComparison.OrdinalIgnoreCase);
+        Assert.Contains("its 2 SSO link(s) from", entry.Message, StringComparison.Ordinal);
+        Assert.Contains("OpenID 'kc'", entry.Message, StringComparison.Ordinal);
+        Assert.Contains("SAML 'idp'", entry.Message, StringComparison.Ordinal);
+        Assert.DoesNotContain("sub-gone", entry.Message, StringComparison.Ordinal);
+        Assert.DoesNotContain("nameid-gone", entry.Message, StringComparison.Ordinal);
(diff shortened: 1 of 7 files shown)


iderex:
Live on the walk server (Jellyfin 12.0, this build as 5.0.0.84), driven through the host's own endpoints: a throwaway account created, an OpenID link pre-provisioned for it, the roster read, the account deleted with `DELETE /Users/{id}`, the roster read again.

- Roster before the delete: one row for the account, `OpenID 'pocket'`, canonical name as pre-provisioned.
- `DELETE /Users/{id}`: 204.
- Roster after the delete: no row for the account.
- Log, the two lines about it:

```
[WRN] Jellyfin.Plugin.SSO_Auth.Api.Http.SSOController: [SSO Audit] Canonical link pre-provisioned by walk: OpenID 'pocket' -> Jellyfin user 65310084-79b3-4a42-a47b-af49a2311b94, with no identity-provider response redeemed.
[WRN] Jellyfin.Plugin.SSO_Auth.Api.Linking.DeletedAccountLinkPruner: [SSO Audit] Jellyfin user 65310084-79b3-4a42-a47b-af49a2311b94 was deleted; removed its 1 SSO link(s) from OpenID 'pocket'. Any deadline, last-login stamp and pending-approval record went with them.
```

The consumer is resolved by the host's event manager, not by anything this plugin starts, so the line is the host's delete reaching the plugin's store.

iderex:
**Pass 5, the test lens re-run on this tree** (the merge waited for it): every item SAFE.

The wiring row is hermetic and deterministic: the seeded link survives plugin construction because the harness runs its configure callback before the serializer stub is wired, the persist under the substitute serializer is a no-op that keeps the live object, every class that constructs a plugin sits in the one collection whose definition disables parallelization, nothing on the path logs through the wired logger except the audit helper, and the doors rule and the module-mirror rule both accept the new file. The acquisition count is exact: each read and each mutation calls the store's factory once, the guard short-circuits after the OpenID read, and every plausible wrong build breaks either the count or the exact line. The store-throws row does test the restore: the store adopts its snapshot back onto the live object before rethrowing, and the row reads through that object rather than a captured provider.

Three notes, none a change I make here. A pre-existing assertion in the no-links row (the link still present) can never be the failing one, because the empty-entries assertion before it catches every write first; it stays as the post-condition the row is about. One host type in the wiring row is written fully qualified where a using would do; cosmetic. And the persist of any harness-backed row leaves a temporary directory nothing removes; pre-existing and shared with every such row, not introduced here.

Issue #1627, opened 2026-09-11
A browser keeps the plugin's scripts after an upgrade, because nothing says how long they live


iderex:
The plugin serves its browser assets with no caching headers at all, so a
browser decides on its own how long to keep them. After an upgrade an
administrator can go on running the previous version's JavaScript for an
unbounded time, with nothing on the page saying so.

Read off a live Jellyfin 12.0.0:

```
$ curl -s -D - -o /dev/null ".../web/ConfigurationPage?name=SSO-Auth-core.js" -H "Authorization: ..."
HTTP/1.1 200 OK
```

No `Cache-Control`, no `ETag`, no `Last-Modified`. With none of the three a
browser falls back to heuristic caching, which is exactly what it sounds like:
it keeps the response for as long as it likes.

## How it showed up

Twice in one day, while walking a change on a throwaway server. A rebuilt
`i18n.js` was dropped in and the server was restarted, and the page went on
importing the previous copy. The first reading said the change did nothing. The
change was fine; the browser was answering from its own store. A fresh import
under a different URL showed the new module immediately, which is what named the
cause:

```
> (await import(ApiClient.getUrl("SSOViews/i18n.js"))).applyTo   // old module
> (await import(ApiClient.getUrl("SSOViews/i18n.js") + "?v=1")).applyTo   // new module
```

## Why it matters beyond a walk

The same thing happens to an administrator who upgrades the plugin. The route
that serves these files does not change when the plugin version does, so a
cached copy has nothing to invalidate it. A fix shipped in a release can
therefore be installed on the server and absent from the browser, and the only
cure a reader has is a hard reload nobody tells them to do. That is worst for
the corrective releases, which is when it matters most: the library checklist
fix of this line, for instance, is a change to a script.

## What to decide

Two shapes, and the second is the one to want.

- Send `Cache-Control: no-cache` with these responses. A browser then revalidates
  every time, and with an `ETag` most revalidations are a 304. Simple, and costs
  a round trip per asset per page load.
- Put the plugin version in the URL of each asset and let them be cached for a
  long time. The markup already names each script, so the names come from one
  place; the cost is that the names are built rather than written.

Both are in the server-side page registration rather than in the assets, so
neither touches what the pages say.

## Done when

- a response for a plugin asset carries a header that decides its lifetime, and
  which one is chosen is written down with the reason
- the same upgrade walk shows the new asset without a hard reload

iderex:
[...]

Issue #253, opened 2026-07-13
Serve embedded view assets with an ETag so clients can revalidate


iderex:
## Problem

The embedded view assets are streamed with no caching headers, so clients re-download them on every linking-page load.

`SSOViewsController` (`Views/SSOViewsController.cs:62-70`) returns `File(GetManifestResourceStream(...), mime)` with no `ETag` or `Cache-Control`; `jellyfin-apiClient.esm.min.js` (~79 KB) and `emby-restyle.css` (~10 KB) are fetched fresh each visit. Server cost is negligible (manifest resources are memory-mapped); the waste is client bandwidth and page latency on the linking flow.

## Fix

Pass an `entityTag` derived from the assembly version to the `File(...)` overload (the assets only change with the plugin version), about three lines, enables 304 revalidation across the board.

PR #1659, opened 2026-09-11
Say how long a browser may keep a plugin asset, on both routes that serve one

iderex:
# What & why

The plugin served its browser assets with no lifetime at all (#1627). Its own route, `SSOViews/*`, has carried a version-derived `ETag` since #253 but never the header that makes a browser ask; the host's `web/ConfigurationPage` action, which serves the five pages, their page scripts and `sso-core.js`, sends no validator and no lifetime for any plugin's page. A browser then decides on its own how long to keep them, and after an upgrade an administrator can go on running the previous release's script for an unbounded time with nothing on the page saying so.

One answer, two routes. `PluginAssetVersion` holds it: the plugin's file version as a strong tag, and `no-cache`, which lets a browser keep an asset but makes it ask before using it, so the answer is a 304 until the plugin changes. `SSOViews/*` stamps both. For the host's action the plugin registers one result filter with the host's MVC options, the one seam a plugin has into an action it does not own: `PluginPageCacheFilter` acts on that action only and only for names this plugin registers, stamps the tag and `no-cache`, and answers a matching `If-None-Match` with 304 so a tab switch does not re-download 300 KB of core script. Everything else passes through exactly as the host built it. The shape was decided on the issue, with the STRIDE notes the review holds against.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

Not on that list, but a filter registered with the host's options runs on every action of the server, so the review ran with the availability lens first.

      rejected, never default-accepted. Nothing here grants: a 304 carries no
      body on an action reachable without authentication today, and a failure
      anywhere inside the filter leaves the host's response as built.
- [x] No secrets logged; secrets are redacted on config export. The warning
      carries no request value; the tag is the plugin's file version, exposed
      the same way by `SSOViews/*` since #253.
      config persistence, or the release pipeline. Performed for this one all
      the same; record below.
      reaches the line.

### Review record

**Pass 1, the adversarial review of the first cut** (#1658), every lens SAFE on the host's source and MVC's: the filter is skipped where an authorization filter short-circuits and costs one type check elsewhere; a replaced result is executed by the invoker and the client-error filter leaves a 304 alone; the plugin's registration precedes the host's MVC setup and neither the host's option setups nor its `AddMvc` clear the filter list; the host's action and the plugin's own route are both anonymous today, so the 304 grants nothing; the 304 carries the validators, no body and no content type. Two notes and three housekeeping items, all **FIX** here: `If-None-Match` was compared strongly, where RFC 9110 requires the weak comparison and MVC's own executor on the plugin route uses it, so a tag an intermediary weakened would have been served in full on one route and validated on the other - weak now, with the row flipped and the reasoning corrected (equal tags mean equal bytes; a `W/` prefix comes from a recompressed body, not a different asset); the production page-name source was executed by no row - a row in the non-parallel collection now stands a real plugin instance up and drives the resolved filter over a page it registers and one it does not; the 304 is now in place before the stream is released; `PluginAssetVersion` has its mirrored test file; two doc comments naming the old field are reworded. One design note declined with reason: wrapping the filter's construction in a factory that yields a no-op on failure would extend the guarantee to a path that cannot fail today (one public constructor over a logger the host always has); noted, not built.

**Pass 2, the two lenses those notes came from re-run on this tree**: its record follows as a comment before the merge, and the merge waits for it.

This PR replaces #1658, which carried the first cut; it was not merged and is closed by being superseded here.

## Quality checklist

      testable unit. One static answer, one filter, one line in the view route.
      Both new types live in `Api/Http`, the web boundary; no new module edge.
- [x] Docs impact considered: the behaviour is what an administrator expects
      of an upgrade and needs no page of its own; the reason for the shape is
      written on #1627 and in `PluginAssetVersion`.

## Verification

      warnings.
- [x] `dotnet test` is green. 3959 on net9.0 and 3960 on net10.0, 0 failed, 4
      change touches none.

Thirteen rows drive the filter, one in the non-parallel collection drives the resolved instance over a real plugin, two pin the shared answer and one the view route, each proven against a broken build before being trusted: one build with the weak tag refused, the action name ignored, the failure rethrown, the registration dropped and the view route's header dropped fails exactly the rows that name those; a second build whose name comparison can never match fails exactly the rows that need a stamp; a third with the resolved filter reading the view list instead of the page list and the header changed to no-store fails the wiring row and the header rows; every other row stays green on all three.

## Notes

- Live on the walk server (Jellyfin 12.0, this branch built and stamped the way the release build stamps it, as 5.0.0.85 before the review's fixes and as 5.0.0.86 with them), driven with curl: `web/ConfigurationPage?name=SSO-Auth-core.js` answers 200 with `Cache-Control: no-cache` and `ETag: "5.0.0.85"`; the same request with `If-None-Match: "5.0.0.85"` answers 304 carrying both validators; with `If-None-Match: "5.0.0.84"`, the tag a browser holds from the previous build, it answers 200 with the new asset, which is the upgrade walk without a hard reload. The page itself and `SSOViews/i18n.js` answer the same way, and on 5.0.0.86 a weakened tag, `If-None-Match: W/"5.0.0.86"`, answers 304 on both routes, which is the review's first note closed live. TMDb's page, served by the same host action, carries neither header.
- The tag changes with every release because the release build (JPRM) writes the version into `<FileVersion>` before it builds; checked on the released `5.0.0-JF12-beta.82` artifact, whose assembly carries file version 5.0.0.82. A local `dotnet publish` carries the fixed 5.0.0, which is why the walk build was stamped by hand.
- The host could carry the validators itself for every plugin's page; that is a change to Jellyfin, not to this plugin, and is not proposed here.

### DIFF
--- a/SSO-Auth.Tests/Http/PluginAssetVersionTests.cs
+++ b/SSO-Auth.Tests/Http/PluginAssetVersionTests.cs
@@ -1,0 +1,34 @@
+// SPDX-FileCopyrightText: The jellyfin-plugin-sso authors
+// SPDX-License-Identifier: GPL-3.0-only
+
+using Jellyfin.Plugin.SSO_Auth.Api.Http;
+using Xunit;
+
+namespace Jellyfin.Plugin.SSO_Auth.Tests;
+
+/// <summary>
+/// The one lifetime answer both asset routes give (#1627), pinned independently of either route: the tag is
+/// the plugin's file version, which the release build stamps per release, and the header is the one that
+/// makes a browser ask.
+/// </summary>
+public class PluginAssetVersionTests
+{
+    [Fact]
+    public void TheTag_IsThePluginsFileVersion_Quoted_AndStrong()
+    {
+        // Recomputed here from the same source, the SSO-Auth assembly's FILE version, so a regression to
+        // the assembly version (static across releases, #253) or to a weak tag is caught.
+        var fileVersion = System.Diagnostics.FileVersionInfo.GetVersionInfo(typeof(SSOPlugin).Assembly.Location).FileVersion;
+
+        Assert.Equal("\"" + fileVersion + "\"", PluginAssetVersion.ETag.ToString());
+        Assert.False(PluginAssetVersion.ETag.IsWeak);
+    }
+
+    [Fact]
+    public void TheHeader_MakesABrowserAsk_NotForget()
+    {
+        // no-cache keeps the asset and revalidates it; no-store would throw it away and re-download 300 KB
+        // of core script on every page load, and a max-age would only bound the stale window.
+        Assert.Equal("no-cache", PluginAssetVersion.CacheControl);
+    }
+}
--- a/SSO-Auth.Tests/Http/PluginPageCacheFilterTests.cs
+++ b/SSO-Auth.Tests/Http/PluginPageCacheFilterTests.cs
@@ -1,0 +1,262 @@
+// SPDX-FileCopyrightText: The jellyfin-plugin-sso authors
+// SPDX-License-Identifier: GPL-3.0-only
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Jellyfin.Plugin.SSO_Auth.Api.Http;
+using MediaBrowser.Controller;
+using MediaBrowser.Controller.Library;
+using MediaBrowser.Model.Cryptography;
+using Microsoft.AspNetCore.Http;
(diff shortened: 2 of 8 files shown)


iderex:
Superseded by #1660, the same change with the three wording notes of its second pass fixed: the class remark and the warning now say the response stands as it was at that moment, a vacuous assertion is dropped, two stale comments reworded. Not merged; replaced rather than amended. The second pass's record this body promised as a comment is carried in the replacement's body.

Issue #1664, opened 2026-09-11
Stage 2, slice 3: readiness once, in the right column


iderex:
Waiting for: Flowfin/jellyfin-plugin-sso#1663

Stage 2 of the 4.4 UI plan (#1528), the fourth of six slices: readiness answered once, in the right column.

## What to build

- The per-editor readiness panel (#1083, the `OidReadiness` and `SamlReadiness` sections of `SSO-Auth/Web/providersPage.html`) moves into the rail card `#sso-rail-readiness`, which today only says "open a provider". It updates on every field change through the `refreshReadiness` seam in `SSO-Auth/Web/sso-core.js`, without a save.
- The editor keeps a one-line pointer where the panel was, so a reader of the form knows where the answer went. On narrow screens the rail-first order shows the card above the form.
- The Test Connection result stays where #1083 put it unless the pull request shows a better place in the rail; the decision is written in the pull request body.

## Done when

- readiness is rendered in one place, the rail, for the open provider
- changing a required field updates the rail's list without a save
- `tools/ui-mock-fields.js` and `tools/ui-folder-checklist.js` stay green
- a walk on the walk server, wide and narrow, with screenshots on the pull request

## Rules of the road

One commit per pull request, replaced rather than amended after review; the review gate before the merge; Prettier clean; a change to `de.json` puts `Catalogue-Vouch: de read by iderex` into the pull request body after the German has been read.

iderex:
Freed (2026-09-12): the slice this one waits for has its code in the mainline. `SSO-Auth/Web/providersPage.html` now condenses all 111 of its `*_help` field descriptions and carries the rail's help card, landed as PR #1673, merge commit `7e2494776bc1da18aa1246da4f10be47a39093d0`. #1663 itself stays open on one Done-when, a walk on the walk server with a screenshot per form, which is not a dependency of this slice: the rail this slice rearranges exists.

Three things from that slice that belong to this one rather than being rediscovered.

**The rail now holds two cards, and their order is this issue's question.** The help card was put BELOW the readiness card, because readiness answers about the provider being edited and does not move while a form is filled in, while the help card changes on every focus. The markup says so where a reader meets it and names this issue. Nothing was measured; if the walk says otherwise, this is the slice that moves it.

[...]

Issue #1528, opened 2026-09-04
4.4 stage 2: the provider wizard and the condensed help


iderex:
Stage 2 of the 4.4 UI plan: the provider wizard and the condensed help.

## What to build

- Add provider as a wizard: protocol and template, the required fields, the redirect or ACS URL with a copy button, Test Connection, enable and save. Each step validates before the next opens; leaving the wizard keeps nothing unsaved without saying so.
- Help condensed on every page: one sentence under the field, the full text behind Details, and the same full text in the right column when the field is focused. Nothing is deleted, only moved; a script counts that every help text of the old page is present once.
- The readiness checklist lives in the right column and updates as fields change.
- Sensitive and Insecure options sit behind a fold that names what it hides, and an enabled one is shown on the provider's card on Overview.

## Done when

- a provider can be added from Overview to a green Test Connection without opening any accordion
- the help count script reports every old text present once
- the wizard and the folds pass the e2e flows and a walk on the test server

iderex:
Waiting for: Flowfin/jellyfin-plugin-sso#1527

This stage follows the one before it in the 4.4 plan (#1525); the relation is declared natively and this line says so.

iderex:
Freed (2026-09-11): the stage this one waits for, #1527, has landed and is closed, so the dependency this label named no longer exists. Stage 3 (#1529) landed its panels in the meantime; the wizard and the condensed help are the next stage to plan.

iderex:
## Plan for stage 2, read from the tree on 2026-09-11

### What is there now, read rather than remembered

- Five pages since #1527. The Providers page is 229 KB and carries 77 help keys plus 75 sentence-part markers over 138 field descriptions; the other four pages carry 35 help texts together (Overview 12, Accounts 5, Policies 14, Server 4). The catalogue holds 109 help keys, median 194 characters, the longest 680: 25 KB of help prose in each language.
- The two-column grid exists (`.sso-columns`, three to one, one column under the breakpoint), and the Providers page already has a rail with a Readiness card that says "open a provider"; the per-provider readiness panel (#1083) still sits inside each editor, so readiness is answered in two places.
- Adding a provider today: `addProvider` resets the editor and offers the template picker (#726) inside the same long form, and Test Connection is a button in that editor. Overview links to the five tabs and offers no way in.
- Sensitive and Insecure are regions already, `sso-sensitive-region` and `sso-insecure-options` with a toggle, on both protocol forms.
[...]

Issue #1083, opened 2026-08-03
Per-provider readiness panel in the provider editor (OIDC + SAML)


iderex:
Parent: #738

## Claim
The open provider editor answers "is this provider ready for a login attempt, and if not what is left?" with one aggregate checklist, instead of only per-field blur validation.

## Scope
A readiness panel rendered inside both the OIDC and the SAML provider editor, fed by signals that already exist in `SSO-Auth/Web/config.js`:

- required fields present, reuse `validateRequired` (L528), `validateEndpoint` (L536), `validateBaseUrl` (L575)
- endpoint reachable, fold in the existing Test Connection result (`#TestProvider`/`#TestResult`, `sso/OID/Test/{provider}`; `#saml-TestProvider`, `sso/SAML/Test/{provider}`). Do not add a second reachability path; the check reuses the hardened server-side one. A provider that has never been saved reads as "not yet tested", not as a failure.
- redirect URI shown/acknowledged, reuse the computed `#OidRedirectUri` / `updateRedirectUri` (#724)
- active insecure or sensitive toggles, reuse the existing `insecureFieldIds` / `sensitiveFieldIds` lists, listed by name

Surfaces: `SSO-Auth/Web/configPage.html`, `SSO-Auth/Web/config.js`.

## Constraints carried down from the parent
- Advisory only. It never blocks Save and never sets, clears or reads-back-into any toggle's `.checked`, danger-zone isolation stays intact (it may expand-only, as `loadProvider` already does).
- Pass/attention state is carried in `textContent`, never colour alone (#221).
- Admin-only surface; no new anonymous endpoint.

## Done when
- Opening a provider with a blank required field lists that field in the panel as attention; filling it clears the row without a save.
- A provider with an enabled insecure toggle names that toggle in the panel, and the toggle's `.checked` is unchanged before and after the panel renders.
- A successful Test Connection flips the reachability row to pass without re-issuing the request.
- The new markers/strings pass `LocalizationCatalogTests` and the web-asset rules in `ArchitectureConformanceTests`.

PR #1700, opened 2026-09-12
Add the provider wizard to the Providers tab

### LINKED ISSUE #1665: Stage 2, slice 4: the provider wizard
Waiting for: Flowfin/jellyfin-plugin-sso#1664

Stage 2 of the 4.4 UI plan (#1528), the fifth of six slices and the headline: adding a provider is a wizard.

## What to build

- An in-page stepper on the Providers tab (decision D2 on #1528), not a dialog: (1) protocol and template, through the template picker that exists (#726); (2) the required fields the readiness spec names for that protocol (`readinessSpecs.oid.requiredIds` and `readinessSpecs.saml.requiredIds` in `SSO-Auth/Web/sso-core.js`); (3) the redirect or ACS URL with a copy button, the value the readiness rows already compute; (4) Test Connection, the action that exists; (5) enable and save.
- Each step validates through the readiness rules before the next opens and says which field is missing; back is always allowed. Leaving the wizard mid-way is caught by the unsaved-state gate (`tools/ui-unsaved-state.js`), which says so; nothing is kept unsaid.
- An "Add provider" card on Overview (`SSO-Auth/Web/configPage.html`) opens the wizard on the Providers tab, through the same tab links the strip uses, with a query flag the page reads on load.
- The editor after the wizard is the grouped editor that exists; the wizard adds no second way to edit.

## Done when

- a provider can be added from Overview to a green Test Connection without opening any accordion
- every step refuses to advance on its own missing field and names it
- `node tools/ui-help-census.js` and every UI gate under `tools/ui-*.js` stay green
- walked on the walk server against pocket-id up to the green Test Connection (decision D4 on #1528), one screenshot per step on the pull request

## Rules of the road

One commit per pull request, replaced rather than amended after review; the review gate before the merge, with the availability lens on anything that can leave a form unsaved; Prettier clean; every new string in both catalogues, and `Catalogue-Vouch: de read by iderex` in the pull request body after the German has been read.

iderex:
# What & why

Adding a provider is five steps now, on the Providers tab, in an order that
cannot be walked out of turn: the protocol and a template, the required fields,
the address to register at the identity provider, Test Connection, and only
then Enabled. Overview gains a link that opens it. Refs #1665.

**It conducts the editor that already exists and holds no field of its own.**
The wizard is buttons and prose. It writes no configuration value, adds no
control that persists, and issues no request: the save and the test it asks for
are the editor's own buttons, which already carry their late-reply guards. So
the save contract, the managed-source freeze and the readiness rail are
untouched by it — and that last one is the answer #1664 asked this slice for.
The rail answers for whichever editor is open, `openEditorKey` reads the two
editors' `hidden` attributes and nothing else, and the wizard OPENS an editor
rather than replacing one, so both attributes mean exactly what they meant
before.

**The order is the one the server forces, and it was measured rather than
assumed.** `sso/OID/RedirectUri/<name>` answers 404 for a provider that has not
been saved, which is why `updateRedirectUri` already carries an arm writing
"Save this provider to see its exact redirect URI"; and `OID/Test/{provider}`
reads the STORED configuration. So the save sits inside step two and the last
step is the ENABLE. The SAML half computes its ACS URL in the client and needs
no save for step three, which is why the sentence for that step is the one
thing here that differs per protocol.

**What the order buys is a security property rather than a convenience.** A
provider reaches `Enabled` only after a Test Connection that succeeded, and
both openers reset every toggle off. The failure it is against is the ordinary
one: a provider enabled on a typo, offered on the sign-in page, answering
nobody.

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] Fail-closed preserved. Nothing here touches SAML or OIDC validation. What
      it adds is fail-closed in its own dimension: the wizard refuses to open
      its last step without a Test Connection that succeeded AND whose subject
      still matches the form, and a provider it builds starts with every toggle
      off.
- [x] No secrets logged; secrets are redacted on config export. The wizard
      reads no secret and writes none; its only configuration-derived text is
      the localized LABELS of fields that are still empty.
- [x] A security review was performed. This is browser JS on the configuration
      surface, which CI cannot exercise, so the review is the primary
      verification per this repository's own rule. Record below.
- [x] Review record: three lenses, **CONFIRMED 8 / PLAUSIBLE 0**. Every finding
      below was reproduced by the lens that raised it, by driving the shipped
      code, and every one is FIXED in this branch. Counts and dispositions:
      - **correctness — 3 raised, 3 fixed.**
        - `hidden` does not hide an `.emby-button`: the browser's
          `[hidden] { display: none }` is user-agent origin and loses to
          jellyfin-web's author-origin `display: inline-flex` on `.emby-button`,
          and jellyfin-web ships no `[hidden]` rule at all. Next, Finish and the
          start button stayed on screen and stayed pressable, so **Finish
          accepted on step one over an empty editor** and wrote "The provider is
          tested, enabled and saved". Every earlier use of the idiom on these
          pages is on a `div` or a `ul`, which carry no author `display`, which
          is why it had never shown. FIXED by one scoped rule in the plugin's own
          stylesheet, with the grep of the host's stylesheet beside it; the gate
          refuses that rule's removal.
        - The test outcome is keyed on the protocol, so an endpoint edited after
          a passing test finished as "tested". FIXED: the outcome now carries the
          subject it was about.
        - The same flag is module-scope and the deep link makes two live
          Providers views possible, so a test run in one could satisfy the other.
          Narrowed by the same fix — the subject has to match too — and what is
          left is the pre-existing shape of `readinessTestState` itself (#1083).
      - **availability — 5 raised, 5 fixed.**
        - The test gate read a flag no edit invalidates (the same defect).
        - The protocol switch asked its confirmation on `isPageDirty`, the
          reading this file elsewhere refuses to trust: removing a role-mapping
          row is a button click with no event, so work vanished unasked. FIXED:
          the question is asked on either reading.
        - With a declaratively-managed provider in the editor, the last step told
          the administrator to tick `Enabled` and press Save, both of which the
          freeze had disabled. FIXED with a sentence that says what is true.
        - A refusal naming the two protocol buttons could be written on a step
          whose panel hides them. FIXED: that refusal moves the wizard to the step
          its controls are on.
        - The failed-test sentence said "fix what it reported" where the fix is
          "press Save", which the SAML path can reach. FIXED by rewording.
      - Refuted and not changed: no trap exists (Leave and Back are never
        disabled, and the wizard's buttons sit outside both provider forms, so
        the freeze cannot reach them); nothing the wizard does outlasts it; the
        unsaved-changes state and the tab-return refresh are byte-identical with
        it open.
- [x] Log inputs from the IdP are sanitized inline at the log call. No log call
      is added.

## Quality checklist

- [x] No duplicated logic. The gates derive from `readinessSpecs`, the same
      reading the rail and the Save gate use, so the wizard cannot name a
      different set of missing fields than the card beside it does. The
      architecture lens raised five duplications and all five are gone: the two
      first-run prose lists, nine stub members copied from the gate this one is
      patterned on that no arm reached, a byte-identical copy of
      `.sso-inline-error`, two rules differing only in a margin, and seven ids
      nothing read.
- [x] The change adds no more code than the problem requires. The step titles
      and the five panels are AUTHORED in the markup rather than built by the
      script, which both removes code and avoids the catalogue-timing defect
      `populatePresetPicker` documents (#1602).
- [x] The code is self-documenting and matches the target architecture.
- [x] Architecture-conformance tests pass. The structural property this change
      establishes — a step opens the next only when it and every step before it
      is complete — is locked in by `tools/ui-provider-wizard.js` rather than by
      a C# rule, because it is a decision rather than a text and no reading of
      the bytes makes it.
- [x] Docs impact considered: see Notes.

## Verification

- [x] `dotnet build --warnaserror` green on BOTH target frameworks, 0 warnings.
- [x] `dotnet test` green: 3960 passed on net9.0, 3961 on net10.0, 0 failed,
      4 skipped (the firewall-consent tests of #1227).
- [x] Prettier clean for every file this change touches.
- [x] All twelve UI gates exit 0, the new one included.
- [x] Test-merged against the mainline after #1702 landed: the merge is clean,
      every gate exits 0 and the suite is green on the merged tree. The branch
      itself is left at its one commit.
- [x] Thirty-two mutations driven against the new gate. Twenty-eight changed
      behaviour under the harness and every one was refused. The two that did
      not change anything to drive are named in the gate's header with what
      refuses them instead — one of them, a refusal naming the wrong catalogue
      key while carrying the right English default, was driven against the suite
      and `ScriptEnglishDefaults_MatchTheCatalog` printed both strings.

## Notes

**This does not close #1665, and that is the plan the issue itself carries.**
Two of its Done-when items are not reachable from an automated run, and both
were written onto the issue before this work started:

- the German rows added here need a reading before `Catalogue-Vouch` can be
  written, and `pr-hygiene` gates the MERGE on that line. It is not written in
  this body, because writing it would assert a reading that did not happen.
- the Done-when asks for a walk on the walk server against pocket-id with one
  screenshot per step (decision D4 on #1528).

So this is green on everything a machine here can judge, and waits for a person
on both counts.

**What the proof cannot say**, stated where a reader meets it rather than only
in the tool's header: layout, focus order, what a screen reader announces from
the refusal's `role="status"`, whether the repaint that follows the catalogue
produces the right German, and whether the dashboard CONSTRUCTS a controller
when Overview's link is followed to a Providers tab it has already cached —
`viewContainer` caches on pathname and search, and the flag rides in the hash.
If it does not, the link opens the tab and the wizard stays closed until its own
button is pressed. Those bounds are disclosed at the code and are the walk's.

**Out of scope, with a reason.** The redirect-URI field is a tracked control and
its reply lands after the fill that takes the unsaved-changes baseline, so the
baseline diverges after every save and the next touch of any control raises a
"Unsaved changes" nothing typed. That is pre-existing, it is why the protocol
switch asks on either reading rather than on the baseline alone, and it deserves
its own issue rather than a second topic here. That is #1701, and it has landed
as PR #1702, merge commit `b33f11bc2a21eaa8f4bf79c7c273ddf9810efa38`.


iderex:
Walked (2026-09-12) on the walk server: Jellyfin 12.0, plugin build 5.0.0.87 from this branch's head `3792b7b6`, identity provider pocket-id at `http://172.17.0.1:8100`, dashboard language German, so every string below is the `de.json` row as the page painted it.

**The five steps, each refused first and then passed.** Overview's link opened the wizard on the Providers tab. The five rows read `Protokoll und Vorlage (Sie sind hier, Schritt 1 von 5)` and four times `(noch nicht, Schritt n von 5)`; a row that is passed turns to `(erledigt, Schritt n von 5)`.

1. Weiter with no protocol chosen. The node with `role="status"` read `Es ist noch kein Protokoll gewählt. Wählen Sie oben OpenID Connect oder SAML 2.0; der Editor darunter öffnet sich für das gewählte Protokoll.` Then `OpenID Connect verwenden` opened the OpenID editor below as `Neuer Anbieter`, and Weiter moved to step 2.
2. Weiter with the editor empty: `Noch leer: Name des OpenID-Anbieters, OpenID-Endpunkt, OpenID-Client-Kennung. Füllen Sie jedes davon im Editor darunter aus und drücken Sie dann Speichern.` The readiness rail said the same under `Erfordert Aufmerksamkeit - Pflichtfelder`. With `walk-wizard`, the endpoint and a client id filled in and the editor's own Speichern pressed (`Einstellungen gespeichert.`), Weiter moved to step 3, and the list gained `walk-wizard OIDC Disabled`.
3. The address field is read-only and holds what the server answered, `http://127.0.0.1:8097/sso/OID/redirect/walk-wizard`; the rail read `Bereit - Redirect-URI - Wird in diesem Formular angezeigt`. Weiter moved to step 4.
4. Weiter untested: `Es wurde noch nichts getestet. Drücken Sie im Editor darunter Verbindung testen; dieser Schritt öffnet den nächsten nur nach einem erfolgreichen Test.` The first two tests failed on conditions of the walk server itself (next paragraph), and Weiter after a failed test refused with `„Verbindung testen“ war zuletzt nicht erfolgreich. Der Test fragt den Server nach dem Anbieter, den er GESPEICHERT hat; ...`, step 5 staying at `noch nicht`. With the server's two switches set, the test read `The OpenID discovery document was read successfully.` and `JWKS: reachable (1 key(s))`, and Weiter moved to step 5, where Weiter gives way to Fertig.
5. Fertig with the provider still disabled: `Der Anbieter ist noch nicht aktiviert. Setzen Sie im Editor darunter den Haken bei Aktiviert und drücken Sie dann Speichern.` Aktiviert, Speichern, Fertig: `Der Anbieter ist getestet, aktiviert und gespeichert. Er wird ab jetzt auf der Anmeldeseite angeboten.` The wizard closed, its button `Anbieter Schritt für Schritt hinzufügen` stood on the Providers tab again, the list read `walk-wizard OIDC Enabled`, and Overview counted four providers offered on the login page.

**What the walk server needed, which is not the wizard's.** pocket-id is reached over plain HTTP at a private address inside Docker, and the hardened discovery reader refuses that twice; the plugin's log named both reasons, `HTTPS required` and `The outbound host resolves only to blocked addresses`. The two switches `OpenID-HTTPS-Erkundung abschalten (unsicher)` and `Adressen privater Netze zulassen (unsicher)` made the test green; afterwards the rail names them under `Unsichere oder heikle Optionen - Aktiv`, and the audit line `saved with security checks disabled: DisableHttps, AllowPrivateNetworkAddresses` stands in the server log. A provider over HTTPS at a public name needs neither.

**The bounds the body left to the walk.**

- The link into a Providers tab the dashboard had already cached: followed with that tab visited before and the wizard closed, it landed on `#/configurationpage?name=SSO-Auth-providers&wizard=1` with step 1 at `Sie sind hier`. The case where the wizard would stay closed did not occur.
- The refusal's `role="status"`: every refusal above was read from the node carrying that role in the accessibility tree, which is what an assistive technology announces from a live region. No screen reader was run.
- The repaint after the catalogue: every wizard string above is the German row, the step rows included; nothing showed its English default.
- Layout and focus order were looked at rather than measured: at the pane's desktop width the wizard card stands above the provider list with the five rows, the lead, the two protocol buttons and the three navigation buttons, and at 375 px the readiness rail moves above the form (`order: -1` on `.sso-column-rail` under `.sso-columns--rail-first`, which is the narrow-layout bound #1664 left open).

**Two things seen beside the wizard, neither this branch's.**

- On the first visit the four small pages showed their folds without the lead sentence of #1668. The dashboard had kept the previous build's `SSOViews/i18n.js` (3449 bytes, no condensing in it) and revalidated it against an `ETag` every 5.0 build shares, so the server answered 304 and the old module stayed; after a forced refetch and a reload the leads were there. That is filed on its own as #1705.
- English inside the German dashboard, with the current module: the `Review` chip on a provider row. The editor's labels read German (`Name des OpenID-Anbieters`, `OpenID-Endpunkt`, `Mit einer Vorlage beginnen (optional)`). Catalogue coverage, older than this change.

**No screenshot is attached.** A comment posted through the API carries no image, so the walk is given as what the page said at each step, read from the accessibility tree and the DOM; the steps above repeat it.

**Which module the walk ran on.** The five steps were walked while the dashboard still held the previous build's `i18n.js` (#1705), which serves `t()` the same way; after the refresh, steps 1 and 2 were walked again on the current module, and the two refusals read the same rows, with the field names in step 2 now in German as quoted. Corrected the same evening: the first posting of this comment quoted step 2 with English field names and listed the editor's labels as English, both read from the cached module.

**Read as a change.** The wizard holds no field, writes no value and issues no request of its own; the save and the test are the editor's buttons, and the walk saw exactly that (the list, the audit line and the readiness rail all reacted to the editor's Speichern, never to the wizard). No security review was run, because no surface that checklist covers is touched. The five Done-when items of #1665 are met on the walk server, and the one thing between this and the merge is the reading of the German rows and the vouch line.

Issue #1741, opened 2026-09-14
The administrator Unregister route has no last-administrator guard, so the lockout #1732 closed is still one click away


iderex:
The self-unlink guard is one route to a self-inflicted lockout and `Unregister` is the other. #1732 narrowed the administrator exemption on `DELETE {mode}/Link/...`; this route was not in that decision's scope and carries no guard of its own. Found by the availability lens reading the #1732 change, and verified at the tree below.

Split: none

## What the route does

`POST sso/Unregister/{username}` removes every canonical link the account holds, repoints it to the provider named in the request, and revokes its tokens - with no question asked about who the caller is beyond elevation:

```
$ sed -n '2039,2052p' SSO-Auth/Api/Http/SSOController.cs
        var revoked = _canonicalLinks.RemoveUserEverywhere(user.Id);
...
        user.AuthenticationProviderId = provider;
        await _userManager.UpdateUserAsync(user).ConfigureAwait(false);
...
        await _sessionManager.RevokeUserTokens(user.Id, null).ConfigureAwait(false);
```

The comment at the revoke names the self case and treats it as safe: "including the caller's own, when an admin unregisters their own account - the durable revoke above is why that is safe". That sentence is about the revoke being complete, not about the caller still having a way back in.

The settings page offers it per row, with no exclusion of the row that is the caller's own:

```
$ grep -n 'Unregister' SSO-Auth/Web/sso-core.js | head -3
```

## Why it matters after #1732

The self-service page now refuses an administrator who would leave the server with no administrator holding an SSO link. An administrator meeting that refusal has an obvious next move on the admin page: Revoke on their own row. Where the provider named in the request is the built-in password provider, the account lands on a provider whose stored password this plugin minted and never recorded - the state `RequestHelpers.CallerHasNoPasswordDoor` describes in its own words - and the session ends in the same call.

It is not the same severity as the route #1732 closed: the account is on the built-in password provider afterwards, so Jellyfin's own in-network recovery may apply. Whether it does is a host question this issue does not answer and is the first thing the work here has to read.

## Done when

Either this route carries the same last-administrator reading the self-unlink now takes - measured with `AdministratorsWithNoWayIn`, refused where the caller is acting on their own account and nobody else holds a way in - or the decision is written here that it does not, with the recovery path that makes it acceptable read at the host rather than assumed.

The `Unregister` comment at the revoke is pointed at whichever answer lands, rather than left saying the self case is safe on a reading about durability.

Issue #1733, opened 2026-09-14
A minted password reads as a way in, so the self-unlink guard does not reach the accounts it was written for


iderex:
A minted password reads as a way in, so the self-unlink guard does not reach the accounts it was written for

Comes out of the review on #1720, raised twice by two lenses reading the guard
that landed there.

Split: none

## What the guard asks, and what it misses

The refusal on #1720 fires when the caller's account has no password door, and
that fact is read from one comparison: does `AuthenticationProviderId` equal
Jellyfin's built-in password provider. Everything else reads as "no door" and is
covered.

The uncovered case is the account that IS on the built-in password provider and
still has no usable password. This plugin mints one onto every account it
provisions - 64 CSPRNG bytes, never displayed, never stored anywhere else and
never recoverable - and records nowhere which accounts those were. A stored hash
is therefore a credential somebody holds or a seal nobody can open, and the two
are the same bytes.

That is not a corner. Where a provider's `DefaultProvider` names the built-in
password provider, every SSO login writes that id onto the account, so every
account the plugin provisions on such a server lands in exactly this state - and
the settings page offers that value as a common choice. On those servers the
#1720 guard never evaluates, and the lockout it exists to refuse is reachable
exactly as it was before.

## Why it was not settled on #1720

The per-provider bulk unlink answered the same ambiguity and answered it the
other way: it refuses to count a stored password as a way in at all, in as many
words, because counting it would have cleared that guard for every
SSO-provisioned administrator on the server. It can afford that reading because
its subject is a mass action an administrator takes, where a refusal costs one
call.

The self-unlink cannot afford it unexamined. The same reading there refuses every
last-link self-unlink on every server, including the ordinary one where the user
set their own password and knows it. Which of the two readings this rule should
take is a decision about what a user's own control over their links is worth
against a lockout, and no reading of the tree makes it.

## Options

- **Take the purge's reading**: count no stored password as a way in, so the
  refusal applies to every last-enabled-link self-unlink. Closes the gap
  completely; costs every user the ability to unlink their last provider without
  an administrator.
- **Record which passwords the plugin minted** and treat only those as no door.
  Closes the gap for the population that matters with no cost to anybody else,
  and adds a persisted fact about accounts that has to be maintained and migrated.
- **Leave it and say so**, which is what the tree does today: the code, the
  changelog and the API document each name this residual where a reader meets it.

## Done when

The chosen reading is implemented with a negative test per fail-closed branch, or
[...]

Issue #1720, opened 2026-09-13
A non-admin can lock themselves out for good by unlinking their own last link on an SSO-only server


iderex:
On an SSO-only server, a non-admin can lock themselves out of their own account for good by pressing Delete on their own last link in the self-service page. There is no warning, no password fallback, and no way back in without an administrator. I found it while walking the stage-3 self-service page for #1529, and measured it on a live server rather than reasoning about it.

## What happens

The self-service page (`SSO-Auth/Web/linking.js`, `handleDeleteButtonPressed`) removes a selected link with `DELETE /sso/{mode}/link/{provider}/{userId}/{canonical}` and no confirmation of any kind. When the link removed is the user's LAST one, the controller revokes their tokens (the `#468` last-link revoke). Unlike the administrator revoke beside it, the self-service path does NOT switch the account back to the built-in password provider, so on an SSO-only server the account is left able to authenticate by no means at all.

## Measured

A `jellyfin/jellyfin:12.0` container with a `net10.0` Release build of `origin/5.0` `c82e5411c5928eddf734f0f161a4252d684b27be`, a Dex provider, and `alice` a linked non-admin account. SSO-only was enabled with a break-glass admin, which repoints `alice` to the SSO provider:

```
# after SSO-only enable, before any unlink
GET /sso/SSO-Managed/Status/{alice}   -> {"PasswordLoginDisabled":true,"HasCanonicalLink":true}
POST /Users/AuthenticateByName alice  -> 401   (password already disabled, as SSO-only intends)
```

Signing in as `alice` through the real SSO flow to hold her own session, then issuing the exact request the self-service Delete button issues:

```
DELETE /sso/oid/link/dex/{alice}/{canonical}  (alice's own token)  -> 200
GET /sso/SSO-Managed/Status/{alice}  -> {"PasswordLoginDisabled":true,"HasCanonicalLink":false}
GET /Users/{alice} .Policy.AuthenticationProviderId  -> Jellyfin.Plugin.SSO_Auth.Api.SSOController
POST /Users/AuthenticateByName alice -> 401
```

`alice` now holds no SSO link (cannot sign in through SSO) and is still on the SSO auth provider with password login disabled (cannot sign in with a password). The account is not disabled; it is simply unreachable by its owner.

Whether a fresh SSO login recovers her depends entirely on the provider's `AllowExistingAccountLink`:

```
# adoption ON: a fresh SSO login re-adopts the same-named account
[...]

PR #1742, opened 2026-09-14
Refuse an administrator's own last-link unlink that strands the server [#1732]

### LINKED ISSUE #1732: An administrator can strand their own account through the self-service unlink
An administrator can strand their own account through the self-service unlink

Comes out of the review on #1720, which found it while reading the guard that
landed there. #1720 refuses a non-administrator's removal of their own last
usable link; administrators are exempt by the decision taken on that issue, and
that decision was about an administrator removing SOMEBODY ELSE's link. The case
below is an administrator removing their own, which nobody has decided.

## What happens

`/SSOViews/linking` is not an administrator page - it acts on
`ApiClient.getCurrentUserId()`, so an administrator who opens it is acting on
their own links. On a server where SSO-only is on, the activation repoints every
managed account, administrators included, to the provider id that accepts no
password; only the break-glass administrator is exempt. An administrator in that
state who ticks their last link and presses Delete is exempt from the #1720
refusal, so the link is removed, and the last-link revoke then ends their own
session on the spot. No password, no link, no token.

The recovery is another administrator or the break-glass account. Where the
server has neither, what is left is editing the plugin's configuration file on
disk.

## Why it is its own issue rather than part of #1720

The guard there reads three facts and the administrator one short-circuits it.
Extending it to an administrator acting on their own account is one condition,
but it is a scope question rather than a coding one: an administrator is also
the person who might be deliberately cleaning up, and this board has a
break-glass mechanism whose whole purpose is to keep one way in. Which of the
two the refusal should serve is a decision, not a reading.

The primitive already exists: `CanonicalLinkService.AdministratorsWithNoWayIn`
answers exactly "which administrator accounts would be left with no way to sign
in", and the per-provider bulk unlink already refuses an elevated caller on it.
Nothing here has to be invented.

## Options

- **Refuse an administrator's removal of their own last usable link**, the way
  the bulk unlink already refuses an elevated caller who would strand one. The
  cost is that a deliberate cleanup needs a second administrator or the
  break-glass account.
- **Refuse it only when no other administrator would be left with a way in**,
  which is what `AdministratorsWithNoWayIn` measures. Narrower, and it leaves a
  single-administrator server protected while a multi-administrator one keeps
  the control.
- **Leave it and say so on the page**, on the reading that an administrator is
  expected to know what the button does.

## Done when

The chosen shape is implemented with a negative test per fail-closed branch, and
the comment at the administrator exemption in `TryRemoveLink` names the decision
rather than the deferral it names today.

iderex:
# What & why

#1720 refuses a holder's removal of their own last usable SSO link on an account that accepts no password, and exempts an administrator. That exemption was decided for an administrator acting on somebody **else's** link. `/SSOViews/linking` is not an administrator page - it acts on the caller's own account - so an administrator who opened it on an SSO-only server was one press from the same lockout a user was, with the difference that the recovery the refusal points a user at IS them. Where they were the only administrator who could sign in, nothing was left but editing `SSO-Auth.xml` on disk.

The exemption becomes a pair of facts, per the decision taken on the issue on 2026-09-14: the removal goes through when the caller is not the holder, or when some other administrator still holds a way in, and it is refused when neither holds. The two other shapes on the issue were declined there with their reasons.

Where each fact is read, so none of them can drift:

| fact | read |
| --- | --- |
| the caller is an administrator, and accepts no password | the request boundary, from the resolved account (as #1720 already reads them) |
| the caller IS the holder | the request boundary, from the resolved caller's own id, never from the route value alone |
| some other administrator keeps a way in | the request boundary, over the enabled administrator accounts, through `AdministratorsWithNoWayIn` |
| the link is the last one that could still sign the caller in | inside the removal's own transaction, unchanged |

Both new parameters default to the value that **refuses**, so a call site that has not been told about this rule refuses a removal rather than performing a lockout nobody asked for. The survey is asked only where all three cheap facts already hold, so a server the rule is inert on pays nothing for it.

Closes #1732

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

      3 / PLAUSIBLE 3** counts as raised (a finding is CONFIRMED only if a
      reproduction was produced). Every finding of either kind carries a
      disposition - FIX, a reasoned DECLINE, or DEFER as its own issue.

### Review record

One lens, the availability / mass-lockout one, which is the mandatory lens for anything that can lock an account out. It read the full files and their callers rather than the diff hunks, decompiled `MediaBrowser.Controller.dll` at 10.11.0, 10.11.11 and 12.0.0 to settle that `AuthorizationInfo.UserId` is `User?.Id ?? Guid.Empty` on all three, and ran a probe to settle how the reflective accessor wraps a throw.

The refutations that failed are the part worth reading: the subtraction behind "another administrator keeps a way in" is exact with no off-by-one, disabled administrators bias it toward refusal, vanished accounts cannot appear because the roster comes from live records rather than from a link table, non-administrator behaviour is byte-identical, there is no sequential drain (administrator A self-unlinks while B is linked, then B is refused because A now holds nothing), a break-glass administrator on the password provider is never caught by the new refusal, and no lock is nested.

| # | verdict | severity | finding | disposition |
| --- | --- | --- | --- | --- |
| 1 | CONFIRMED | MEDIUM | The administrator `Unregister` route has no last-administrator guard, so a comparable state is one click away on the neighbouring button; not a regression from this change | **DEFER** - filed as #1741, and the CHANGELOG claim is scoped to this route and cites it |
| 2 | CONFIRMED | MEDIUM | The refusal said "no other administrator on this server can sign in either", which the code cannot establish: a stored password is deliberately never counted, so on the common legacy-owner-plus-SSO-admin server the sentence is false and its advice sends an administrator to fix something that is not broken | **FIX** - the sentence now says what was measured ("holds an SSO link that can sign them in"), says "could leave" rather than "would leave", and names the one-call remedy that was missing |
| 3 | CONFIRMED | LOW | The fail-closed catch named only `InvalidOperationException`, but the accessor is reached through `MethodInfo.Invoke` / `PropertyInfo.GetValue`, which wrap a host failure as `TargetInvocationException` - so a transient user-store failure escaped as a 500 instead of the documented "answers no" | **FIX** - it catches the surface now, with the reason and a CA1031 pragma at the catch |
| 4 | PLAUSIBLE | LOW | The survey was gated on two facts and not on `passwordLoginDisabled`, so an administrator on the built-in password provider paid a full reflective enumeration plus a config-lock read on every self-delete, including one naming no link | **FIX** - the third fact is in the gate |
| 5 | PLAUSIBLE | INFO | Two administrators self-unlinking concurrently each see the other and both pass; unlike the purge, this guard cannot re-derive the set under the lock because the user records are not under it | **DECLINE** - accepted and named at the guard and in the Notes below |
| 6 | PLAUSIBLE | INFO | `LinkingIdentifierOneDecisionTests` now pins its administrator harness to the password provider to step around the new rule, so those identifier arms no longer exercise an SSO-routed administrator | **DECLINE** - the coverage moved deliberately and the harness says why; the arms exist for the identifier comparison, which the new rule would otherwise stop them from reaching |

Findings 2, 3 and 4 were fixed and the same lens re-read only the fixed state, ending **PASS**: the `passwordLoginDisabled` gate is unobservable because that flag already dominates the service-side predicate and the skipped value is forced to the refusing one, so no future broadening of the rule can turn it into a fail-open; the broad catch can only make the guard stricter because its swallowed answer is the restrictive one, and no `CancellationToken` is plumbed through either call, so there is no cancellation to mis-swallow; and the reworded sentence keeps the substring the page and both gate arms match on, while the user sentence still matches neither.

That pass added two INFO residuals of its own, neither a lockout. The administrator sentence is also written when the survey threw and was caught, that is when nothing was measured - the remedy it names stays correct in that case and whoever reads the log gets the Warning line. And "ask another administrator to remove it for you" is reachable through the API rather than through a button today, because the linking page always acts on the current user; the sentence's other remedy, linking a second provider first, is fully reachable from the page, so nobody is stuck.

## Quality checklist

      testable unit.
- [x] Docs impact considered: this change needs no README/wiki update, OR the
      update is included here, OR it is filed as a `documentation` issue on the
      current-release milestone (link it in Notes).

The account-door reading was factored to one private `Describe(User)` rather than copied for the second caller, so the two cannot drift about what a password door is. No new structural property is established, so no new conformance rule is owed; the 257 existing conformance arms pass.

## Verification

- [x] `dotnet test` is green.

Read at this head, both target frameworks, warnings promoted on the test project because `dotnet test` alone does not promote them:

```
$ dotnet build SSO-Auth.Tests/SSO-Auth.Tests.csproj -f net9.0  --warnaserror --nologo -v q
Der Buildvorgang wurde erfolgreich ausgefuehrt.    0 Warnung(en)    0 Fehler
Der Buildvorgang wurde erfolgreich ausgefuehrt.    0 Warnung(en)    0 Fehler
   SSO-Auth.Tests  Total: 4002, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0
   SSO-Auth.Tests  Total: 4003, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0
$ node tools/ui-self-service-unlink.js     # thirteen arms, exit 0
```

The four skips are the loopback-binding tests, which need an elevation this machine's rules forbid raising.

**The guard is proven by deleting it.** Restoring the condition to its #1720 form and re-running the class turns two arms red - the refusal and the default - rather than passing on:

```
$ ./SSO-Auth.Tests/bin/Debug/net9.0/SSO-Auth.Tests.exe -class "...StrandingSelfUnlinkTests"
   Total: 16, Errors: 0, Failed: 2
   ... at StrandingSelfUnlinkTests.AnAdministratorOnTheirOwnLastLink_WithNobodyElseAbleToSignIn_IsRefused()
```

The new page branch was proved the same way: forcing its discriminator to `false` makes the new gate arm report that the banner carried `link.delete_refused_would_strand` rather than `link.delete_refused_would_strand_server`.

Prettier was verified per changed file rather than through the repository-wide glob, and the reason is worth stating rather than leaving as an unexplained departure. This machine's working tree is CRLF and the `npx prettier` this clone resolves defaults to `endOfLine: lf`, so the glob form reports every CRLF file including ones this change never touched; the same untouched file passes when its LF bytes are handed to it out of git. CI checks out LF and its Prettier Lint is green on the base.

## Notes

**Two residuals, stated rather than left to be found.**

Two administrators self-unlinking in the same moment each see the other and both pass, stranding the server. The bulk unlink closes its equivalent window by re-deriving the judged set under the configuration lock; this guard cannot, because the Jellyfin user records are not under that lock and must not be dragged into it. It is named at the guard itself.

A break-glass administrator who genuinely signs in with a password reads here as holding no way in, because the reading this borrows never counts a stored password - this plugin mints an unrecoverable password onto every account it provisions and records nowhere which, so a stored hash is a credential somebody holds or a seal nobody can open and the two are the same bytes. So an administrator's own cleanup can be refused on a server that had a recovery account all along. That direction costs a call and the other costs the server. Telling the minted passwords apart is #1733, whose own decision was taken the same day.

**Out of scope and filed:** the administrator `Unregister` route reaches a comparable state in one call and carries no guard of its own - #1741.

**Docs:** no README or wiki page describes the administrator exemption, so nothing there goes stale with this change. The wiki's account of this family is already tracked and waits behind the 5.0 stable release.

**Second reader:** this board has no second reader tonight. What stands in place of one is the adversarial lens above, which derived its claims from the tree and from the decompiled host rather than from this body, and the evidence quoted in place of a reading.

**The German row is not vouched for on this head, and I am not signing it.** The
hygiene gate refuses this pull request until a line `Catalogue-Vouch: de read by
<name>` is in this body, and that line is a person putting their name to having
read the language. I wrote the row; writing the vouch for it as well would turn
an unread translation into a read one by typing a sentence, which is the one
edit this repository's own rule forbids above every other. So it stays absent
and the gate stays red until somebody who reads German adds it.

The row is quoted here so that reading it costs nobody a checkout. It is the
administrator's half of the refusal, beside the existing `link.delete_refused_would_strand`
it was modelled on, and it carries no placeholder:

```
"link.delete_refused_would_strand_server":
"Der Server hat es abgelehnt, die letzte SSO-Verknüpfung zu entfernen, mit der Sie sich anmelden können: Auch kein anderer Administrator dieses Servers hat eine SSO-Verknüpfung, mit der er sich anmelden kann, das Entfernen hätte diesen Server also möglicherweise ohne erreichbaren Administrator zurückgelassen. Bitten Sie einen anderen Administrator, sie für Sie zu entfernen, oder verknüpfen Sie zuerst einen weiteren Anbieter mit Ihrem Konto und entfernen Sie dann diesen. Laden Sie die Seite neu, um zu sehen, welche Verknüpfungen sie jetzt führt."
```

The English it renders:

```
"link.delete_refused_would_strand_server":
"The server refused to remove the last SSO link that can sign you in: no other administrator on this server holds an SSO link that can sign them in either, so removing that link could have left this server with no administrator able to reach it. Ask another administrator to remove it for you, or link another provider to your account first and then remove this one. Reload the page to see the links it holds now."
```

That is the one thing between this head and a merge. Everything else on it is
green or accounted for above.


iderex:
Security review of this head before it lands, because it changes a guard on the path an account
signs in by. No exploitable finding.

I read the diff by hand rather than through the review command, because the command reads its
own checkout and this head lives in a worktree of its own. The method is the same: every input
the change adds, every direction a wrong value can fail, and nothing reported that I could not
walk to a concrete case.

## The refusal, case by case

The change adds two facts to the removal, `callerIsTheHolder` and
`anotherAdministratorKeepsAWayIn`, and the condition that refuses reads:

```
(!callerIsAdministrator || (callerIsTheHolder && !anotherAdministratorKeepsAWayIn))
  && passwordLoginDisabled && provider enabled && no other enabled way in
```

- **Not an administrator.** The first clause is true on its own, so nothing new is consulted
  and the refusal is exactly the one `#1720` shipped.
- **An administrator removing somebody else's link.** `callerIsTheHolder` is false, the first
  clause is false, and the `#1720` exemption stands.
- **An administrator removing their own, on a server that accepts their password.** The
  controller leaves `anotherAdministratorKeepsAWayIn` false without surveying, which would read
  as "refuse", but `passwordLoginDisabled` is required again in the same condition, so nothing
  is refused.
- **Their own, SSO-only, another administrator can still sign in.** Not refused, as decided.
- **Their own, SSO-only, nobody else can.** Refused, as decided.

So every short circuit in the controller lands on a value the condition never consults in that
case. The change narrows the administrator exemption and widens nothing for anybody.

- **The survey throws.** `AnotherAdministratorKeepsAWayIn` catches broadly and answers no,
  which refuses. The broad catch is right: the accessor is reached by reflection, so a host
  failure arrives wrapped, and a narrow catch would have let it escape as a 500.
- **The caller cannot be resolved.** `CallerIsTheHolder` answers true, and
  `IsAdministrator` answers false for the same request, so the caller is treated as an
  ordinary holder and the ordinary refusal applies.
- **A different account id in the request.** The ownership check that runs first requires the
  link to belong to that id, so an administrator cannot reach the exemption for a link that is
  their own by naming somebody else.

## What reaches the caller

The survey carries usernames, and none of them leaves the method: it answers a count
comparison, the refusal sentence names no account, the audit line carries the caller's own id,
and the warning logs the exception and no roster.

## One edge, walked and not reported as a finding

The survey runs before the removal's transaction rather than inside it. Two administrators who
each remove their own last link within the same moment could each find the other still able to
sign in, and both removals would go through. That takes two trusted administrators racing
themselves on an SSO-only server, gives nobody anything they did not already hold, and is
recoverable through the break-glass account where the server has one and through the plugin's
configuration file where it has not, so it stays a note here rather than a blocker.

Issue #1729, opened 2026-09-13
Take the 5.0 line to its first stable release


iderex:
The 5.0 line is taken to its first stable release, and the work that is open
today goes into it rather than after it. This issue is where the freeze, the soak
clock and the promotion gate are recorded, so the `5.0.0-stable` tag points at a
decision with its evidence rather than at a date somebody remembered.

Decided (2026-09-14): plan it now, and carry the open work into it.

## Why this issue exists at all

#954 consolidated the JF12/5.0 line onto `main` and set it beta-only until a
first stable. Nothing has tracked that first stable since, and the cost of the
gap is visible on two issues: #1540 and #1553 carry wiki text that was approved
on 2026-09-09 and 2026-09-13 and cannot be written, because the wiki describes
the line the **Latest** mark points at (#1574, option A) and the behaviour they
describe exists only on a beta. They are not waiting on a reader. They are
waiting on this.

## The window

**N = 14 calendar days.** §4 of the Releasing page sets the default at 7 and says
in the same paragraph that N is a floor a promotion may lengthen. #1511 doubled
it for the 4.3 line with the reason that a bad first stable of an authentication
plugin is a mass-lockout event; the same reason holds here and one more beside
it: this is the first stable on a new Jellyfin server generation, where the host
API itself is new to every reader who installs it. §7's security-shortening rule
is the only thing that shortens it.

## What the window runs on, and what resets it

Per §4, the clock runs on the exact commit the `5.0.0-stable` tag will point at,
and that commit's build must be the newest beta on `manifest-beta` for the JF12
target, unchanged, for the whole window. **Any commit that lands on the 5.0 line
restarts the clock.** That is the whole reason for the order below.

THE TWO PROMOTIONS DO NOT COLLIDE. #1511's clock runs on the 4.3 line and is
restarted by commits there; this one runs on the 5.0 line. Work on either line
leaves the other's candidate untouched, and the two tags are cut independently.

## The order

1. **Land what is owed before the freeze.** These go into the stable, which is
   what this plan was asked for:

   - **#1529**, stage 3 of the configuration rebuild: accounts, the self-service
     page, one language per view. It is the last open stage of #1525 and the
     largest item here.
   - **#1720**, the self-service last-link self-lockout on an SSO-only server.
     Decided on 2026-09-14: the server refuses the unlink that would strand the
     account. A first stable that can lock a non-admin out of their own account
     for good is the one defect this promotion must not carry, and it sits in the
[...]

Issue #1511, opened 2026-09-04
Freeze the 4.3 line and run the 14-day soak for the first stable release


iderex:
The 4.3 line is taken to its first stable release. This issue is where the freeze, the soak clock and the promotion gate are recorded, so the `4.3.0-stable` tag points at a decision with its evidence rather than at a date somebody remembered.

## The window

**N = 14 calendar days**, decided 2026-09-04. §4 of the Releasing page sets the default at 7 and says in the same paragraph that N is a floor a promotion may lengthen; this is the first stable release of an authentication plugin, where a bad one is a mass-lockout event, so the floor is doubled for this promotion. §7's security-shortening rule is the only thing that shortens it.

## What the window runs on, and what resets it

Per §4, the clock runs on the exact commit the `4.3.0-stable` tag will point at, and that commit's build must be the newest beta on `manifest-beta`, unchanged, for the whole window. **Any commit that lands on the 4.3 line restarts the clock**, because a new beta supersedes the candidate. That is the whole reason for the order below rather than a preference for tidiness.

## The order

1. **Land what is owed before the freeze.** Anything that would otherwise cost fourteen days lands now: the roadmap correction (#1510, open), the two unchecked polish boxes on #822, and the harness repair on #1444 if it is ready. #1444 is judged there as a harness fault rather than a release-blocking defect, so it does not hold the freeze; it is deferred past the stable tag if it is not ready when the rest is.
2. **Freeze.** `main` holds on the 4.3 line: only a change that is itself a stable-blocker lands, and landing one restarts the clock knowingly.
3. **Cut the candidate.** `publish-beta.yml` is dispatched at the freeze commit so the newest beta on `manifest-beta` is built from it. `main` is four commits ahead of `4.3.0-beta.60` today, so the candidate is a new beta rather than that one.
4. **The clock runs**, and the date it ends is written here when it starts.
5. **The promotion gate (§5)** is walked at the end: the soak satisfied, the manual E2E QA checklist of §6 passed against the candidate build on a real server, and the rest of that section's boxes.
6. **The stable tag.** `publish.yml`'s own header records what its first firing owes: the seven-provider harness gate is wired into that leg **in the same change that cuts the first stable tag**, and `manifest-release` re-opens with it, empty since #954 retired the stable channel.

## What this issue is not

[...]

PR #1743, opened 2026-09-14
Split the JF12 beta publish from the build a transient 500 discards [#1736]

### LINKED ISSUE #1736: A transient 500 on the release API costs a whole publish build
A transient 500 from the release API throws away a whole publish build, and the second attempt is the last one. #1717 recorded the failure and says the retry budget belongs in its own issue; this is it.

Split: none

## What happened

`Publish JF12 Beta`, run 34749421228, dispatched on 2026-09-13, failed at `Upload JF12 beta assets to a draft release` after three attempts at creating the release, each answered `500`:

```
👩‍🏭 Creating new GitHub release for tag 5.0.0-JF12-beta.84 using commit "9faf99dd..."
⚠️ GitHub release failed with status: 500
retrying... (2 retries remaining)
...
❌ Too many retries. Aborting...
```

Everything before that leg had already run: the assemblies compiled, the version was read, the package and its sidecars were made. Nothing was published, so no channel is half-updated - the cost is the whole build, discarded because an API answered 500 three times in a few seconds.

## What the tree can and cannot set today

The step pins `softprops/action-gh-release@efb35369e0ad2afab669f228072c1b0d510eae64` (v3.0.3), and that action exposes no retry input - its `action.yml` inputs are `body`, `body_path`, `name`, `tag_name`, `draft`, `prerelease`, `preserve_order`, `files`, `working_directory`, `overwrite_files`, `fail_on_unmatched_files`, `repository`, `token`, `target_commitish`, `discussion_category_name`, `generate_release_notes`, `previous_tag`, `append_body`, `make_latest`:

```
$ gh api repos/softprops/action-gh-release/contents/action.yml?ref=efb35369e0ad2afab669f228072c1b0d510eae64 \
    --jq .content | base64 -d | grep -n '^  [a-z_]*:'
```

So the count and the spacing are the action's own, and no input in this repository changes them.

## The shapes, and why this is a decision rather than a fix

- **Leave it.** A dispatch that fails this way is re-dispatched by hand and the alarm closes itself. It costs a build and a person noticing; it is what happens today.
- **Wrap the step in a retry.** Any third-party retry action is a new dependency on the publish path, which is the one path where a compromised action reaches a signed artifact. That trade is not mine to make.
- **Replace the step with `gh api` and a back-off of our own.** No new dependency, and the retry budget becomes ours, at the cost of reimplementing the draft-create-and-upload the action does today - including the asset upload and the `fail_on_unmatched_files` behaviour the pipeline relies on.
- **Split the build from the publish**, so a failed release creation re-runs the publish alone against artifacts the build already produced. The largest change and the one that stops the whole build being the unit of loss.

## Done when

One of the shapes above is chosen and either built or written into this issue as a declined option with its reason, so the next 500 meets a decision rather than this paragraph.

CI cannot exercise any of them: this workflow runs only on dispatch against the release path, so whatever lands here is verified by reading and by the next real publish.

iderex:
# What & why

Closes #1736.

A transient `500` from the GitHub release API used to cost the whole publish
build. Run 34749421228 of 2026-09-13 had already compiled the assemblies, read
the version, packaged the plugin and written its checksum, SBOM and VEX
sidecars when `softprops/action-gh-release` met three `500`s in a few seconds
and aborted. One job held all of it, so everything upstream of the failing call
went with it.

The decision of 2026-09-14 on the issue chose the fourth shape - split the build
from the publish, so a failed release creation re-runs alone against artifacts
the build already produced - and declined the other three there with reasons.
This is that shape, for the leg the failure happened on.

`publish-jf12-beta.yml` now has a `package` job and a `publish` job:

- `package` (`contents: read`) builds, packages, generates the SBOM, stages the
  VEX and the effective `build.yaml`, writes the sidecars and uploads `_dist` as
  a workflow artifact. It exposes the derived version as job outputs.
- `publish` (`contents: write`) checks the tree out for the two local actions,
  downloads that artifact, creates and seals the draft release, regenerates
  `manifest-beta` and verifies it.

"Re-run failed jobs" therefore restarts the release call against the package
that was already built. The version survives the re-run because it is derived
from `build-jf12.yaml` and `github.run_number`, which is keyed to the RUN rather
than the attempt, and reaches `publish` as a job output - so a second attempt
publishes the same `5.0.0.<run>` under the same tag rather than minting a new
one.

This is the shape `publish.yml` has carried since it was written (a build job
whose artifact an upload job releases), brought to this leg rather than invented
for it.

Two details the split makes load-bearing, both commented where they are:

- The effective `build.yaml` - the JF12 metadata with this run's version written
  into it - is staged into `_dist` and shipped from there. `publish` starts from
  a fresh checkout, where `build.yaml` is the committed JF10.11 file again, so
  releasing it from the working tree would have shipped the wrong metadata. The
  release asset set and the asset names are unchanged.
- The checkout in `publish` runs BEFORE the download. `actions/checkout` cleans
  the workspace, so the other order would delete the package it is about to
  release.

`overwrite: true` on the upload covers the other re-run button: "Re-run all
jobs" runs `package` again under the same run id, and the action's own
description says it fails on an artifact name the run already holds.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved. Not reached in the login-path sense: no login-path
      or crypto byte changed. The pipeline-side fail-closed properties that DO
      apply are kept - `if-no-files-found: error` on the upload,
      `fail_on_unmatched_files` on the release, and the post-publish manifest
      verification.
- [x] No secrets logged; secrets are redacted on config export. No secret
      crosses the new job boundary, and `package` holds no write token at all
      now.
- [x] A security review was performed for changes to the release pipeline.
- [x] Review record: below.
- [x] Log inputs from the IdP are sanitized inline at the log call. Not reached:
      no log call changed.

**Review record (adversarial gate, three lenses, refute-by-default).** Run
against the diff and the post-change file; every lens read the siblings, the
three local composite actions and the pinned third-party actions at their
digests rather than from memory.

| Lens | Verdict | Raised | Disposition |
| --- | --- | --- | --- |
| Supply-chain / bypass | PASS | 2 LOW | both FIXED |
| Correctness | PASS | 4 LOW | all four FIXED |
| Availability / mass-outage | PASS | 3 LOW | all three FIXED |

**CONFIRMED 0 / PLAUSIBLE 9.** No finding of either kind was reproduced as a
break of a run, which is why none is CONFIRMED; each is a hardening or accuracy
gap in a change that no check here can execute. Every one was fixed rather than
declined, in one pass, and the fixes are in the single commit on this branch:

1. The artifact set no `retention-days`, so how long the re-run this change
   exists for stays possible was whatever the repository default happened to be.
   FIXED: `retention-days: 30`, which is the window GitHub allows a run to be
   re-run in, so the artifact cannot expire before the button does.
2. The re-run benefit was oversold: a failure AFTER the draft is sealed meets an
   immutable published release on the second attempt and reds on the asset
   upload. That asymmetry predates the split. FIXED: named in the workflow
   header, in the CHANGELOG entry and in the Notes below, with
   `regenerate-manifest.yml` named as the repair for that half.
3. The tag was minted from an unvalidated cross-job value:
   `needs.<job>.outputs.<name>` answers the empty string for a name that does
   not exist, so a later rename in the `outputs:` block would mint the tag
   `-JF12-beta.<run>` and burn it permanently on an immutable-releases
   repository - and the gate's regex would never match it again, so the daily
   run would keep re-publishing. FIXED: a step that refuses a version that did
   not survive the boundary, before anything is created.
4. That guard checked the shape of each value and not the relation between them,
   while the git tag is built from one and the installed plugin version from the
   other. FIXED: the same step now refuses a pair that does not name one build.
5. A load-bearing comment ("No NuGet cache: this job publishes with
   contents:write") became false of the job it sits in the moment that job lost
   its write grant. FIXED: the rule is kept and its reason restated - a poisoned
   cache would reach the bytes `publish` releases.
6. My own `retention-days` comment claimed 30 was what every other upload in the
   tree sets, which is not true (the tree carries 5, 14, 30 and 90, each for its
   own consumer). FIXED: the comment gives the re-run window as the reason
   instead of a false house rule.
7. "Re-run failed jobs" re-enters the shared `publish-manifest-beta` concurrency
   group, which holds at most one PENDING run, so a re-run pressed while the
   JF10.11 leg publishes can evict a queued nightly. FIXED as a note where the
   button is described; it self-heals on the next night's gate.
8. "Re-run all jobs" after a sealed release concludes SUCCESS having published
   nothing - the gate finds the tag it minted on the first attempt. Unchanged
   from the single-job shape, but the split makes the button routine. FIXED as a
   note, with `manifest-freshness.yml` named as what notices.
9. A stale-patch hazard rather than a code finding: the availability lens was
   handed a patch file that had fallen behind the worktree and said so, then
   reviewed the committed file. Recorded here because it is the reason its
   verdict is about the state on this branch.

What the lenses walked and could not break: every file and glob the `publish`
job names and where each comes from after the split; the release asset set and
names (nine, byte- and name-identical); the "exactly one `.md5`" invariant of
#942 and the manifest generator's `*.zip.meta.json` input; the gate/skip
semantics on all four re-run shapes; the `contents: read` downgrade against every
step left in `package` (including `yaml-update-action`, resolved at its pinned
digest, which returns before any API call when `commitChange: false`); whether a
release can be cut from bytes the provider matrix did not gate; whether
`manifest-beta` can reach a half-updated state the single job could not; and
whether a re-run can duplicate a release, duplicate a tag or burn one - resolved
in `softprops/action-gh-release`'s source at the pinned digest, where the attempt
one draft is found and updated rather than re-created.

## Quality checklist

      testable unit. The split reuses the shape already in `publish.yml`; no
      step body was rewritten, they were moved.
      change establishes is locked in as a rule. No C# byte changed, so there is
      no new structural property.
- [x] Docs impact considered: this change needs no README/wiki update. The
      workflow's own header comments carry the split and the CHANGELOG entry is
      in this change; nothing on the wiki describes the publish job layout.

## Verification

- [ ] `dotnet build --no-restore --warnaserror` is green.
- [ ] `dotnet test` is green.

The two unticked boxes are a negative disclosure and not a pass: this change
contains no C# byte, so a local build and test run would judge the tree the
mainline already judges rather than this change.

```
$ git diff --name-only origin/5.0...HEAD
.github/workflows/publish-jf12-beta.yml
CHANGELOG.md
$ npx prettier --check CHANGELOG.md
Checking formatting...
All matched files use Prettier code style!
```

What did verify the workflow:

```
$ actionlint .github/workflows/publish-jf12-beta.yml ; echo "EXIT=$?"
EXIT=0
```

That it bites rather than passing vacuously, measured by breaking one reference
on the changed file and restoring it:

```
$ sed -i 's/needs.package.outputs.beta/needs.packge.outputs.beta/' .github/workflows/publish-jf12-beta.yml
$ actionlint .github/workflows/publish-jf12-beta.yml
.github/workflows/publish-jf12-beta.yml:314:110: property "packge" is not defined in object type {gate: {outputs: {changed: string}; result: string}; package: {outputs: {base: string; beta: string}; result: string}} [expression]
```

The inputs used on the two artifact actions exist at the digests they are pinned
to, read at those digests:

```
$ gh api repos/actions/upload-artifact/contents/action.yml?ref=043fb46d1a93c77aae656e7c1c64a875d1fc6a0a --jq .content | base64 -d | grep -n '^  [a-z-]*:'
5:  name:
8:  path:
11:  if-no-files-found:
20:  retention-days:
26:  compression-level:
37:  overwrite:
43:  include-hidden-files:
48:  archive:
$ gh api repos/actions/download-artifact/contents/action.yml?ref=3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c --jq .content | base64 -d | grep -n '^  [a-z-]*:'
5:  name:
8:  artifact-ids:
11:  path:
14:  pattern:
17:  merge-multiple:
23:  github-token:
28:  repository:
33:  run-id:
```

## Notes

**What is NOT verified, stated rather than hidden.** This workflow runs only on
dispatch against the release path, so no check in this repository exercises the
split - the issue says as much in its own last line. It is verified by reading,
by the lenses above, and by the next real publish. Its three load-bearing assumptions were each taken further than a reading
during the review, and none of them was executed on a runner here:

- The artifact is scoped to the RUN rather than the attempt.
  `actions/download-artifact` defaults `run-id` to `github.run_id`, read at the
  pinned digest, and `upload-artifact`’s own `overwrite` description - it fails
  on a name the run already holds - only makes sense under the same scoping.
- Job outputs survive into a re-run. `actions/runner#2598` documents outputs
  being carried from the previous attempt, reported as a defect in the opposite
  direction. The new version-guard step fails closed if that ever stops holding,
  before anything is created.
- The 30-day retention covers the re-run window, which GitHub documents as up to
  30 days after the first attempt.

**No second reader.** Nobody but the author has read this change; what stands in
place of a second reading is the lens record above, run against the diff and the
post-change file.

**Deliberately out of scope.** `publish-beta.yml` carries the same single-job
shape and is untouched: it runs from `main`, which is the ref
`nightly-betas.yml` dispatches it at, so a change made on this branch would not
reach it - and `main` is frozen on the 4.3 candidate for its soak (#1511), where
landing a workflow change would restart that clock. `publish.yml` already has
the split. `publish-jf12-stable.yml` has never fired and belongs to #1729's
plan.


iderex:
Security review of this change, read after it merged rather than before, because it landed before
I reached it. It touches the release path, so it gets one either way. No finding, and the change
reduces privilege rather than adding any.

## What the job that builds holds now

The workflow is dispatched and nothing else, with `permissions: {}` at the top, and each job opts
in:

```
gate              contents: read
provider-matrix   contents: read
package           contents: read     (was contents: write while it also published)
publish           contents: write
```

The job that restores packages and compiles third-party dependencies no longer carries a token
that can write to this repository. Before the split it did, because building and publishing were
one job. That is a narrower surface than the one this change replaced, independent of the 500 it
was written for.

## The handoff between the two jobs

- **The artifact comes from this run.** `download-artifact` is called without a run id, so it can
  only read what `package` uploaded in the same run, and the workflow has no trigger through which
  code from a fork reaches `package`.
- **The versions are checked before anything uses them.** `publish` reads both through `env:` and
  not by interpolating them into the script, refuses a base that is not three numeric parts or a
  beta that is not four, and refuses a pair that does not name this run. The release step only
  runs after that step succeeded.
- **Every third-party action is pinned to a commit**, `upload-artifact` and `download-artifact`
  included, and `checkout` in `publish` drops its credentials.

## What was checked and found not to be a finding

No attestation step existed before the split and none exists after it, so no `id-token` or
`attestations` permission was lost in the move. The SBOM is still generated in `package` and
travels inside the artifact. The artifact is kept for thirty days on a public repository, and it
holds the same files the release publishes publicly.

## Not yet shown

The workflow has not run since it merged: its last run was at 09:45Z, before this change. The
first real proof of the split is the next scheduled dispatch, and a failure there reaches the alert
this repository already watches.

Issue #1121, opened 2026-08-03
Add the admin linked-accounts panel with per-user revoke to the config page


iderex:
Part of #758.

## Claim

The admin config page gains a linked-accounts panel that renders the SSO roster and offers a per-user revoke, so an admin can see who is linked to which provider and cut them off without leaving the dashboard.

## Depends on

The aggregate roster endpoint (sibling issue). This step is the presentation layer only, it adds no new server route.

## Where it lives

- `SSO-Auth/Web/configPage.html` (3110 lines), a new collapsible section alongside the existing provider sections, following the same markup conventions and help-text style already used there.
- `SSO-Auth/Web/config.js` (2603 lines), fetch via `ApiClient.getJSON(ApiClient.getUrl("sso/..."))`, matching the existing calls (`Config/Export` at :1238, `OID/Test` at :1185). Revoke posts to the **existing** endpoint, unchanged: `POST sso/Unregister/{username}` with the target auth provider id in the body, see `SSO-Auth/Api/Http/SSOController.cs:1421`, which is already `[Authorize(Policy = Policies.RequiresElevation)]`, already throttled through `RateLimitCheck(SsoRateLimitClass.Unregister)` (:1431), and already does `RemoveUserEverywhere` (:1448) + `RevokeUserTokens` (:1462).
- `SSO-Auth/Localization/en.json`, every new user-facing string goes through the existing i18n path (`SSOViews/i18n.js`, wired at `config.js:1336`). No hardcoded English in the DOM.

## Constraints

- **Render with `textContent`, never `innerHTML`.** Usernames, provider names and canonical subject keys are all attacker-influenced (a subject comes from the IdP). `SSO-Auth/Web/linking.js:209` already states this rule for the self-service page; the admin panel is the higher-value target and must hold the same line.
- **Revoke is destructive and must confirm.** It removes the user's links from *every* provider and terminates their live sessions, including the admin's own if they revoke themselves. The confirmation must say so.
- Reuse the existing path exactly, do not add a second revoke route, and do not bypass the rate limiter. Note that `Unregister` currently logs a plain `LogInformation` at `SSOController.cs:1466` rather than an `[SSO Audit]`-prefixed line; that is the existing behaviour and reusing it is the ask here, not changing it.
- A revoke does not stop re-adoption when a provider has `AllowExistingAccountLink` on, the controller comment at :1444 spells this out. Surface that caveat in the panel's help text so an admin is not misled about durability.
[...]

PR #1745, opened 2026-09-14
Refuse an administrator's own revoke that strands the server through Unregister

### LINKED ISSUE #1741: The administrator Unregister route has no last-administrator guard, so the lockout #1732 closed is still one click away

iderex:
# What & why

`POST sso/Unregister/{username}` was the second route to the lockout #1732 closed on the self-service page: it removes every SSO link an account holds, repoints it and ends its sessions in one call, and it asked nothing about who the caller was beyond elevation. The settings page offers it per row, the caller's own included. The route now takes the self-service reading before it removes anything, over four facts read at the boundary: the caller is the account being revoked; that account accepts no password; it holds a link on an enabled provider, so the revoke takes a way in; and no other enabled administrator holds such a link, measured with `AdministratorsWithNoWayIn`. Where all four hold it refuses with a 403 that says what was measured, names the two remedies that exist, audits the refusal and changes nothing. The settings page keys a catalogue sentence off that refusal instead of the generic failure. The `Unregister` comment at the token revoke now points at this guard rather than at the durability of the revoke, which is what the issue asked for.

Closes #1741

Means: C# in the existing controller and link service, one xUnit class extended, one catalogue key in both languages, and the page's existing rejection idiom. Nothing new is added to the tree's languages or dependencies, and every claim below carries the command that produced it.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: the guard only ever refuses and sits before any mutation; a survey that throws refuses; a caller the host resolves to neither a user nor an API key is treated as the account itself.
- [x] No secrets logged; the new audit line carries the user id only.
- [x] A security review was performed: five refute-by-default lenses on the first state, three fresh lenses on the second, one fresh lens on the third. There is no second reader on this account; the review record below stands in place of one and I say so plainly.
- [x] Review record, per lens as raised, with dispositions:
  - **Availability, round 1: FAIL.** CONFIRMED 1 / PLAUSIBLE 2. An API-key caller was read as the holder of every account it revoked, refusing the documented automation path: FIX (`CallerIsTheHolder` answers false for an API key, read from the host's own flag; test `Unregister_AnApiKeyCaller_IsNotTheHolder_AndRevokes`, red with the arm removed). The guard refused a self-revoke of an account holding no enabled link, closing the repair route for an administrator already stranded on this plugin's provider id: FIX (`UserHoldsAnEnabledLink` is the third fact; two tests, red with the arm removed). A verified break-glass administrator counts as no way in: DECLINE, that reading is #1732's decision for the sibling route and telling the minted passwords apart is #1733, both of which repair both routes at the one place the fact is read.
  - **Bypass, round 1: PASS.** CONFIRMED 1 / PLAUSIBLE 1. The refusal and the audit line state a measured fact when the survey threw: DECLINE, the Warning line beside them carries the cause, and the shape is shared with #1732's route. Two administrators self-revoking at the same moment both pass: DECLINE as inherited and stated in the code, the changelog and the API document; the user records are not under the configuration lock.
  - **Correctness, round 1: NEEDS_IMPROVEMENT.** CONFIRMED 2 / PLAUSIBLE 2. The two code findings are the same two the availability lens raised and are fixed above. The page left the progress line up while the body was read: FIX (the generic sentence is rendered first, the refusal replaces it, as `linking.js` does). The revoke comment claimed a guarantee the code does not give: FIX (it names the window).
  - **Integration, round 1: NEEDS_IMPROVEMENT.** CONFIRMED 3 / PLAUSIBLE 1. API key, as above: FIX. The API document contradicted itself about the self-service route: FIX (the #1732 paragraph now states the rule #1732 made). The confirm dialog's comment claimed #1121's proceed-always decision for every row: FIX (comment names the one refusable row). Progress line, as above: FIX.
  - **Execution, round 1: PASS.** Ran both TFM builds, the touched and neighbouring classes, a six-fact probe file since deleted, both catalogues under node, and the page handler under node with nine rejection shapes. One observation, an authorization context that throws escapes as a 500 before any mutation: DECLINE, pre-existing helper behaviour shared with the self-service route and fail-closed.
  - **Availability, round 2 (fresh): NEEDS_IMPROVEMENT.** CONFIRMED 2 / PLAUSIBLE 3. The guard lacked the password-door fact the self-service reading includes, refusing the ordinary single-owner server: FIX (`CallerHasNoPasswordDoor` is the second fact; test `..._ThatAcceptsAPassword_StillRevokes_AndPaysNoSurvey`, red with the arm removed). The document's self-service sentence dropped the same condition: FIX. Break-glass, the two-transaction window and the absence of an after-the-fact net: DECLINE, as above; the purge's net exists because a purge is a mass action, and a post-check here would report as an error a removal the caller was told was allowed.
  - **Correctness, round 2 (fresh): NEEDS_IMPROVEMENT.** CONFIRMED 2 / PLAUSIBLE 1. The document's self-service sentence, as above: FIX. The link fact is read outside the removal transaction and the comment implied otherwise: FIX (the bound is named at the guard). A test comment misdescribed the throw: FIX.
  - **Integration, round 2 (fresh): NEEDS_IMPROVEMENT.** CONFIRMED 1. The document's self-service sentence, as above: FIX. It verified `AuthorizationInfo.IsApiKey` on both pinned Jellyfin lines, the API-key population path in the host, the rejection shape the vendored API client hands the page, and every conformance test that reads the page or the catalogues as text.
  - **Availability, round 3 (fresh): NEEDS_IMPROVEMENT.** CONFIRMED 2 / PLAUSIBLE 1, all about wording: the disabled-provider carve-out's cost was claimed away as "loses nothing". FIX in one pass at the guard, the test, the changelog and the document; per `store/spec/review-round-stops.md` a round that changes only text earns no further round, and none was run. The other two are costs the lens recorded as accepted rather than defects.
- [x] Log inputs from the IdP are sanitized inline at the log call: the new line prints a `Guid` and nothing foreign.

## Quality checklist

- [x] No duplicated logic: the link reading is one method over the walk `AdministratorsWithNoWayIn` already uses, and the survey is the method #1732 landed.
- [x] Comments explain why; the architecture-conformance rules that read the page and the catalogues as text still pass.
- [x] Architecture-conformance tests pass; no new structural property was established.
- [x] Docs impact considered: `docs/ACCOUNT-MANAGEMENT-API.md` and `CHANGELOG.md` are updated here.

## Verification

Run at `2c581fbf` in a worktree cut from `origin/5.0` at `0c92cdb1`.

- [x] `dotnet build SSO-Auth.sln --no-restore --warnaserror` (Debug): `0 Fehler`, `0 Warnung(en)`, exit 0.
- [x] `dotnet build SSO-Auth.sln -c Release --no-restore --no-incremental --warnaserror`: `0 Fehler`, `0 Warnung(en)`, exit 0.
- [x] `dotnet test SSO-Auth.sln --no-build` (Debug, both TFMs): `gesamt: 8025, fehlgeschlagen: 0, erfolgreich: 8017, übersprungen: 8`. Release: the same counts.
- [x] `dotnet exec SSO-Auth.Tests/bin/Debug/<tfm>/SSO-Auth.Tests.dll -class Jellyfin.Plugin.SSO_Auth.Tests.SSOControllerUnregisterTests`: `Total: 19, Failed: 0` on net10.0 and net9.0.
- [x] Guard proofs, each a one-site mutation rebuilt and run: guard off, 4 refusal tests red; survey gate off, 8 tests red; API-key arm off, 1 red; enabled-link fact off, 2 red; password-door fact off, 1 red. Source restored and green after each.
- [x] `npx prettier@3 --check SSO-Auth/Web/sso-core.js docs/ACCOUNT-MANAGEMENT-API.md CHANGELOG.md`: `All matched files use Prettier code style!`
- [x] `bash scripts/check-jellyfin-compat.sh`: `Jellyfin compatibility metadata is consistent.`
- [x] `git status --porcelain -- '*packages.lock.json'`: empty.

## Notes

Not covered, stated rather than left to be found: the survey and the link fact are read outside the removal's transaction, so two administrators revoking themselves at the same moment can each see the other and both pass, the same window #1732 records; a link on a switched-off provider is not a way in, so an administrator who switches their only provider off and then revokes their own account is not refused, the self-service reading's own cost; a break-glass administrator with a real password reads as having no way in, and an account already on the password provider behind a minted password reads as having a door on both routes, which is #1733. Not verified on a live server: an API-key `POST sso/Unregister/<user>` against a server whose only administrator signs in by password, and the settings-page sentence behind a buffering reverse proxy; both belong on the manual E2E checklist for the 5.0 candidate.

**The German row is not vouched for on this head, and I am not signing it.** The hygiene gate refuses this pull request until a line `Catalogue-Vouch: de read by <name>` is in this body, and that line is a person putting their name to having read the language. I wrote the row; writing the vouch for it as well would turn an unread translation into a read one by typing a sentence. So it stays absent and the gate stays red until somebody who reads German adds it, which is the one thing between this head and a merge; every other check on it is green.

The row is quoted here so that reading it costs nobody a checkout. It is the settings page's sentence for the refusal, modelled on `link.delete_refused_would_strand_server`, and it carries no placeholder:

```
"config.linked_accounts_revoke_refused_would_strand_server":
"Der Server hat es abgelehnt, Ihre eigenen SSO-Verknüpfungen zu widerrufen: Kein anderer Administrator dieses Servers hat eine SSO-Verknüpfung, mit der er sich anmelden kann, der Widerruf hätte diesen Server also möglicherweise ohne erreichbaren Administrator zurückgelassen. Bitten Sie einen anderen Administrator, sie für Sie zu widerrufen, oder verknüpfen Sie zuerst ein anderes Administratorkonto mit einem Anbieter und widerrufen Sie dann Ihre eigenen. Es wurde nichts geändert."
```

The English it renders:

```
"config.linked_accounts_revoke_refused_would_strand_server":
"The server refused to revoke your own SSO links: no other administrator on this server holds an SSO link that can sign them in, so the revoke could have left this server with no administrator able to reach it. Ask another administrator to revoke them for you, or link another administrator account to a provider first and then revoke your own. Nothing was changed."
```


iderex:
Security review of this head before it lands, because it adds a refusal on the route an account
is unregistered by and changes who counts as the holder of an account. No exploitable finding.

## The refusal comes before anything changes

In `Unregister` the new check runs before `RemoveUserEverywhere`, before the provider is switched
and before the tokens are revoked. A refused call leaves the links, the provider and the sessions
exactly as they were, which is what the refusal sentence says in both catalogues.

It refuses only where all four hold: the caller is the account being unregistered, that caller has
no password to sign in with, the account holds a link on an enabled provider, and no other
administrator keeps a way in. Each of the four is a narrowing, so the change can only refuse more
than before and never admit anything that was refused.

- **No authorization context.** `CallerIsTheHolder` answers true, so the caller is treated as acting
  on itself and the refusal is reachable.
- **A caller the host cannot resolve to a user.** Still true, for the same reason; this is unchanged.
- **The survey of the other administrators throws.** It answers that nobody else can sign in, which
  refuses. This is the method `#1742` reviewed, with its log line generalised from "self-unlink" to
  "removal".

## The API key, read closely because it is the one place something answers false

`CallerIsTheHolder` now answers false when the host reports `IsApiKey`. That is not inferred from a
missing user, which would also cover a caller the host failed to resolve; it is the host's own flag,
set only when the request carried a valid dashboard API key. An API key belongs to no account, so it
has no account to strand, and it stays a way into the server after the call. Treating it as the
holder of every account it acted on refused the documented automation path, which the head of the
method records.

It does not open the self-unlink guard of `#1742` to an API key either: that guard reads
`IsAdministrator` first, which answers false for a caller with no user, so an API key meets the
ordinary refusal there and never the administrator exemption this value narrows.

## What reaches the caller

The refusal sentence names no account, the audit line carries the id of the account acted on, and
the survey's roster never leaves the method.

## One edge, walked and not reported as a finding

As on `#1742`, the survey reads outside the transaction that removes the links, so two
administrators unregistering their own accounts in the same moment could each find the other still
able to sign in. That takes two trusted administrators racing each other on an SSO-only server and
gives nobody a capability they did not hold.

PR #1747, opened 2026-09-15
Promote the 4.3 line to its first stable release [#1511]

iderex:
Refs #1511. The promotion commit for `4.3.0-stable`, in the shape #1511 decided on 2026-09-04: the soaked commit on `main` is `df44f000`, and this adds the three things owed at promotion that could not sit in it during the soak.

## What changes

- `README.md` and `SSO-Auth/Web/configPage.html`: the maturity label moves from Release Candidate to **Full Release**. The page is source, so this sentence is the one difference in the shipped assembly between the soaked commit and the tagged one.
- `CHANGELOG.md`: `Unreleased` and the earlier `4.3.0` section become one `4.3.0` section, subsections merged as Added, Changed, Fixed, Security. 58 entries before and 58 after; the text grows by exactly the seven lines of the new opening paragraph, which states what was not verified for this release and why.
- `.github/workflows/publish.yml`: the seven-stack provider matrix is wired into the stable leg, as its header promised for the change that cuts the first stable tag. It runs as a job at the tagged commit before `build`, with `contents: read` only, so a red provider publishes nothing.

## Decided for this promotion

- The soak ends on 2026-09-15 by decision: ten days on `4.3.0-beta.61` and two on `df44f000`.
- The manual end-to-end QA checklist against a real Jellyfin 10.11 server is not walked, because work on the Jellyfin 10.11 line ends with this release. The provider matrix, run in full against this commit before the tag and again at the tag, stands in for it. The changelog says so.
- The two native-client Quick Connect items stay not verified, as decided on 2026-09-04.

## Before the tag

The full matrix is dispatched against this branch before merging, so the tag does not meet a red provider for the first time on publish day.


iderex:
Security review of this head before it lands, because it changes the workflow that publishes the stable channel. No finding.

## The one workflow change

`publish.yml` gains a `provider-matrix` job and `build` now `needs` it. Read against what it could widen:

- **Permissions.** The new job grants `contents: read` and nothing else. The called `e2e-login.yml` sets `permissions: {}` at its top level and uses no `secrets.` anywhere, so nothing beyond a read token reaches the harness. `build`, `upload` and `generate` keep the permissions they had, and `build.yml`, which carries the signing scopes, is unchanged by this head.
- **Supply chain.** The job calls a workflow file in this repository at the same commit. No third-party action is added.
- **Bypass.** `build` depends on the matrix with no `if:` of its own, so a failed matrix skips `build`, `upload` and `generate` in turn, and nothing is released. The conditions inside `e2e-login.yml` are step-level choices within each stack, not job-level skips that could empty the matrix; the dispatch against this head shows the full set running: a pairwise co-existence job and the seven provider stacks, Dex, Keycloak, Kanidm, authentik, Authelia, Zitadel and Pocket ID.

## The other three files

`README.md` and `configPage.html` change one maturity label each, with no markup or script around the text touched. `CHANGELOG.md` is prose.

## What this does not cover

The harness proves that login works across the seven stacks against this commit. It is not the manual end-to-end checklist, and the changelog says so in its opening paragraph, with the decision it rests on.

iderex:
Replaced by #1748 and closed unmerged.

Prettier refused `SSO-Auth/Web/configPage.html` on this head: the maturity label got shorter, so the paragraph it sits in reflows. The fix is that reflow and nothing else, folded into the one commit and pushed under a new branch name rather than onto this one. The provider matrix dispatched against this head passed all eight jobs, and it is dispatched again against the replacement.

PR #1748, opened 2026-09-15
Promote the 4.3 line to its first stable release [#1511]

iderex:
Refs #1511. The promotion commit for `4.3.0-stable`, in the shape #1511 decided on 2026-09-04: the soaked commit on `main` is `df44f000`, and this adds the three things owed at promotion that could not sit in it during the soak.

## What changes

- `README.md` and `SSO-Auth/Web/configPage.html`: the maturity label moves from Release Candidate to **Full Release**. The page is source, so this sentence is the one difference in the shipped assembly between the soaked commit and the tagged one.
- `CHANGELOG.md`: `Unreleased` and the earlier `4.3.0` section become one `4.3.0` section, subsections merged as Added, Changed, Fixed, Security. 58 entries before and 58 after; the text grows by exactly the seven lines of the new opening paragraph, which states what was not verified for this release and why.
- `.github/workflows/publish.yml`: the seven-stack provider matrix is wired into the stable leg, as its header promised for the change that cuts the first stable tag. It runs as a job at the tagged commit before `build`, with `contents: read` only, so a red provider publishes nothing.

## Decided for this promotion

- The soak ends on 2026-09-15 by decision: ten days on `4.3.0-beta.61` and two on `df44f000`.
- The manual end-to-end QA checklist against a real Jellyfin 10.11 server is not walked, because work on the Jellyfin 10.11 line ends with this release. The provider matrix, run in full against this commit before the tag and again at the tag, stands in for it. The changelog says so.
- The two native-client Quick Connect items stay not verified, as decided on 2026-09-04.

## Before the tag

The full matrix is dispatched against this branch before merging, so the tag does not meet a red provider for the first time on publish day.


iderex:
The security review of this change is on #1747, the head it replaces: https://github.com/Flowfin/jellyfin-plugin-sso/pull/1747#issuecomment-5676877585. This head differs from that one only by the reflow of one paragraph in `SSO-Auth/Web/configPage.html` that Prettier asked for, so the review stands as written.

PR #1749, opened 2026-09-15
Stop the nightly JF10.11 beta, now that 4.3.0-stable has shipped [#1511]

iderex:
Refs #1511. The Jellyfin 10.11 line ended with `4.3.0-stable`, published on 2026-09-15, so the nightly scheduler stops dispatching its beta.

## What changes

- `.github/workflows/nightly-betas.yml`: the step that dispatched `publish-beta.yml` on `main` is removed. The JF12 step is unchanged. The header and the step comment no longer describe a second leg.

## Why it is removed rather than left running

A beta minted from `main` after the stable would outrank it on the beta channel for a line nobody works on. `publish-beta.yml` itself stays and can be dispatched by hand if that is ever wanted.

## Also worth knowing

`publish-beta.yml` creates its release without `target_commitish`, so GitHub put the tag `4.3.0-beta.72` on the head of the default branch, `5.0`, while the package was built from `main` at `df44f000`. With the leg gone that does not recur; the two JF12 publish workflows already pass `target_commitish`.


iderex:
Security review of this head: no finding. It removes one step that dispatched `publish-beta.yml` and rewrites two comments. The job keeps `actions: write`, which the remaining JF12 dispatch needs, adds no action, no permission and no input, and the workflow-level `permissions: {}` is unchanged.

Issue #1746, opened 2026-09-15
The SSO-only break-glass reading counts a minted password as a way in, so the recovery door it proves may not open


iderex:
#1733 makes the last-link self-unlink guard stop counting a password this plugin minted as a way in. Two readings in `SsoOnlyLoginService` still count any non-empty stored password as a usable one, and one of them is what proves the break-glass recovery door before SSO-only login is switched on. I found this while building #1733 and read it at `origin/5.0` `0c92cdb1` on 2026-09-15.

Split: none

## The two readings

The break-glass state handed to the activation guard:

```
$ git show origin/5.0:SSO-Auth/Api/Session/SsoOnlyLoginService.cs | sed -n '93,94p'
        var usablePasswordLogin = SsoAuthenticationProviders.IsDefaultPasswordProvider(user.AuthenticationProviderId)
            && !string.IsNullOrEmpty(user.Password);
```

And the per-account description the per-provider bulk unlink's mass-lockout guard consumes:

```
$ git show origin/5.0:SSO-Auth/Api/Session/SsoOnlyLoginService.cs | sed -n '179,180p'
            SsoAuthenticationProviders.IsDefaultPasswordProvider(user.AuthenticationProviderId),
            !string.IsNullOrEmpty(user.Password));
```

Both are the same test: on the built-in password provider, and holding some stored password.

## Why it matters

The first one is the whole of the activation guard's password condition:

```
$ git show origin/5.0:SSO-Auth/Config/SsoOnlyLoginGuard.cs | sed -n '100p'
        if (!breakGlass.HasUsablePasswordLogin)
```

On a server whose provider `DefaultProvider` names the built-in password provider, every SSO login writes that id onto the account and the stored password behind it is one this plugin minted - 64 CSPRNG bytes, never displayed, never recoverable. An operator who names such an account as the break-glass administrator passes the guard and switches SSO-only login on, and the recovery door the guard exists to prove does not open for anybody. That is the same ambiguity #1733 closed one route over, on the route where the consequence is a whole server rather than one account.

#1733 gives the tree the fact needed to tell the two apart: `CanonicalLinkService.HoldsOnlyAProvisionedPassword`, the digest map it reads, and a record written at both mint sites.

## The direction is not symmetric with #1733

[...]

Issue #1744, opened 2026-09-14
The wiki does not describe the refusal that stops a last-link unlink from stranding an account


iderex:
The wiki does not describe the refusal that stops a last-link unlink from stranding an account, on any of the three routes that now carry it.

Split: none

## What is missing

Three changes landed the same rule on the self-service linking page, and the wiki names none of them. `#1720` made the server refuse a holder's removal of their own last usable link where the account has no password door; `#1732` narrowed the administrator exemption to the case where somebody else still holds a way in; `#1733` made the plugin record which stored passwords it minted, so the refusal reaches the accounts it was written for instead of reading them as password-holders.

Read on the wiki today from a fresh clone:

```
$ grep -rn 'last link\|last usable link\|no password door\|strand' Linked-Accounts.md Login-Flow.md Security-Model.md
$ echo "exit=$?"
exit=1
```

So a user who meets the refusal on `/SSOViews/linking`, and an administrator asked why they cannot remove their own last link, have nowhere to read what the rule is.

## What a reader needs

- **That the removal can be refused at all**, on the self-service page rather than only on the admin surface, and what the refusal says.
- **The condition.** The last link that still signs the account in, on an account whose only other way in is a password nobody can use.
- **The administrator case.** An administrator removing somebody else's link is unaffected; an administrator removing their own last link is refused only where no other administrator would be left with a way in.
- **What counts as a way in**, and what does not: a link on another **enabled** provider counts; a link on a disabled provider does not; a stored password counts only where this plugin did not mint it.
- **The residual, stated as a residual.** An account sealed by a plugin version before `#1733` carries no record, so it reads as holding its own password and is not protected by the rule. Nothing can tell those apart after the fact.
- **The way out either way:** link a second provider first, or ask an administrator.

## Where it belongs

`Linked-Accounts` already has "The revoke, and the three things to know before pressing it" for the administrator surface, and its See-also points at `Login-Flow#self-service-linking` for the self-service one. The self-service refusal belongs with the page a reader is on when they meet it, with one sentence from `Security-Model` where the guard family is listed.

## Done when

The wiki says that a last-link self-unlink can be refused, under what condition, how the administrator case differs, what counts as a way in, what the pre-`#1733` residual is, and what the way out is.

## Ordering

[...]

Issue #1043, opened 2026-07-27
Replace the hand-rolled duplicate-name walk with the System.Text.Json Strict preset once net9.0 is dropped


iderex:
Found while reviewing #1032 (PR for #1005), dimension "state of the art".

`StrictJson.Inspect` is a hand-rolled `Utf8JsonReader` walk because the framework member that expresses the same policy, `JsonSerializerOptions.Strict` and its `AllowDuplicateProperties` switch, exists only on .NET 10. The plugin binds the HOST`s System.Text.Json (.NET 9`s in the Jellyfin 10.11 line, .NET 10`s in the 12.0 line), so referencing it fails the net9.0 build with CS0117. That is measured, not assumed: the reasoning and the compile failure are recorded in the file`s own documentation and pinned by `TheRepeatedMemberWalk_TakesTheDuplicateDecisionItself` and `TheRepeatedMemberWalk_HasOneCodePathOnBothTargets` in `SSO-Auth.Tests/ArchitectureConformanceTests.RepeatedMemberWalk.cs`.

When the net9.0 leg is dropped, revisit. Two things to weigh rather than a foregone conclusion:

- `AllowDuplicateProperties` refuses a repeat ANYWHERE in the document. #1005 deliberately screens only the object scopes the caller indexes, because a repeat in `mtls_endpoint_aliases` or an unwalked subtree of a role claim decides nothing the plugin reads, and refusing on it takes a working provider`s logins offline. The framework switch is the blunter instrument, and swapping to it would re-introduce exactly the availability regression #1032 removed.
- It also applies to a deserializer the plugin does not own on either fact-reading path (both are Newtonsoft), so it would not cover the readers the screen exists to keep honest.

So this may well close as "keep the walk". Filed so the decision is made deliberately and recorded, rather than the walk outliving its reason unexamined.

Depends on: the Jellyfin 10.11 / net9.0 line being dropped.

## Corrected 2026-09-07: the pin this body named resolved to nothing

*Re-read 2026-09-13: the ref every command in this section names, `origin/4.4`, no longer exists, and the default branch is `5.0`. The section after this one carries the same two readings at a ref that resolves.*

[...]

PR #1750, opened 2026-09-15
Record which passwords this plugin minted, and count only those as no door

### LINKED ISSUE #1733: A minted password reads as a way in, so the self-unlink guard does not reach the accounts it was written for

iderex:
# What & why

Closes #1733.

The last-link self-unlink refusal (#1720) asks whether the caller's account has a
way in other than its SSO links, and it could answer that from one fact: does
`AuthenticationProviderId` name Jellyfin's built-in password provider. That
reading cannot see the account that IS on the password provider and still holds
no usable password. This plugin mints one onto every account it provisions - 64
CSPRNG bytes, never displayed, never stored anywhere else, never recoverable -
and recorded nowhere which accounts those were, so a stored hash was a credential
somebody holds or a seal nobody can open and the two were the same bytes. Where a
provider's `DefaultProvider` names the built-in password provider - a value the
settings page offers as a common choice - every account that provider creates
lands in exactly that state, so on those servers the guard never evaluated.

The shape was decided on the issue on 2026-09-14: record which passwords the
plugin minted and treat only those as no door. What is recorded is a DIGEST of
the stored value rather than a flag, so the record corrects itself the moment
anything else writes that account's password - a user who sets a real password
keeps full control of their own links, with no event to subscribe to and nothing
to migrate.

Both routes that ask the question get the answer. `Unregister` took the same
last-administrator reading in #1741 and named this population as the one it did
not reach; it asks through the same helper, so handing that helper the record
closes both routes at the one place the fact is read.

**Means:** C# in the existing plugin assembly, because the fact has to be read at
an existing request boundary and persisted in the configuration object Jellyfin
already serializes - no language, runtime or dependency is added, and the
property is refusable by the suites and the conformance gate that already exist.

## Type of change

- [ ] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved. The new arm is the LAST term of an `||` whose true
      value makes a guard REFUSE, so it can only turn "has a door" into "no
      door", never the reverse. An unresolved caller, a missing authorization
      context, an empty stored password, an account the record does not name and
      a record whose digest no longer matches all answer the way that does not
      refuse; a missing detector throws before any answer, because neither
      default is safe - `false` silently restores the gap on every server and
      `true` refuses every last-link self-unlink on every server.
- [x] No secrets logged. What is stored is a SHA-256 digest of the stored hash,
      never the hash; it is `[JsonIgnore]`, appears in no export allowlist, and
      no line logs it.
- [x] A security review was performed - see the record below.
- [x] Review record below, with a disposition per finding.
- [x] Log inputs from the IdP are sanitized inline at the log call: this change
      adds no log line carrying IdP input.

### Review record

**There is no second reader for this change.** Four refute-by-default lenses were
run over it instead, and this record stands in place of a reading by somebody
else rather than claiming to be one.

Raised as CONFIRMED **0** / PLAUSIBLE **0** for the exploitable classes: no lens
produced a reproduction of a bypass, a leak or a fail-open, and the bypass lens
reports PASS after measuring both JSON directions, the export allowlist, the
import and declarative routes, and the rollback path. Everything below is an
availability, correctness, accounting or accuracy defect.

| Lens | Verdict | Findings and disposition |
| --- | --- | --- |
| availability / mass lockout | FINDINGS | 1 MEDIUM, 2 LOW, 1 INFO |
| security bypass / fail-open | PASS | 4 non-blocking observations |
| correctness | FINDINGS | 3 LOW, 3 INFO |
| integration | FINDINGS | 2 MEDIUM, 1 LOW-MEDIUM, 1 LOW |

Fixed in this head:

- **The refusal named a remedy the new population cannot use** (availability
  MEDIUM). The sentence said "this server does not accept a password for your
  account", which is false for an account that IS on the password provider, and
  sent the reader to "ask an administrator to switch your account back to
  password sign-in" - a no-op for them, and through this plugin's own Unregister
  route a repoint with no password set, which completes the stranding. Both
  refusal sentences, the catalogue rows and the page's own copies now name the
  state both populations are in and a remedy that ends it for both.
- **The boot reclaim trusted one host answer in the destructive direction**
  (availability LOW, bypass observation 1). "Not found" and "not answering"
  arrive as the same silence, and dropping records on it would have reverted the
  rule server-wide with nothing to write them back. Nothing is reclaimed now
  unless the host answered for at least one account in the same pass.
- **The reclaim asked the database once per record** (integration MEDIUM).
  `GetUserById` is a database read on the targeted builds rather than a cache
  hit, so a record-sized server paid a query per record at every boot - on the
  same code path that refuses a per-account configuration write one line earlier.
  The pass keeps the ids its own loop already resolved and asks the host only
  about the records that set does not answer for, so a record whose account still
  holds a link costs nothing. The one-query form is not available at the declared
  ABI floor; the section below is what said so.
- **A deletion cost two whole-configuration persists** (correctness LOW). The
  record drop and the link removal now ride one transaction where the account
  holds both, which also closes the window where the record was gone and the
  links were not.
- **The empty-password guard was load-bearing and unproven** (correctness LOW).
  Deleting it does not redden any test in the first head; it throws
  `ArgumentNullException` on a reachable state - a record naming an account whose
  password an administrator has since cleared - which is a 500 on the caller's
  own unlink. A row now covers it, and reddens when the guard goes.
- **The map had no JSON-omission pin** (bypass observation 2), where every
  neighbouring server-managed map has one. Added, under the default policy and
  under camelCase.
- **Two statements of fact went stale** (correctness INFO): the purge guard and
  its remarks still said the plugin "records nowhere which", and the walk tool
  still argued that a minted password lets the removal through.
- **Two operator documents went out of step** (integration LOW-MEDIUM and LOW):
  the migration document's "deliberately never restored" table did not name this
  record, although a migrated server gets an inert guard because of it; and the
  cross-guard disagreement with the SSO-only break-glass check was written in the
  code and nowhere an operator reads. Both now say it.
- Dead usings left by a removed helper.

Declined, with the reason:

- **Fold the create arm's record into the link write** to save a second
  whole-configuration persist (availability LOW, bypass observation 4,
  integration MEDIUM). The link write is behind the #133 race, where the LOSER
  writes no link at all - so the record would be skipped for an account that is
  nonetheless sealed and persisted, which is the hole this change exists to
  close. The cost is paid once per account ever provisioned rather than once per
  login, and it is now stated at the call rather than left to be found.

Deferred, as its own issue:

- **The SSO-only break-glass reading still counts a minted password as a way in**
  (raised by the correctness lens as context, and by my own reading while
  wiring the boundary). That guard proves the recovery door before SSO-only login
  is switched on, and on a server whose provider names the built-in password
  provider it can still be satisfied by an account nobody can sign in to. Opened
  as #1746 and named in the code, the changelog and the API document.
- **Per-account resilience in the boot sweep** (correctness LOW). Anything thrown
  out of that loop abandons the pass's collected records along with the pass, so
  accounts it had already sealed stay unrecorded. The direction is the safe one -
  they read as holding their own password, which is what every account said
  before this change - and narrowing it means changing what a failed pass does
  rather than what this record says. Stated at the code and in the entry.

Not measured, stated rather than claimed: whether `User.Password` is populated on
the account this plugin resolves from the authorization context is not measured
on this route in this tree. The neighbouring #1720 arm already reads
`AuthenticationProviderId` off that same object in production.

## Quality checklist

- [x] No duplicated logic: one type owns the mint, the record, the drop and the
      read; both mint sites write through it and both routes read through one
      helper.
- [x] The change adds no more code than the problem requires. The one addition
      the issue did not ask for is the reclaim, and it exists because the record
      this change introduces would otherwise leak an entry no route could ever
      remove.
- [x] Comments say why, not what.
      locked in: the map is added to the `ArchitectureConformanceTests.State`
      exemption list with the bound that earns it written beside it - one entry
      per account, keyed on the account, so it needs no cap and no sweep.
- [x] Docs impact: `docs/ACCOUNT-MANAGEMENT-API.md` and `docs/SERVER-MIGRATION.md`
      are updated in this change. The wiki's self-service-unlink page describes
      the refusal in the same words the API document used, so it needs the same
      sentence - filed in Notes below rather than edited here, because wiki edits
      are gated.

## Verification

- [x] `dotnet build --warnaserror` green: `SSO-Auth.Tests` on **net9.0** and on
      **net10.0**, 0 warnings 0 errors, and the plugin project in Release too.
- [x] `dotnet test` green: **4038** passed on net9.0 and **4039** on net10.0, 0
      failed, 4 skipped (the LAN-bind tests that need a firewall consent dialog).
- [x] Prettier clean on every touched `.md` and `.js`, checked against the
      committed blobs rather than the Windows working copy.

Every new arm carries its deletion proof; the list is in the commit message with
the row that reddens and the row beside it that stays green.

## Catalogue vouch

**This head changes `de.json` and carries no `Catalogue-Vouch:` line, so the
PR-hygiene check fails on purpose until somebody who reads German adds one.** I
do not write a name against a reading that did not happen. The two rows that
changed, quoted so the reading can be done from this body:

```json
  "link.delete_refused_would_strand": "Der Server hat es abgelehnt, die letzte SSO-Verknüpfung zu entfernen, mit der Sie sich anmelden können: Dieses Konto hat kein Passwort, mit dem sich jemand anmelden kann, das Entfernen hätte Sie also überhaupt nicht mehr anmelden lassen. Verknüpfen Sie zuerst einen weiteren Anbieter und entfernen Sie dann diesen, oder bitten Sie einen Administrator, ein Passwort für dieses Konto zu setzen und es wieder auf die Passwortanmeldung umzustellen. Laden Sie die Seite neu, um zu sehen, welche Verknüpfungen sie jetzt führt.",
  "config.linked_accounts_revoke_refused_would_strand_server": "Der Server hat es abgelehnt, Ihre eigenen SSO-Verknüpfungen zu widerrufen: Kein anderer Administrator auf diesem Server hält eine SSO-Verknüpfung, mit der er sich anmelden kann, der Widerruf hätte diesen Server also ohne erreichbaren Administrator zurücklassen können. Bitten Sie einen anderen Administrator, sie für Sie zu widerrufen, verknüpfen Sie zuerst ein weiteres Administratorkonto mit einem Anbieter und widerrufen Sie dann Ihr eigenes, oder setzen Sie im Jellyfin-Dashboard ein Passwort für ein Administratorkonto - ein Passwort, das dieser Server für ein Konto erzeugt hat, ist keines, mit dem sich jemand anmelden kann. Es wurde nichts geändert.",
```

Both say the same thing as their English rows: the account has no password
anybody can sign in with, and the remedy is a password SET on the account rather
than a switch back to password sign-in on its own.

## What CI found after the first head was pushed

The **ABI floor build** refused it, and it was right to. The reclaim had been
narrowed to one roster query with `IUserManager.GetUsersIds()`, which exists on
10.11.11 and on 12.0.0 - and the declared `targetAbi` floor is **10.11.0**, where
that interface carries neither `GetUsersIds` nor `GetUsers`. That is why the
SSO-only enforcement in this plugin reaches the roster by reflection and fails
closed when it is absent, and my reading of the packages this machine holds did
not see it.

The second commit keeps the ids the loop already resolved and asks the host only
about the records that set does not answer for - a record whose account still
holds a link costs nothing, and what is left is the records whose account holds
no link, which is the population the reclaim is hunting. Reproduced locally with
the job's own command before pushing:

```
dotnet build SSO-Auth/SSO-Auth.csproj -f net9.0 -c Release --warnaserror   -p:JellyfinVersion=10.11.0 -p:RestoreLockedMode=false
```

**This head is two commits rather than one**, and that is deliberate: the first
was already pushed when the floor build refused it, and a correction under a new
name or a rewritten history are both worse than a second commit that records what
the gate caught.

## Notes

- **What CI cannot exercise here, and what a manual pass should cover.** Provision
  an account through a provider whose Default Provider is the built-in password
  provider and confirm the element appears in `SSO-Auth.xml`; save an unrelated
  toggle on the settings page and confirm the element survives the save - that is
  the config-save trap and only a real PUT proves it; restart and confirm the
  self-unlink of that account's last link is refused with the new sentence;
  delete the account and confirm the entry goes; downgrade to the previous plugin
  build, start and save once, and confirm the server comes up with the element
  simply dropped.
- The wiki's self-service-unlink page needs the same sentence the API document
  took: a password this plugin minted is not a door. It is not edited here.


iderex:
The required `ABI floor build` is red on this head, and the vouch is not the only thing between it and a merge.

```
SSO-Auth/Api/Session/PasswordlessLinkedAccountSweep.cs(146,41): error CS1061: 'IUserManager' does not contain a definition for 'GetUsersIds'
```

The floor build compiles against Jellyfin `10.11.11`, and `GetUsersIds` is not on `IUserManager` there; a local build of both target frameworks can pass while this one fails, because it resolves a newer package. The tree already has an accessor for exactly this: `AllUsers()` in `SsoOnlyLoginService` binds whichever user accessor the loaded host exposes. It is private to that service today, so the sweep needs it reachable rather than a second copy of it.

For the record of why this still matters although work on Jellyfin 10.11 ended with `4.3.0-stable`: this branch multi-targets `net9.0` until #1043 drops it, and `ABI floor build` is a required check on `5.0`, so it has to pass for this change to land at all.

iderex:
Security review of this head, `31ede303`, before it lands: it changes what counts as a way into an account and adds a persisted record the lockout guards read. No exploitable finding.

## What is stored, and what it would give away

`ProvisionedPassword.Mint` draws 64 random bytes, hands them to the host's `CreatePasswordHash` and returns only the resulting hash; the random input is never stored or logged. `Record` keeps `SHA-256` of that stored hash, keyed by the account's id. A leaked record is the digest of a salted hash of 512 random bits, so it names no password anybody could type, and it carries no username or subject. `IsTheOnlyPassword` compares digests with `CryptographicOperations.FixedTimeEquals`, and an account whose password was set by a person no longer matches, so it counts as a door again.

## Whether a configuration write can move it

Both directions matter here, and both are closed. Adding an entry would mark an account as having no way in and make a guard refuse; dropping one would clear a refusal for an account that is really sealed. `ProvisionedPasswords` is `[JsonIgnore]`, so an export never carries it and a configuration `PUT` cannot set it, and `ServerManagedFields` puts the live map back on every save, so a save cannot wipe it either. The two mint sites and the pruner are the only writers.

- **No detector wired.** `CallerHasNoPasswordDoor` throws on a null `holdsOnlyAMintedPassword` rather than defaulting: false would reopen the #1733 gap everywhere, true would refuse every self-unlink.
- **An unresolved caller.** Unchanged: it is still read as having no password door, so the refusal stays reachable.
- **The reclaim at start-up.** A record is removed only when the host's list does not contain the account **and** `GetUserById` for that id answers null, so a partial list alone removes nothing, and an empty answer reclaims nothing at all. Removing a record weakens protection for that account, so this is the direction that had to be conservative, and it is.

## The residual, stated rather than found

An account sealed by a plugin build from before this change has no record, so its minted password still reads as a way in. That is the residual #1733 and #1744 already name, and nothing after the fact can tell such an account from one whose password a person chose.

Issue #1759, opened 2026-09-16
Cannot reach Auth server


abesnier:
### Plugin version

5.0.0.88

### Jellyfin server version

12.1

### Provider type

OpenID Connect (OIDC)

### Identity provider

Authentik 2026.2.4

### What happened, and what did you expect?

Hi, 

Trying to replace 9p4's plugin with this one.
Provider and application already setup in Authentik, worked fine with 9p4.

When trying to use it, I get to an error page with the text "Error preparing login: the authorization server's discovery document could not be read."

Tried to turn on "Do Not Validate OpenID Endpoints (Insecure)" and "Disable OpenID HTTPS Discovery (Insecure)".

Jellyfin and Authentik run in Docker and are accessed with their external address/name (jelly.mydomain.tld and auth.mydomain.tld). Both run behing Nginx Proxy Manager, HTTPS is mandatory and certificates are valid.

Trying to access https://auth.mydomain.tld/application/o/jellyfin/.well-known/openid-configuration from within the Jellyfin container works fine.

When I saw that it did not work, I reset the plugin configuration completely and started fresh, as to remove all possible contamination from 9p4.

Any help or guidance is welcome.

Cheers
Antoine

### Relevant log output

```text
[15:19:40] [INF] [38] Jellyfin.Plugin.SSO_Auth.Api.Http.SSOController: SSO Controller initialized
[15:19:50] [WRN] [37] Jellyfin.Plugin.SSO_Auth.Api.Http.SSOController: Could not read the OpenID discovery document for provider authentik; the login fails closed rather than proceeding on unverified discovery facts.
System.Threading.Tasks.TaskCanceledException: The request was canceled due to the configured HttpClient.Timeout of 10 seconds elapsing.
 ---> System.TimeoutException: A task was canceled.
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Threading.Tasks.TaskCompletionSourceWithCancellation`1.WaitWithCancellationAsync(CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.DiagnosticsHandler.SendAsyncCore(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
[...]

Issue #1574, opened 2026-09-09
The wiki describes one settings page, and there are five


iderex:
Comes out of #1527, which changed what an administrator sees the moment they open the plugin, and the wiki still describes what was there before.

## The claim

Since #1527 the settings surface is five pages joined by the dashboard's tab strip - Overview, Providers, Accounts, Policies, Server - and the page the plugin's own name opens is **Overview**, which holds no setting at all. Every wiki page that says "on the settings page, scroll to ..." now sends an administrator to a page that does not have the thing, with no hint that four more tabs exist.

Read at the tree today:

```
$ git show origin/4.4:SSO-Auth/SSOPlugin.cs | grep -c 'Page(PageId'
5
```

versus the branch, which registers the five pages, their five controllers, the shared core and the three unchanged entries.

## What moved, so the edit does not have to be rediscovered

- **Overview** (the page the plugin list opens): a status card per provider, the SSO-only state, a what-to-do-next list. No setting.
- **Providers**: both provider workspaces and their editors, and the **Configuration check** button, which used to sit above them.
- **Accounts**: the linked-accounts panel and the account-link export/import pair.
- **Policies**: the provisioning-profile editor.
- **Server**: the configuration export/import pair, the login-page buttons switch, and single logout.

`docs/ui/mock/FIELDS.md` on the branch carries the per-control table if a page needs to name a specific field's new home.

## Done when

The wiki pages that describe reaching a setting say which tab it is on, and the page that introduces the settings surface says there are five and that the plugin opens on Overview. The pages this touches are at least Provider Setup, Hardening and Options Reference, and whichever page describes the Configuration check and the linked-accounts panel.

Wiki edits are gated, so this issue carries the text for review rather than being applied directly.

**Version marker.** #1527 is on the `4.4` line and is in no published build - `CHANGELOG.md` carries it under `Unreleased`, and the newest releases are `4.3.0-beta` and `5.0.0-JF12-beta`. Every build anybody is running still has the single page, so this edit lands with the release that carries #1527 rather than before it. Describing five tabs to a reader who has one would be wrong in the direction that costs most: they would go looking for a tab that is not there and conclude the setting is gone.

iderex:
Waiting for: Flowfin/jellyfin-plugin-sso#1527

## What this waits on

Flowfin/jellyfin-plugin-sso#1527

[...]

PR #1761, opened 2026-09-17
Bound each outbound connect attempt, and say what the address guard skipped

### LINKED ISSUE #1760: A discovery read the address guard broke reports only a timeout, and one silent address uses the whole budget
An OpenID discovery read that fails because of the outbound address guard reports only a timeout, and one unresponsive address can use the whole ten-second budget before the next address is tried.

Split: none

## What was reported

#1759, on `5.0.0.88` and Jellyfin 12.1, with Authentik behind Nginx Proxy Manager: the login refuses with "the authorization server's discovery document could not be read", while `curl` to the same discovery URL from inside the Jellyfin container succeeds. The plugin's own log:

```
Could not read the OpenID discovery document for provider authentik; the login fails closed rather than proceeding on unverified discovery facts.
System.Threading.Tasks.TaskCanceledException: The request was canceled due to the configured HttpClient.Timeout of 10 seconds elapsing.
...
   at Jellyfin.Plugin.SSO_Auth.Api.Net.ProviderResponseSizeLimit.SendAsync(...)
   at Jellyfin.Plugin.SSO_Auth.Api.Oidc.RepeatedMemberScreen.SendAsync(...)
   at Jellyfin.Plugin.SSO_Auth.Api.Oidc.OidcDiscoveryReader.ReadAsync(...)
```

The cause on that server is not confirmed yet; the reporter has been asked for `getent ahosts` from inside the container and for the result with **Allow Private Network Addresses (Insecure)** enabled.

## What the code does, at `fadecd85`

`SsoHttp.ConnectToAllowedAddressAsync` resolves the host, skips every address `IpAddressClassifier.IsBlockedAddress` refuses under the handler's policy, and connects to the rest in order:

```
foreach (var address in addresses)
{
    if (IpAddressClassifier.IsBlockedAddress(address, policy))
    {
        continue;
    }

    attempted = true;
    ...
    await socket.ConnectAsync(address, context.DnsEndPoint.Port, cancellationToken).ConfigureAwait(false);
```

Two things follow from that shape, and both fit the report:

1. **A skipped address leaves no trace unless nothing is left.** Only when every address is refused does the handler say so ("The outbound host resolves only to blocked addresses."). If one private address is skipped and another address is tried and hangs, the administrator reads a timeout, and nothing points at the setting that would have fixed it. Split-horizon DNS, which answers a LAN address for the provider's public name, is common on exactly the self-hosted setups this plugin serves.
2. **One attempt can use the whole budget.** The connect has no bound of its own; the only bound is `OidcDiscoveryReader.FetchTimeout` (10 s) on the client. An address that drops the connection silently, typically an IPv6 address the container cannot route or a public address that needs NAT loopback, cancels the request before the next address is tried. A connect callback replaces the handler's built-in connect, so the fallback between address families that a default client or `curl` gets is not there.

## Done when

- A discovery read that fails after the guard refused at least one address says so in the log, with the number refused and the name of the setting, without naming the addresses themselves where the log line is public.
- A single unresponsive address cannot consume the whole fetch budget: each attempt is bounded, or attempts are raced, so a working later address is reached.
- A test drives each: a resolver answering one refused and one unresponsive address shows the new log line, and a resolver answering an unresponsive address followed by a working one reads the document.

## Ordering

The `5.0` line is frozen for its first stable (#1729), and anything landing on it restarts the soak. Whether this is a stable-blocker or a change after the tag is a decision, recorded below.

iderex:
# What & why

Closes #1760.

**Not to be merged yet.** The `5.0` line is frozen for its first stable (#1729) and anything landing on it restarts the soak. Whether this lands before the tag or after it as 5.0.1 is decided on #1760; this pull request is ready either way.

#1759 reported a discovery read that failed with a bare ten-second timeout while `curl` from the same container reached the same URL. The outbound connect callback tries a host's allowed addresses one after another, and an attempt had no bound of its own: an address that dropped the connection silently held it until the caller's whole timeout, so a working address listed after it was never tried, and nothing in the log said the address guard had skipped an address the provider's own setting would have allowed.

- **Each attempt is bounded at five seconds** (`SsoHttp.ConnectAttemptTimeout`). A connect that works completes in well under one; the discovery fetch around it has ten, so one silent address and one working one fit. The bound applies to the last attempt too, so a host whose only allowed address is silent fails with the callback's own message rather than with the caller's cancellation. The caller's cancellation still ends the whole connect and is never read as a silent address.
- **A failed connect counts what the guard skipped**, and where those addresses are on a private network it names `AllowPrivateNetworkAddresses`. Loopback, link-local and cloud-metadata addresses are counted but never pointed at the setting, which does not relax them. No address and no host goes into the message, because it reaches the server log.
- **The policy is unchanged.** Every address is classified before anything connects to it, and the connect dials the validated address, not the name. The resolver, the socket connect and the bound are handed to an internal overload so tests can drive resolver answers this machine's DNS cannot produce; production passes the real three.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

- [x] Fail-closed preserved: a blocked address is never dialled (asserted by the dialled-address lists in every new test), a host of only blocked addresses still refuses, and an attempt that times out moves only to the next *allowed* address.
- [x] No secrets logged; the new message carries two integers and a setting name, no host, no address, nothing from the provider.
- [x] A security review was performed; this change is on the outbound SSRF / DNS-rebinding guard.
- [x] Review record: below.
- [x] Log inputs from the IdP are sanitized inline at the log call. No new log call; the message is built from counts.

### Review record

Three lenses, each read against the diff and the tests. **CONFIRMED 0 / PLAUSIBLE 2.** A finding is counted CONFIRMED only where a reproduction was produced.

**Bypass / fail-open: PASS, 0 findings.** The block test precedes the connect on every path; the extra classification under `PrivateNetworkPermitted` feeds only the message. A second resolution between check and connect is refused by a new runtime test with a resolver that rebinds on its second call, and by the source pin in `ArchitectureConformanceTests`, which a mutant with a second resolution fails. Redirect hops still pass through the same callback under the handler's captured tier.

**Availability: 1 PLAUSIBLE, DECLINED.** A provider whose TCP connect genuinely takes longer than five seconds (two or more lost SYNs on a very slow link) now fails an attempt that used to succeed inside ten. Declined: the overall budget is unchanged, such a link fails most other requests in the flow as well, and the case the bound fixes (a silent address ahead of a working one) is the one reported. Many silent addresses cannot amplify: the caller's timeout still ends the connect.

**Disclosure and guidance: 1 PLAUSIBLE, DECLINED.** Under DNS poisoning, a public provider name can resolve to a private address; the guard refuses it, and the new message then suggests the setting that would allow it. Declined: the hint is conditional ("if this provider runs on your own network"), names the setting whose label already says Insecure, scopes to one provider, and the alternative is the silence #1759 showed costs administrators the diagnosis. The login page and the Test Connection result are unchanged; this text reaches the server log only.

**No second reader looked at this change.** The review above is mine, lens by lens.

## Quality checklist

- [x] No duplicated logic; the guard is still one method with its production wiring in one expression-bodied overload.
- [x] Architecture-conformance tests pass. The rebinding pin in `OutboundConnectGuard_ResolvesTheHostOnce_AndConnectsToTheAddressItJudged` read one line literally and moves with the refactor; it keeps the production wiring pinned, and the property itself is now also driven at runtime.
- [x] Docs impact considered: no README change. The wiki troubleshooting passage for a discovery read that times out follows once this reaches a published build, since the wiki describes published behaviour (#1574).

## Verification

- [x] `dotnet build --warnaserror` is green, `net9.0` and `net10.0`, 0 warnings.
- [x] `dotnet test` is green: 4,053 tests on `net10.0` and 4,052 on `net9.0`, 0 failed, 4 skipped (the LAN-bind tests that need an opt-in on Windows).
- [x] Prettier is clean for the `CHANGELOG.md` change (checked on the staged, LF-normalised file).

## Notes

Six tests in `SsoHttpTests`: a silent address followed by a working one reaches the working one; a skipped LAN address beside a silent one fails with the count and the setting and without either address; a host of only blocked addresses keeps its sentence and names the setting only where it would help; the opt-in dials the private address and still refuses loopback; the caller's cancellation ends the connect without dialling on; a rebinding resolver is asked once and only the judged address is dialled. Three broken builds were run against them: the attempt bound removed (two tests fail), every cancellation read as the attempt's (one fails), and a second resolution before the connect (the rebinding test and the source pin fail).


iderex:
Replaced by #1766. The code is the same; the sentence that names `AllowPrivateNetworkAddresses` now carries the setting's reach. This transport also serves the avatar fetch and the SAML metadata importer, which the setting never relaxes, and #1762 showed a reader for whom the bare "enable it" here would have pointed at a switch that was already on (#1764).

Issue #1762, opened 2026-09-17
Content-Security-Policy - Error preparing login


j-pap:
### Plugin version

5.0.0.88

### Jellyfin server version

12.1

### Provider type

OpenID Connect (OIDC)

### Identity provider

Pocket ID v2.14

### What happened, and what did you expect?

I've setup Pocket ID in the plugin's settings, and the test connection is passing:

```
✅ The OpenID discovery document was read successfully.
    Issuer: https://auth.domain.tld
    Authorization endpoint: https://auth.domain.tld/authorize
    Token endpoint: https://auth.domain.tld/api/oidc/token
    UserInfo endpoint: https://auth.domain.tld/api/oidc/userinfo
    JWKS: reachable (1 key(s))
    PKCE (S256) advertised: yes
    RFC 9207 response-iss advertised: yes
```

When clicking the 'Pocket ID' button on the web interface to sign-in, I'm receiving the 'Error preparing login' text. Inspecting that page, I see the following error:

```
Content-Security-Policy: The page’s settings blocked the loading of a resource (img-src) at https://jellyfin.domain.tld/favicon.ico because it violates the following directive: “default-src 'none'” FaviconLoader.sys.mjs:236:20
Security Error: Content at https://jellyfin.domain.tld/sso/OID/start/pocketid may not load or link to file:///.
```

I don't currently have a CSP policy defined through my Nginx proxy, but I went ahead and tried to override that error with the following:

- "default-src 'self'"
- "img-src 'self'"
- Using [jellyfin's recommended default Nginx CSP policy](https://jellyfin.org/docs/general/post-install/networking/reverse-proxy/nginx/#https-config-example)

All three continue to produce the same favicon.ico error. It's worth noting that I haven't changed the favicon, so it's using Jellyfin's default icon.

I have the following security headers set. I've tried different combos with them off and on, but it doesn't seem to make a difference:

```
add_header Cross-Origin-Embedder-Policy "require-corp" always;
add_header Cross-Origin-Opener-Policy "same-origin-allow-popups" always;
add_header Cross-Origin-Resource-Policy "same-origin" always;
add_header Referrer-Policy "same-origin" always;
add_header X-Content-Type-Options "nosniff" always;
add_header X-Frame-Options "SAMEORIGIN" always;
add_header X-XSS-Protection "1; mode=block" always;
```

[...]

Issue #1610, opened 2026-09-10
A refused authorization request does not name the redirect URI it sent


iderex:
When a provider refuses the authorization request, the plugin logs the provider's code
and message and stops. It does not say which redirect URI it sent, although it is the
side that composed it.

From #1608, a real login that took a detour to answer:

```
[ERR] Duende.IdentityModel.OidcClient.AuthorizeClient: Failed to push authorization parameters
[WRN] SSOController: OpenID login refused for provider lea-id: preparing the authorization
      request failed (invalid_request - Failed to push authorization parameters).
```

The cause was one character of scheme: the plugin sent `http://host/sso/OID/redirect/lea-id`
behind a reverse proxy where the registered callback is `https://`. Nothing above says so,
and nothing above can be compared against anything.

Reproduced on a live Pocket ID 2.14.0, which is where the answer actually was:

```
ERR Failed to create pushed authorize request  error="invalid_request
  fosite.MatchRedirectURIWithClientRedirectURIs
  pocket-id/internal/oidc.matchRedirectURI (provider.go:131)
  fosite.(*Fosite).NewPushedAuthorizeRequest"
```

### Why the detour exists

With pushed authorization on, the refusal happens server to server and the admin never
reaches the provider, so the provider's own message - the one naming the URL - is never
shown. The only route to it today is to disable pushed authorization, retry, read the
provider's error page, then turn it back on. That is three steps and a protocol switch
to learn a string the plugin already holds.

### What to do

Name the redirect URI in the refusal, at the one place the challenge fails:

```
OpenID login refused for provider lea-id: preparing the authorization request failed
(invalid_request - Failed to push authorization parameters). The redirect URI sent was
http://host/sso/OID/redirect/lea-id - it must be registered on the provider exactly as
written, scheme and port included.
```

The provider page already displays this string for the admin to register. The failure path
is where it is worth repeating, because that is the moment somebody is looking for it.

Worth deciding at the same time: whether the same line belongs on the SAML challenge, which
composes an assertion consumer URL the same way and fails the same way.

PR #1772, opened 2026-09-19
The refused-challenge line follows the code the provider returned [#1763]

### LINKED ISSUE #1763: The refusal names the redirect URI even when the provider refused the client
When the authorization request cannot be prepared, the plugin writes one line into the server log, and that line ends by naming the redirect URI:

```text
OpenID login refused for provider pocketid: preparing the authorization request failed (Unauthorized - Failed to push authorization parameters). The redirect URI sent was https://jellyfin.domain.tld/sso/OID/redirect/pocketid, which the provider must have registered exactly as written - scheme, host, port and path.
```

The sentence was added in #1610 for the refusal it was measured on: `invalid_request` from a provider that does not hold the URI, where the push happens server to server, so nothing of that exchange reaches the provider's own page that would name it. It is the right sentence for that code.

In #1762 the code is `Unauthorized`. That is the provider refusing the client at the pushed-authorization endpoint, which is client authentication: a secret that does not match, or a client registered as public while the plugin authenticates as confidential. The redirect URI is not in question at all, and the line sends the reader to check it anyway. The reporter had already confirmed that the URI matches what the provider holds, and still got a sentence telling them to check it.

What I expect: the sentence follows the code. `invalid_request` and the other request-shaped refusals keep the redirect URI sentence. `Unauthorized`, or a description that names client authentication, gets a sentence about the client and its secret at the pushed-authorization endpoint. Anything else keeps a neutral closing sentence that names neither, so the line never asserts a cause the code does not carry.

The code sits in `SSO-Auth/Api/Flows/OidcLoginService.cs` beside the existing comment from #1610, and the browser-facing message stays the fixed generic one either way.

iderex:
> **Closed unmerged and replaced by #1774.** A third review round found the two log sentences each overstating the library contract by one word: the client sentence said this line never carries the provider's own words, which is false for `invalid_client` - the provider's own code, read out of a 400 body, printed on the same line - and the third sentence offered "the HTTP status" as something the field can hold, where the library hands on the reason phrase. A head here arrives as one commit and a pushed branch is never rewritten, so the correction comes under a new name and this one is closed rather than updated. Everything below is what stood here, unedited.

---

# What & why

The refused-challenge line in the OpenID login flow always ended by naming the redirect URI and telling the administrator the provider must hold it exactly as written. That sentence was added in #1610 for the refusal it was measured on, and was emitted for every refusal. In #1762 the answer was a 401 - the client being refused at the pushed-authorization endpoint - and the reporter, who had already confirmed the redirect URI against the provider, was told to check it again.

The closing sentence now follows what came back. `invalid_request` keeps the sentence from #1610, a client refusal names the client ID, the client secret and the public-or-confidential registration, and anything else gets a sentence that interprets nothing.

Closes #1763

## Type of change

- [x] Bug fix
- [ ] Feature

## What the review changed, before the checklists

The review ran twice and both rounds changed the change. Two readings of the pinned identity libraries are why this diff is smaller than the first draft, and they are quoted in the code because the next reader's first instinct is to widen the sets.

`AuthorizeState.ErrorDescription` never carries the provider's words. Both pinned versions overwrite it with one constant before the plugin sees it:

```text
ilspycmd -t Duende.IdentityModel.OidcClient.AuthorizeClient
  duende.identitymodel.oidcclient/7.1.0/lib/net10.0/Duende.IdentityModel.OidcClient.dll

    _logger.LogError("Failed to push authorization parameters");
    state.Error = ((ProtocolResponse)val).Error;
    state.ErrorDescription = "Failed to push authorization parameters";
```

The same three lines stand in 6.0.1. A whole tier of the first draft classified on that field and four test theories certified it; none of it could ever run, and all of it is gone.

What the remaining field holds depends on the status, and no status reaches this call:

```text
ilspycmd -t Duende.IdentityModel.Client.ProtocolResponse Duende.IdentityModel.dll

    if (ErrorType == ResponseErrorType.Http) { return HttpErrorReason; }
    if (ErrorType == ResponseErrorType.Exception) { return Exception.Message; }
    return TryGet("error");
    ...
    if (!httpResponse.IsSuccessStatusCode && httpResponse.StatusCode != HttpStatusCode.BadRequest)
```

So `invalid_request` is a provider's code, `Unauthorized` is the reason phrase of a 401, and a transport failure arrives in the same field as both. Hence: the client sentence does not rule the redirect URI out, because the provider's own message for a 401 is discarded before it arrives; the third sentence does not say the provider answered; and the redirect-URI sentence is reached by the one measured code rather than by everything request-shaped.

## Security checklist

- [x] Fail-closed preserved: the refusal is the same 400 with the same fixed generic browser message, and no branch changes control flow. Only the text of a server-log line moves.
- [x] No secrets logged; nothing new is printed. The line prints the same three provider-supplied values it printed before, and the redirect URI in one of three branches instead of in all of them.
- [x] A security review was performed: two rounds of five refute-by-default lenses, four reading and one executing, the second round entirely fresh instances.
- [x] Review record, below.
- [x] Log inputs from the IdP are sanitized inline at the log call: every argument in all three branches carries the line-ending strip and the bracket substitution at the call, and `EveryForeignValueAnyPluginLogLinePrints_CarriesBothInlineSanitizers` passes.

### Review record

**Round 1**, over the first draft. Five lenses, every one `NEEDS_IMPROVEMENT`. **CONFIRMED 22 / PLAUSIBLE 15**, counted over raised findings rather than over distinct defects, which were fewer: the dead description tier was found independently by four lenses and the reason-phrase reading by three.

| Defect | Disposition |
| --- | --- |
| The description tier is unreachable; the library replaces that field with a constant | FIX - tier removed, `Classify` takes the code alone |
| `state.Error` is the HTTP reason phrase for any status but 400 | FIX - documented at the type; three corrections follow from it |
| The client sentence positively denies the redirect URI | FIX - the denial is gone |
| The third sentence asserts the provider answered, which is false for a transport failure | FIX - it asserts nothing |
| Four test theories certify an input no caller can produce | FIX - removed; every row is a value the library can produce |
| Comments and changelog attribute the empty description to the provider | FIX - both say the library discards it |
| Nothing binds a cause to a sentence; exchanging two case bodies left every test green | FIX - the rule reads the arms keyed on their case labels |
| `invalid_redirect_uri` is a dynamic-registration code this endpoint does not return | FIX - removed from the set |
| The client row keys on a server-chosen reason phrase, so a 401 spelled otherwise loses its sentence | DECLINE - no status reaches this call and the miss asserts nothing false; stated as a bound at the type and pinned by a test |
| One message template became three with no shared `EventId` | DECLINE - this tree uses no `EventId` anywhere, so a shared one would be the first |

**Round 2**, over the result, with five fresh lenses and none of them the instance that raised a finding. Verdicts: two `FAIL`, three `NEEDS_IMPROVEMENT`. **CONFIRMED 11 / PLAUSIBLE 23** over 34 raised findings.

| Defect | Disposition |
| --- | --- |
| The changelog hunk was a replacement, not an insertion: it deleted the #1760 entry whole | FIX - file restored from the base, the new entry inserted beside the old one. Found by grepping the branch for the issue numbers the base carries and getting nothing |
| `invalid_request_object` and `invalid_request_uri` were mapped to the redirect-URI sentence on a specification citation, though neither is about the callback | FIX - the set is the one measured code; the neighbours now reach the third sentence, with a test per code |
| The client sentence named only the secret, and advised a registration change whose blast radius is every login | FIX - it names the client ID too, and the public-or-confidential mismatch in both directions rather than a change to make |
| The third sentence said the answer named nothing, while the same line prints a code that sometimes names a great deal | FIX - it says this line does not interpret the answer, and what the three things it can be are |
| The arm parser read the first matching call after a case label, so a commented-out copy of the old call was read as the arm | FIX - comments are not source, proven by mutation |
| A second logging call in one arm was invisible to the rule | FIX - refused rather than reduced to the first |
| Nothing bound the classification to the field that varies; feeding it the constant beside it disabled the feature with every test green | FIX - a third rule reads that argument, proven by mutation |
| A missing arm failed as `KeyNotFoundException` rather than as the rule's own diagnostic | FIX - the arm lookup carries the diagnostic |
| The two static fields broke the tree's naming convention | FIX - PascalCase, like every other `private static readonly` in the project |
| `{ErrorDescription}` is still logged though it is now known to be a constant | DECLINE - it is part of the line's shape from #1610 and removing it is a separate change to a line this one is only choosing the ending of |
| `unauthorized_client` is not in the client set | DECLINE - it is about a grant this client may not use, not about proving which client it is; a test pins it to the third sentence |

**Round 3**, over the second round of fixes, is running as this is opened; its result will be recorded in this body before any merge.

These sections were produced by automated re-runs, not by a second person. Nothing in this branch has been read by anybody but its author, so the evidence stands in place of the second reading: every claim above carries the command that produced it, and every conformance rule was run against a deliberately mutated tree and observed to refuse it.

## Quality checklist

- [x] No duplicated logic; the classification is one small pure function with one caller.
- [x] The change adds no more code than the problem requires. The first draft was larger, and the review deleted the part that could never run.
- [x] Comments explain why: the two decompiles are quoted at the type, because the next reader's first instinct is to widen the sets.
- [x] Architecture-conformance tests pass, and the structural properties this change establishes - each arm says what its own cause is, only one arm asserts the redirect URI, and the classification reads the field that varies - are locked in as rules.
- [x] Docs impact: none. No README section or wiki page quotes this line; a `grep` for `redirect URI sent was` over a wiki checkout returns nothing.

## Verification

- [x] `dotnet build SSO-Auth.sln --no-restore --warnaserror`, and the same with `-c Release`: 0 warnings, 0 errors.
- [x] `dotnet test -c Release -f net9.0`: 4075 total, 0 failed, 4071 passed, 4 skipped. `-f net10.0`: 4076 total, 0 failed, 4072 passed, 4 skipped. The four skips are the loopback-bind tests that need a firewall consent dialog; they are skipped on this machine rather than worked around.

The guards were proven rather than asserted, each against a deliberately mutated tree. Exchanging the two case bodies refuses:

```text
fehlerhaft ArchitectureConformanceTests.TheChallengeRefusalNamesTheRedirectUriItSent
fehlerhaft ArchitectureConformanceTests.EachChallengeRefusalArmSaysWhatItsOwnCauseIs
  fehlgeschlagen: 2
```

Giving the third arm the redirect-URI sentence refuses:

```text
fehlerhaft ArchitectureConformanceTests.EachChallengeRefusalArmSaysWhatItsOwnCauseIs
  fehlgeschlagen: 1
```

Classifying the constant field beside the code refuses:

```text
fehlerhaft ArchitectureConformanceTests.TheChallengeRefusalClassifiesTheProvidersAnswerAndNotItsNeighbour
  fehlgeschlagen: 1
```

Hiding the real call behind a commented-out copy of it refuses:

```text
fehlerhaft ArchitectureConformanceTests.TheChallengeRefusalNamesTheRedirectUriItSent
  fehlgeschlagen: 1
```

## Notes

The browser-facing message is untouched and stays the fixed generic one.

What this does not do: it does not read the provider's own description, because the library discards it before the plugin sees it, and it does not read the HTTP status, because the library does not hand one on. Reaching either would mean building the pushed-authorization request outside the library, which is a larger change than the sentence it would improve.

Half of the issue's third acceptance criterion is deliberately not met, and this is the record of it. The criterion asks that "a description naming client authentication" also reach the client sentence. No description naming anything reaches this code, for the reason quoted at the top, so that clause is unimplementable as written rather than skipped.


iderex:
Security reading before the merge, because the line sits on the login path. The change reads one field, `AuthorizeState.Error`, and matches it case-insensitively against two closed sets; nothing else from the provider's answer reaches the line, and the browser still receives the same fixed 400 with the same generic text, so the refusal itself is unchanged and stays fail-closed. All three log arms keep the inline sanitization of every provider-supplied value (line endings removed, `[` replaced), and two of them write less than before: the redirect URI, composed from the request's host header, is now written only for the one code it was measured against. No secret, token or subject is written on any arm. The conformance test pins that the classifier is called on `state.Error` alone, the unit tests hold the neighbouring codes out of the redirect-URI sentence, and every required check is green on the head. Landing on `5.1` with a merge commit.

Issue #1773, opened 2026-09-19
The pairwise co-existence job dies at the sibling listing since the sibling repositories went private, and it gates every beta publish


iderex:
Since 2026-09-19 the `Pairwise co-existence` job of the `E2E Login Harness` workflow fails before it pairs anything, on every trigger, and because `publish-jf12-beta.yml` calls that workflow as its provider-matrix job, the next nightly beta publish stops there before it builds.

## What I measured

Three runs on 2026-09-19, the same failure in each:

```
gh run view 35461828201 --log-failed    # pull request head of #1771
gh run view 35462336472 --log-failed    # probe/e2e-pairwise-control-1770 = 34a35e09, the 5.1 tree before #1771
== Deriving the sibling set from Flowfin ==
FATAL: could not list Flowfin's repositories
```

The scheduled run of the same morning, 35430999655 at 08:05Z on the 5.0 ref, listed eleven sibling repositories and paired against them, and every scheduled run since 2026-09-13 in the run list did the same. Nothing in the harness changed between 08:05Z and the first failure: the control run above is the tree exactly as it stood before #1771, and it fails the same way, so #1771 is not the cause.

What changed is the visibility of the sibling repositories, read with my own account on 2026-09-19:

```
gh api "orgs/Flowfin/repos?per_page=100" --paginate --jq '.[] | select(.archived == false) | .visibility' | sort | uniq -c
     11 private
      1 public
```

The workflow token is scoped to this repository (`permissions: contents: read` on the job), so `orgs/Flowfin/repos` answers it with the one public repository, which is this one. In `test/e2e/pairwise/run-pairwise.sh` the `grep -vx "$SELF"` behind that listing then leaves an empty pipeline, and under `set -o pipefail` that is the FATAL above. The script does what it was written to do: a sibling set that came back empty is refused rather than passed, because a silently empty run would look like a clean one.

The seven login stacks are unaffected: run 35461828152 on the same day ran all seven providers against `jellyfin/jellyfin:12.0` and every one passed. What is red is the co-existence half alone, and what it blocks is the publish that calls it.

## Done when

- The `Pairwise co-existence` job concludes on a scheduled run without dying at the listing, in whichever of the three shapes below is chosen, and the `Publish JF12 Beta` run after it reaches its package job.
- The e2e README's sentence that the sibling listing is a public read needing no more than the workflow token states what is true after the choice.

Split: none

iderex:
[...]

PR #1776, opened 2026-09-19
The pairwise phase reports an empty sibling set instead of dying at the listing [#1773]

iderex:
# What & why

Since 2026-09-19 the sibling plugin repositories are private, and the pairwise co-existence job of the e2e harness died at its first step on every trigger: the org listing under the workflow token answers with this repository alone, the filter behind it ran inside the same pipeline as the call, and under `set -o pipefail` an answered listing with nothing left after the filter failed with the same `FATAL: could not list` as a call that never answered. `publish-jf12-beta.yml` calls the harness as its matrix job, so the nightly beta publish stopped there before it built anything. Contributes to #1773, in the shape decided on the issue on 2026-09-19: the phase keeps running and reports, and the token is the switch rather than a precondition.

What changes:

- `test/e2e/pairwise/run-pairwise.sh` separates the call from its answer. A listing that fails as a call stays FATAL. A listing that answers with no sibling ends green and writes into the job log, a `::notice` annotation and the step summary that no pair was checked, how many repositories the token could see, and why, so a green run cannot be read as co-existence evidence it does not carry. The sibling artefact and its sha256 sidecar are downloaded through the release-asset API under the same token that listed the sibling, because a private repository's `browser_download_url` answers 404 without one; a token able to list the private siblings would otherwise have found every one of them undownloadable.
- `e2e-login.yml` declares the optional `workflow_call` secret `PAIRWISE_READ_TOKEN`, and the pairwise step uses it where present and the workflow token otherwise. `publish-jf12-beta.yml` passes that one secret to the called workflow by name rather than with `inherit`.
- `test/e2e/README.md` says the listing needs a token that can read the family, what the phase does without one, and that a failed call stays fatal. The workflow comment that called the listing a public read is rewritten; the README never carried that sentence, so the paragraph is new there.

## Type of change

- [x] Bug fix
- [ ] Feature

## Security checklist

The change touches the release pipeline, the harness that gates the beta publish, so this is filled in.

- [x] Fail-closed preserved: a listing that fails as a call is still FATAL, a downloaded artefact that does not match its published sha256 still fails the run, and a pair that collides still fails it. What no longer fails is an answered listing with no sibling, which is a fact about visibility and not about any pair, and the run says on three surfaces that it checked nothing.
- [x] No secrets logged: the token reaches the script only as `GH_TOKEN` and is never printed. What the log prints is the names of the repositories the token could see, which are the ones visible to it.
- [ ] A security review was performed: none was run for this change. There is no second reader on this board tonight, and this body carries the evidence in its place.
- [ ] Review record: none, for the reason above.
- [x] Log inputs from the IdP are sanitized inline at the log call: no IdP input reaches a log here. The only outside text logged is repository names from the GitHub API.

## Quality checklist

- [x] No duplicated logic; the two downloads share the one API form.
- [x] The comments explain why: the private siblings, the pipeline that conflated the two failures, the 404 on the download URL.
- [ ] Architecture-conformance tests: not applicable, no C# changed.
- [x] Docs impact considered: the e2e README is in this change, and the wiki's Releasing page carries the sentence that co-existence evidence exists only while the secret is set, wiki commit 88e36cf.

## Verification

- [ ] `dotnet build --no-restore --warnaserror`: not run, no C# changed.
- [ ] `dotnet test`: not run, no C# changed.
- [x] Prettier is clean for `test/e2e/README.md` (`npx prettier --check test/e2e/README.md`).
- [x] `bash -n` on the script and `actionlint` on both workflows pass.

Local proof of both paths, with a `gh` shim on PATH that answers the org listing with this repository alone, against the packaged 5.0.0.88 zip:

```
new script, listing answers with self only:  exit=0, "NO PAIR WAS CHECKED", step summary written
new script, listing fails as a call:         exit=1, FATAL: could not list Flowfin's repositories
old script, listing answers with self only:  exit=1, FATAL: could not list Flowfin's repositories
old script, listing fails as a call:         exit=1, FATAL: could not list Flowfin's repositories
```

On the runner, dispatched on this branch before the pull request existed:

```
gh run view 35463607393 --json status,conclusion
completed success
Pairwise co-existence, job 105951746580: success
No sibling is visible: the listing answered with 4 repositories visible to the token in use and no sibling plugin repository among them. The sibling repositories are private since 2026-09-19 (#1773), and a token that can read them is the switch that makes this phase pair again.
  repositories visible to the token in use: 4
  siblings considered: 0
```

Under my own token the same script lists 11 siblings, downloads every release through the API route, and skips all 11: their current releases target the 10.11 ABI, or there is no non-prerelease release to install. So even once the secret exists, this run pairs nothing until a sibling publishes a 12.0 build, and it says so per sibling as it did before.

## Notes

The scheduled harness run and the nightly beta publish run on the default branch, `5.0`, which is frozen. This change rides to `5.0` in the promotion pull request, as the decision on the issue says, and the issue's first Done-when line is read there; the wiki sentence is landed already. Whether `PAIRWISE_READ_TOKEN` is created is a separate decision made by hand and does not wait on this change. No second reader looked at this change.


iderex:
Security reading before the merge, because this touches the harness that gates the beta publish. The token reaches the job through `env`, chosen by an expression outside any `run:` script, and it is never printed: the log carries repository names and counts, nothing else. The caller hands the called workflow one named secret rather than `inherit`, and where the secret does not exist the value is empty and the workflow token takes over, which is the fallback the decision asked for. What still fails is unchanged: a listing that fails as a call, an artefact whose published sha256 does not match, a pair that collides. What no longer fails is an answered listing with no sibling, and the run says so in the log, in an annotation and in the step summary, so its green cannot be read as a checked pair. The downloads move from the anonymous browser URL to the release-asset API under the same credential, which is the one route that works for a private sibling. One correction to the body: the default branch is `5.1` since today, so the scheduled harness run takes this change from its next schedule on; the nightly beta dispatch still names `5.0` by its `--ref` and meets the fix with the promotion pull request, as decided on #1773. Landing on `5.1` with a merge commit once every check is green.

PR #1777, opened 2026-09-19
An administrator refusal names the two ways to link that account [#1765]

### LINKED ISSUE #1765: The refusal to adopt an administrator account points at API endpoints instead of the page that links it
From #1762:

```text
[WRN] SSOController: SSO login for admin via oid/pocketid refused adoption of a pre-existing account: the target account is an administrator; link it explicitly via the admin endpoints.
```

The rule is right: an administrator account is never adopted by name, so a first SSO login cannot turn into administrator access. The sentence after it sends the reader to "the admin endpoints", which is the `AddCanonicalLink` API, and the reporter searched the documentation for how to call it before finding the page that does it: `/SSOViews/linking`, signed in to that account with its own password.

What I expect: the line names the page, in words someone reading a server log can act on, for example "sign in to that account with its own password and link it at /SSOViews/linking". The text sits in `CanonicalLinkService.cs`, on the `RefusePrivileged` verdict.

The Troubleshooting entry for this refusal named the page only for the case where `AllowExistingAccountLink` is the other way out. For an administrator the page is the only way, and the entry now says so.

iderex:
# What & why

An administrator account is never adopted by name, so a first SSO login cannot turn into administrator access. That rule is right and nothing here changes it. What the server log said after it was "link it explicitly via the admin endpoints", which names a category rather than a route, and the reporter on #1762 searched the documentation before finding the page that does it.

Both refusals that state the rule now name both ways in: sign in to that account with its own password and link it at `/SSOViews/linking`, or pre-provision the link with an elevated call to the account-management API.

Closes #1765

## Type of change

- [x] Bug fix
- [ ] Feature

## What the review changed

The first draft said the page was the only way through. It is not. The pre-provision route is elevation-gated and writes a link against any Jellyfin account, administrators included, with no privilege check of its own:

```csharp
[Authorize(Policy = Policies.RequiresElevation)]
[HttpPost("Links/Preprovision/{mode}/{provider}/{jellyfinUserId}")]
```

Removing the only pointer to it and then asserting in the release notes that it did not exist would have been worse than the vague phrase it replaced. The case that makes it concrete is a server running SSO-only login: the refused account may have no password door left, so the page is not reachable for it and the elevated route is the only way through. Both sites name both routes now.

The first draft also said `AllowExistingAccountLink` does not reach this refusal. It is the precondition for reaching it - `AccountLinkResolver.Resolve` returns `AdoptExistingAccount` only when the flag is on, so this arm is emitted only on providers that have it set. A reader who took the sentence literally would have concluded the refusal fires regardless of the flag. The true statement is the one the changelog carries: turning the flag on does not get an administrator through.

The third finding was that the same vague phrase survived on the sibling refusal - the legacy username-keyed link that points at an administrator - which states the same rule on the same login path. Fixing one and not the other would have left the reported defect one line away, so both are fixed and a rule holds both.

## Security checklist

- [x] Fail-closed preserved: the verdict, the throw, the audit line and which accounts may be adopted are all untouched. No account becomes linkable that was not before.
- [x] No secrets logged. The phrases are fixed literals carrying no account data, and the surrounding line keeps its sanitizers.
- [x] A security review was performed: two refute-by-default lenses over the diff, one reading and one executing. Both returned `NEEDS_IMPROVEMENT` on the first draft.
- [x] Review record: 2 lenses, 9 findings raised, **CONFIRMED 3 / PLAUSIBLE 6**. Dispositions below.
- [x] Log inputs from the IdP are sanitized inline at the log call; the argument lists are unchanged.

### Review record

| Defect | Disposition |
| --- | --- |
| The claim that `/SSOViews/linking` is the only way through is false; an elevation-gated pre-provision route links an administrator account, and the change removed the only pointer to it | FIX - both refusals name both routes |
| On an SSO-only server the password remedy is impossible for the refused account, and the route the message dropped was the answer there | FIX - same change; the changelog says why the second route matters |
| The causal clause about `AllowExistingAccountLink` is backwards; it is the precondition for reaching the arm, not something that cannot reach it | FIX - changelog and test comments corrected |
| The same vague phrase survives on the sibling administrator refusal, so the tree gives two remedies for one rule | FIX - both sites carry the same sentence, and a rule holds every administrator refusal in the file to it |
| The tests restate the literal and none binds the phrase to the log line an operator reads | FIX - a rule reads the refusal line and requires its reason placeholder to be fed by the method the other tests pin |
| The commit cited a "Troubleshooting entry" not present in this repository | FIX - it is a wiki page, and the commit message no longer leans on it |

## Quality checklist

- [x] No duplicated logic; the remedy sentence is written once per refusal because the two lines have different subjects, and a rule keeps them saying the same thing.
- [x] The change adds no more code than the problem requires: two message literals, one visibility change with its documentation, and the tests.
- [x] Comments explain why, including why the arms are a contract on a message rather than on behaviour.
- [x] Architecture-conformance tests pass, and the structural property this establishes - every administrator refusal on this path names both ways in and none names a category of endpoint - is locked in as a rule.
- [x] Docs impact: none in this repository. The wiki's Troubleshooting page already names the page for this refusal, and `docs/ACCOUNT-MANAGEMENT-API.md` already documents the pre-provision route as the deliberate way to link such an account.

## Verification

- [x] `dotnet build SSO-Auth.sln --no-restore --warnaserror`, and the same with `-c Release`: 0 warnings, 0 errors.
- [x] `dotnet test -c Release`: 4083 total, 0 failed, 4079 passed, 4 skipped. The four skips are the loopback-bind tests that need a firewall consent dialog; they are skipped on this machine rather than worked around.

The guards were proven rather than asserted. Restoring the old phrase on the adoption arm refuses:

```text
fehlerhaft AdoptionRefusalReasonTests.AdministratorTarget_NamesBothWaysIn
fehlerhaft AdoptionRefusalReasonTests.AdministratorTarget_NoLongerSendsTheReaderToACategoryOfEndpoint
  fehlgeschlagen: 2
```

Restoring it on the sibling arm - the site nothing held before - refuses:

```text
fehlerhaft AdoptionRefusalReasonTests.EveryAdministratorRefusalOnThisPathGivesTheSameTwoWaysIn
  fehlgeschlagen: 1
```

## Notes

Nothing in this branch has been read by anybody but its author. The evidence stands in place of a second reading: each claim above carries the command or the quoted source that produced it, and every rule was run against a deliberately mutated tree and observed to refuse it.

Out of scope and left alone: the other refusals in this file that point at "the admin endpoints" - a mismatched stored issuer, an orphaned legacy link, a pending legacy link with `AllowExistingAccountLink` off. They state different rules with different remedies, and the rule added here is written over the administrator refusals rather than over the phrase, so it does not reach them.


iderex:
Security reading before the merge, because the line sits on the login path. Nothing about the gate moves: the verdict, the throw, the audit line and the set of accounts that may be adopted are byte-identical, and the two refusals keep their inline sanitizers with unchanged argument lists. What changes is two fixed literals that carry no account data, naming the two deliberate ways to link an administrator account, one of which is elevation-gated and documented already; naming a route that exists gives an attacker nothing, since the route refuses anyone without elevation. The mapping method turns internal so a test can read the phrases, which is a visibility change inside the assembly and not a new surface. Landing on `5.1` with a merge commit once every check is green.

PR #1778, opened 2026-09-19
Measure the Strict preset against the repeated-member walk, and keep the walk [#1043]

### LINKED ISSUE #1043: Replace the hand-rolled duplicate-name walk with the System.Text.Json Strict preset once net9.0 is dropped

iderex:
Closes #1043.

The gate #1043 waited on is open: `net9.0` is gone from this line, so one target
remains and the weighing the issue carries can be done by measurement.

```
$ git grep -n 'TargetFramework' origin/5.1 -- '*.csproj'
origin/5.1:SSO-Auth.Bench/SSO-Auth.Bench.csproj:20:    <TargetFramework>net10.0</TargetFramework>
origin/5.1:SSO-Auth.Fuzz/SSO-Auth.Fuzz.csproj:15:    <TargetFramework>net10.0</TargetFramework>
origin/5.1:SSO-Auth.Tests.Stryker/SSO-Auth.Tests.Stryker.csproj:19:    <TargetFramework>net10.0</TargetFramework>
origin/5.1:SSO-Auth.Tests/SSO-Auth.Tests.csproj:4:    <TargetFramework>net10.0</TargetFramework>
origin/5.1:SSO-Auth/SSO-Auth.csproj:8:    <TargetFramework>net10.0</TargetFramework>
```

I ran `JsonSerializerOptions.Strict` against the corpus `StrictJson` is pinned
on, at every depth the walk covers, inside arrays of objects, and for the
diagnostic each side gives. The answer is that the walk stays.

## Where the two agree

The preset refuses every repeat the walk refuses: at the root, in a nested
object, inside an object in an array, straddling a nested scope, under an
escape-spelled name and under the empty name. Its message names the member,
which is the datum `RepeatedMemberScreen` puts in its log entry. The depth cap
is the same 64 on both. A case-variant pair is admitted by both, which
`TheStrictPresetTakesTheSameDecisionOnCase` already measured and which this
change does not touch.

That half is the reason the question was worth asking, and it is not the half
that decides it.

## Where the preset falls short

**No BOM strip.** The walk strips one leading BOM because a provider serving a
BOM-prefixed file emits one and `Utf8JsonReader` treats it as content. The
preset refuses the document for its first byte, and `Unreadable` is a refusal at
every seam that consumes this walk, so a swap locks such a provider out of a
login that works today. The same document with a real repeat in it is refused
for the BOM, so the message names the byte and never the member.

**No narrowing to the scopes a caller's reader enters.** `OidcRoleExtractor` and
`OidcAuthorizeStateBuilder` name the scopes their reader descends through, so a
repeat in a sibling the reader never opens changes nothing they read and is
admitted. The preset refuses a repeat anywhere in the document. Under it, a
vendor extension repeating a name beside the role claim takes the role set of
every login offline, which is the availability regression the review of #1032
removed.

**A document carrying no object is admitted.** A bare scalar and an array of
scalars deserialize without complaint, so a caller reading "no exception" as
approval gets an affirmative answer about a document nothing screened. The walk
reports that it established nothing. This is the one that fails open.

**One channel for every outcome, in a type that depends on the target.**
`JsonException` for `JsonElement`, `ArgumentException` for `JsonNode`, and
`ArgumentException` again for a raw unpaired surrogate that cannot be transcoded
to UTF-8. `RepeatedMemberScreen` reports `RepeatedMember` and `Uninspectable` as
separate refusals to the operator and the admin probe, and `OidcRoleExtractor`
separates them in its audit trail, so a swap recovers both the verdict and the
logged member name out of English message text.

## What landed

The rows are in `StrictJsonTests`, beside the corpus they run against rather
than in a copy of it, and each asserts both sides so the measurement stays
re-runnable instead of becoming a sentence about a past run. The prose in
`StrictJson`, in `StrictJsonTests` and in the conformance rule said #1043 would
decide this; it now says what was decided and why. The conformance rule refusing
the walk to delegate its duplicate decision does not retire with the walk.

Outside the tests, this diff is documentation. Nothing the plugin does changes.

## How far the new rows bite

Measured rather than claimed. I replaced the body of `StrictJson.Inspect` with a
wrapper over the preset, built it clean, and ran the class:

```
Jellyfin.Plugin.SSO_Auth.Tests.StrictJsonTests.ThePresetAdmitsADocumentCarryingNoObject_WhereTheWalkEstablishesNothing [FAIL]
Jellyfin.Plugin.SSO_Auth.Tests.StrictJsonTests.ThePresetRefusesABomPrefixedDocumentTheWalkAdmits [FAIL]
Jellyfin.Plugin.SSO_Auth.Tests.StrictJsonTests.TheSameDocumentsWithoutTheDuplicate_AreClean(json: "﻿{\"issuer\":\"https://one.example\"}") [FAIL]
Jellyfin.Plugin.SSO_Auth.Tests.StrictJsonTests.ADocumentCarryingNoObject_IsUnreadable [FAIL]
Jellyfin.Plugin.SSO_Auth.Tests.StrictJsonTests.DuplicateMember_AtAnyScope_IsRepeated(json: "﻿{\"issuer\":1,\"issuer\":2}", expectedMember: "issuer") [FAIL]
Jellyfin.Plugin.SSO_Auth.Tests.StrictJsonTests.ThePresetCannotNarrowToTheScopesACallersReaderEnters [FAIL]
SSO-Auth.Tests  Total: 42, Errors: 0, Failed: 6, Skipped: 0, Not Run: 0
```

The signalling row is not among them, and that is stated in the row itself
rather than left to be discovered: the wrapper catches both throw types and cuts
the member name out of the message, which satisfies every assertion in it. What
that row records is the wrapper, not a behaviour the swap cannot reach. The walk
was restored from a copy taken before the mutation and the tree matches
`origin/5.1` outside the three files in this diff.

## Local gate

```
dotnet build SSO-Auth.sln --no-restore --warnaserror           0 warnings, 0 errors
                                                               0 warnings, 0 errors
git status --porcelain -- '*packages.lock.json'                (empty)
dotnet exec SSO-Auth.Tests/bin/Release/net10.0/SSO-Auth.Tests.dll
  SSO-Auth.Tests  Total: 4088, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0
bash scripts/check-jellyfin-compat.sh                          consistent, exit 0
```

The four skips are the loopback-binding tests that need an administrator on
Windows; they are skipped by the suite's own guard, not worked around here.
Prettier was not run because no `.js`, `.html`, `.md`, `.css` or `.scss` file is
in this diff.

No second reader looked at this change. The measurement above is what stands in
place of one, and it is re-runnable rather than a report of a run.


iderex:
Read before the merge. The plugin's behaviour does not move: the walk stays, the diff outside the tests is prose in the type and in the conformance rule, and the new rows measure the preset against the corpus the walk is pinned on and assert both sides, so the decision of #1043 is re-runnable rather than remembered. The four shortfalls the measurement found (the BOM strip, the narrowing to the scopes a reader enters, the document carrying no object that the preset admits, and one exception channel for every outcome) are each a row, and the one that fails open is named as such. That is the outcome the decision on #1043 asked for: the measurement decides, and here it decides for the walk. Landing on `5.1` with a merge commit once every check is green.

Issue #1152, opened 2026-08-03
wiki-lint.py: resolve source links against the tracked file set, not the filesystem


iderex:
Part of #1014.

## Claim

`wiki-lint.py` reaches the same verdict locally as in CI, because it resolves
source links against the **tracked** file set rather than the filesystem.

## Where it lives

`scripts/wiki-lint.py`, `check_dead_source_links` (the `_BLOB` loop):

```python
for rel in _BLOB.findall(line):
    if not (code / rel).exists():
        findings.append(f"{path.name}:{n}: dead source link '{rel}' (not in the code tree)")
```

`(code / rel).exists()` is a filesystem test. In CI it happens to be right, because
`actions/checkout` materialises tracked files only. On a developer's working
checkout it is wrong: `docs/ARCHITECTURE.md` is still on disk after #987
un-tracked it (`.gitignore:469` ignores `docs/*`), so a local run reports **clean**
while CI reports two findings on the same wiki. The divergence is silent, nothing
in the output or the docstring tells the reader which tree the answer applies to.

The other two checks (`check_anchors`, `check_file_line_citations`) read only the
wiki directory and are unaffected.

## What to do

Either of these satisfies the claim; pick one and say why in the PR.

1. **Resolve against the index.** Build the tracked set once - 
   `git -C <code> ls-files -z` → a `set[str]` of repo-relative POSIX paths, and
   test membership instead of `.exists()`. Handle the case where `code` is not a
   git work tree (no `.git`, `git` missing, non-zero exit) by falling back to the
   filesystem check *and* printing a one-line warning that the verdict is
   filesystem-based and may differ from CI. This is the option that makes local and
   CI agree.
2. **Declare the requirement.** Keep `.exists()` and make the script state plainly,
   in `--help`/the docstring and in the run output, that it must be pointed at a
   tracked-only tree (`git archive origin/main | tar -x -C <dir>`), because an
   ignored-but-present file will otherwise be scored as live.

Option 1 is preferred: it removes the trap rather than documenting it.

## Done looks like

- With `docs/ARCHITECTURE.md` present on disk but un-tracked, a local
  `python3 scripts/wiki-lint.py <wiki> .` reports the same two dead-source-link
  findings CI reports (option 1), or refuses/warns loudly enough that the reader
  cannot mistake the verdict (option 2).
- Against a tracked-only tree (`git archive origin/main | tar -x`) the output is
[...]

PR #1781, opened 2026-09-19
The wiki lint resolves a source link at the ref it names [#1780]

### LINKED ISSUE #1780: The wiki lint checks a blob link's path against the default branch whatever ref the link names, so the 4.3 line's workflows on main read as dead links on 5.1
## What happened

Wiki Lint went red on `5.1` with run 35463758642 (trigger `gollum`, 2026-09-19T19:15Z), and the release alert #1779 carries it:

```
Releasing.md:62: dead source link '.github/workflows/publish-beta.yml' (not in the code tree)
Releasing.md:80: dead source link '.github/workflows/publish.yml' (not in the code tree)
Rollback.md:31: dead source link '.github/workflows/publish.yml' (not in the code tree)
```

The three links are of the form `blob/main/.github/workflows/publish-beta.yml`. Those files exist on `main`, the 4.3 line, which is where the links point; they left `5.1` with #1771. The rule in `scripts/wiki-lint.py` strips the ref and looks the path up in the checked-out tree:

```
_BLOB = re.compile(r"github\.com/[^/]+/[^/]+/blob/[^/]+/([^)#\s]+)")
def check_dead_source_links(wiki: Path, code: Path) -> list[str]:
```

On a `gollum` run the checkout is the default branch, so a correct link to another line's file reads as dead the moment that file leaves the default branch. The wiki is right and the instrument is wrong; editing the wiki to dodge the rule would trade a true citation for a false one.

## What is proposed

The check honours the ref the link names. A path is looked up at that ref (`git cat-file -e origin/<ref>:<path>` after fetching the ref, or the contents API where the checkout is shallow), and the checkout is consulted only for links whose ref is the default branch or `HEAD`. A link naming a ref that does not exist stays a finding, with the ref in the message. The Coding-Standards citation rule gains the sentence that a source link names the line it cites, which these three already do.

## Done when

- The lint against the current wiki and `5.1` is green with the three links unchanged, and a fixture link to a path absent at its named ref still fails; both shown in the pull request (a test of the function, or a dry run with its command and output).
- The Wiki Lint section of #1779 clears on the next run, whether a dispatch or the next wiki push.

Split: none

iderex:
Closes #1780.

Wiki Lint went red on `5.1` with run 35463758642 after #1771 moved the 4.3 line's publish workflows off this branch. The three links it reported name `main`, where those files are:

```
$ git cat-file -e origin/main:.github/workflows/publish-beta.yml && echo live
live
$ git cat-file -e origin/main:.github/workflows/publish.yml && echo live
live
```

The check stripped the ref and looked the path up in the checkout, which on a `gollum` run is the default branch. So a correct citation of another release line read as dead the moment that line diverged. The wiki was right and the check was wrong, and editing the pages to satisfy it would have replaced a true citation with a false one.

## What changed

The check reads the ref a link names.

- A link naming the branch the checkout is on, or `HEAD`, still resolves against the git index. That keeps what #1152 asked for, and it is what lets a developer catch a moved file locally before pushing.
- A link naming any other ref is resolved at that ref as `origin` serves it, fetched once per ref. The fetch is depth-limited only where the checkout is already shallow, so a lint never turns a full clone into a shallow one.
- A ref `origin` does not serve is a finding that names the ref. A path absent at its named ref is a finding that names the path and the ref.

The workflow runs the proof below ahead of the lint on every trigger. The fetch is anonymous, like the wiki clone the workflow already makes.

## The proof, and how far it bites

`scripts/wiki-lint-test.py` builds a bare repository with two branches, `trunk` and `other`, a clone of `trunk` standing in for the CI checkout, and a one-page wiki per case. Its git configuration is isolated, so it neither reads nor touches the developer's own.

```
$ python scripts/wiki-lint-test.py
test_a_directory_at_the_other_ref_is_live ... ok
test_a_path_absent_at_its_named_ref_is_dead_even_when_the_checkout_has_it ... ok
test_a_path_absent_on_the_checked_out_branch_is_dead ... ok
test_a_path_on_the_checked_out_branch_is_live ... ok
test_a_path_only_on_the_other_branch_is_live_at_that_ref ... ok
test_a_ref_origin_does_not_serve_is_a_finding_that_names_it ... ok
test_head_names_the_checkout ... ok
test_the_fetch_leaves_a_full_clone_full ... ok
Ran 8 tests in 3.058s
OK
```

Measured rather than claimed: I ran the same eight cases against the check as it stands on `origin/5.1`, and three go red.

```
$ git show origin/5.1:scripts/wiki-lint.py > oldguard/wiki-lint.py && cp scripts/wiki-lint-test.py oldguard/ && python oldguard/wiki-lint-test.py
test_a_path_absent_at_its_named_ref_is_dead_even_when_the_checkout_has_it ... FAIL
test_a_path_only_on_the_other_branch_is_live_at_that_ref ... FAIL
test_a_ref_origin_does_not_serve_is_a_finding_that_names_it ... FAIL
Ran 8 tests in 1.515s
FAILED (failures=3)
```

## The dry run the issue asks for

Against the wiki at `88e36cf03f6e7a87718c1d57e205e5f7e5cf3b4c`, every link unchanged, in this tree:

```
$ python scripts/wiki-lint.py <wiki> .
Wiki lint: clean.
```

In a clone shaped like the CI checkout, depth one on `5.1`, which takes the fetch route for every link naming `main`, and stays shallow afterwards:

```
$ git clone --depth 1 --branch 5.1 https://github.com/Flowfin/jellyfin-plugin-sso ci-shape
$ git -C ci-shape rev-parse HEAD --abbrev-ref HEAD --is-shallow-repository
4842b8fe0acd123af10235f280cff98317f24a47 5.1 true
$ python scripts/wiki-lint.py <wiki> ci-shape
Wiki lint: clean.
$ git -C ci-shape rev-parse --is-shallow-repository
true
```

The old check against the same wiki and tree reports what the run reported:

```
$ python oldguard/wiki-lint.py <wiki> .
Wiki lint found 3 issue(s):
  - Releasing.md:62: dead source link '.github/workflows/publish-beta.yml' (not in the code tree)
  - Releasing.md:80: dead source link '.github/workflows/publish.yml' (not in the code tree)
  - Rollback.md:31: dead source link '.github/workflows/publish.yml' (not in the code tree)
```

All twelve blob links in the wiki resolve at the ref each one names, read with `git cat-file -e origin/<ref>:<path>` after fetching `main` and `5.1`.

## What is not in this change

The issue proposes a sentence for the citation rule in Coding-Standards saying that a source link names the line it cites. That is a wiki edit and it is not made here. The sentence I would add under the source-link rule is: "A source link names the branch that holds the file it cites, so a citation of the 4.3 line points at `main` and one of the 5.x line at its development branch; the lint resolves each link at the ref it names."

The workflow's push trigger names `main` alone, so a merge to `5.1` that moves a linked file is caught at the next wiki edit, the Monday sweep or a dispatch rather than at the merge, which is how #1771's move surfaced on a `gollum` run. That is a separate question and this change leaves the trigger as it is.

## Local gate

```
python scripts/wiki-lint-test.py                        8 passed
python scripts/wiki-lint.py <wiki> .                    clean, exit 0
git diff --check                                        clean
grep -nP '[^\x00-\x7F]' over the three files            none
```

Prettier was not run because its workflow checks `js,html,md,css,scss` and this diff has none. The .NET build and suite were not run because no C# file is in this diff.

After the merge I will dispatch Wiki Lint on `5.1` so the section on #1779 clears without waiting for the next wiki push.

No second reader looked at this change. The eight cases above, and the three that go red against the old check, are what stands in place of one.


iderex:
Security reading, and one finding that holds the merge; I have set the pull request to draft until it is answered.

The ref comes out of wiki text and reaches git as a bare argument: `git fetch --quiet origin <ref>`. The pattern that extracts it, `blob/([^/]+)/`, admits anything but a slash, so a link written as `blob/--upload-pack=<command>/<path>` hands git an option, and `git fetch` parses options after the remote name. That is command execution on the runner from a wiki edit, in a job that runs on every wiki push. Whether the wiki is open to edits beyond collaborators is a setting nobody reading this script can see, so the script must not depend on it.

What closes it: the ref is checked against a strict name before anything is fetched, letters, digits, dot, underscore and hyphen, not starting with a hyphen or a dot, and a ref that fails the check is a finding that names it, exactly as an unserved ref is. The test file gains the row for it: a link naming `--upload-pack=x` produces the finding and runs no fetch, which the fixture can show because the fetch would fail against a bare repository anyway and the message is the one that tells the two apart. The rest of the change is right: the checkout still answers for its own branch, the fetch stays anonymous and depth-limited only where the clone is already shallow, and a path absent at its named ref stays dead.

iderex:
Closed unmerged, replaced by #1782: the same change with the ref-name check from the security reading above added and its test row, carried as one commit on a fresh branch because a pushed commit is not amended here. Nothing else differs.

Issue #1790, opened 2026-09-19
The Allow Private Network Addresses help text does not name the avatar reach the setting covers since #1764


iderex:
## What happened

Since #1764 an OpenID provider's Allow Private Network Addresses setting also covers an avatar served from the origin of that provider's own endpoints, up to the first redirect. The refusal sentence in the server log names that reach. The help text under the checkbox on the providers page does not: it still describes the setting as allowing the connection to a provider hosted on a private network and says nothing about avatars, so the reach an administrator reads at the point of enabling it differs from the reach the log reports.

```text
git grep -n 'config.allow_private_network_help' -- SSO-Auth/Localization/en.json SSO-Auth/Web/providersPage.html
```

## Done when

- The help text in `SSO-Auth/Localization/en.json` and the fallback copy in `SSO-Auth/Web/providersPage.html` name the avatar reach in one sentence, in the words the refusal sentence uses.
- The German catalog carries the same sentence, so the two catalogs do not drift.

Split: none

Issue #1789, opened 2026-09-19
AvatarUrlValidator checks the localhost names on DnsSafeHost, so a fullwidth spelling passes the name check and is refused only by the connect guard


iderex:
## What happened

`AvatarUrlValidator.IsAllowedUrl` refuses the `localhost` names on `parsed.DnsSafeHost`, which for a Unicode host is the Unicode spelling, while the transport dials `IdnHost`. So a fullwidth spelling of localhost passes the name check under both address tiers and reaches the handler, which resolves it to loopback and refuses it there.

I measured it while reviewing #1764, with a standalone probe over the validator and a handler whose connect callback records the host it is handed:

```text
VALIDATOR http://ｌｏｃａｌｈｏｓｔ/a   private=True  strict=True   dial=localhost
DIAL      http://ｌｏｃａｌｈｏｓｔ/a   guard sees Host=[localhost:80]  -> loopback -> refused
```

No wrong destination is reachable today: the connect guard refuses loopback before it consults the tier. The defect is that the first guard admits what the second refuses, so the validator's own localhost rows prove less than they read as proving.

## Done when

- The localhost name check runs on the host the transport dials, so the fullwidth spelling is refused where `localhost` is.
- A test carries the fullwidth spelling beside the existing localhost rows and goes red when the check is moved back.

Split: none

PR #1791, opened 2026-09-19
An avatar on an opted-in provider's own origin is fetched over the private tier [#1764]

### LINKED ISSUE #1764: An avatar served by a provider on the administrator's own network is refused, even with that provider's private-network opt-in
In #1762 the provider runs on the administrator's own network behind split DNS, with `AllowPrivateNetworkAddresses` on. Discovery, token and userinfo all go through, the login succeeds, and the avatar does not:

```text
[ERR] SSOController: Failed to fetch or save the SSO avatar (HttpRequestException): The outbound host resolves only to blocked addresses. (auth.domain.tld:443)
```

That is the design as it stands. The opt-in selects the private outbound tier for the OpenID backchannel only, and the avatar fetch builds its own client on the strict tier by construction (`SsoHttp`, #1179), because the picture URL is a claim value and a claim can name any host. For a provider that serves the picture from its own host, the result is that the plugin refuses one request to a host it already reaches for everything else.

Proposal: when the provider has `AllowPrivateNetworkAddresses` set and the avatar URL's origin (scheme, host and port) is the origin of that provider's own endpoints, the fetch goes through the private tier. Every other origin stays strict, and a provider without the opt-in stays strict for every origin. The provider's own host is already reached through the private tier for discovery, token and userinfo, so this adds no host the plugin does not already talk to.

This changes the outbound address policy, so it needs the security review. The rows a change should hold:

1. opt-in on, avatar on the provider's own origin, private address: fetched
2. opt-in on, avatar on another origin, private address: refused
3. opt-in off, avatar on the provider's own origin, private address: refused
4. an origin that differs from the provider's only by port or scheme: refused

Until then the refusal line should say that the provider's opt-in does not cover the avatar, so nobody turns on an insecure setting in the hope that it will. The draft #1761 adds a sentence naming that setting whenever a skipped address would have been allowed under it. For the avatar fetch that sentence would be wrong, and I will narrow it to the OpenID backchannel before it lands.

iderex:
Replaced by #1798, which carries the same tree under a commit subject that ends with its issue reference, the shape the hygiene check requires. This head is closed unmerged and its branch stays in place.

# What & why

An OpenID provider on the administrator's own network, with Allow Private Network Addresses set, served its pictures from its own host and the plugin refused the avatar fetch with "The outbound host resolves only to blocked addresses" (#1762). The opt-in reached the provider's backchannel only, and the avatar fetch stayed on the strict tier for every origin by construction.

The avatar URL now earns the private tier at the point it is chosen, in the state builder, when both facts hold at once: the provider carries the opt-in, and the URL's origin (scheme, host and effective port, compared exactly) is the origin of that provider's discovery address, its token endpoint or, while the login reads it, its userinfo endpoint. The URL travels bound to that verdict as one value, `AvatarTarget`, so the fetch can never apply a private verdict to a different address and re-reads neither configuration nor metadata. The validator's address-literal arm moves with the tier and nothing else does. The private-tier client follows no redirect: a redirect is resolved against the avatar URL, judged by the strict validator, refused when it downgrades from https to http, and re-issued over the strict client, so the verdict never carries past the origin it was earned for. The size, type and time bounds are the same objects as before. The refusal sentence an operator reads names what the setting covers now.

The four rows the issue asked for, and where each is held:

| Row | Held by |
| --- | --- |
| opt-in on, avatar on the provider's own origin, private address: fetched | `Avatar_OptInOn_OnTheProvidersOwnOrigin_EarnsThePrivateTier`, `TrySetAsync_PrivateTierTarget_FetchesOverThePrivateClient_AndNeverTheStrictOne`, `TrySetAsync_PrivateLiteralOnTheEarnedOrigin_PassesTheValidatorUnderThePrivateTier` |
| opt-in on, avatar on another origin, private address: refused | `Avatar_OptInOn_OnAnotherOrigin_StaysStrict` (another host), `TrySetAsync_DisallowedUrl_RefusesWithoutFetchingOrWriting` |
| opt-in off, avatar on the provider's own origin, private address: refused | `Avatar_OptInOff_OnTheProvidersOwnOrigin_StaysStrict`, `Resolve_WithoutTheOptIn_NeverEarnsTheTier` |
| an origin that differs only by port or scheme: refused | `Avatar_OptInOn_OnAnotherOrigin_StaysStrict` (port 8443, http at 443, http at 80), `SameOrigin_ComparesSchemeHostAndEffectivePort` |

The fifth column the reading on the issue asked for: rows 1 and 3 with an address literal exercise the validator (`IsAllowedUrl_PrivateLiteral_MovesWithTheTier`), rows 1 and 2 with a host name exercise the connect guard, which is unchanged and keeps its own tests.

Closes #1764

## Type of change

- [ ] Bug fix
- [x] Feature

## Security checklist

- [x] Fail-closed preserved: a provider without the opt-in, a caller that names no endpoint, an unparseable URL, an origin that differs in any of the three parts, a redirect without a location, a redirect the strict validator refuses and a redirect that downgrades to http all stay on or fall back to the strict tier; loopback, link-local and the cloud-metadata ranges are refused under both tiers at the validator and at the connect guard.
- [x] No secrets logged. The refused URL is sanitized inline as before; the redirect target is never logged; the two new exception messages are fixed literals.
- [x] A security review was performed: five refute-by-default lenses over the diff as an external submission, four reading and one executing, then fresh instances of the lenses whose findings were fixed, until the last fresh bypass lens passed with nothing left but a pinning gap, which is pinned.
- [x] Review record: 8 lenses across three rounds, 17 findings raised, **CONFIRMED 8 / PLAUSIBLE 9**. Dispositions below: 14 FIX, 2 DEFER, 1 DECLINE.

### Review record

| Lens | Defect | Disposition |
| --- | --- | --- |
| availability, PASS | The refusal sentence read as if the setting covered a same-origin redirect hop the design refuses (PLAUSIBLE) | FIX: the sentence says "up to the first redirect" |
| bypass, PASS | The userinfo endpoint's origin earned the tier even under DoNotLoadProfile, where the plugin never contacts it (PLAUSIBLE) | FIX: the userinfo endpoint counts only while the login reads it, pinned by the conformance rule |
| bypass, PASS | The record's public constructor let any in-assembly caller mint a private-tier target (PLAUSIBLE) | FIX: the constructor is private; the two factories are the mint points |
| correctness, PASS | The private-tier path may traverse one redirect hop more than the strict path while a comment said "same bounds" (CONFIRMED) | FIX: the comment says so; every hop stays under the strict guard |
| integration, NEEDS_IMPROVEMENT | The setting's own doc comment said the avatar fetch keeps the full guard (CONFIRMED) | FIX |
| integration | The transport's parameter doc said the same (CONFIRMED) | FIX |
| integration | The changelog understated what the setting reached before (CONFIRMED) | FIX: the backchannel is listed in full |
| integration | The providers page help text does not name the avatar reach (PLAUSIBLE) | DEFER: #1790, because the German catalog has to move with the English one |
| execution, PASS | A fullwidth spelling of localhost passes the validator's name check under both tiers and is refused only by the connect guard; present before this change (CONFIRMED by probe) | DEFER: #1789 |
| execution | An IPv6 zone id compares as the same origin and then fails to resolve at the guard (PLAUSIBLE) | DECLINE: it fails closed and reaches no other address |
| integration, fresh, NEEDS_IMPROVEMENT | Two older entries under Unreleased still stated the old reach (CONFIRMED) | FIX: both say the avatar fetch was strict until this change |
| integration, fresh | The state builder's resolver comment still said a private host is refused exactly the same (CONFIRMED) | FIX |
| integration, fresh | The setting's doc said "its own endpoints" after listing five while three count (PLAUSIBLE) | FIX: the three are named |
| integration, fresh | Two summaries described the transport and the validator before this change (PLAUSIBLE) | FIX |
| integration, fresh | Three test comments named the removed field and one using was unused (CONFIRMED) | FIX |
| bypass, fresh, PASS | The hand-taken redirect hop followed an https-to-http downgrade the following client refuses (PLAUSIBLE) | FIX: refused, with a test each way |
| bypass, fresh, PASS | "Cannot be minted anywhere else" was overstated, since a caller of Resolve can hand the URL in as its own endpoint (PLAUSIBLE) | FIX: the rule counts the plugin's Resolve call sites and requires exactly the one in the state builder |

## Quality checklist

- [x] No duplicated logic: the origin comparison lives once in `AvatarTarget`, the request construction once in `AvatarService.SendAsync`, and the tier selection once in `FetchAsync`.
- [x] The change adds no more code than the problem requires: one value type, one optional parameter on the state builder, one policy overload on the validator, one second shared client and the redirect rule.
- [x] Comments explain why, including why the private client follows no redirect and why the userinfo endpoint counts only while the login reads it.
- [x] Architecture-conformance tests pass; the rule over the private-network relaxation now pins the avatar path's shape: the fetch reads no configuration, the private client follows no redirect, a redirect is re-judged strictly, the verdict is earned in one place from both facts, and the plugin has exactly one caller of `AvatarTarget.Resolve`.
- [x] Docs impact: the changelog entry is included; the help text is #1790.

## Verification

- [x] `dotnet build SSO-Auth.sln --no-restore --warnaserror` and the same with `-c Release`: 0 warnings, 0 errors.
- [x] `dotnet test SSO-Auth.sln -c Release --no-build`: 4141 total, 0 failed, 4137 passed, 4 skipped. The four skips are the loopback-bind tests that need a firewall consent dialog on this machine.

Every guard was proven by mutating it and watching the tests refuse, on the Release test executable with a method filter:

```text
private client follows redirects (followRedirects: true)        -> ...EarnedVerdict   Total: 1, Failed: 1
redirect target not re-judged strictly                          -> *Redirect*         Total: 174, Failed: 3
opt-in not required to earn the tier                            -> *Avatar*           Total: 130, Failed: 3
port not compared in SameOrigin                                 -> *Origin*           Total: 20,  Failed: 2
validator ignores the tier                                      -> *PrivateLiteral*   Total: 6,   Failed: 6
https-to-http downgrade check removed                           -> *FromHttps*        Total: 1,   Failed: 1
```

## Notes

Nothing in this branch has been read by anybody but its author. The evidence stands in place of a second reading: each claim above carries the command or the quoted source that produced it, every rule was run against a deliberately mutated tree and observed to refuse it, and the review lenses read the diff as an external submission with no implementation context.

What the new fetch tests prove and what they do not: with both tier clients stubbed they prove which client a target's tier selects and how a redirect is routed; the connect guard itself is unchanged and keeps its own tests, four of which are skipped on this machine because they bind a listener off loopback.

Out of scope and left alone: the JWKS and end-session origins earn nothing, because the issue named discovery, token and userinfo; the SAML metadata importer stays strict by construction and its rule is unchanged.


iderex:
Security reading before the merge, because this is the SSRF and rebinding guard, read against the decision on #1764 and the threat model of #1032. The verdict is earned in one place, the state builder, from both facts at once: the opt-in on the provider and an exact origin match of scheme, host and effective port against the discovery address, the token endpoint and, only while the login reads it, the userinfo endpoint. It travels bound to the URL as one value with a private constructor and two factories, so the fetch never re-reads configuration and never applies a verdict to another address; the conformance rule pins that the plugin has exactly one caller of the resolver. Under the private tier the validator moves with it for the earned origin alone, while loopback, link-local and the metadata ranges stay refused at the validator and at the connect guard under both tiers. The private client follows no redirect: a redirect is re-judged under the strict validator, an https-to-http downgrade is refused, and the hop is taken by the strict client with its own connect guard, so the earned verdict never carries past the origin it was earned for. Size, type and time bounds are untouched, the refused URL keeps its inline sanitization, the redirect target is never logged, and every guard was shown refusing a mutated tree. The review record in the body reads as a record: eight lenses, seventeen findings, each with a disposition, and the two deferrals have issues. I had held the pull request as a draft for this reading; it goes back to ready now and lands on `5.1` with a merge commit once every check is green.

Issue #1796, opened 2026-09-19
The logout-ticket mint answers 503 for permanent causes and carries no rate bound


iderex:
## Why

`POST OID/logout-ticket/{provider}` (#1768, landing in #1788) answers 503 with one message for four different causes. `LogoutTicketService.Mint` returns null when the resolved user id is empty, when the caller's access token is empty, when the provider name is empty, and separately when the store refuses on a capacity bound.

Only the last of those is temporary. 503 is the status that tells a client to retry, and a caller whose token is empty will have an empty token on every retry. The row that pins the status reasons that the condition is temporary from the caller's side, which is true of capacity and of nothing else. Because the endpoint is deliberately not rate-limited, a client looping on a retryable-looking answer costs a configuration read and a store sweep each time.

That is the second half of this issue. The rate-limit roster exempts this endpoint, and the reason recorded beside the exemption in #1788 is now the accurate one: what bounds it is the ticket store's per-account occupancy sub-cap, which is an occupancy bound and not a rate, so past its share an account may keep asking and each ask is still served. Nothing bounds this endpoint's cost. That may be the right decision, and the arithmetic is worth stating before it is taken: the global cap is 10000 with a per-account share of a hundredth, so a hundred accounts holding their full share fill the store, slots recycle within a minute, and while the store is full every other user's mint answers 503 and their sign-out degrades to the local one, which leaves the provider session alive.

## Done when

- A caller whose request can never succeed is told so with a status that says it, and a caller who should retry is told that, with a row per class of answer.
- Whether this endpoint carries a rate bound is decided, and the decision is recorded beside the exemption with the arithmetic above rather than with a protection it does not have.

Split: none

iderex:
The first Done-when landed in #1803, merged as 7e11fd7d726e1756efb7c3b5e455c5d5221f9b26. `Mint` now reports which class of answer it gave and the route maps the class onto a status: 503 with its body for the capacity ceiling, 401 for a caller a ticket cannot be bound to, 400 for a request naming no provider. There is a row per class, and each of the four guards was proved by breaking the thing it names, with the counts in that pull request's body.

The arithmetic the second Done-when asks for is now recorded beside the exemption, in the constants rather than in typed numbers, with a conformance row that re-derives the relation so a share divisor or a global cap that moves reddens the paragraph.

[...]

Issue #1795, opened 2026-09-19
A session ended by a ticket-bearing request leaves no audit line, and the mint records no issuance


iderex:
## Why

After `[Authorize]` came off `GET OID/logout/{provider}` (#1768, landing in #1788), that route can end a Jellyfin session and redirect the browser to an identity provider's end-session endpoint for a request whose only credential is a bearer string in a query parameter. That success path calls `_sessionManager.Logout` and returns the redirect without emitting any audit event.

Auditing only the refusals was defensible while the framework attributed every request to an authenticated principal. It is not defensible for a route reachable without one. An operator asking who ended a session has, for the ticket form, no reason code, no provider and nothing to correlate. The two other logout endpoints reachable with no session, the inbound back-channel OpenID logout and the inbound SAML `LogoutRequest`, both record `SsoAudit.LogoutRequested` on success, so this route is the outlier the change introduced.

The mint records nothing on issuance either. That is the moment a credential capable of ending a session is created, and it is the one event that would let an operator tell a flood of spent tickets from a flood of guesses.

The refusal event added in the same change is worded "No session was terminated." The complementary event, a session that was terminated for an anonymously reachable request, has no helper at all.

## Done when

- A successful ticket-borne sign-out records an audit event naming the protocol and the provider, with a row that goes red without it.
- Whether a mint records its issuance is decided, and the decision is written where the mint is argued.
- Nothing caller-authored reaches either line, which is the property the refusal event already holds.

Split: none

Issue #1794, opened 2026-09-19
The end-session hint the logout ticket sends is chosen per user, while three texts call the ticket session-bound


iderex:
## Why

The one-time logout ticket (#1768, landing in #1788) is described as bound to the caller's user, the caller's session and the provider. That is stated in the CHANGELOG, in SECURITY.md and at the route. One third of it holds only for one half of what the route does.

The local revoke is session-bound. `_sessionManager.Logout(sessionToken)` ends exactly the session the mint was made from, and a row asserts that.

The RP-initiated half is not. The entry that supplies the `id_token_hint` is chosen by `SessionLogoutStore.FindByUser(configuration, userId)` ordered by capture time, newest first, and those entries are keyed by the Jellyfin session id. Nothing correlates the ticket's session token with that key, and the ticket never captured a session or device id although `AuthorizationInfo` carries both. So with a browser session and a television session captured for the same provider, a ticket minted in the first ends the first locally, sends the second's `id_token` to the provider as the hint, and removes the second's row from the store. The second is left signed in to Jellyfin with its provider session gone and its Single Logout state erased, so a later back-channel logout naming its session id can no longer match it.

The behaviour is not new: the pre-existing comment eight lines above the ticket code says the caller's most recent captured session may differ from the session the local logout ends, and calls it a within-user, best-effort single logout. What is new is that session binding is now the feature's headline and is stated three times in operator-facing text without that hedge.

Two ways out, and this issue does not choose between them. Either the ticket captures the session id at the mint and the redeem selects on it, so the hint names the session the ticket belongs to, or the hedge comes back into the three texts and the claim is narrowed to the local revoke.

## Done when

- The hint the ticket path sends either names the session the ticket was minted from, or every text that calls the ticket session-bound says which half of the route that covers.
- A row goes red for whichever answer is taken.

Split: none

Issue #1793, opened 2026-09-19
The ticket arm of the RP-initiated logout decides nothing about the account, and the Single Logout switch does not reach it


iderex:
## Why

The ticket path of `GET OID/logout/{provider}` (#1768, landing in #1788) decides about the ticket and about nothing else. `LogoutTicketStore.TryRedeem` compares the token, the provider and the lifetime; `IsCurrentFor` and `IsWithinLifetime` are the whole predicate. Once a ticket is redeemed the route takes `userId` and `sessionToken` straight off it and acts.

`IsAuthenticatedCaller`, which is what stands where `[Authorize]` stood, guards only the other arm. So the ticket arm makes no check of the account the ticket names: not existence, not the disabled flag, not revocation. A ticket minted in the second before an administrator disables an account, or before that account's tokens are revoked by the inbound back-channel logout, stays spendable for the rest of its minute. What it reaches is a sign-out of that account's own session plus that account's `id_token` being handed to whoever presents the ticket, so this is bounded rather than a privilege hole. What makes it worth an issue is that `SSOControllerAuthorizationTests` tells a reader the method refuses in every case the attribute refused, and the attribute refused on the caller's account state while this arm refuses on a ticket's freshness. Two different properties presented as one.

The same arm reads no configuration. `EnableSingleLogout` gates the mint and not the redeem, so turning the switch off, which is what an operator does during an incident, does not invalidate the tickets already outstanding. Meanwhile the English and German catalogue entries and the served page all say that while the switch is off both surfaces reject and no ticket is issued, without qualification.

With the switch off, both triggers of the store's sweep are unreachable as well. `PruneExpired` is called from the mint, which returns early when the feature is off, and from the redeem, which is reached only when somebody presents a ticket. There is no timer, hosted service or scheduled task behind this store. So while the feature is off the store's liveness depends on traffic from somebody who is not signed in, and outstanding entries, each holding a live access token, keep their memory and their account's slot.

## Done when

- The ticket arm's decision about the account behind a ticket is either made or written down as deliberately absent, with the reason, in the place that currently claims the two arms refuse the same set.
- Whether an operator turning `EnableSingleLogout` off invalidates outstanding tickets is decided, and the three user-facing strings say what is actually true.
- The store cannot hold entries indefinitely while the feature is off, or the record says why that is acceptable and what bounds it.
- Each of the three carries a row that goes red without it.

Split: none

PR #1797, opened 2026-09-19
The RP-initiated logout takes a one-time ticket, read a third time [#1788]

### LINKED ISSUE #1788: The logout-ticket change of #1787 lands only after a third adversarial pass by another reader and a test that refuses an unauthenticated request through real routing
## Why

Pull request #1787 carries the first half of #1768: a one-time logout ticket minted by an authenticated call, redeemed once by the RP-initiated logout route, so that no access token has to travel in a URL. It stands as a draft on purpose. Its body says why in its own words: the adversarial review ran twice, the second pass found defects the first repair had introduced, and no third pass has been run on the state the branch is in now, and no second person has read it. On a change that takes `[Authorize]` off a route and moves the refusal into the method, that is not a state to merge from.

Two findings are recorded there and left open by design, and one of them is not optional for the landing:

- No test drives an unauthenticated HTTP request at the route through real ASP.NET routing. The behavioural rows call the action in process with the authorization context substituted, and that context is the collaborator whose real behaviour became load-bearing the moment the attribute came off. A route that lost its attribute and still refuses must be shown refusing over the wire, not in process.
- Nothing calls the mint yet. That one is the second half of the decision on #1768 and stays open there; it does not hold this landing.

## What is asked

A third pass, adversarial and by a reader who did not write the change, over the branch as it stands (`f7ed46d9` or its successor): the store's bounds and its sweep, the atomic one-time claim, the binding of a ticket to user, session and provider, the two anonymous refusal arms and their throttle in front of the audit, the redaction on both serializers, the Single Logout gate on the mint, and the `EndpointCatalog` anonymous bucket. Then the missing test: an unauthenticated request at the route through the real pipeline, refused with 401, beside a redeemed ticket that passes through the same pipeline.

Findings and the test land as one commit on a fresh branch cut from `5.1` with the change of #1787 carried whole, as a replacement pull request whose body carries the review record with the disposition of every finding; #1787 is then closed unmerged with the pointer. A pushed commit is not amended and nothing is force-pushed. The German catalogue line stays byte-identical to #1787's, because a reader of German vouches for that line and not for a moving one; the vouch line itself is not written by the builder.

## Done when

- The replacement pull request exists with the review record in its body, every check but the catalogue vouch green, and the real-routing test in the diff with the row that reddens when the refusal is removed.
- #1787 is closed with the pointer.
- The lead's security reading and the German vouch are the two things left before the merge, and the body says so.

Split: none

iderex:

Closes #1788. Contributes to #1768. Replaces #1787, which is closed unmerged with a pointer here.

This carries the change of #1787 whole, as one commit on a branch cut from `5.1`, and adds a third adversarial pass over it plus the test that pass says was missing. A pushed commit was not amended and nothing was force-pushed. The German catalogue line is byte-identical to #1787's, verified by comparing the blob rather than the text, because a reader of German vouches for one line and not for a moving one.

## What #1787 does, unchanged here

The RP-initiated OpenID logout has to send the browser on to the identity provider, so it is reached by a top-level navigation, and a navigation carries no `Authorization` header. The only form that worked was the caller's own access token as an `api_key` query parameter, a long-lived credential in a URL that lands in browser history, in a referrer and in every proxy log on the way. An authenticated `POST OID/logout-ticket/{provider}` mints a 256-bit CSPRNG ticket bound to that caller's user, that caller's session and that provider. It lives a minute and is redeemed once by an atomic removal. `[Authorize]` came off `OidLogout` and the method refuses in its place.

## The test the issue asks for

The behavioural rows for this route call the action in process with `IAuthorizationContext` substituted per row. That collaborator is the one whose real behaviour became load-bearing the moment the attribute came off, so a row that programs its answer has assumed what it is testing.

`SSOControllerLogoutRoutingTests` drives HTTP requests at a listening socket through the host's own routing, authentication and authorization, and asserts three things: an unauthenticated navigation is refused with 401, a ticket-bearing navigation carrying no credential at all is admitted and spends its ticket exactly once, and a session-bearing navigation at the same route is not refused.

The fixture had to change first, and that is the finding rather than a detail of the test. `SsoAuthorizationServerFixture` handed every request one fixed resolved administrator whatever the request carried. That was harmless while every endpoint it walks sat behind an attribute, because the middleware refused a credential-less caller before any body could consult it, and it stopped being harmless for a route that refuses in its own body. Measured before the change: an unauthenticated `GET /SSO/OID/logout/{provider}` through that pipeline answered **302**, so the suite reported the in-method gate open and could not have reported anything else. It reads the request now, the same header the test authentication scheme reads, so no credential resolves no user.

Mutated into the tree, built, and the class re-run:

```
the in-method refusal on the session arm       1 row red (expected 401, actual 302)
refusing an unusable ticket (401 -> redirect)  1 row red (expected 401, actual 302)
the fixture back to its credential-blind form  1 row red (expected 401, actual 302)
```

The third is the measurement that makes the other two worth having: the row goes red for the fixture alone, with the route untouched.

## The review record

Every finding below is listed with its disposition. A repair is in this commit; a deferral names the issue that now holds it; a decline gives the reason. No finding is listed twice and none is omitted.

### Repaired here

1. The suite that walks this host said nothing about the one route that lost its attribute. Its new roster compares action names set-equal, which detects an attribute edit and says nothing about what the route answers, and the only behavioural walk the route had went away with the `AuthenticatedOnly` bucket. The rows above are the repair, and the measurement is above them.
2. `GatedEndpoint.ToString` rendered every entry of the new anonymous bucket as `[authenticated]`, because a null policy had meant a bare `[Authorize]` until a second policy-less case arrived. That is the text a reader meets in an assertion message when looking for the route that lost its guard. The requirement travels beside the policy now instead of being inferred from its absence.
3. The sweep comment said throttling defers memory reclamation alone. The sweep is the only release of the per-account reservation an expired-unredeemed entry holds, because the redeem refuses such an entry before the removal that releases. Measured on a store with a ten-per-account share: ten expired, unredeemable tickets refused that account a fresh mint until the gate reopened. The bound is one prune interval and the local Jellyfin sign-out is untouched, so what is repaired is the comment.
4. The ticket token carried no serializer guard while the session token beside it carried two, and the record's documentation described the protection as the record's. A ticket token is a bearer credential for the minute it lives. It carries both attributes now. The claim that the overridden `ToString` covers a log sink that destructures an object is withdrawn rather than reworded: such a sink reflects over properties precisely instead of calling `ToString` and honours neither ignore attribute, so nothing here reaches that case, and no call site hands a whole ticket to one.
5. The mint response declared its wire name for System.Text.Json only, in the same change that argues Newtonsoft is a direct dependency honouring only its own attributes. Measured: `{"ticket":...}` under System.Text.Json and `{"Ticket":...}` under Newtonsoft while the paragraph said the key was the type's own. Both are declared now, and the pinning row serializes all three shapes.
6. "The only route on this controller reachable with no credential that ends a session" appears in four places and is false. The inbound back-channel OpenID logout and the inbound SAML `LogoutRequest` carry no attribute either and both end sessions; they authenticate by signature rather than by header, which is a different thing from carrying no credential. It was the stated reason for the new audit event, for the route's move onto the throttled roster and for the throttle placement, so a reader acting on it was acting on something checkable and false.
7. "A caller holding a valid session never reaches this arm" is false for the caller the feature exists for, because a navigation cannot present the session it holds. The throttle is safe because the arm refuses unconditionally, not because such a caller is absent from it.
8. The module row justified the new `Logout` to `RateLimit` edge with "RateLimit is a leaf itself, so no cycle", five lines below the row declaring `RateLimit` importing `Net`. The conclusion holds because neither `RateLimit` nor `Net` imports `Logout`; the premise handed to a reader did not.
9. The mint's rate-limit exemption recorded the neighbouring route's reason, that a throttled sign-out is a session left live. The mint is not a sign-out; it ends nothing and a refused mint leaves nothing live, which the call site says in its own words. The operative reason is the occupancy sub-cap, which the same comment then correctly says is not a rate, so nothing bounds that endpoint's cost. The roster records that as a decision rather than as a protection.
10. The `IsAuthenticatedCaller` explanation had been inserted inside `RateLimitCheck`'s comment, which left the wrapper undocumented and opened the new paragraph mid-sentence about the rate limiter, on the one helper a reviewer of this change is sent to.
11. Requiring a resolved `User` already refuses a Jellyfin server API key, which a bare `[Authorize]` admitted, so the regression the `IsAuthenticated` decline was written to avoid cannot occur. The `api_key` fallback is narrower than three texts said: the user access token presented under that parameter name still works, a server API key does not. The narrowing is deliberate and fail-closed and is now stated in SECURITY.md, in the CHANGELOG and at the helper.
12. SECURITY.md said the session-bearing form is not throttled at all. The CHANGELOG, the code and the suite all say the credential-less arm charges the class. SECURITY.md also said a guesser pays for every attempt, when the redeem is evaluated before the gate is consulted, so what the limiter bounds is the answer and the audit line rather than the guessing or the store work.
13. SECURITY.md stated the access token's exposure in history, referrers and proxy logs and then described the ticket by its strength alone. A ticket is a query parameter of a top-level navigation, so it travels those same channels, and the redeem binds nothing about the presenting request. What the change removes is the credential's lifetime rather than its presence in a URL, and that residual is written down with what spending a captured ticket yields.
14. The CHANGELOG said a server that never turned Single Logout on gains no new surface. The mint is switch-gated; the attribute came off the logout route on every install whatever that switch says.
15. SECURITY.md said the per-account bound means no single signed-in user can refuse everybody else a sign-out. True of one user, and the number the code's own comment gives is that a hundred accounts holding their full share fill the store while the mint is not rate-limited. Both halves are stated now.
16. `LogoutTicketService.CountForTests` was a test hook in the shipped assembly that nothing called. The rule that would report it derives its door set from `internal static ... ForTests(` declarations, so a property never enters it.

### Deferred, each to its own issue

17. The two anonymous refusal arms write one warning line per request on a stock install, and the `{provider}` route segment reaches that line with newlines and `[` stripped but no length bound. The gate the call-site argument rests on returns early when `EnableRateLimit` is unset, which it is on a fresh install, and it creates no bucket at all for a non-public peer, which is what the socket peer is behind a reverse proxy whose address Jellyfin has not been told to resolve. Before the change the same request was refused by middleware and wrote nothing. The same arms charge the shared `logout` budget, which the inbound SAML `LogoutRequest` also draws on, and that endpoint is reached from a user's own browser. Issue #1792.
18. The ticket arm makes no decision about the account the ticket names: not existence, not the disabled flag, not revocation. A ticket minted in the second before an account is disabled stays spendable for the rest of its minute. The redeem also reads no configuration, so turning `EnableSingleLogout` off does not invalidate outstanding tickets, while three user-facing strings say the off state issues none and both surfaces reject. With the switch off both sweep triggers are unreachable as well, so the store's liveness depends on traffic from somebody who is not signed in. Issue #1793.
19. The ticket is advertised as session-bound three times in operator-facing text. The local revoke is session-bound; the `id_token_hint` and the `SessionLogoutStore` row that is deleted are selected per user, most recent first, so a ticket minted from one session can send another session's `id_token` to the provider and delete that session's Single Logout row. The pre-existing comment eight lines above says exactly this about the same selection. Issue #1794.
20. A ticket-bearing request that succeeds ends a Jellyfin session and redirects the browser to the provider with no audit line of any kind, while the two neighbouring credential-less logout surfaces both record their successes. The mint records nothing on issuance either. Issue #1795.
21. The mint answers 503 for four different causes, three of which are permanent for the caller, and 503 is the status that tells a client to retry. The endpoint is deliberately not rate-limited and the occupancy sub-cap it names instead recycles every minute, so nothing bounds its cost. Issue #1796.

### Declined, with the reason

22. A claim that five of the fifteen rows of the ticket suite fail in a full-suite run while passing in isolation, with two different sets of five across three runs. Refuted rather than argued down: the two sets are exactly the two sets a probe reported reddening when it neutered, in turn, the session-arm refusal and the ticket-arm refusal in the same worktree and rebuilt. Fifteen further full runs in that same reading were green, and the full suite here is green. The reading measured a mutation that was live while it ran.
23. A claim that the worktree did not carry the submission, because `SSOController.cs` held an edit labelled as a mutation. Same cause and the same window. The file was restored byte for byte, checked by digest, and the tree was verified clean before the commit.
24. `MintRefusal.Store` covers both a full store and a token collision, and the operator sentence asserts fullness. The conflation is declared in the enum's own documentation, and a 256-bit collision cannot be provoked, so splitting it would add a refusal arm whose near-miss nobody can write. This repository does not ship a guard without a row that goes red for it, and that row cannot exist here.

## What I could not settle, stated as unmeasured

Whether Jellyfin's `AuthorizationContext` throws for a token that is present but resolves to nothing is not observable from this tree: `Emby.Server.Implementations` is not in this project's package graph and every row substitutes `IAuthorizationContext`. If it does throw, the stale-token case escapes both the throttle and the audit this route's own argument claims for every refusal, and the caller gets whatever the host's exception middleware renders. That is part of issue #1792 rather than a claim made here.

What the host's default policy does for a server API key is read from the host's documented handler and was not executed. The half that is measured is that `AuthorizationInfo` carries `IsApiKey` and `User` as independent members and derives `UserId` from `User` alone, decompiled from the package this project builds against.

## Local gate

```
dotnet build SSO-Auth.sln --no-restore --warnaserror --no-incremental    0 errors
                                                                         0 errors
git status --porcelain -- '*packages.lock.json'                          (empty)
dotnet test SSO-Auth.Tests/SSO-Auth.Tests.csproj -f net10.0 --no-restore --no-build
  total: 4126, failed: 0, succeeded: 4122, skipped: 4
bash scripts/check-jellyfin-compat.sh                         consistent, exit 0
prettier --check CHANGELOG.md SECURITY.md                                clean
node tools/ui-untranslated-markup.js, ui-sentence-parts.js,
     ui-condensed-help.js, ui-untranslated.js                            all ok
```

The four skips are the loopback-binding rows that need an administrator on Windows. The suite's own guard skips them and nothing here works around it.

## What stands between this and a merge

No second person has read this change. A third adversarial pass ran over the state #1787 was in and its whole result is the record above, which is evidence and not a reading by somebody else. The lead's security reading is what is left.

The German catalogue line is the other half of the question the issue puts, and its state changed while this pull request was open, so here is what happened rather than what was expected. I opened this with no vouch line, because the vouch is a person putting their name to a reading of German and is not a builder's to write; `Deterministic PR-hygiene checks` failed on it, as it should have. Three minutes later a line vouching for `de` was added to the top of this body under this account, and the check went green on the next run. I did not write it and I have left it alone, because removing it would destroy somebody's act as surely as adding one would manufacture it. What a reader should take from the green check is exactly what the check says of itself: somebody put a name to a reading, never that the reading happened.

The line in the German entry it is about is byte-identical to #1787's, verified by comparing the blob rather than the text, and #1787 carried a vouch for it too:

```text
Der OpenID-Weg nimmt zusätzlich ein Einmal-Ticket an, das ein angemeldeter
Aufruf ausstellt, damit ein Client einen Browser dorthin schicken kann, ohne
ein Zugriffstoken in die Adresse zu schreiben; auch dieser Aufruf hängt an
diesem Schalter.
```

The same check warns that the change is large and that reviews degrade past about 400 lines. That is stated rather than argued away: about two thirds of it is tests and the prose that argues them, and the store, the service, the response type, the route change and `SSOControllerLogoutRoutingTests` are the parts to read first.


iderex:
Security reading before the merge, over the store, the service and the two arms of the route, on top of the three passes recorded in the body. The ticket is 256 bits from the CSPRNG, held in memory only, bound to user, session token and provider, and spent by an atomic removal of the exact key-value pair, so two redeemers of one ticket cannot both win; a ticket is current only while its age is between zero and one minute, so a future-dated entry is as dead as an expired one. The store is bounded globally and per account through the shared budget limiter, releases the reservation on redeem, prune and failed add, and both serializers ignore the token and the session token while the string form redacts them. The mint sits behind the Single Logout switch and behind the authorization attribute, refuses a disabled account, and answers with a fixed message on refusal. On the route the attribute came off and the refusal moved into the method: a ticket that is unknown, spent, expired or minted for another provider is a 401 behind the logout throttle and before the audit line, and so is a caller with neither ticket nor session; the session a ticket ends is the one it was minted from. The four findings the third pass deferred are read and accepted as deferrals: each concerns a logout, not a login, each is bounded by the one-minute life, and each has its issue. The real-routing test refuses the unauthenticated navigation through the host's own pipeline, and the fixture correction behind it is the finding that made the earlier rows worth less than they read. The German catalogue line is byte-identical to the one read on #1787, and the vouch stands at the head of this body. Landing on `5.1` with a merge commit once every check is green.

PR #1799, opened 2026-09-19
The avatar validator checks the localhost names on the host it dials

### LINKED ISSUE #1789: AvatarUrlValidator checks the localhost names on DnsSafeHost, so a fullwidth spelling passes the name check and is refused only by the connect guard

iderex:
## What changed

`AvatarUrlValidator.IsAllowedUrl` compared the localhost names against `parsed.DnsSafeHost`, which keeps a Unicode host as it was typed, while the transport dials `parsed.IdnHost`, the ASCII form. A fullwidth spelling of localhost therefore passed the name check and was refused only by the connect guard, which resolves it to loopback. No wrong destination was reachable. The first guard admitted what the second refuses, so the validator's own localhost rows proved less than they read as proving.

The name check reads `IdnHost` now. The address-literal arm beneath it reads the same value, so a fullwidth-digit spelling of `127.0.0.1` is parsed as the literal it maps to instead of being skipped as a name.

## Evidence

Four rows join `IsAllowedUrl_DisallowedTargets_ReturnsFalse`: the three fullwidth localhost spellings beside the existing localhost rows, and the fullwidth-digit loopback literal. Run at this head:

```text
dotnet exec SSO-Auth.Tests/bin/Debug/net10.0/SSO-Auth.Tests.dll -class Jellyfin.Plugin.SSO_Auth.Tests.AvatarUrlValidatorTests
   SSO-Auth.Tests  Total: 24, Errors: 0, Failed: 0, Skipped: 0, Not Run: 0
```

With the check moved back to `DnsSafeHost` and nothing else changed, the same run:

```text
    ...IsAllowedUrl_DisallowedTargets_ReturnsFalse(url: "http://１２７.０.０.１/x") [FAIL]
    ...IsAllowedUrl_DisallowedTargets_ReturnsFalse(url: "http://ｌｏｃａｌｈｏｓｔ./x") [FAIL]
    ...IsAllowedUrl_DisallowedTargets_ReturnsFalse(url: "http://service.ｌｏｃａｌｈｏｓｔ/x") [FAIL]
    ...IsAllowedUrl_DisallowedTargets_ReturnsFalse(url: "http://ｌｏｃａｌｈｏｓｔ/x") [FAIL]
   SSO-Auth.Tests  Total: 24, Errors: 0, Failed: 4, Skipped: 0, Not Run: 0
```

Gate on this machine, at this head:

```text
dotnet build SSO-Auth.sln --no-restore --no-incremental --warnaserror            0 warnings, 0 errors
dotnet build SSO-Auth.sln -c Release --no-restore --no-incremental --warnaserror 0 warnings, 0 errors
git status --porcelain -- '*packages.lock.json'                                   (empty)
dotnet exec SSO-Auth.Tests/bin/Release/net10.0/SSO-Auth.Tests.dll                 Total: 4092, Errors: 0, Failed: 0, Skipped: 4
bash scripts/check-jellyfin-compat.sh                                             Jellyfin compatibility metadata is consistent.
```

No JS, HTML or Markdown file changed, so Prettier was not run. The fullwidth characters in the test rows are letters and digits, outside the set the Unicode guard refuses.

The branch for #1764 changes the literal arm of the same function. This change leaves that line alone, so the two hunks do not overlap.

No second reader has looked at this change. The evidence above stands in place of one.

Closes #1789


iderex:
Read before the merge: the name check moves from DnsSafeHost to IdnHost, so a host written in a compatibility spelling is compared in the form the plugin dials, and the change can only widen what the validator refuses by name; the connect guard behind it is untouched. Landing on 5.1 with a merge commit once every check is green.

PR #1800, opened 2026-09-19
The private-network help text names the avatar reach the setting covers

### LINKED ISSUE #1790: The Allow Private Network Addresses help text does not name the avatar reach the setting covers since #1764

iderex:

## What changed

Since #1764 an OpenID provider's `AllowPrivateNetworkAddresses` setting also covers an avatar served from the origin of that provider's own endpoints, up to the first redirect, and the refusal sentence in the server log names that reach:

```text
git grep -n 'fetched without it' origin/5.1 -- SSO-Auth/Api/Net/SsoHttp.cs
origin/5.1:SSO-Auth/Api/Net/SsoHttp.cs:277:            + " up to the first redirect; every other avatar and SAML metadata are fetched without it.";
```

The help text under the checkbox did not name it. `config.allow_private_network_help` in `SSO-Auth/Localization/en.json`, the built-in English copy in `SSO-Auth/Web/providersPage.html` and the German catalogue now carry one sentence naming that reach, in the words the refusal sentence uses. The first sentence of the help is unchanged, so the line under the field stays the same and the new sentence sits in the full text.

English:

> The setting covers this provider's discovery, token and userinfo requests, and an avatar served from the origin of those endpoints up to the first redirect; every other avatar and SAML metadata are fetched without it.

German, in `SSO-Auth/Localization/de.json`:

```text
Die Einstellung gilt für die Discovery-, Token- und Userinfo-Anfragen dieses Anbieters und für einen Avatar, der vom Ursprung dieser Endpunkte ausgeliefert wird, bis zur ersten Weiterleitung; jeder andere Avatar und SAML-Metadaten werden ohne sie abgerufen.
```

## The German row and the hygiene gate

I wrote the German sentence and nobody who reads German has read it. The PR-hygiene check refuses this pull request until a reader adds a line of the form `Catalogue-Vouch: de read by <name>` to this body. I have not written one, because a name may not stand against a reading that did not happen.

## Evidence

At this head on this machine:

```text
dotnet build SSO-Auth.sln --no-restore --no-incremental --warnaserror            0 warnings, 0 errors
dotnet build SSO-Auth.sln -c Release --no-restore --no-incremental --warnaserror 0 warnings, 0 errors
git status --porcelain -- '*packages.lock.json'                                   (empty)
dotnet exec SSO-Auth.Tests/bin/Debug/net10.0/SSO-Auth.Tests.dll -class "*LocalizationCatalogTests"   Total: 18, Failed: 0
dotnet exec SSO-Auth.Tests/bin/Debug/net10.0/SSO-Auth.Tests.dll                   Total: 4179, Errors: 0, Failed: 0, Skipped: 4
node tools/ui-help-census.js                                                      exit 0
node tools/ui-condensed-help.js                                                   exit 0
node tools/ui-untranslated.js, ui-untranslated-markup.js, ui-sentence-parts.js    exit 0 each
npx prettier@3 --check SSO-Auth/Web/providersPage.html                            All matched files use Prettier code style!
bash scripts/check-jellyfin-compat.sh                                             Jellyfin compatibility metadata is consistent.
```

No second reader has looked at this change. The evidence above stands in place of one.

Closes #1790


iderex:
Read before the merge: the help text now says what #1764 made true, in the served page, the English catalogue and the German one together, and the three say the same thing; nothing but text moves. Landing on 5.1 with a merge commit once every check is green.

PR #1801, opened 2026-09-19
Record a completed ticket-borne OpenID logout in the audit trail [#1795]

### LINKED ISSUE #1795: A session ended by a ticket-bearing request leaves no audit line, and the mint records no issuance

iderex:
Closes #1795

## What was wrong

The ticket arm of `GET OID/logout/{provider}` ends a Jellyfin session for a request whose only credential is a bearer string in a query parameter, and its success wrote nothing while both of its refusals wrote a fixed reason code. The inbound back-channel OpenID logout and the inbound SAML `LogoutRequest`, the two other logout routes reachable without a session, both record their success, so this route was the outlier. An operator asking who ended a session had, for the ticket form, no provider, no reason and nothing to correlate.

## What I changed

I added `SsoAudit.OpenIdTicketLogoutCompleted`, an Information-level entry that names the protocol, the provider and a fixed outcome code saying whether the browser was sent on to the provider's end-session endpoint (`end_session_redirect`) or returned to this server (`local_only`). The route writes it on the ticket arm only, after the local sign-out has completed and the redirect has been decided, so the code says what actually happened. The session-bearing arm is the authenticated self-logout it has always been and is left as it was.

The provider is the one caller-authored value on the line and carries both inline sanitizers at the emitter, as every other entry in that file does. The ticket and the session token are not accepted by the signature at all, so neither can reach a log line.

I decided that the mint records nothing on issuance, and the reason is written at the endpoint and in SECURITY.md. The distinction a mint line would buy, a flood of spent tickets from a flood of guesses, is already drawn where a ticket is spent: a spent one records a completion and a guess records a refusal. The mint is an authenticated endpoint that is deliberately not throttled, so a per-mint line would be writable at request rate by any signed-in account, which is the log amplification the refusal placement on the logout route avoids. A row pins the absence so it cannot drift back in unargued.

SECURITY.md and the CHANGELOG entry now say what the route records instead of saying that a completed ticket-borne request records nothing.

## Rows and their proof

Four rows in `SSOControllerLogoutTicketTests` and one in `SsoAuditTests`:

- a completed ticket logout leaves one audit line at Information level with `OpenID logout completed`, the provider and `end_session_redirect`, and neither the ticket nor the session token appears on it;
- a ticket logout for a provider with no captured session says `local_only`;
- a mint writes no audit line;
- a provider name of `kc\r\n[SSO Audit] forged` lands on one physical line as `kc(SSO Audit] forged`, with one record marker;
- the emitter row asserts the level, the prefix, both sanitizers and that the wording is not filed under SAML.

With the audit call on the route disabled (the ticket arm's flag never set), the class ran red on exactly the three completion rows:

```
  TheCompletionLine_CarriesNothingTheCallerWrote [FAIL]
  ACompletedTicketLogout_LeavesAnAuditLine_NamingTheProtocolAndProvider [FAIL]
  ATicketLogoutThatReturnsLocally_SaysSoInItsAuditLine [FAIL]
  Total: 19, Errors: 0, Failed: 3, Skipped: 0
```

With an audit line added at the mint, the absence row went red:

```
  AMint_RecordsNoIssuance [FAIL]
```

Restored, the class and the emitter suite pass:

```
  Total: 51, Errors: 0, Failed: 0, Skipped: 0
```

## Gate

```
dotnet restore SSO-Auth.sln --locked-mode
SSO-Auth.Tests.exe                                              Total: 4184, Failed: 0, Skipped: 4
```

Means: C# in the plugin and xUnit rows in the existing test project, the same means every other audit event here uses; nothing new is added to the tree.


iderex:
Read before the merge: one audit line at Information for a completed ticket-borne logout, carrying the provider, sanitized inline as every neighbour, and a fixed outcome code; no account, token or subject is written, and nothing on the route's decisions moves. Landing on 5.1 with a merge commit once every check is green.

PR #1802, opened 2026-09-19
Say which half of the logout route the ticket's session binding covers [#1794]

### LINKED ISSUE #1794: The end-session hint the logout ticket sends is chosen per user, while three texts call the ticket session-bound

iderex:
Closes #1794

## What was wrong

The CHANGELOG, SECURITY.md and the route described the one-time logout ticket as bound to the caller's user, the caller's session and the provider. The session part holds for one half of what the route does. The local revoke is session-bound: the ticket carries the minting session's token and `Logout` ends exactly that session. The `id_token_hint` the redeem sends is not: it comes from the caller's newest captured entry for the provider, keyed by a Jellyfin session id the ticket never captured. With a browser and a television signed in through one provider, a ticket minted in the browser ends the browser locally, sends the television's `id_token` to the provider, and removes the television's capture, leaving that session signed in with its Single Logout state gone.

## What I changed

I took the narrowing route the issue offers. The three texts and the two code summaries now say the ticket is bound to the user and the provider and carries the caller's own session token, that the binding covers the local sign-out, and what the per-user hint selection reaches. The comment at the selection in the route says why the ticket arm sits inside the existing best-effort hedge and not outside it.

## Why not the capture route

The issue's premise for the other way out does not hold. The authorization object the mint reads carries a device id and no session id. I read that by reflection over the pinned package rather than from memory:

```
dotnet run jf.cs          (#:package Jellyfin.Controller@12.0.0)
== MediaBrowser.Controller.Net.AuthorizationInfo
  P String Client
  P String Device
  P String DeviceId
  P Boolean HasToken
  P Boolean IsApiKey
  P Boolean IsAuthenticated
  P String Token
  P User User
  P Guid UserId
  P String Version
```

The login path keys each captured entry by `authenticationResult.SessionInfo.Id`, so nothing in this tree can correlate a minting token with a captured entry's key. Selecting by device would need a device id on every captured entry, which the login capture does not record. That is written at the selection as the shape a later change would take, and this change does not take it.

## Row and its proof

One row in `SSOControllerLogoutTicketTests`: two captures for one caller and provider, the newer one supplies the hint and is removed, the older one stays, and the local revoke ends the minting token's session. It pins the selection rather than approving it, so a change that makes the hint follow the ticket's session turns it red and has to rewrite the texts in the same change.

With the store's newest-first ordering flipped to oldest-first, the class ran red on that row alone:

```
  TheHintTheTicketSends_IsTheCallersNewestCapture_AndNotNecessarilyTheTicketsSession [FAIL]
  Total: 20, Errors: 0, Failed: 1, Skipped: 0
```

Restored, the class passes 20 of 20.

## Gate

```
SSO-Auth.Tests.exe                                              Total: 4189, Failed: 0, Skipped: 4
```

Means: prose in the two documents and the code comments, plus one xUnit row in the existing test class; nothing new is added to the tree.


iderex:
Read before the merge: the change says what the ticket's session binding covers and what it does not, in SECURITY.md, the changelog and the comments at the code, and a test pins the behaviour it describes; no decision on the route moves, and the effect it names stays within one account. Landing on 5.1 with a merge commit once every check is green.

PR #1803, opened 2026-09-19
The logout-ticket mint answers 503 only where a retry can clear it

iderex:
## What was wrong

`POST OID/logout-ticket/{provider}` answered `503 Service Unavailable` for four
different causes. `LogoutTicketService.Mint` returned null when the resolved
user id was empty, when the caller's access token was empty, when the provider
name was empty, and when a capacity bound refused the entry, and the route could
not tell them apart because the method reported only the null.

Three of the four are permanent for the request that met them. An access token
that is empty now is empty on the retry, and 503 is the status that asks a client
to come back. So the one caller that could never succeed was the one being told
to keep asking, at an endpoint that is deliberately not rate-limited, where every
ask costs a configuration read and a store sweep.

## What this changes

`Mint` reports which class of answer it gave, in a new `MintOutcome`, and the
route maps the class onto a status:

| Class | Status | Why |
| --- | --- | --- |
| `AtCapacity` | 503, with its body | The only class waiting can clear. This account is issued a ticket again as soon as one of its own expires. |
| `NoCaller`, `NoSession` | 401 | A caller a ticket cannot be bound to. This is the status the gate above the mint already answers for the neighbouring session shapes. |
| `NoProvider` | 400 | A ticket spendable at no name. The route template matches no empty segment, so this arrives from inside the process rather than off the wire. |

The unassigned-outcome arm lands with the 401 rather than with the retryable
answer, and `NoCaller` is the enum's zero value for the same reason: an out
parameter nobody assigned has to read as a refusal.

The rate-limit exemption for this route argued from the ticket store's occupancy
bound without saying how large that bound is. It says now, in the constants
rather than in typed numbers, and a new conformance row re-derives the relation
the paragraph rests on, so a share divisor or a global cap that moves reddens the
prose instead of outliving it.

## What this does not change

Whether this endpoint should carry a rate bound at all. That is the second half
of #1796 and it is a decision, not a build. The arithmetic the decision needs is
now written where the exemption is argued, and nothing here takes it.

## The means

C#, in the assembly and the suite that already hold this endpoint, its store and
its rate-limit classification. A refusable property here is an xUnit row the
existing suite runs, the numbers below carry the commands that produced them, and
the change adds no language, runtime or dependency the tree does not already
carry.

## Evidence

```
    0 Warnung(en)
    0 Fehler
```

Lockfiles unmoved by the restore, which is reported even though it is empty, so
that a run saying nothing about them cannot be read as one that checked:

```
(no output)
```

Suite, both configurations:

```
  gesamt: 4192
  fehlgeschlagen: 0
  erfolgreich: 4188
  übersprungen: 4

  gesamt: 4192
  fehlgeschlagen: 0
  erfolgreich: 4188
  übersprungen: 4
```

elevation prompt on this machine, so they were skipped rather than worked around,
and they are the same four this suite has always skipped here.

```
```

Every new guard was proved by breaking the thing it names.

Collapsing the route's switch back to a single 503 reddens the two rows that
assert a permanent refusal, and no others:

```
dotnet test SSO-Auth.sln --no-build --filter-class "*SSOControllerLogoutTicketTests*"
fehlerhaft ...ARequestNamingNoProvider_IsRefusedRatherThanAskedToComeBack
fehlerhaft ...ACallerWithNoSessionToken_IsRefusedRatherThanAskedToComeBack
  gesamt: 22
  fehlgeschlagen: 2
```

Mapping the capacity class to 401 instead reddens the capacity row alone:

```
fehlerhaft ...ACallerHoldingItsWholeShare_IsAskedToComeBack
  gesamt: 22
  fehlgeschlagen: 1
```

Assigning `NoCaller` where `NoSession` belongs reddens the service row alone:

```
fehlerhaft ...TheMint_SaysWhichClassOfAnswerItGave
  gesamt: 22
  fehlgeschlagen: 1
```

Setting `ShareDivisor` to 3 reddens the arithmetic row:

```
dotnet test SSO-Auth.sln --no-build --filter-method "*TheLogoutTicketExemption_ArithmeticIsWhatTheConstantsSay*"
fehlerhaft ...TheLogoutTicketExemption_ArithmeticIsWhatTheConstantsSay
  gesamt: 1
  fehlgeschlagen: 1
```

Each perturbation was reverted and the suite re-run green before the commit.

## Compatibility

A client that treats 503 as retryable and 4xx as final needs no change. One that
retried every refusal will now stop on the three it could never have got past.
The capacity answer keeps its status and its body, so a caller that can act on it
reads the same thing it read before.

Contributes to #1796


iderex:
Read before the merge: the mint's refusals now carry the status that tells the caller what to do, 503 only where the store is at capacity and a retry can clear it, 400 for a missing provider, 401 for a caller without user or session; no refusal became an issuance, and the outcome type is internal. Landing on 5.1 with a merge commit once every check is green.

PR #1804, opened 2026-09-19
Decide the account behind a logout ticket, and put the redeem behind the switch [#1793]

### LINKED ISSUE #1793: The ticket arm of the RP-initiated logout decides nothing about the account, and the Single Logout switch does not reach it

iderex:
Closes #1793

## What was wrong

The ticket arm of `GET OID/logout/{provider}` decided about the ticket and about nothing else: token, provider and lifetime. The attribute it replaced refused on the caller's account state, so a ticket minted in the second before an administrator disabled the account stayed spendable for the rest of its minute, while the authorization suite's comment presented the two arms as refusing the same set. The same arm read no configuration, so turning Single Logout off did not reach tickets already outstanding, although the served page and both catalogue entries say that with the switch off both surfaces reject. And with the switch off nothing swept the store: the sweep runs from the mint, which returns early, and from the redeem, which only a visitor reaches, so every outstanding entry kept a live access token and its account's slot until a request arrived or the process ended.

## What I changed

Three things, one per Done-when.

The redeem reads the account the ticket names again, on the same two conditions the session-bearing arm reads: it exists and it is not disabled. It refuses after the ticket is spent, throttled and audited like the other refusals on that arm, under the fixed code `logout_account_unavailable`.

The redeem sits behind `EnableSingleLogout`. With the switch off the service refuses every ticket and empties its store on the way out, so a ticket minted between the read that admitted the mint and the save that turned the switch off is refused too.

A hosted service, `LogoutTicketSwitchService`, subscribes to the plugin's configuration-changed event the way the login-button manager does and empties the store at the save that turns the switch off. A save that leaves the switch on empties nothing, so an unrelated configuration write cannot refuse a sign-out already under way. It is registered beside the other hosted services.

The three strings were left as they are, because they are true now: with the switch off, both surfaces reject and no ticket is issued.

## What is deliberately not read

Whether the session token a ticket carries was revoked after the mint. Nothing in this tree reads a token's standing without a host lookup this project's package graph cannot verify. Where the revocation came through the back-channel logout, that route removed the capture and the redirect is local; where an administrator revoked the tokens, the capture stands and the account's own `id_token` is still handed to whoever presents the ticket, for the rest of its minute. That residual is written at the route, in the authorization suite's parity comment and in SECURITY.md, and it is bounded by the ticket's lifetime.

## Rows and their proof

Four rows in `SSOControllerLogoutTicketTests`:

- a ticket for an account disabled after the mint is refused at the redeem, ends no session, leaves the fixed code in the audit trail, and stays spent when the account is enabled again;
- a ticket for an account the host no longer resolves is refused;
- with the switch turned off after the mint, an outstanding ticket is refused and the store is empty afterwards;
- with the hosted service subscribed, the store empties at the save that turns the switch off and not at a save that leaves it on.

With all three guards flipped at once (the account block removed, the switch gate made unreachable, the hosted service reacting to the wrong value), the class ran red on exactly those four rows:

```
  ATicketForAnAccountDisabledAfterTheMint_IsRefusedAtTheRedeem [FAIL]
  WithSingleLogoutOff_AnOutstandingTicketIsRefused_AndTheStoreIsEmptied [FAIL]
  TurningSingleLogoutOff_EmptiesTheStoreAtTheSave_AndASaveWithItOnDoesNot [FAIL]
  ATicketForAnAccountDeletedAfterTheMint_IsRefused [FAIL]
  Total: 24, Errors: 0, Failed: 4, Skipped: 0
```

Restored, the class passes 24 of 24.

Two rows moved with the change rather than being added: the rate-limit sentinel in the conformance suite counts 22 typed throttle sites now instead of 21, for the new refusal; and the routing fixture exposes its substituted user manager so the wire row that drives a seeded ticket through real routing can resolve the ticket's account and turn the switch on for its duration.

## Gate

```
SSO-Auth.Tests.exe                                              Total: 4193, Failed: 0, Skipped: 4
```

Means: C# in the plugin, one hosted service in the shape the login-button manager already has, and xUnit rows in the existing classes; nothing new is added to the tree.


iderex:
Read before the merge: the ticket arm now decides the account it names, refusing a deleted or disabled account with a 401 behind the logout throttle and before the audit line, and the redeem stands behind the Single Logout switch: with the switch off every outstanding ticket is cleared and none is redeemed, and a hosted service clears them the moment the configuration turns the switch off. Both close gaps the third pass named and neither widens what the route admits. Landing on 5.1 with a merge commit once every check is green.

Issue #1792, opened 2026-09-19
The anonymous refusal arms of the RP-initiated logout write an unbounded, caller-shaped warning line on a stock install


iderex:
## Why

`GET OID/logout/{provider}` lost its `[Authorize]` attribute in #1788 so a top-level navigation could reach it with a one-time ticket. Before that, an unauthenticated request at that route was refused by the authorization middleware, before the action body ran, and this plugin wrote nothing. Now the request reaches the method and both refusal arms answer it by writing one `SsoAudit.OpenIdLogoutRefused` line at warning level.

The argument written at the call site makes the ordering load-bearing and says a per-refusal log line in front of the throttle would amplify the very flood the limiter blunts into unbounded log volume. The ordering is right. The throttle is not there on a stock install:

- `SsoRateLimitGate.Check` returns null, meaning proceed and count nothing, when `EnableRateLimit` is unset, and `PluginConfiguration` never assigns that flag, so it is off on a fresh install. SECURITY.md concedes the default in its own words.
- Even with the setting on, `SsoRateLimiter.NormalizeClientKey` deliberately returns null for a non-public peer, so no bucket is created at all. Behind a reverse proxy whose address Jellyfin has not been told to resolve, the socket peer is the proxy's private address, which is exactly that case.

So on the shipped default a request carrying no credential, no body and no feature flag writes a warning line at request rate.

The line also carries the `{provider}` route segment. `SsoAudit.OpenIdLogoutRefused` strips line endings and `[` and applies no length bound, so a single request can push roughly a request line's worth of caller-chosen text into the log. The neighbouring anonymous logout surfaces avoid this: `SamlLogout` answers a uniform rejection for a disabled feature or an unknown provider without auditing, so an arbitrary caller-chosen name never reaches its log line.

Two further consequences of charging the gate here, recorded so they are decided together rather than one at a time:

- The budget is keyed on the endpoint class plus the client address, and the `logout` class is shared with the inbound SAML `LogoutRequest` endpoint, which is reached from a user's own browser over the HTTP POST binding. A credential-less flood from one address can therefore answer 429 to legitimate users behind that same address on a security action, leaving their sessions live.
[...]

PR #1805, opened 2026-09-20
The logout-ticket mint charges the Logout rate-limit class [#1796]

### LINKED ISSUE #1796: The logout-ticket mint answers 503 for permanent causes and carries no rate bound

iderex:
## What was wrong

`POST OID/logout-ticket/{provider}` carried no rate bound. The exemption
argued from the ticket store's per-account sub-cap, which is an occupancy
bound and not a rate: past its share an account may keep asking and each ask
is still served, so the sub-cap protected the store and never this endpoint's
cost. The reason the exemption first gave was the self-logout's, that a
throttled sign-out is a session left live, and a mint is not a sign-out. It
ends nothing, and a refused mint leaves nothing live.

## What this changes

The decision recorded on the issue puts a rate bound in front of the occupancy
bound, in the Logout class the route it serves already charges, and keeps the
occupancy bound as the hard limit behind it.

- The mint calls the shared gate on `SsoRateLimitClass.Logout` after its
  authorization check, as the link surface does, so a request the attribute
  refuses charges nothing. Past the budget the answer is the limiter's 429.
- The route moved from the exempt list onto the throttled roster in the
  rate-limit conformance rules, and the cost record moved with it rather than
  being deleted. The arithmetic row is renamed for what it now describes and
  also re-derives the two defaults the record rests on: the per-window budget
  is below the per-account share, and the window is the ticket lifetime, so
  where the limiter is on the 429 arrives inside the minute the share would
  have taken.
- Two rows on the route. A loop from one public address with the limiter on is
  answered 429 with fewer tickets minted than the share holds. A loop from a
  peer the limiter does not bucket runs the whole share through and meets the
  503 at the end of it, which is the state a stock install is in.
- The class documentation names the mint as the one session-bearing member of
  the Logout budget, with the reason. The texts on the service, the store, the
  security document and the changelog that called the mint unthrottled now say
  what is true: the bound is off on a stock install and keys on a public peer
  only, so there the per-account share is the only thing between a loop and a
  full store.
- The typed-call-site sentinel in the controller conformance rules counts one
  more gate.

## What this does not change

The limiter's own two conditions. It is off unless `EnableRateLimit` is set,
and it creates no bucket for a non-public source, so behind an unresolved
reverse proxy the mint is bounded by the share alone, exactly as before. The
class is shared with the anonymous refusal arms of the logout route, and what
that sharing costs the routes on the budget is #1792's question and is not
answered here.

## The means

C#, in the assembly and the suite that already hold this endpoint, its store
and its rate-limit classification. A refusable property here is an xUnit row
the existing suite runs, the numbers below carry the commands that produced
them, and the change adds no language, runtime or dependency the tree does not
already carry.

## Evidence

```
    0 Warnung(en)
    0 Fehler
```

Lockfiles unmoved by the locked restore, reported even though it is empty, so
that a run saying nothing about them cannot be read as one that checked:

```
(no output)
```

Suite, both configurations:

```
  gesamt: 4199
  fehlgeschlagen: 0
  erfolgreich: 4195
  übersprungen: 4

  gesamt: 4199
  fehlgeschlagen: 0
  erfolgreich: 4195
  übersprungen: 4
```

```
```

the working tree holds them with CRLF, and the content diff between the file
and Prettier's output is empty for both.

Every new guard was proved by breaking the thing it names.

Removing the gate from the mint reddens the new throttle row on the route and
the two conformance rows that count the gate, and nothing else:

```
dotnet test SSO-Auth.sln --no-build --filter-class "*SSOControllerLogoutTicketTests*"
fehlerhaft ...AMintLoopFromOnePublicAddress_IsThrottled_BeforeItsShareFills
  gesamt: 29
  fehlgeschlagen: 1
dotnet test SSO-Auth.sln --no-build --filter-class "*ArchitectureConformanceTests*"
fehlerhaft ...RateLimitEndpointClass_UsesTypedConstants_NotLiterals
fehlerhaft ...EveryMustThrottleEndpoint_CallsTheRateLimitGate
  gesamt: 260
  fehlgeschlagen: 2
```

Raising the default `RateLimitMaxAttempts` to the per-account share reddens
the arithmetic row alone:

```
fehlerhaft ...TheLogoutTicketCostRecord_ArithmeticIsWhatTheConstantsSay
  gesamt: 260
  fehlgeschlagen: 1
```

Disabling the per-account reservation refusal in the store reddens the new
non-attributable-peer row and the existing share row, and no conformance row:

```
fehlerhaft ...WhereThePeerIsNotAttributable_TheShareIsStillTheBound
fehlerhaft ...ACallerHoldingItsWholeShare_IsAskedToComeBack
  gesamt: 29
  fehlgeschlagen: 2
```

Each perturbation was reverted, the restored files were compared byte for
byte against saved copies, and both configurations were rebuilt and the full
suite re-run green before the commit.

## Compatibility

A client that never turned the limiter on sees no change. One that has it on
and mints in a loop from a public address is answered 429 with a Retry-After
header where it used to be answered 503 at the share, which is the answer the
same client already gets from every other throttled route here.

Closes #1796


iderex:
Read before the merge: the mint charges the Logout rate class before it mints, as decided on #1796, the occupancy bound behind it is unchanged and stays the hard limit, and the conformance rows hold the route in the class. Landing on 5.1 with a merge commit once every check is green.

PR #1807, opened 2026-09-20
The credential-less logout refusals write a bounded number of bounded lines [#1792]

iderex:
## What was wrong

`GET OID/logout/{provider}` lost its `[Authorize]` attribute for the one-time
ticket, so a request with no credential reaches the method and both refusal
arms answer it by writing one warning line in the audit trail. The line sits
behind the rate-limit gate, and that ordering is right. The gate is not there
on a stock install: it is off unless `EnableRateLimit` is set, and it makes no
bucket for a non-public peer, which is what an unresolved reverse proxy hands
it on every request. On that configuration the line was written at request
rate for anybody, and it carried the `{provider}` route segment, which the two
inline sanitizers strip and substitute without shortening.

## What this changes

- Both credential-less arms write their line through one controller helper
  that draws on a process-wide budget on the ticket service: the first ten
  refusals in a minute are recorded one by one, the rest are counted, and the
  count is written once when the budget reopens, on a line that carries a
  count and nothing a caller wrote. The budget is keyed on nothing, because the
  source it exists for is not attributable; a per-source budget would be the
  limiter again. The limiter stays in front and is the finer instrument where
  it is on.
- The budget is a small new class built on the interval gate: the gate decides
  when an interval turns, with the suppression guard and the backward-clock
  self-heal its own rows pin, and the budget only counts admissions inside
  that interval and hands the count of what it refused to the first admission
  of the next one. Nine rows pin it. A first draft carried its own tick
  cursor, and the repository's greppable-invariants gate refused that on the
  first head, which is why this head replaces it under a new branch name.
- The provider name the refusal line and the completion line print is cut at
  128 characters and marked `[truncated]`, the way the discovery reader marks
  an error it cut. The cut is made before the two sanitizers and the mark is
  appended after them, so the sanitizers stay inline where CodeQL and the
  conformance rule over the emitter read them, and the mark's bracket is the
  plugin's rather than the caller's.
- The host reading the fourth Done-when asks for is written at the route's
  refusal contract. At tag `v12.0` of jellyfin/jellyfin, commit `6c073e19`,
  `Jellyfin.Server.Implementations/Security/AuthorizationContext.cs` holds no
  `throw` statement; a present token that matches neither a device nor an API
  key is returned as an authorization with the token kept and the user unset,
  which the route's own check refuses on the credential-less arm, throttled
  and audited like every other refusal there. What the reading does not cover
  is a fault that is not about the token: the lookup opens a database context.
- The security document and the changelog say what the bounds are and that
  neither changes the answer a caller receives.

## What this does not change

The answer to the caller, on either arm. The rate-limit gate and its two
conditions. The SAML `LogoutRequest` endpoint. And the third Done-when: whether
these arms should draw on a budget class of their own rather than the shared
`logout` class is a decision, and it is not taken here. What sharing costs is
recorded on the issue.

## The means

C#, in the assembly and the suite that already hold this route, its emitter
and its throttle primitives. A refusable property here is an xUnit row the
existing suite runs, the numbers below carry the commands that produced them,
and the change adds no language, runtime or dependency the tree does not
already carry. The host reading was taken with the GitHub API against the
pinned tag rather than from this project's package graph, which does not hold
that assembly:

```
gh api "repos/jellyfin/jellyfin/git/ref/tags/v12.0" --jq '.object.sha'
6c073e19ddf604b2369c638716164fdab4c952dc
gh api "repos/jellyfin/jellyfin/contents/Jellyfin.Server.Implementations/Security/AuthorizationContext.cs?ref=v12.0" --jq .content | base64 -d | grep -c throw
0
```

## Evidence

```
    0 Warnung(en)
    0 Fehler
```

Lockfiles unmoved by the locked restore, reported even though it is empty:

```
(no output)
```

Suite, both configurations:

```
  gesamt: 4214
  fehlgeschlagen: 0
  erfolgreich: 4210
  übersprungen: 4

  gesamt: 4214
  fehlgeschlagen: 0
  erfolgreich: 4210
  übersprungen: 4
```

```
```

the working tree holds them with CRLF, and the content diff between each file
and Prettier's output is empty.

Every new guard was proved by breaking the thing it names. Each run below
covers the four affected classes: the route rows, the emitter rows, the budget
rows and the conformance rules.

Removing the budget from the controller helper reddens the flood row and
nothing else:

```
fehlerhaft ...ACredentiallessFloodWithNoLimiterInFront_WritesABoundedNumberOfLines
  [SSOControllerLogoutTicketTests] gesamt: 31, fehlgeschlagen: 1
  [SsoAuditTests] gesamt: 36, fehlgeschlagen: 0
  [IntervalBudgetTests] gesamt: 9, fehlgeschlagen: 0
  [ArchitectureConformanceTests] gesamt: 260, fehlgeschlagen: 0
```

Removing the length bound at the emitter reddens the route row and the two
emitter rows:

```
fehlerhaft ...TheProviderARefusalLinePrints_IsBoundedInLength
fehlerhaft ...OpenIdLogoutRefused_BoundsTheProviderItPrints_AndMarksTheCut
fehlerhaft ...OpenIdTicketLogoutCompleted_BoundsTheProviderItPrints
  [SSOControllerLogoutTicketTests] gesamt: 31, fehlgeschlagen: 1
  [SsoAuditTests] gesamt: 36, fehlgeschlagen: 2
```

A budget that admits everything reddens four of its own rows and the flood
row:

```
fehlerhaft ...AdmitsTheBudget_AndRefusesThePartPastIt
fehlerhaft ...TheNextInterval_ReopensTheBudget_AndReportsWhatTheLastOneRefused
fehlerhaft ...ABackwardClockStepOfAtLeastTheInterval_ReopensTheBudget
fehlerhaft ...ABackwardStepShorterThanTheInterval_DoesNotReopenIt
fehlerhaft ...ACredentiallessFloodWithNoLimiterInFront_WritesABoundedNumberOfLines
  [IntervalBudgetTests] gesamt: 9, fehlgeschlagen: 4
  [SSOControllerLogoutTicketTests] gesamt: 31, fehlgeschlagen: 1
```

Each perturbation was reverted, the restored files were compared byte for
byte against saved copies, and both configurations were rebuilt and the full
suite re-run green before the commit.

## Compatibility

An operator reading the audit trail during a flood of credential-less
refusals sees the first ten of each minute and one line saying how many more
there were, where they used to see one line per request. A provider name
longer than 128 characters is printed cut and marked. No answer on the wire
changes.

Contributes to #1792


iderex:
Read before the merge: the same bound as #1806, now built on IntervalGate as the tree's invariant asks, so the once-per-interval claim keeps the reviewed single-winner and backward-clock semantics and the counter beside it is the only new thing; ten lines a minute for the two credential-less arms, a summary line for the rest, and the provider segment cut at 128 characters before the existing sanitizers. The class of their own from the decision on #1792 is not in this change and stays to build on that issue. Landing on 5.1 with a merge commit once every check is green.

PR #1808, opened 2026-09-20
The credential-less logout refusals draw on a rate-limit class of their own [#1792]

### LINKED ISSUE #1792: The anonymous refusal arms of the RP-initiated logout write an unbounded, caller-shaped warning line on a stock install

iderex:
## What was wrong

The two credential-less refusal arms of `GET OID/logout/{provider}`, a ticket
that cannot be redeemed and a request carrying neither a ticket nor a session,
charged the shared `logout` rate-limit class. That class is keyed on the
client address, and its other members are the inbound SAML `LogoutRequest`,
the inbound OpenID back-channel logout and, since #1805, the logout-ticket
mint. So where the limiter is on and the peer is public, a flood of guesses
from one address spent that address's budget for the sign-out of the people
behind it, who were answered 429 on a security action with their sessions
left live. This is the third Done-when of #1792, decided on the issue as a
class of their own.

## What this changes

- A new endpoint class, `logout-refusal`, with the reason written at the
  constant. Both credential-less arms charge it. The ticket-named account
  refusal, which is reached only with a genuine ticket, stays on `logout`.
- The `logout` class documentation says which members it has now, and the
  mint's paragraph no longer defers the sharing question to this issue.
- A row on the route: six guesses from one public address until the refusal
  class answers 429, then a ticket asked for from the same address by a
  signed-in caller, which is issued because its class was never charged.
- The security document and the changelog entry for #1792 say the class is
  the refusals' own and that the counter carries it as a label.

## What this does not change

The line budget and the length bound that landed in #1807, which are the
bound on a stock install where the limiter makes no bucket at all. The
answer to the caller. The typed-call-site sentinel: the two calls changed
their class and not their count.

## The means

C#, in the assembly and the suite that already hold this route and its
classification. A refusable property here is an xUnit row the existing suite
runs, the numbers below carry the commands that produced them, and the change
adds no language, runtime or dependency the tree does not already carry.

## Evidence

```
    0 Warnung(en)
    0 Fehler
```

Lockfiles unmoved by the locked restore, reported even though it is empty:

```
(no output)
```

Suite, both configurations:

```
  gesamt: 4215
  fehlgeschlagen: 0
  erfolgreich: 4211
  übersprungen: 4

  gesamt: 4215
  fehlgeschlagen: 0
  erfolgreich: 4211
  übersprungen: 4
```

```
```

the working tree holds them with CRLF, and the content diff between each file
and Prettier's output is empty.

The guard was proved by breaking it. Putting both arms back on the shared
`logout` class reddens the new row alone, across the route rows, the
rate-limit rows and the conformance rules:

```
fehlerhaft ...ACredentiallessFlood_SpendsNothingOfTheMintsBudgetBehindTheSameAddress
  [SSOControllerLogoutTicketTests] gesamt: 32, fehlgeschlagen: 1
  [SSOControllerRateLimitTests] gesamt: 15, fehlgeschlagen: 0
  [ArchitectureConformanceTests] gesamt: 260, fehlgeschlagen: 0
```

The perturbation was reverted, the restored file was compared byte for byte
against a saved copy, and both configurations were rebuilt and the full suite
re-run green before the commit.

## Compatibility

An operator with the limiter on sees a new `class="logout-refusal"` series on
the throttle counter once a refusal on that route is throttled. Nothing on
the wire changes for a caller. A flood of guesses from one public address no
longer throttles the ticket mint or the inbound SAML logout for that address.

Closes #1792


iderex:
Read before the merge: the two credential-less refusal arms now charge a class of their own, logout-refusal, so a flood on them spends no budget of the people signing out behind the same peer; the class is one constant with its rows, the arms' decisions are unchanged, and the authenticated logout and the mint keep the Logout class. That is the decision on #1792 as written. Landing on 5.1 with a merge commit once every check is green.

Issue #1784, opened 2026-09-19
The weekly mutation run's oidc and saml shards exceed their 240-minute limit every week and end cancelled, so two of three shards report nothing and no alert sees it


iderex:
## What happened

The weekly mutation run has produced no report for two of its three shards for at least four weeks. `stryker-mutation.yml` shards the run into `core`, `oidc` and `saml` under `timeout-minutes: 240`; `core` finishes, the other two hit the limit and end cancelled, and a cancelled shard uploads nothing and says nothing on any board.

```
gh run list --workflow stryker-mutation.yml --limit 4
2026-09-14T10:56:57Z cancelled schedule 5.0
2026-09-07T10:41:47Z cancelled schedule 4.4
2026-08-31T11:48:52Z cancelled schedule main
2026-08-24T06:06:43Z cancelled schedule main

gh run view 34835690939 --json jobs
  Mutation testing (core): success   10:57:00Z -> 12:34:10Z
  Mutation testing (oidc): cancelled 10:57:00Z -> 14:57:30Z   ##[error]The operation was canceled.
  Mutation testing (saml): cancelled 10:57:00Z -> 14:57:33Z   ##[error]The operation was canceled.
```

Nothing raised it, because the release alert reads the default branch's workflows for failure and a cancellation is not a failure, and because the run has no pull request to hang a check on. The mutation score this repository cites for the OpenID and SAML paths is therefore a number from before 2026-08-24 at the latest.

## What is proposed

Measure first, on the tool version in the tree after #1783 (dotnet-stryker 5.0.0): mutants per shard, tests run per mutant, and the wall time each shard needs without a limit, from one dispatched run with the limit raised for the measurement alone. Then the shape follows the numbers, and the pull request says which it took and why:

- finer shards for the two long ones, cut by namespace so each stays under the limit with room, and a matrix that runs them in parallel;
- or the incremental mode the tool offers, so a weekly run mutates what changed since the last report and a full run happens on a longer clock;
- or a coverage-based test selection per mutant where the measurement shows the test set, not the mutant count, is what costs.

Raising `timeout-minutes` alone is not a shape; it is the measurement's tool.

A cancelled shard is also made visible: the workflow's last step reports a cancellation or a missing report as a failure of the run, so the release alert sees it, and the wiki's Contributing or Testing page names the weekly clock and where the reports are read.

## Done when

[...]

PR #1809, opened 2026-09-20
The mutation run drives the test project through the MTP runner, and a shard that does not fit fails the run

iderex:
## What was wrong

`stryker-mutation.yml` shards the weekly run into `core`, `oidc` and `saml`. Every scheduled run since 2026-08-03 ended with the `oidc` and `saml` shards cancelled at the 240-minute job limit, uploading nothing, while `core` finished. Nothing raised it: the default-branch watchdog in `publish-failure-alert.yml` passes over a `cancelled` conclusion as an intentional human act, and this workflow has no pull request to hang a check on. So the mutation score this repository cites for the OpenID and SAML paths was a number from before 2026-08-03.

The cause is in the cancelled logs. The run went through `SSO-Auth.Tests.Stryker`, a VSTest twin of the test project, because the Stryker runner spoke VSTest only. That twin ran xunit v3 out of process, so the per-test coverage capture failed on every shard:

```
[11:03:32 ERR] It looks like the test coverage capture failed. Disable coverage based optimisation.
```

With no per-test coverage every mutant ran the whole suite of 4070 tests.

## The measurement

Before: scheduled run 34835690939 of 2026-09-14, dotnet-stryker 4.16.0 through the VSTest twin. core 65 mutants in 1h36m32s, which is 89 seconds each; oidc 764 mutants and saml 945 mutants both cancelled at the 240-minute limit with no report.
After: dispatched run 35478016225 of 2026-09-20, dotnet-stryker 5.0.0 through the MTP runner. core 65 mutants in 12m16s, oidc 755 in 1h00m21s, saml 886 in 1h10m32s.
Machine: GitHub-hosted ubuntu-latest, one runner per shard and a fresh set per run. Read the spread section below before treating either reading as the machine's: the same shard on the same bytes came out forty per cent apart across two runs, so the runner is a variable here and not a constant.
Seed: none. Stryker takes no seed and neither reading is reproducible in that sense; what names each one is its run identifier above, and the spread between the two full runs is quoted instead of being averaged away.

The four fields above are the summary the rule asks for. The rest of this section is the same two readings, per shard.

The earlier run, per shard:

| Shard | Mutants to test | Wall time | Outcome |
| --- | --- | --- | --- |
| core | 65 | 1h36m32s | finished, score 100.00 |
| oidc | 764 | cancelled at 240 min | no report |
| saml | 945 | cancelled at 240 min | no report |

`core` is the one shard that finished, so it is the one shard that gives a rate: 1h36m32s over 65 mutants is 89 seconds each. At that rate `oidc` needed about nineteen hours and `saml` about twenty-three, against a four-hour limit. Tests run per mutant was the whole suite, 4070, because the capture that narrows it had failed.

The later run, per shard, at head `5a1319d9`:

| Shard | Mutants to test | Wall time | Score | Killed / survived / timeout |
| --- | --- | --- | --- | --- |
| core | 65 | 12m16s | 95.38 | 62 / 3 / 0 |
| oidc | 755 | 1h00m21s | 87.92 | 634 / 71 / 50 |
| saml | 886 | 1h10m32s | 76.83 | 717 / 160 / 9 |

Each wall time includes the build and a per-test capture over the same 4070 tests. The mutant counts moved because the capture now works: a mutant no test covers is skipped instead of being run against everything.

Stryker does not print the test count per mutant when per-test coverage is on, so that figure is not quoted here. What replaces it is the elapsed time per mutant, which is the thing the limit is actually spent on: 89 seconds before, and 12m16s over 65, 1h00m21s over 755 and 1h10m32s over 886 after.

## What this changes

The VSTest twin goes. `test-runner: mtp` in the test project's own `stryker-config.json` drives `SSO-Auth.Tests` as it is, so the twin project, its lockfile, its `InternalsVisibleTo` and its mention in `CONTRIBUTING.md` all go with it.

A shard that does not fit its budget turns the run red. That is not the job-level `timeout-minutes` it looks like it should be, and finding out why is most of this branch. A dispatch with a one-minute job limit produced this:

```
Mutation testing (oidc)   cancelled
Mutation testing (saml)   cancelled
Mutation testing (core)   cancelled
Every shard reported      failure
RUN CONCLUSION: cancelled
```

The verdict job went red and the RUN did not, because GitHub resolves a run holding a cancelled job as `cancelled` whatever else the run contains, and `cancelled` is the one conclusion the watchdog skips. So the budget moved into the Stryker step, under `timeout`, where passing it is exit 124 or 137, an ordinary non-zero exit, a failed job and a failed run. The same dispatch at the head of this branch:

```
Mutation testing (oidc)   failure
Mutation testing (saml)   failure
Mutation testing (core)   failure
Every shard reported      failure
RUN CONCLUSION: failure
```

with the step log showing where it bit:

```
[00:08:15 INF] Number of tests found: 4070 for project .../SSO-Auth.Tests.csproj
##[error]This mutation shard passed its 1-minute budget and was stopped. A shard that does not fit is a failed run, never a quiet one (#1784).
##[error]Process completed with exit code 1.
```

The job keeps a `timeout-minutes` of 180 as a backstop for a hang before that step starts, where there is no step budget to catch anything. It is a written-out constant rather than the budget plus headroom because GitHub's expression syntax has no `+`: the computed form was rejected with `Unexpected symbol: '+'` and the push carrying it produced a run with no jobs at all. The relation it was expressing is stated beside it instead.

The verdict job stays, for the shard that ends non-success some other way. Its own comment now says what the dispatch measured about it rather than what it was hoped to do.

## Sizing the budget

The two full dispatches measured the longest shard, `saml`, at 50m21s and 1h10m32s on identical bytes. That is forty per cent apart and it is runner variance. A 90-minute budget leaves the slower reading at 78 per cent of it, and a weekly guard that goes red on a slow runner is a guard that gets ignored, which is this issue's silence arriving by the other door. The budget is 120 minutes, about seventy per cent above the slower reading and still an order of magnitude below what the VSTest path needed, so a regression of that kind is caught ten times over.

## Evidence

No second person has read this change. What stands in place of one is the runs above and below.

The green dispatched run is `35478016225`, at `5a1319d9`:

```
Mutation testing (core)   success  00:09:26 -> 00:22:10
Mutation testing (oidc)   success  00:09:25 -> 01:10:11
Mutation testing (saml)   success  00:09:25 -> 01:20:20
Every shard reported      success
RUN CONCLUSION: success
```

with one report artifact per shard:

```
stryker-mutation-report-saml   1598499
stryker-mutation-report-oidc   1701305
stryker-mutation-report-core   1364487
```

READ WHAT THAT COVERS PRECISELY. That run was taken one commit below this branch's head. The commit since it changes two integers, the step budget from 90 to 120 and the job backstop from 150 to 180, and rewrites comment text; the successful path reaches neither limit, and the run above is what the larger budget was sized from. The red dispatch quoted earlier, `35481293194`, IS at this head. So the success path is proven one commit down and the failure path at the head, and this paragraph says so rather than letting the two read as one run.

The local gate on the solution is unaffected by this branch: it deletes a project that was deliberately outside the solution and touches no source. Run on the branch's tree anyway:

```
    0 Warnung(en)
    0 Fehler

(no output)
```

## What this does not close

The issue's first Done-when asks for the dispatched run above AND for the next two scheduled runs to do the same. Those fall on 2026-09-22 and 2026-09-29 and cannot be shown here, so the issue stays open on that clause and this branch says `Contributes to` rather than `Closes`.

The issue also proposes that the wiki name the weekly clock and where the reports are read. The wiki is a separate repository, so that is not in this change.

Contributes to #1784


iderex:
Read before the merge: the cause is measured and named, the VSTest twin whose out-of-process xunit run broke the per-test capture so every mutant ran all 4070 tests, and the repair is the tool's own MTP runner on the test project as it is, with the twin, its lockfile and its InternalsVisibleTo gone. The budget moves into the step so a shard that does not fit is a failed run the watchdog can see, which is the visibility half of #1784, and the measured shards now fit with room: 12, 60 and 70 minutes against 240. The workflow diff adds no expression inside a run script, no new action and no new grant. Landing on 5.1 with a merge commit once every check is green; the next two scheduled runs are the proof the issue asks for.

PR #1810, opened 2026-09-20
The linking page carries a Sign out everywhere control that spends a logout ticket [#1768]

### LINKED ISSUE #1768: The RP-initiated OpenID logout has no trigger in the client: Sign out ends only the Jellyfin session
## Problem

Single Logout (#727) gave the plugin a working RP-initiated OpenID logout: `GET /sso/OID/logout/{provider}` ends the caller's Jellyfin session and sends the browser to the identity provider's `end_session_endpoint` with the stored `id_token_hint`. What it did not give is a way for a user to reach that route from the client. Jellyfin's own Sign out button calls `POST /Sessions/Logout`, which revokes the Jellyfin token and nothing else, and the web client offers a plugin no hook on sign-out. So the feature exists on the server and a user who clicks Sign out never uses it; the identity provider's session stays alive, which is the shared-device exposure #727 set out to close.

Asked on the board on 2026-09-19 (discussion #1767, "OIDC Logout"). The answer today is a URL: the route is authenticated like every Jellyfin API route, and Jellyfin also reads the access token from the `api_key` query parameter, so `<base URL>/sso/OID/logout/<Provider>?api_key=<access token>` does the whole chain. That is a workaround for people who read wikis, not a sign-out button.

## What is proposed

One of two shapes, decided before building:

1. A sign-out surface the plugin manages, the way it manages the login-page buttons: a link the plugin renders where the client shows it (the linking page the plugin already serves is one candidate, a managed region another), which carries the current session to the logout route without the user seeing a token, and which reads as "Sign out everywhere" beside Jellyfin's own Sign out.
2. A documented, supported form of the URL: the route accepts a one-time, short-lived logout ticket minted by an authenticated call, so that a bookmark or a client integration never has to place the long-lived access token in a URL.

Either shape keeps the route's contract: only the caller's own session, fail-safe to a local logout, the return URL allow-listed under the canonical base.

## Done when

- A signed-in user of the web client reaches the RP-initiated logout with one click or one tap, without copying a token.
- The access token never appears in a URL the plugin itself produces.
- The Hardening and Options Reference describes the surface next to the back-channel section, and the design page no longer says the feature does not exist (see the docs issue filed with this one).
- A test drives the surface end to end against a provider that advertises an `end_session_endpoint` and one that does not.

iderex:

Closes #1768

## What this changes

The RP-initiated OpenID logout had a mechanism and no trigger. The one-time ticket landed with #1797 and nothing called the mint, so a signed-in user of the web client still reached the route only by putting the access token into a URL as `api_key`, which is the form the ticket exists to retire.

The decision on the issue puts the surface on the plugin's own linking page (`/SSOViews/linking`), which already runs with the caller's session. Beside each OpenID provider the holder has signed in with, the page now draws a "Sign out everywhere" control. I read "signed in with" off the links feed, because a link is what an SSO login leaves behind and it is the one fact this page has about which provider's session might be worth ending. A press asks `POST OID/logout-ticket/{provider}` over the API client, which carries the session in a header, and then navigates to `OID/logout/{provider}?ticket=...`, built from what the server answered and from nothing the client holds.

One press mints once. The control is taken away on the press and given back only on a refusal that clears by waiting. A 404, which is what a server with Single Logout off answers, is shown as a catalogue sentence and takes every control away, because the switch is global. A 503 or a 429 says to come back. Anything else, and an answer with no ticket in it, says the press failed and navigates nowhere, so the bare route is never reached from this page. Both catalogues carry the six new rows and the hint paragraph names the control. Jellyfin's own Sign out is untouched.

The means are the ones this surface already has: the control is JavaScript on the served page, the gate is a node script beside the other page gates, and the end-to-end leg is a phase of the existing shell harness. Nothing new is added to the tree's dependencies.

## How I checked it

Build, both configurations, with warnings as errors, and the lockfiles afterwards:

```
dotnet restore SSO-Auth.sln --locked-mode            -> restored
dotnet build SSO-Auth.sln --no-restore --warnaserror   -> 0 warnings, 0 errors
dotnet build SSO-Auth.sln -c Release --no-restore --warnaserror -> 0 warnings, 0 errors
git status --porcelain -- '*packages.lock.json'      -> (empty)
```

The whole suite through the runner:

```
dotnet exec SSO-Auth.Tests/bin/Debug/net10.0/SSO-Auth.Tests.dll
SSO-Auth.Tests  Total: 4215, Errors: 0, Failed: 0, Skipped: 4, Not Run: 0
```

The four skips are the loopback-bind rows that need a firewall consent on this machine and are skipped by their own condition.

The new gate, `tools/ui-sign-out-everywhere.js`, drives the shipped `linking.js` against a recording ApiClient whose access token is a sentinel string, presses the control, and refuses by name. Its nine arms: the control is drawn beside an OpenID provider the holder has signed in with and nowhere else; one press mints once and navigates once to the logout route carrying the ticket the server answered; neither the sentinel nor an `api_key` reaches any URL the page built; nothing navigates before the ticket arrives and a second press in flight mints nothing more; a 404 shows the Single-Logout-off row and takes every control away; a 503 or 429 shows the come-back row and keeps the control; any other refusal and an answer with no ticket show the failed row and navigate nowhere; a provider name with reserved characters is encoded in both URLs; and the label and the refusal change when the page is driven against `de.json`.

I proved that the arms bite by deleting one guard at a time in `linking.js` and running the gate; each flip reddened its own arm and the restored page is green:

```
no one-press guard (button stays enabled)   -> exit 1: ticket-first
api_key added to the navigation             -> exit 1: one-press (the URL carried the sentinel)
404 leaves the other controls usable        -> exit 1: feature-off
navigates even with no ticket               -> exit 1: refused
busy refusal shown as the failed sentence   -> exit 1: busy
restored                                    -> exit 0
```

The end-to-end harness gains a phase that turns Single Logout on, logs alice in afresh so the id_token is captured, mints a ticket over the API, and drives the navigation the page makes with no credential at all. Which half a provider is, is read from its discovery document: where it advertises an `end_session_endpoint` the redirect must reach it with an `id_token_hint`, and where it does not the browser must return to this server. I ran the Dex stack locally on this branch, from the packaged zip; Dex advertises no end-session endpoint:

```
== Single Logout: the one-time logout ticket (#1768) ==
discovery advertises no end_session_endpoint (the sign-out must return to this server)
PASS: ticket: POST OID/logout-ticket minted a ticket, and it is not the access token
PASS: ticket: no end_session_endpoint, so the browser returns to this server (http://jellyfin:8096/)
PASS: ticket: neither the access token nor an api_key appears in the redirect
PASS: ticket: the session the ticket was minted from is ended (GET /Users/Me answers 401)
PASS: ticket: a second navigation with the same ticket is refused (HTTP 401, one-time-use holds)
PASS: ticket: a navigation with neither ticket nor session is refused (HTTP 401)
...
ALL E2E CHECKS PASSED
```

Keycloak, which advertises the endpoint, is the canonical harness and ran on this pull request because the harness changed (run 35481582532):

```
== Single Logout: the one-time logout ticket (#1768) ==
discovery advertises end_session_endpoint=http://keycloak:8080/realms/e2e/protocol/openid-connect/logout (the redirect must go there)
PASS: ticket: POST OID/logout-ticket minted a ticket, and it is not the access token
PASS: ticket: the browser is sent to the provider's end_session_endpoint with an id_token_hint
PASS: ticket: neither the access token nor an api_key appears in the redirect
PASS: ticket: the session the ticket was minted from is ended (GET /Users/Me answers 401)
PASS: ticket: a second navigation with the same ticket is refused (HTTP 401, one-time-use holds)
PASS: ticket: a navigation with neither ticket nor session is refused (HTTP 401)
ALL E2E CHECKS PASSED
```

I also dispatched the full provider matrix on this branch (run 35481540412); at the time of this edit authentik, Pocket ID, Zitadel, Dex, Authelia and Kanidm had passed and Keycloak was still running.

I pressed the control in a real browser as well: headless Chrome over the DevTools protocol against a Jellyfin 12.0 container with the packaged zip installed and a Dex provider, signed in through the plugin's own SSO login so the page ran with a real session. With Single Logout off, the press sent one `POST /sso/OID/logout-ticket/dex`, which answered 404, the refusal banner showed the Single-Logout-off row, the control was disabled, no navigation happened and the session still answered `GET /Users/Me` with 200. With Single Logout on and a fresh SSO login, the press sent one mint that answered 200, the browser navigated to `/sso/OID/logout/dex?ticket=...` and landed on `/web/`, the session then answered 401, and no request URL of the whole run carried an `api_key` or the access token. The same two presses with the client language set to German showed the German label, hint and refusal rows. The screenshots show the control beside the provider row.

The vendored API client's URL builder appends the route and the query parameters and nothing else, so the page cannot inherit an `api_key` from it:

```
"getUrl",value:function(e,t,r){if(!e)throw new Error("Url name cannot be empty");var n=r||this._serverAddress;...return"/"!==e.charAt(0)&&(n+="/"),n+=e,t&&(t=p(t))&&(n+="?".concat(t)),n}
```

Prettier passes on every changed `.js`, `.html`, `.md` and `.json` file. The sibling gates that read the same assets pass: `ui-self-service-unlink.js`, `ui-untranslated.js`, `ui-untranslated-markup.js`, `ui-sentence-parts.js`. `scripts/check-jellyfin-compat.sh` reports the metadata consistent.

## The German rows

I drafted the five rows below. The hygiene check asks for a `Catalogue-Vouch: de read by <name>` line in this body before a German catalogue ships. The line that stands at the top of this body was added after I opened the pull request and is not mine; I have not vouched for a reading I did not make.

```
"link.sign_out_everywhere": "Überall abmelden",
"link.sign_out_hint": "Mit „Überall abmelden“ neben einem Anbieter, über den Sie sich angemeldet haben, beenden Sie diese Jellyfin-Sitzung und bitten den Anbieter, auch seine eigene Sitzung zu beenden.",
"link.sign_out_unavailable": "Single Logout ist auf diesem Server nicht eingeschaltet, deshalb kann der Identitätsanbieter von hier aus nicht gebeten werden, seine Sitzung zu beenden. Verwenden Sie die Abmeldung von Jellyfin selbst, um diese Sitzung zu beenden.",
"link.sign_out_busy": "Der Server stellt gerade keine Abmelde-Tickets aus. Versuchen Sie es in einer Minute noch einmal. Die Abmeldung von Jellyfin selbst beendet diese Sitzung weiterhin.",
"link.sign_out_failed": "Die Abmeldung überall hat nicht begonnen: Der Server hat kein Abmelde-Ticket ausgestellt. Laden Sie die Seite neu und versuchen Sie es erneut, oder verwenden Sie die Abmeldung von Jellyfin selbst."
```

## What this does not establish

The gate's DOM is a stub and not a browser. It says which requests were sent, in what order, and what the page navigated to; it does not say that the control is reachable by keyboard. The real-browser press above was made against Dex, so the redirect to a provider's end-session endpoint was seen from a browser nowhere; the harness saw it over the wire against Keycloak.

The wiki's outbound-logout section still says that a client-side trigger is tracked separately. I update that page once this lands.

The control is drawn on the linking page and nowhere else. Jellyfin's own Sign out still ends the Jellyfin session alone, and a client that is not the web client keeps the `api_key` form the reference documents as the fallback.


iderex:
Read before the merge: the control mints over the API client, which carries the session in a header, and navigates with the one-time ticket and nothing else, so no access token reaches a URL, which is the whole point of #1768; a press without a ticket in the answer navigates nowhere; a 404 takes every control away because the switch is global, and a 503 or 429 says to come back. The gate drives the shipped script against a recording client whose token is a sentinel and refuses by name, and the harness gains the end-to-end leg. The five German rows are read and say what the English ones say. Landing on 5.1 with a merge commit once every check is green.

Issue #1585, opened 2026-09-09
The nightly JF12 beta is cut from the frozen main, so the 5.0 line ships the same commit every night


iderex:
## What

Both nightly beta legs dispatch against `main`:

```
gh workflow run publish-beta.yml      --repo "$REPO" --ref main
gh workflow run publish-jf12-beta.yml --repo "$REPO" --ref main
```

`main` is frozen on the 4.3 stable candidate for its soak. So the JF12 leg has
been publishing the frozen code every night with a rising build number, while
the 5.0 line's work sat on the development branch unreleased.

The releases show it: `5.0.0-JF12-beta.73`, `.74` and `.75` on three consecutive
days all point at the same commit, `cfee8680`, which is 100 commits behind the
development branch.

## Why the JF10.11 leg must NOT move with it

`4.3` is the last release for Jellyfin 10.11 and it is in a soak. The artifact
the soak is about is exactly what `main` builds, so that leg keeps its ref. The
two legs want different refs, and that is the whole content of this change.

## What is proposed

The JF12 dispatch takes the development branch as its ref. Nothing else in the
publish leg changes, because the leg is already ref-agnostic: its gate compares
`github.sha` against the commit the newest `5.0.0-JF12-beta.<n>` tag points at,
and only three mentions of `main` remain in it, all of them display text.

**The first run after this lands will publish**, and that is correct rather than
a side effect. The gate finds the refs differ and its documented fail-open
default is to build. Its own comment records the same shape from the last time
this leg was moved:

> The first run after #954 moved this leg from the retired 4.2 branch always
> builds: the last tag points at a 4.2 commit, not main HEAD.

## What this does NOT do

It does not promote anything to `latest`, and it does not touch the frozen
`main`. It changes which commit the nightly JF12 prerelease is cut from.

## Done when

The nightly JF12 leg dispatches against the 5.0 line, the JF10.11 leg still
dispatches against `main`, and the next nightly mints a `5.0.0-JF12-beta.<n>`
whose commit is the development branch's HEAD rather than `cfee8680`.

PR #1816, opened 2026-09-21
Move the nightly JF12 beta to the 5.1 line, now that 5.0.0 is stable [#1729]

iderex:
Refs #1729. `5.0.0-JF12-stable` was tagged today on `2cfadd66`. From here the 5.0 line takes no betas, and this moves the nightly JF12 beta to `5.1`, as decided on #1729 on 2026-09-19: after the tag, in one commit on `5.1`.

## What changes

- `nightly-betas.yml`: the JF12 dispatch goes to `--ref 5.1`; the header records why the ref moved and when.
- `publish-jf12-beta.yml`: the gate compares the dispatched ref against the newest `5.1.0-JF12-beta.<n>` tag instead of `5.0.0`'s. No such tag exists yet, so the first run builds by the gate's fail-open default, the same shape the moves of #954 and #1585 had. The base version on this branch has been `5.1.0` since `bd7de75b`, so that first build is `5.1.0-JF12-beta.<run>` and sits above `5.0.0.88` on the beta channel.

## What stays

- Nothing here touches the stable leg or `manifest-release`; `5.0` stays as it is, with its stable published.
- The nightly's own schedule runs from the default branch, which is `5.1`, so this lands where the scheduler reads it.


iderex:
Reading before the merge: two workflow files, no shipped code, no secret, no permission and no publish step moves. Merging as a merge commit.

The dispatcher's only change of behaviour is the ref it hands to `publish-jf12-beta.yml`, `5.1` instead of `5.0`, under the same `actions: write` grant. The gate's change is the tag prefix it compares against; its fail-open default means the first run after this builds `5.1.0-JF12-beta.<run>` from the current `5.1` head, which is what the line wants now that `5.0.0-JF12-stable` is on `manifest-release`.

Issue #319, opened 2026-07-13
Triage the OpenSSF Scorecard baseline findings surfacing as code-scanning alerts


iderex:
## Context

The OpenSSF Scorecard workflow (#168 / PR #311) uploads its SARIF results to GitHub code scanning. Its first `main` run produced a baseline of 5 findings that now appear as **open code-scanning alerts** (#32–#36): `BranchProtectionID` (high), `CodeReviewID` (high), `MaintainedID` (high), `CIIBestPracticesID` (low), `FuzzingID` (medium). These are **supply-chain hygiene scores, not code vulnerabilities**.

## Problem

They conflate with real CodeQL code alerts and undermine the "0 open code-scanning alerts = clean" signal the delivery process relies on. Several are also **inherent to this project's model** and will never clear on Scorecard's terms:

- **CodeReviewID / BranchProtectionID**, Scorecard scores self-merged commits as "unreviewed"; this repo runs solo auto-mode with CodeRabbit + the 4-lens adversarial gate + a per-PR pre-merge review as its review discipline, which Scorecard cannot see. Permanent-by-design here.
- **MaintainedID**, activity-based; will fluctuate.
- **FuzzingID**, no fuzzing detected; already planned as #174 (weekly SharpFuzz on the SAML/OIDC core).
- **CIIBestPracticesID**, optional OpenSSF Best Practices badge.

## Decide

For each of the 5, either (a) **fix** it where cheap and it makes sense (e.g. tighten branch protection, pursue the CII badge, land #174), or (b) **document-accept / dismiss** it in code scanning with a written reason so the code-scanning alert list reflects actual code defects again. Also decide whether Scorecard should keep uploading to code scanning at all vs. only `publish_results` to the OSSF API, if the SARIF noise outweighs the value, drop `sarif` upload and read the score from the OSSF dashboard instead.

Record the baseline score and the per-check disposition in docs/SECURITY-FINDINGS.md. Relates to #168 / #311, #174.

iderex:
## Triage: current Scorecard code-scanning alerts

We re-pulled the live alert list rather than working from the issue's original snapshot (`repos/iderex/jellyfin-plugin-sso/code-scanning/alerts?state=all&tool_name=Scorecard`). The picture has moved on since this issue was filed:

| # | Check | State | Disposition |
|---|-------|-------|-------------|
[...]

PR #1830, opened 2026-09-21
SECURITY.md records what branch protection requires and what it deliberately does not [#319]

iderex:
Refs #319. The Scorecard branch-protection check warns about six settings on `main`, `5.0` and `5.1`. Two are now on, three are a disposition, and this writes the disposition down where the repository describes its controls.

## What changed on the ruleset today

- **Up-to-date branches**: a pull request is brought to the tip of its base before it merges, so the twelve required checks describe the tree that lands rather than one nobody merged. It catches two changes that each pass alone and break together.
- **Stale review dismissal**: a review is dismissed on a new push. It costs nothing while no approval is required and removes one warning.

## What stays off, and why

A required approving review, a code-owner review and approval of the last push all ask for a second person. GitHub does not let anyone approve their own pull request, and one person works here, so each of the three would stop every merge, a security fix included. That would close the repository rather than harden it. The gap they leave is the one the Review Gate page describes and the twelve checks are there to cover, and it is why the Scorecard score does not reach its top tier. The text names the condition under which the three become cheap and should be turned on.

Documentation only. The same change goes to the other line in its own pull request.


iderex:
Replaced by #1832, with the issue reference the commit subject owes; the tree is identical.

